PR #41, opened 2020-11-13
Mosaic#39 add --debug flag

denismaxim0v:


src/pty_bus.rs
```diff
@@ -34,6 +34,26 @@ fn _debug_log_to_file (message: String) {
...
+    use std::fs::OpenOptions;
+    use std::io::prelude::*;
+    let mut path = PathBuf::new();
+    path.push(
+        [
+            String::from("/tmp/mosaic-logs/mosaic-"),
+            pid.to_string(),
+            String::from(".log"),
+        ]
+        .join(""),
```
TheLostLambda, line 47: Just a very tiny Rust nitpick, but for joining with an empty string, there is [the `.concat()` method](https://doc.rust-lang.org/std/primitive.slice.html#method.concat) that might be a bit more idiomatic :)

TheLostLambda (commented):
This is looking pretty good to me! I left one nitpick and it might be nice to have some tests, but I'll be honest, tests for this seem a bit challenging to write.

Regardless, thanks a ton for writing this up! I'll probably defer to @imsnif to make the final call though :)

denismaxim0v:
Great, let me know what you end up with

imsnif:
@denismaxim0v - this looks great to me. It will also be extremely helpful in debugging compatibility issues.
I played with this around a little on my machine and it definitely does the right thing. One can use these logs to debug issues by simply doing `cat <logfile>` inside a mosaic pane (and outside for comparison). The mosaic-logs folder is re-created on startup. All looks good.

@denismaxim0v - I think if you just apply @TheLostLambda's suggestion, we can merge this.

Thanks for this!

PR #47, opened 2020-11-15
Add variable to track if the pane is in fullscreen mode or not

### LINKED ISSUE #46: Fullscreen command should be blocked when there is only 1 pane open
To reproduce:
* Start the app
* Toggle full-screen (currently with ctrl-e)
* Open another pane - see that things start to get messy (eg. no borders are drawn)

To fix this, the full-screen command should be a no-op when there is only one pane open.

henil:
Fixes #46


src/screen.rs
```diff
@@ -92,6 +92,7 @@ pub struct Screen {
     panes_to_hide: HashSet<RawFd>,
     active_terminal: Option<RawFd>,
     os_api: Box<dyn OsApi>,
+    is_fullscreen: bool,
```
imsnif, line 95: Can we change the name to fullscreen_is_active? I feel it'll be a little clearer.

imsnif (changes requested):
Great and very quick work on this!
I left one small nitpick about naming.

Other than that, do you think we can add a test for this? You can look at the other tests for inspiration... something like "add a new terminal, toggle it to full screen, add another terminal and see that it does the right thing"?
Our tests use snapshot testing, so they take a "snapshot" of the app each time it renders. We use cargo insta for this. There's no docs for it yet, so feel free to hit me up here or on Discord and I'd be happy to walk you through it.

src/screen.rs
henil: agreed. I am not so good with names XD.

henil:
(rebased to upstream master to resolve conflict)

imsnif:
Test's looking good! To get it to run, you need to add the new file to mod.rs in the same folder (https://github.com/mosaic-org/mosaic/blob/main/src/tests/integration/mod.rs)
When it runs for the first time, it will fail (because there are no snapshots for it). You can use `cargo insta` (https://docs.rs/insta/1.2.0/insta/) to review the snapshots and add them to this PR.

I won't be 100% available today (we're going into lockdown tomorrow and there's a lot to do before), but I'll try to take a look later today or worst case tomorrow.

imsnif (approved):
Looks great! Thanks for the changes.
Note that right now this causes some extraneous rendering. `toggle_active_terminal_fullscreen` includes a call to `self.render()` on its own. So we do that, render and then we render again with the split. This is really okay though and not noticeable at all. I think we can address this at another time.

Merging this.

PR #57, opened 2020-11-18
Total layout percent 100 validation

5c077m4n:


src/tests/integration/layouts.rs
```diff
@@ -49,3 +49,45 @@ pub fn accepts_basic_layout() {
     assert_snapshot!(next_to_last_snapshot);
     assert_snapshot!(last_snapshot);
 }
+
+#[test]
+#[ignore]
```
5c077m4n, line 54: @imsnif I ignored these tests as they pass when they run on their own but fail when they run together with all of the others (locally)

src/tests/integration/layouts.rs
```diff
@@ -49,3 +49,45 @@ pub fn accepts_basic_layout() {
...
+    let mut opts = Opt::default();
+    opts.layout = Some(PathBuf::from(
+        "src/tests/fixtures/layouts/parts-total-more-than-100-percent.yaml",
+    ));
+
+    start(Box::new(fake_input_output.clone()), opts);
+}
+
+#[test]
+#[ignore]
```
5c077m4n, line 75: ^^^

src/tests/utils.rs
```diff
@@ -36,18 +36,30 @@ pub fn get_output_frame_snapshots(output_frames: &[Vec<u8>], win_size: &Winsize)
...
-    pub const RESIZE_UP: [u8; 10] = [11, 0, 0, 0, 0, 0, 0, 0, 0, 0]; // ctrl-k
-    pub const MOVE_FOCUS: [u8; 10] = [16, 0, 0, 0, 0, 0, 0, 0, 0, 0]; // ctrl-p
-    pub const RESIZE_LEFT: [u8; 10] = [8, 0, 0, 0, 0, 0, 0, 0, 0, 0]; // ctrl-h
-    pub const RESIZE_RIGHT: [u8; 10] = [12, 0, 0, 0, 0, 0, 0, 0, 0, 0]; // ctrl-l
-    pub const SPAWN_TERMINAL: [u8; 10] = [26, 0, 0, 0, 0, 0, 0, 0, 0, 0]; // ctrl-z
-    pub const QUIT: [u8; 10] = [17, 0, 0, 0, 0, 0, 0, 0, 0, 0]; // ctrl-q
-    pub const SCROLL_UP: [u8; 10] = [27, 91, 53, 94, 0, 0, 0, 0, 0, 0]; // ctrl-PgUp
-    pub const SCROLL_DOWN: [u8; 10] = [27, 91, 54, 94, 0, 0, 0, 0, 0, 0]; // ctrl-PgDown
-    pub const CLOSE_FOCUSED_PANE: [u8; 10] = [24, 0, 0, 0, 0, 0, 0, 0, 0, 0]; // ctrl-x
+    /// ctrl-b
```
henild, line 39: Any particular reason for putting them on a new line? For me, they seem to be more readable when they are on the same line.

henil, line 39: Any particular reason for putting them on a new line? For me, they seem to be more readable when they are on the same line.

src/tests/integration/layouts.rs
henil: This is happening due to multiple threads running and causing a race condition. Try increasing the duration from this line to see if this helps. https://github.com/mosaic-org/mosaic/blob/6f168c5ef235bf743668fa8f58d5d55344887b63/src/tests/fakes.rs#L39

src/tests/utils.rs
5c077m4n: It's for IDEs, the triple slash in the prev line is just like jsdoc

src/tests/integration/layouts.rs
5c077m4n: Thank you :)

5c077m4n: @henil The problem here is that the `panic` happens inside a single thread and so does not stop the main one - so the program just hangs. I've tried to use `panic::set_hook` and `process::exit` inside it but that exits the entire test process... Is there a clean way of closing the parent thread form its child just for this test or do you think that it should just be left for unit tests?

henil: I am not sure about that too. @imsnif any ideas?

src/tests/utils.rs
khs26: Could we avoid this change for now? I've got a (slightly) more substantial change coming and will incorporate this here.

roee-shapira-namogoo: Sure

imsnif:
Hey, so I'm going to address everything here including the different threads in this PR.

1. The panic indeed happens because one thread panics. It's not ideal, but it does indicate a problem. I think the problem here is that we're missing the layout fixture `src/tests/fixtures/layouts/parts-total-more-than-100-percent.yaml`. We also seem to be missing the validation for the 100% total itself... could it be that this PR is missing some local changes?
2. I see you already removed what @khs26 asked for, thanks! I think it's definitely best to address this in the other PR. :)

There are some problems with atomicity in the tests and in getting stdin that might cause some panics. I'm working on them in a different PR. These shouldn't happen in the CI though, so if something fails in the CI it's an actual problem.

roee-shapira-namogoo:
@imsnif the errors that I had was where caused because the panic in child thread did not cause the main thread to panic too, so it just kept on running (and so causing all other tests to fail too).
I tried to catch this panic (`panic::set_hook` & `catch_unwind`) but then I could only exit the process completely, and not just the main thread.

imsnif:
Aha, okay - I understand. Then maybe we can move the layout validation outside the start function, so that it happens before we open all the threads? What do you think?

roee-shapira-namogoo:
@imsnif I think that all validations should happen before we start doing any real work, both for easier debugging (less logs) and for a faster response

roee-shapira-namogoo:
@5c077m4n <- is me, this is just my work account

imsnif:
Cool, so I guess that would solve the panic issue here, right?

roee-shapira-namogoo:
Exactly

src/main.rs
```diff
@@ -98,7 +99,11 @@ pub fn start(mut os_input: Box<dyn OsApi>, opts: Opt) {
     let command_is_executing = CommandIsExecuting::new();
 
     delete_log_dir().unwrap();
-    delete_log_file().unwrap();
+
+    if opts.clear_log_file {
+        // TODO: Remove this once it is decided that it goes into the logs dir
```
roee-shapira-namogoo, line 104: @henil Was it you who wanted to move all of the log files to one directory?

src/main.rs
```diff
@@ -129,54 +134,39 @@ pub fn start(mut os_input: Box<dyn OsApi>, opts: Opt) {
         opts.debug,
     );
 
+    if let Some(layout_path) = opts.layout {
```
roee-shapira-tikal, line 137: @imsnif Why was this in the event loop in the thread in the first place? Am I missing something?

src/main.rs
roee-shapira-tikal, line 104: @henil Was it you who wanted to move the logging files all to one directory?

src/main.rs
imsnif: The original thought was that we either
1. Have a layout, in which case we need to create panes as the layout specified.
2. Don't have a layout in which case we need to create one pane.

I think if we want to do the validation outside, we can pass an Option<Layout> instead of using the Opt and it'll do the same thing. (or use `and_then` on opt.layout if you're feeling fancy)

src/main.rs
henil: Yes, but it was decided it should be done in different PR

roee-shapira-tikal: Yep, just wanted to make sure that you noticed this

src/main.rs
roee-shapira-tikal: The validation is happening now in the `Layout` constructor, do you want to move it to here?

roee-shapira-tikal: @imsnif Fancy is my middle name

imsnif: Whatever you feel would work best

roee-shapira-tikal: @imsnif Since the layout is from a static file I don't think that we need to re-parse it in every iteration of the event loop.
I think that parsing it once will be better for both the logic and performance - unless the initial layout/file content can change then it *has* to be in the loop.

roee-shapira-tikal:
@imsnif Ready!

src/main.rs
henil: This has been decided and I am opening a PR for it. So I think this check and `clear_log_file` from the `Opt` struct should be removed.

henil (changes requested):

src/layout.rs
```diff
@@ -87,6 +94,34 @@ fn split_space(space_to_split: &PositionAndSize, layout: &Layout) -> Vec<Positio
...
+fn validate_layout_percentage_total(layout: &Layout) -> bool {
+    let total_percentages: u8 = layout
+        .parts
+        .iter()
+        .map(|part| {
+            let split_size = part.split_size.as_ref();
+            match split_size {
+                Some(SplitSize::Percent(percent)) => *percent,
+                None => {
+                    // TODO: if there is no split size, it should get the remaining "free space"
```
imsnif, line 106: <3

src/layout.rs
```diff
@@ -108,6 +143,26 @@ pub struct Layout {
...
+                File::open(&path).expect(&format!("cannot find layout {}", &path.display()));
+
+            let mut layout = String::new();
+            layout_file
+                .read_to_string(&mut layout)
+                .expect(&format!("could not read layout {}", path.display()));
+            let layout: Layout = serde_yaml::from_str(&layout)
+                .expect(&format!("could not parse layout {}", path.display()));
+
+            if !validate_layout_percentage_total(&layout) {
```
imsnif, line 158: What do you think about moving this to its own `self.validate` method? I think this will be the first validation of many.

src/main.rs
```diff
@@ -128,33 +131,20 @@ pub fn start(mut os_input: Box<dyn OsApi>, opts: Opt) {
...
-                            ));
-                            pty_bus.spawn_terminals_for_layout(layout);
-                        }
-                        None => {
-                            pty_bus.spawn_terminal_vertically(None);
-                        }
+                    if let Some(layout) = maybe_layout {
+                        pty_bus.spawn_terminals_for_layout(layout);
+                    } else {
+                        pty_bus.spawn_terminal_vertically(None);
```
imsnif, line 145: Looking good :)

src/utils/logging.rs
```diff
@@ -5,7 +5,7 @@ use std::{
     path::PathBuf,
 };
 
-use crate::utils::consts::{MOSAIC_TMP_LOG_DIR, MOSAIC_TMP_LOG_FILE};
+use crate::utils::consts::*;
```
imsnif, line 8: I actually very much prefer to be explicit about imports. I think it makes it much clearer when you're trying to figure out where something is coming from.

imsnif (changes requested):
Good work! I left some minor nitpicks about the code. If you could take a look at them and also merge from main I think we're good to go.

imsnif (approved):
Very cool! Merging

PR #58, opened 2020-11-18
Added cache to ci build process.

5c077m4n:
May help in the future - but should be discussed as to when this will be bad (stale deps)

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -11,11 +11,17 @@ env:
 
 jobs:
   build:
-
     runs-on: ubuntu-latest
-
     steps:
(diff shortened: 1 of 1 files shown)


imsnif:
Hum... so, the benefit we get out of it is just faster tests/builds, right?
I'm honestly a little wary of introducing it right now. I think this isn't really an issue for us, as CI runs quite fast now. Or is there another reason?

imsnif:
@5c077m4n - I think the time has come for this. I think adding this would help with some flaky tests we've been experiencing. The branch is deleted, so I can't re-open this. Would you like to put this together into a new PR?

PR #60, opened 2020-11-19
Simplify geometry stuff

khs26:
Planning to simplify some of the logic around drawing panes and add some geometry helper functionality.


src/boundaries.rs
```diff
@@ -1,5 +1,19 @@
...
+    use std::fs::OpenOptions;
+    use std::io::prelude::*;
+    let mut file = OpenOptions::new()
+        .append(true)
+        .create(true)
+        .open("/tmp/mosaic-log.txt")
+        .unwrap();
+    file.write_all(message.as_bytes()).unwrap();
+    file.write_all("\n".as_bytes()).unwrap();
+}
```
khs26, line 15: I'm not a terrible enough person to leave this in - will pick up the logging functionality

src/boundaries.rs
```diff
@@ -16,204 +30,117 @@ pub mod boundary_type {
...
-        Some(symbol)
-    } else {
-        combine_symbols(second_symbol, first_symbol)
+        (a, b) => combine_symbols(b, a), // If we didn't find a match, swap them
     }
 }
 
-#[derive(PartialEq, Eq, Hash, Debug)]
-pub struct Coordinates {
+pub struct ScreenCanvas {
```
khs26, line 109: The thinking here was potentially to have a more generic type for "Rectangle that can contain other rectangles and knows some stuff about what needs to be drawn". This may already be covered by terminal_pane or similar, I'm not sure

src/boundaries.rs
```diff
@@ -16,204 +30,117 @@ pub mod boundary_type {
...
-    fn bottom_boundary_y_coords(&self) -> usize {
-        self.y() + self.rows()
-    }
-}
-
-pub struct Boundaries {
     columns: usize,
     rows: usize,
     boundary_characters: HashMap<Coordinates, BoundaryType>,
+    borders: HashMap<EdgeType, usize>,
```
khs26, line 115: I think it should probably also have a `children: Box<Vec<ScreenCanvas>>` or similar too

imsnif:
Just leaving a note saying I haven't forgotten this and will take a closer look sometime this week

khs26:
@imsnif it's very much not ready yet - I've been raising draft PRs early so that people can get a chance to share input. Will convert them to full reviews when ready

imsnif:
> I've been raising draft PRs early so that people can get a chance to share input.

For sure, that's what I was talking about :)

src/boundaries.rs
imsnif: I like everything about this PR except this...
Most multiplexers keep data in this way, and while it's very comfortable for us developers it creates some issues for the users. For example, if we want to resize a pane outside of its parent. This can happen for example where:
* we have 4 rectangles on screen (by splitting horizontally and then splitting each one of the halves vertically)
* we resize the bottom-right pane up
* we then resize it back down to where it was
* we resize the same pane left

To my knowledge, we're the only multiplexer that supports this. Every other multiplexer would drag the pane's sibling with it in at least one of these directions. And I think it's because they keep a hierarchical map of the panes.

PR #62, opened 2020-11-20
refactor(structure): Change structure of `/tmp` directory

henil:


imsnif:
Cool, thanks for doing this! I merged from main because there was a minor conflict.

Issue #25, opened 2020-11-05
Hotkeys are a usability problem


thedrow:
The reason I never use tmux is because I can never remember the hotkeys.
It'd be useful to have a command which reminds you what to do.

In addition, it'd be useful to have commands which you can type to perform an action.

imsnif:
I am very much on board with this. What we have now is mostly just a placeholder.

My initial thoughts are to go with a vim-like modal approach. You press a hot-key to go into "control" mode, you get a visual indication that you're in that mode (eg. the borders become dashes) and you have a brief powerline-like help that tells you what you can do in this mode. What do you think?

thedrow:
I'll need to see a screenshot of the concept, but generally, it sounds a lot better, useability wise.

khs26:
I've started thinking about this, along with adding a split between command and normal mode.

@thedrow, I like the idea of also having an "easy" command mode, where you start typing and get some simple auto-complete of available commands

imsnif:
Hey @thedrow - while we're still working out the kinks on this, we've made a lot of progress in this area recently. If you'd like to check out the latest version from main and let us know what you think that would be great :) There's now a status bar which lets you know the keybindings available to you in real time.

thedrow:
I'll check it out.

PR #63, opened 2020-11-21
Hotkeys - WIP

khs26:
Starting to work on custom hotkey definitions - have the very basic framework of saying what the module's goals are and defining some representations for keys.


khs26:
Might go some of the way to addressing #25 but we also need a sensible default layout

Also worth noting that this is currently based off #42 and should look more reasonable when that's committed.

PR #64, opened 2020-11-21
Hotkeys

khs26:
Starting to work on custom hotkey definitions - have the very basic framework of saying what the module's goals are and defining some representations for keys.

Might go some of the way to addressing #25 but we also need a sensible default layout

Also worth noting that this is currently based off #42 and should look more reasonable when that's committed.


src/input/hotkeys.rs
```diff
@@ -0,0 +1,366 @@
...
+pub struct InputKey {
+    /// Base (keyboard) key
+    base_key: BaseInputKey,
+    /// Modifier keys
+    modifiers: HashSet<ModifierKey>,
+    /// The character sequence sent when this key is pressed
+    char_sequence: Vec<u8>,
+    /// How to display this key to the user
+    user_string: String,
+}
```
khs26, line 284: @TheLostLambda this is my current attempt at an input key struct

PR #65, opened 2020-11-21
Added clippy check.

5c077m4n:


roee-shapira-tikal:
@imsnif not too bad, right?

.github/workflows/rust.yml
```diff
@@ -35,4 +36,16 @@ jobs:
...
+    name: Clippy
+    runs-on: ubuntu-latest
 
+    steps:
+      - uses: actions/checkout@v1
+      - uses: actions-rs/toolchain@v1
+        with:
+            toolchain: stable
+            components: clippy
+      - uses: actions-rs/clippy@master
```
roee-shapira-tikal, line 49: @imsnif This needs to be changed asap.

src/terminal_pane/scroll.rs
```diff
@@ -318,17 +316,12 @@ impl Scroll {
...
-                        y += current_line.wrapped_fragments.len();
-                    }
-                }
-                None => break,
+        while let Some((current_index, current_line)) = indices_and_canonical_lines.next() {
+            if current_index == canonical_line_cursor_position {
+                y += current_line.wrapped_fragments.len() - line_wrap_cursor_position;
+                break;
+            } else {
+                y += current_line.wrapped_fragments.len();
```
khs26, line 324: This seems much nicer

src/screen.rs
```diff
@@ -67,6 +67,7 @@ pub enum ScreenInstruction {
     ClearScroll,
     CloseFocusedPane,
     ToggleActiveTerminalFullscreen,
+    #[allow(dead_code)]
```
khs26, line 70: I'm not sure about adding warning suppressions (similarly for the other allow directives)

.github/workflows/rust.yml
imsnif: Sorry, I'm not sure I understand?

src/screen.rs
imsnif: I feel the same... @5c077m4n - could you share the reasoning here?

imsnif: ah, oops - didn't realize this was already changed

imsnif:
@5c077m4n - looking good! I'm fine with everything here. What else needs to be done? Do you also want to take care of some of the warnings we've got?

.github/workflows/rust.yml
roee-shapira-tikal: The master tag as it's not versioned.
But from what I see it's not really being developed actively

imsnif:
Also, could you merge from main? There are some changes there that might need to run through these tests

src/screen.rs
```diff
@@ -142,7 +142,7 @@ impl Screen {
             let pid = new_pids.next().unwrap(); // if this crashes it means we got less pids than there are panes in this layout
             let mut new_terminal = TerminalPane::new(
                 *pid,
-                self.full_screen_ws.clone(),
+                self.full_screen_ws,
```
roee-shapira-tikal, line 145: @imsnif clippy really does not like the `.clone()` here, is it necessary?

imsnif: I think this is `Copy`, so no need for sure. It's probably a leftover from a refactor.

roee-shapira-tikal: 😎

imsnif:
@5c077m4n - so is this ready?

src/utils/logging.rs
```diff
@@ -7,44 +7,42 @@ use std::{
 
 use crate::utils::consts::{MOSAIC_TMP_LOG_DIR, MOSAIC_TMP_LOG_FILE};
 
-fn atomic_create_file(file_name: &str) {
-    let _ = fs::OpenOptions::new().create(true).open(file_name);
+fn atomic_create_file(file_name: &str) -> io::Result<fs::File> {
+    fs::OpenOptions::new().create(true).open(file_name)
```
imsnif, line 11: @5c077m4n - the previous way was actually intentional. This is "atomic" create file, meaning that it's okay for it to fail if the file already exists.

roee-shapira-tikal: Reverted

roee-shapira-tikal:
This is nearly done, there are just a couple of more minor changes that you might want to look at yourself @imsnif (like `src/main.rs:309:41`)

imsnif:
So... is this ready for review or not?

roee-shapira-tikal:
Yep, now it is

src/terminal_pane/terminal_pane.rs
```diff
@@ -733,12 +735,9 @@ impl vte::Perform for TerminalPane {
...
-                        self.scroll.show_cursor();
-                        self.mark_for_rerender();
-                    }
-                    _ => {}
+                if let Some(&25) = params.get(0) {
+                    self.scroll.show_cursor();
+                    self.mark_for_rerender();
                 };
             }
         } else if c == 'r' {
```
imsnif, line 743: @5c077m4n - I can't comment on the lines below directly for some reason, but can we disable this clippy check? The one with the errors about the identical ifs. I would really like to leave it like this for now. The comments are super important and this block is generally very much in-flux as we keep reverse-engineering the VT100 protocol.

src/utils/logging.rs
```diff
@@ -18,33 +18,31 @@ pub fn debug_log_to_file(message: String) -> io::Result<()> {
...
+        debug_log_to_file(message)
+    } else {
+        Ok(())
     }
-
-    Ok(())
 }
 
-pub fn delete_log_file() -> io::Result<()> {
+pub fn _delete_log_file() -> io::Result<()> {
```
imsnif, line 32: The functions in this file are general utility functions. Can we add exceptions for them so that we don't have to prefix them with underscores? It would just be confusing and annoying while debugging :)

imsnif (changes requested):
Really great work on this! I imagine this was not trivial.

I left some general comments. I'm looking into the stdin error (it might be the cause of some of our troubles). But until then (and on top of my comments in the code): why is the build not failing? It has two errors and several warnings. Shouldn't it be red?

roee-shapira-tikal:
@imsnif the tests are not failing because this clippy github actions is not really being developed, I've looked some other possibility but this still looks better
And about the clippy exceptions there are annotations that can disable the warnings (`#[allow(clippy::...)]`)

imsnif:
> @imsnif the tests are not failing because this clippy github actions is not really being developed, I've looked some other possibility but this still looks better

So... I guess this solution doesn't do the job for us? If we can't see the failures on PRs, then this essentially will not be part of our CI. Or am I missing something?

> And about the clippy exceptions there are annotations that can disable the warnings (`#[allow(clippy::...)]`)

Cool - so can we add them?

imsnif:
Also (if we're going forward with this) we should either address the warnings or have them not appear inside the code-browsing. Because it's hard to review PRs with them.

roee-shapira-tikal:
Sure, I'll do what you asked tonight/tomorrow morning @imsnif

imsnif:
> Sure, I'll do what you asked tonight/tomorrow morning @imsnif

Awesome. But just to stress: we need some sort of failure in our CI for this to be meaningful. I don't mind it even being a hack, like a test-case running clippy or whatever :)

roee-shapira-tikal:
😎

5c077m4n:
@imsnif after some research I found that there are three possibilities to clippy check in the ci:
1. an action that can't check PRs
2. an action that doesn't fail the build
3. an action that does not show the errors on the pr reviews
Other than creating a brand new action I don't really an easy way of using this tool...

imsnif:
I see... thanks for all the research! Does number 3 still check PRs and fail the build though?

5c077m4n:
Yep, that's whats configured for now

imsnif:
Alright, so maybe it can be a good compromise?

5c077m4n:
Great :) in that case this branch is ready for your review @imsnif

imsnif (approved):
Great work as always. Thanks for doing this! I think we will all appreciate it :)

PR #73, opened 2020-11-25
The first steps towards WASM module integration

TheLostLambda:
Hello all! This is a grafting of some of the [WASM module loader](https://github.com/mosaic-org/wasm-module-experiments) into mosaic!

There is a lot still to do, but this is the first step!

From here, I'll be adding

- [ ] A way to load "tiles" from a layout file
- [ ] A way to actually render a plugin to a pane
- [ ] Working on parsing input into a struct usable by the plugin system


TheLostLambda:
Also, it's a bit silly to have put those tasks here, check out #74 for an actual tracking issue!

PR #75, opened 2020-11-27
Add custom panic hook. Print backtrace and thread, error info on panic.

kunalmohan:
feat: print backtrace on panic and ensure proper shutdown.

The implementation works as follows- a custom panic hook collects the backtrace, sends it over to the ipc_thread via socket. The ipc_thread unsets `raw_mode` and prints the backtrace and sends a message to the main thread which joins the active threads and exits.

`Sender<AppInstruction>` does not implement `Sync` and thus could not be captured by the closure in `panic::set_hook()`. We, therefore, have to go through `ipc_thread` for signalling the main thread.

### DIFF
--- a/src/errors.rs
+++ b/src/errors.rs
@@ -1,0 +1,47 @@
+use crate::AppInstruction;
+use backtrace::Backtrace;
+use std::panic::PanicInfo;
+use std::sync::mpsc::SyncSender;
+use std::{process, thread};
+
+pub fn handle_panic(info: &PanicInfo<'_>, send_app_instructions: &SyncSender<AppInstruction>) {
+    let backtrace = Backtrace::new();
+    let thread = thread::current();
+    let thread = thread.name().unwrap_or("unnamed");
+
+    let msg = match info.payload().downcast_ref::<&'static str>() {
+        Some(s) => *s,
+        None => match info.payload().downcast_ref::<String>() {
+            Some(s) => &**s,
+            None => "Box<Any>",
+        },
+    };
+
+    let backtrace = match info.location() {
+        Some(location) => {
+            format!(
+                "\nthread '{}' panicked at '{}': {}:{}\n{:?}",
+                thread,
+                msg,
+                location.file(),
+                location.line(),
+                backtrace
+            )
+        }
+        None => {
+            format!(
+                "\nthread '{}' panicked at '{}'\n{:?}",
+                thread, msg, backtrace
+            )
+        }
+    };
+
+    if thread == "main" {
+        println!("{}", backtrace);
+        process::exit(1);
+    } else {
+        send_app_instructions
+            .send(AppInstruction::Error(backtrace))
+            .unwrap();
+    }
+}
--- a/src/errors.rs
+++ b/src/errors.rs
@@ -10,15 +10,12 @@ pub fn handle_panic(info: &PanicInfo<'_>, send_app_instructions: &SyncSender<App
     let thread = thread.name().unwrap_or("unnamed");
 
     let msg = match info.payload().downcast_ref::<&'static str>() {
-        Some(s) => *s,
-        None => match info.payload().downcast_ref::<String>() {
-            Some(s) => &**s,
-            None => "Box<Any>",
-        },
+        Some(s) => Some(*s),
+        None => info.payload().downcast_ref::<String>().map(|s| &**s),
     };
 
-    let backtrace = match info.location() {
-        Some(location) => {
+    let backtrace = match (info.location(), msg) {
+        (Some(location), Some(msg)) => {
             format!(
                 "\nthread '{}' panicked at '{}': {}:{}\n{:?}",
                 thread,
@@ -28,11 +25,23 @@ pub fn handle_panic(info: &PanicInfo<'_>, send_app_instructions: &SyncSender<App
                 backtrace
             )
         }
-        None => {
+        (Some(location), None) => {
+            format!(
+                "\nthread '{}' panicked: {}:{}\n{:?}",
+                thread,
+                location.file(),
+                location.line(),
+                backtrace
+            )
+        }
+        (None, Some(msg)) => {
             format!(
                 "\nthread '{}' panicked at '{}'\n{:?}",
                 thread, msg, backtrace
             )
+        }
+        (None, None) => {
+            format!("\nthread '{}' panicked\n{:?}", thread, backtrace)
         }
     };
 
--- a/src/main.rs
+++ b/src/main.rs
@@ -12,6 +12,7 @@ mod terminal_pane;
 mod utils;
 
 use std::io::Write;
+use backtrace::Backtrace;
 use std::os::unix::net::UnixStream;
 use std::path::PathBuf;
 use std::sync::mpsc::{channel, Receiver, Sender};
@@ -37,6 +38,7 @@ enum ApiCommand {
     SplitHorizontally,
     SplitVertically,
     MoveFocus,
+    Error(String),
 }
 
 #[derive(StructOpt, Debug, Default)]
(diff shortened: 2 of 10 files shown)


kunalmohan:
Switching to [crossbeam-channel](https://docs.rs/crossbeam-channel/0.5.0/crossbeam_channel/) should help since its `Sender` and `Receiver` implement `Sync` and `Send`. Also, it is better performing and has more features, so might be a good option.

imsnif:
Hey, I have gone over this briefly and it looks solid. Before I try it out and go over it with greater detail:
I'm not very thrilled about going through the ipc_thread on this. I totally get why you did it, but it might get things mixed up a little, especially if we start doing that for other features too.

Do you think this could be solved if you clone the `Sender<AppInstruction>` in a similar way that we do with the other threads? Or is this not possible with a closure?

kunalmohan:
I completely agree that going through `ipc_thread` is not great (especially if its the `ipc_thread` panicking), but unfortunately, cloning the sender won't work. Closure requires it to implement `Sync` trait. Here's the complete error:
```
error[E0277]: `std::sync::mpsc::Sender<AppInstruction>` cannot be shared between threads safely
   --> src/main.rs:146:26
    |
146 |        std::panic::set_hook(Box::new(move |info| {
    |  ___________________________^________-
    | | __________________________|
    | ||
147 | ||         let backtrace = Backtrace::new();
148 | ||         let thread = thread::current();
149 | ||         let thread = thread.name().unwrap_or("unnamed");
...   ||
179 | ||         send_app_instructions_clone.send(AppInstruction::Error);
180 | ||     }));
    | ||_____-^ `std::sync::mpsc::Sender<AppInstruction>` cannot be shared between threads safely
    | |______|
    |        within this `[closure@src/main.rs:146:35: 180:6]`
    |
    = help: within `[closure@src/main.rs:146:35: 180:6]`, the trait `Sync` is not implemented for `std::sync::mpsc::Sender<AppInstruction>`
    = note: required because it appears within the type `[closure@src/main.rs:146:35: 180:6]`
    = note: required for the cast to the object type `dyn for<'r, 's> Fn(&'r PanicInfo<'s>) + Send + Sync`
```

I switched to crossbeam_channel on my local system and it's working. We don't need to go through `ipc_thread` in that case.

imsnif:
I see what you mean. Alright, I hope you don't mind if I play with it a little tomorrow to see if I can find a workaround? If not then we can definitely go with crossbeam.

imsnif:
Hey @kunalmohan - I played with this a little locally and it works for me if we change the app sender/receiver pair to be a sync channel. Specifically like this:

```rust
   use std::sync::mpsc::{channel, Receiver, Sender, sync_channel, SyncSender};

    let (send_app_instructions, receive_app_instructions): (
        SyncSender<AppInstruction>,
        Receiver<AppInstruction>,
    ) = sync_channel(0);

```
I don't think doing that for this channel matters, since it doesn't really matter if other threads are blocked sending app messages because of their nature (mostly telling the app to close or reporting errors). I feel this would be a more lightweight solution than introducing crossbeam as a dependency. What do you think?

kunalmohan:
Sounds good. We could switch to crossbeam later if we feel the need.

src/main.rs
```diff
@@ -136,6 +137,50 @@ pub fn start(mut os_input: Box<dyn OsApi>, opts: Opt) {
     );
     let maybe_layout = opts.layout.map(Layout::new);
 
+    let send_app_instructions_clone = send_app_instructions.clone();
```
imsnif, line 140: Can we move this into the `set_hook` block in a similar way to which we do in the other threads? Just for consistency.

Something like:
```rust
std::panic::set_hook({
    let send_app_instructions = send_app_instructions.clone();
    Box::new(move |info| {
    ...
    });
});
```

src/main.rs
```diff
@@ -136,6 +137,50 @@ pub fn start(mut os_input: Box<dyn OsApi>, opts: Opt) {
...
+                    thread, msg, backtrace
+                )
+            }
+        };
+
+        if send_app_instructions_clone
+            .send(AppInstruction::Error(backtrace.clone()))
+            .is_err()
+        {
+            get_os_input().unset_raw_mode(0);
```
imsnif, line 179: I'm not thrilled about this tbh. If I understand correctly, we can't pass os_input here because it doesn't implement `Sync`, just like we had with the app instructions sender, right?

The issue is that while right now calling this factory function doesn't change a lot, in the future we might want to add certain system startup operations there, and then calling it again here might cause issues.

src/main.rs
```diff
@@ -136,6 +137,50 @@ pub fn start(mut os_input: Box<dyn OsApi>, opts: Opt) {
     );
     let maybe_layout = opts.layout.map(Layout::new);
 
+    let send_app_instructions_clone = send_app_instructions.clone();
+
+    #[cfg(not(test))]
+    std::panic::set_hook(Box::new(move |info| {
+        let backtrace = Backtrace::new();
```
imsnif, line 144: Do you think it could be possible to move this to a different file? Something like `errors.rs`? We'd leave the panic hook here, but call a function that performs all of this?

imsnif (changes requested):
Tried this locally and it looks really great! I'm ecstatic to see normal stack traces again ;P
I left a few comments in the PR, but all in all stellar work! Let's get this merged ASAP.

src/main.rs
kunalmohan: Yes the issue is `os_input` does not implement the `Sync` trait. I added this for the case if the `main` thread panics. But on testing this now, I found that it is not working, so the backtrace is getting displayed in a weird manner in this case. How do you propose I proceed here?

src/errors.rs
```diff
@@ -0,0 +1,47 @@
...
+    let thread = thread::current();
+    let thread = thread.name().unwrap_or("unnamed");
+
+    let msg = match info.payload().downcast_ref::<&'static str>() {
+        Some(s) => *s,
+        None => match info.payload().downcast_ref::<String>() {
+            Some(s) => &**s,
+            None => "Box<Any>",
+        },
+    };
```
imsnif, line 18: I'm not sure I understand what's going on here. Can you help me out?

src/main.rs
imsnif: I think the main thread panicking is a special and rare enough case that we don't need to deal with it right now. It's really okay to just ignore it.

src/errors.rs
imsnif: (ping @kunalmohan because you said you don't always see these :) )

kunalmohan: `payload()` returns the payload attached with the panic, e.g. the string passed to `panic!()` or `.expect()`. It's return entity isn't statically typed but implements `Any` trait as mentioned [here](https://doc.rust-lang.org/std/panic/struct.PanicInfo.html#method.payload). Since it's mostly `&str` or `String` typed (as mentioned in docs), we check for them or return `"Box<Any>"` if it's not the case.

imsnif: I see... why is `Box<Any>` a static string ref though?

kunalmohan: I am not sure what you're asking. We'll have to return a single type (that implements the `Display` trait) from the different match arms, therefore the string ref (because the other arms return that). We can have any other message returned instead of `"Box<Any>"`.

imsnif: Yeah, that was the question. Thanks for the explanation. What do you think about having `msg` be an `Option`, and having an `and_then` instead of this match? Then we can handle the Some/None case in what is currently the `info.location()` match?

Something like:
```rust
    match ((info.location(), msg)) {
        (Some(location), Some(msg)) => { ... },
        (Some(location), None) => { ... },
        (None, Some(msg)) => { ... },
        (None, None) => { ... },
    }
```

kunalmohan: Sounds good.

kunalmohan:
@imsnif I've made the changes. I chose `map()` instead of `and_then()` though (looked more appropriate).

Tests seem to be failing on the CI, they pass on my local machine. Is there a way to rerun them or do I need push some changes for that?

imsnif:
I'm looking into why the tests are failing, tried re-running a few times and no go

imsnif:
Hey @kunalmohan, so apparently making the application channel synchronous caused some annoying race conditions with the tests. I solved this by adding a short sleep before our tests send the QUIT command. (In real world apps this is not a problem, since we'd want the app to exit as quickly as possible when the user quits anyway).
Right now I'm adjusting the time to see what's the shortest duration we can get away with (in order to keep the tests as fast as possible). Are we good with everything else?

src/errors.rs
```diff
@@ -0,0 +1,56 @@
...
+
+pub fn handle_panic(info: &PanicInfo<'_>, send_app_instructions: &SyncSender<AppInstruction>) {
+    let backtrace = Backtrace::new();
+    let thread = thread::current();
+    let thread = thread.name().unwrap_or("unnamed");
+
+    let msg = match info.payload().downcast_ref::<&'static str>() {
+        Some(s) => Some(*s),
+        None => info.payload().downcast_ref::<String>().map(|s| &**s),
+    };
```
imsnif, line 15: Why not make this `None` if it's `None`?

imsnif:
@kunalmohan - Alright, I'm pleased with the sleep time. I just left one minor comment about the implementation we talked about, but I'm good with merging even if you'd prefer to leave it as is (just so I understand where it's coming from). Give me a ping when ready.

src/errors.rs
kunalmohan: Should we not check for `String` at all? I'm not aware of the specific cases where it returns `&str` and `String`. I just want it to be as thorough as we can, so we don't miss any important info. If `&str` is more common and `String` is comparatively rare, I can refactor it to simply return `None` here.

imsnif: Actually, you're right. :) let's keep it this way

Issue #55, opened 2020-11-17
Feature: implement directional change focus


imsnif:
Implement commands for changing the focus of the pane up/down/right/left.
If there are multiple panes in that direction, it should move to the pane with the most overlapping space.

denismaxim0v:
I could pick this up if no one is working on it

imsnif:
That would be great :)

imsnif:
Implemented. Thanks @denismaxim0v !

PR #78, opened 2020-11-30
feat: directional move focus #55

denismaxim0v:


denismaxim0v:
that's still wip, I'll write the tests later on today

henil:
I have tried the branch locally and it's working great for me. Just leaving some nitpicks which came to my mind from the first look.

src/screen.rs
```diff
@@ -1282,6 +1287,154 @@ impl Screen {
...
+                        self.active_terminal = Some(active.pid);
+                        ()
+                    }
+                }
+            }
+            None => {
+                self.active_terminal = Some(active_terminal.unwrap().pid);
+                ()
+            }
+        }
```
henil, line 1398: .

src/screen.rs
```diff
@@ -1282,6 +1287,154 @@ impl Screen {
...
```
henil, line 1435: .

src/screen.rs
```diff
@@ -1282,6 +1287,154 @@ impl Screen {
...
```
henil, line 1361: Same here

src/screen.rs
```diff
@@ -1282,6 +1287,154 @@ impl Screen {
...
```
henil, line 1324: I think this `match` statements should be reduced to `if let`, as it improves readability and that's what `if let` statements are for. And also reduces a level of indent :)

imsnif:
Hey @denismaxim0v - on the surface this looks great. Before I go deeper into it, do you think you can add some tests for these? There are quite some edge cases here that we should probably be testing. (eg. making sure it picks the terminal with the largest overlap in each case, etc.) You can look at some of the resize tests in this repo for inspiration.

denismaxim0v:
> Hey @denismaxim0v - on the surface this looks great. Before I go deeper into it, do you think you can add some tests for these? There are quite some edge cases here that we should probably be testing. (eg. making sure it picks the terminal with the largest overlap in each case, etc.) You can look at some of the resize tests in this repo for inspiration.

Yup, working on it, gonna commit those tomorrow

src/screen.rs
denismaxim0v: Agree, this can be reduced

src/screen.rs
```diff
@@ -1282,6 +1287,130 @@ impl Screen {
...
+                .max_by_key(|(_, (_, c))| c.get_horizontal_overlap_with(&active))
+                .map(|(_, (pid, _))| pid);
+            match next_index {
+                Some(p) => {
+                    self.active_terminal = Some(*p);
+                }
+                None => {
+                    self.active_terminal = Some(active.pid);
+                }
+            }
```
henil, line 1315: Thanks for fixing the previous ones, perhaps these inner ones can too be reduced.

src/screen.rs
```diff
@@ -1282,6 +1287,130 @@ impl Screen {
...
+                .max_by_key(|(_, (_, c))| c.get_vertical_overlap_with(&active))
+                .map(|(_, (pid, _))| pid);
+            match next_index {
+                Some(p) => {
+                    self.active_terminal = Some(*p);
+                }
+                None => {
+                    self.active_terminal = Some(active.pid);
+                }
+            }
```
henil, line 1346: .

src/screen.rs
```diff
@@ -1282,6 +1287,130 @@ impl Screen {
...
```
henil, line 1377: .

src/screen.rs
```diff
@@ -1282,6 +1287,130 @@ impl Screen {
...
+                .max_by_key(|(_, (_, c))| c.get_horizontal_overlap_with(&active))
+                .map(|(_, (pid, _))| pid);
+            match next_index {
+                Some(p) => {
+                    self.active_terminal = Some(*p);
+                }
+                None => {
+                    self.active_terminal = Some(active.pid);
+                }
+            }
```
henil, line 1408: .

henil (changes requested):

src/terminal_pane/terminal_pane.rs
```diff
@@ -164,6 +164,48 @@ impl TerminalPane {
...
+                && (self.get_x() + self.get_columns() >= (other.get_x() + other.get_columns())))
+            || (other.get_x() <= self.get_x()
+                && (other.get_x() + other.get_columns() >= (self.get_x() + self.get_columns())))
+    }
+    pub fn get_vertical_overlap_with(&self, other: &TerminalPane) -> usize {
+        std::cmp::min(
+            self.get_x() + self.get_columns(),
+            other.get_x() + other.get_columns(),
+        ) - std::cmp::max(self.get_x(), other.get_x())
+    }
```
imsnif, line 208: Do you think we can implement these on the `Rect` trait instead of on terminal_pane? That way when we add other pane types (eg. plugins) they will get this for free.

src/tests/integration/move_focus_down.rs
```diff
@@ -0,0 +1,77 @@
...
+    let mut fake_input_output = get_fake_os_input(&fake_win_size);
+    fake_input_output.add_terminal_input(&[
+        &COMMAND_TOGGLE,
+        &COMMAND_TOGGLE,
+        &SPLIT_HORIZONTALLY,
+        &SPLIT_VERTICALLY,
+        &SPLIT_VERTICALLY,
+        &MOVE_FOCUS_UP,
+        &MOVE_FOCUS_DOWN,
+        &QUIT,
```
imsnif, line 64: Correct me if I'm wrong, but I think here both pane candidates are of the same size, aren't they? Maybe we can adjust the size of one of them using a resize command in this input? (this is true to the other tests as well, I think)

imsnif (changes requested):
Very cool! I'm looking forward to this feature :)

Great work. The code looks very clean and the tests are great. I left two relatively minor comments and after addressing them I think we can merge this.

src/tests/integration/move_focus_down.rs
denismaxim0v: Yup, my bad, will push it with a resize

src/terminal_pane/terminal_pane.rs
denismaxim0v: Yup, don't see why it can't be moved, will do

src/tests/integration/move_focus_down.rs
denismaxim0v: Now that I look at it I do the vertical split two times so the panes are 25%, 25%, 50% so it's still testing whether it switches to the most overlapping one

imsnif: You're right. My bad :)

imsnif (approved):
Nice solution with `Self` - I did not know you could do that :)

PR #79, opened 2020-12-01
Fix a race condition, build-times, and warnings

TheLostLambda:
This is a bit a variety PR, but it brings build times down from 53 to 23 seconds (less than half the time) and gets rid of nearly all of the compiler warnings.

Oh, also fixes a race condition with pty_bus and layouts!

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -11,18 +11,14 @@ mod screen;
 mod terminal_pane;
 mod utils;
 
-use std::io::{self, Read, Write};
+use std::io::Write;
 use std::os::unix::net::UnixStream;
 use std::path::PathBuf;
 use std::sync::mpsc::{channel, Receiver, Sender};
-use std::sync::{Arc, Mutex};
 use std::thread;
 
 use serde::{Deserialize, Serialize};
 use structopt::StructOpt;
-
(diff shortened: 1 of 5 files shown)


src/utils/logging.rs
```diff
@@ -23,14 +23,9 @@ pub fn atomic_create_dir(dir_name: &str) -> io::Result<()> {
     }
 }
 
-pub fn debug_log_to_file(message: String) -> io::Result<()> {
```
TheLostLambda, line 26: Found @imsnif 's vim macros :wink:

imsnif: Ouch ;P

src/main.rs
```diff
@@ -257,13 +253,17 @@ pub fn start(mut os_input: Box<dyn OsApi>, opts: Opt) {
     // Here be dragons! This is very much a work in progress, and isn't quite functional
     // yet. It's being left out of the tests because is slows them down massively (by
     // recompiling a WASM module for every single test). Stay tuned for more updates!
-    #[cfg(not(test))]
+    #[cfg(feature = "wasm-wip")]
```
imsnif, line 256: <3

imsnif (approved):

PR #83, opened 2020-12-04
Introduce ErrorContext for tracking errors across threads

### LINKED ISSUE #61: Implement Error handling
Right now we panic when there's an error. Since we do that inside a thread, it's very hard to tell what caused the error. This is especially true if the error we end up getting is an error from another thread that fails to send a message to the panicked thread.

We need some system that would (among other things) allow us to get proper errors across our threads. This will greatly ease development.

kunalmohan:
fix #61 
Tracks and stores function names that send messages across threads. This information will then be passed to `PanicInfo` as payload for displaying at panic.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -861,7 +861,6 @@ checksum = "0debeb9fcf88823ea64d64e4a815ab1643f33127d995978e099942ce38f25238"
 name = "mosaic"
 version = "0.1.0"
 dependencies = [
- "arrayvec",
  "async-std",
  "backtrace",
  "bincode",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -32,9 +32,9 @@ checksum = "4d25d88fd6b8041580a654f9d0c581a047baee2b3efee13275f2fc392fc75034"
 
 [[package]]
 name = "arrayvec"
-version = "0.5.2"
+version = "0.5.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "23b62fc65de8e4e7f52534fb52b0f3ed04746ae267519eef2a83941e8085068b"
+checksum = "cff77d8686867eceff3105329d4698d96c2391c176d5d03adc90c7389162b5b8"
 
 [[package]]
 name = "async-channel"
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -32,9 +32,9 @@ checksum = "4d25d88fd6b8041580a654f9d0c581a047baee2b3efee13275f2fc392fc75034"
 
 [[package]]
 name = "arrayvec"
-version = "0.5.1"
+version = "0.5.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "cff77d8686867eceff3105329d4698d96c2391c176d5d03adc90c7389162b5b8"
+checksum = "23b62fc65de8e4e7f52534fb52b0f3ed04746ae267519eef2a83941e8085068b"
 
 [[package]]
 name = "async-channel"
@@ -861,6 +861,7 @@ checksum = "0debeb9fcf88823ea64d64e4a815ab1643f33127d995978e099942ce38f25238"
 name = "mosaic"
 version = "0.1.0"
 dependencies = [
+ "arrayvec",
  "async-std",
  "backtrace",
  "bincode",
--- a/src/errors.rs
+++ b/src/errors.rs
@@ -3,6 +3,7 @@ use crate::screen::ScreenInstruction;
 use crate::{AppInstruction, OPENCALLS};
 use arrayvec::ArrayVec;
 use backtrace::Backtrace;
+use std::fmt::{Debug, Error, Formatter};
 use std::panic::PanicInfo;
 use std::sync::mpsc::SyncSender;
 use std::{process, thread};
@@ -26,29 +27,29 @@ pub fn handle_panic(
 
     let backtrace = match (info.location(), msg) {
         (Some(location), Some(msg)) => format!(
-            "thread '{}' panicked at '{}': {}:{}\n{:#?}\n{:?}",
+            "{:#?}\n\u{1b}[0;0mError: \u{1b}[0;31mthread '{}' panicked at '{}': {}:{}\n\u{1b}[0;0m{:?}",
+            err_ctx,
             thread,
             msg,
             location.file(),
             location.line(),
-            err_ctx,
             backtrace
         ),
         (Some(location), None) => format!(
-            "thread '{}' panicked: {}:{}\n{:#?}\n{:?}",
+            "{:#?}\n\u{1b}[0;0mError: \u{1b}[0;31mthread '{}' panicked: {}:{}\n\u{1b}[0;0m{:?}",
+            err_ctx,
             thread,
             location.file(),
             location.line(),
-            err_ctx,
             backtrace
         ),
         (None, Some(msg)) => format!(
-            "thread '{}' panicked at '{}'\n{:#?}\n{:?}",
-            thread, msg, err_ctx, backtrace
+            "{:#?}\n\u{1b}[0;0mError: \u{1b}[0;31mthread '{}' panicked at '{}'\n\u{1b}[0;0m{:?}",
+            err_ctx, thread, msg, backtrace
         ),
         (None, None) => format!(
-            "thread '{}' panicked\n{:#?}\n{:?}",
-            thread, err_ctx, backtrace
+            "{:#?}\n\u{1b}[0;0mError: \u{1b}[0;31mthread '{}' panicked\n\u{1b}[0;0m{:?}",
+            err_ctx, thread, backtrace
         ),
     };
 
@@ -62,7 +63,7 @@ pub fn handle_panic(
     }
 }
 
-#[derive(Clone, Debug)]
+#[derive(Clone)]
 pub struct ErrorContext {
     calls: ArrayVec<[ContextType; MAX_THREAD_CALL_STACK]>,
 }
@@ -80,7 +81,17 @@ impl ErrorContext {
     }
 }
 
-#[derive(Debug, Copy, Clone)]
+impl Debug for ErrorContext {
+    fn fmt(&self, f: &mut Formatter) -> Result<(), Error> {
+        writeln!(f, "Originating Thread(s):")?;
+        for (index, ctx) in self.calls.iter().enumerate() {
+            writeln!(f, "\u{1b}[0;0m{}. {:?}", index + 1, ctx)?;
+        }
+        Ok(())
+    }
+}
+
+#[derive(Copy, Clone)]
 pub enum ContextType {
     Screen(ScreenContext),
     Pty(PtyContext),
@@ -88,6 +99,25 @@ pub enum ContextType {
     IPCServer,
     StdinHandler,
     AsyncTask,
+}
+
+impl Debug for ContextType {
+    fn fmt(&self, f: &mut Formatter) -> Result<(), Error> {
+        let purple = "\u{1b}[1;35m";
+        let green = "\u{1b}[0;32m";
+        match *self {
+            ContextType::Screen(c) => write!(f, "{}screen_thread: {}{:?}", purple, green, c),
+            ContextType::Pty(c) => write!(f, "{}pty_thread: {}{:?}", purple, green, c),
+            ContextType::App(c) => write!(f, "{}main_thread: {}{:?}", purple, green, c),
+            ContextType::IPCServer => write!(f, "{}ipc_server: {}AcceptInput", purple, green),
+            ContextType::StdinHandler => {
+                write!(f, "{}stdin_handler_thread: {}AcceptInput", purple, green)
+            }
+            ContextType::AsyncTask => {
+                write!(f, "{}stream_terminal_bytes: {}AsyncTask", purple, green)
+            }
+        }
+    }
 }
 
 #[derive(Debug, Clone, Copy)]
--- a/src/errors.rs
+++ b/src/errors.rs
@@ -1,7 +1,6 @@
 use crate::pty_bus::PtyInstruction;
 use crate::screen::ScreenInstruction;
 use crate::{AppInstruction, OPENCALLS};
-use arrayvec::ArrayVec;
 use backtrace::Backtrace;
 use std::fmt::{Debug, Error, Formatter};
 use std::panic::PanicInfo;
@@ -23,7 +22,7 @@ pub fn handle_panic(
         None => info.payload().downcast_ref::<String>().map(|s| &**s),
     };
 
-    let err_ctx: ErrorContext = OPENCALLS.with(|ctx| ctx.borrow().clone());
+    let err_ctx = OPENCALLS.with(|ctx| *ctx.borrow());
 
     let backtrace = match (info.location(), msg) {
         (Some(location), Some(msg)) => format!(
@@ -63,21 +62,26 @@ pub fn handle_panic(
     }
 }
 
-#[derive(Clone)]
+#[derive(Clone, Copy)]
 pub struct ErrorContext {
-    calls: ArrayVec<[ContextType; MAX_THREAD_CALL_STACK]>,
+    calls: [ContextType; MAX_THREAD_CALL_STACK],
 }
 
 impl ErrorContext {
     pub fn new() -> Self {
         Self {
-            calls: ArrayVec::new(),
+            calls: [ContextType::Empty; MAX_THREAD_CALL_STACK],
         }
     }
 
     pub fn add_call(&mut self, call: ContextType) {
-        self.calls.push(call);
-        OPENCALLS.with(|ctx| *ctx.borrow_mut() = self.clone());
+        for ctx in self.calls.iter_mut() {
+            if *ctx == ContextType::Empty {
+                *ctx = call;
+                break;
+            }
+        }
+        OPENCALLS.with(|ctx| *ctx.borrow_mut() = *self);
     }
 }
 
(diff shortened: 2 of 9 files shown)


kunalmohan:
This is still a WIP. We now need to replace `unwrap()` with `expect()` to pass the info to `PanicInfo`.

Some part of the new code here has introduced more race conditions causing tests to fail intermittently- they pass on increasing time b/w snapshots.

imsnif:
Hey @kunalmohan - good start on this.

I think, as you point out yourself and we start seeing in the implementation, that any solution forcing us to pass the error context very deep into the stack of functions in the thread itself would be very fragile and hard to maintain. This would mean that any developer who wants to add things would need to be aware of the inner workings of this error mechanism. And worst yet - remember to do things differently. :) This is of course true to being unable to use `unwrap` and the like.

When I think about it, the only thing we really need in addition to the backtrace of the final thread is knowing which threads the error passed through and which instructions (be it `ScreenInstruction`, `PtyInstruction`, etc.) it had in those threads. Do you think we might be able to log some sort of `OpenCalls` struct globally and match against it when we display the error? I'm not sure what the exact implementation would be for it, but I have a feeling we would be able to do more-or-less exact matching through information we have in the error backtrace. What do you think? Does this make sense? Did I manage to explain my general idea well enough?

kunalmohan:
Yep, basically logging the `send()` calls is the main idea. As I see it we have two options here-
1. Simply log the `send()` calls with the message they are sending and the function name where the `send()` is called. Here we won't have to edit any function to accept any new argument. But the user will have to manually identify and filter out each call in the sequence that caused the error. Not very good option in my opinion.
2. Log the `send()` call (and other info) along with a unique sequence number assigned at the time of stdin (similar to how we create a new `ErrorContext` object currently in this PR. The sequence number will have to be passed along with the message between the threads. User will have to figure out the last call made in that sequence, which should be possible with the available backtrace. The user can then simply filter out the calls made in that particular sequence using the sequence number of that call. This would require us to edit just the functions that send any message to another thread.

I'm not sure if any of this matches to your idea of globally logging an `OpenCalls` struct. If not, could you elaborate your idea a bit more?

imsnif:
To be honest, I'm more leaning toward option 1. I think most (all?) paths are rather unique and there will mostly be just one "open call" the user will have to choose from. I'd even say to just display the first one and assume there are no others. Or am I missing something and this doesn't make sense?

kunalmohan:
This makes sense. We don't have a large number of threads that the user needs to navigate through so it should not be very difficult. And I think you're correct about the paths being unique so I guess we should go ahead with this.

Just to be very clear (before I start work on this), we'll only log the first open call, i.e. only the `send()` calls in the `stdin_handler` and the `ipc_server` threads. These will be logged to a new file (or the existing `MOSAIC_TMP_LOG_FILE`?). Should this be behind the debug flag or not?

imsnif:
Hmm... maybe I didn't understand you correctly. I was thinking to log the previous calls in memory, and then dump them with the backtrace to screen somehow.

imsnif:
So that in the end we'll have something like
`screen_thread(CloseActivePane)->pty_thread(ClosePane)-><backtrace_of_error_in_pty>`

kunalmohan:
> Hmm... maybe I didn't understand you correctly. I was thinking to log the previous calls in memory, and then dump them with the backtrace to screen somehow.

Right, sorry, it slipped from my mind. I had just one concern with it. We'll use `Arc<Mutex<OpenCalls>>` to share it between threads. I was concerned with cases where Mutex lock gets poisoned in panic. But I guess that's rarely going to happen (and we always can switch to `parking_lot::Mutex`, if it does). So I'll go ahead with this.

> I'd even say to just display the first one and assume there are no others.

Also, could you briefly explain what you meant by this?

imsnif:
@kunalmohan - I think I might have been too general here in my thoughts and caused some confusion. Let's take a step back.

If we:
1. Pass an `ErrorContext` in every call we make
2. Attach the name of the instruction that we send so that this context contains a `Vec` of those instructions which we can attach to a backtrace later if there's an error
3. Keep that context globally in the thread
4. use that context in the panic hook (maybe sending it using another panic hook inside each thread somehow?)
5. in the main panic hook, get:
  * the message if there is one
  * the `ErrorContext`
  * the backtrace

Could we forgo having something global to the whole app shared with an `App<Mutex>` (because that will hurt performance)?

imsnif:
I don't know if it will work, but maybe we can somehow use [catch_unwind](https://doc.rust-lang.org/stable/std/panic/fn.catch_unwind.html) inside each thread to forward the `ErrorContext` to the main thread?

kunalmohan:
@imsnif I guess I found a solution. We'll have a [`LocalKey`](https://doc.rust-lang.org/std/thread/struct.LocalKey.html) for each thread.
```rust
thread_local!(static OpenCalls: RefCell<ErrorContext> = RefCell::new(ErrorContext::new());
```
When a thread sends a message to another thread, we'll clone the current `ErrorContext` of the thread and pass it to the other thread.
When a thread receives a new message, it will replace/update it's current `ErrorContext` in `LocalKey` with the one it has received with the new message. Hence at each point in time, the `LocalKey` of each thread would contain the context of the instruction that it is currently processing.
We can directly access this key in the panic hook and when a thread panics, the hook will automatically select the key of the thread which panicked.
I have tested it [here](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=cefe47f6696a94670feed0da1f73e339).

imsnif:
Interesting. I haven't used `LocalKey` before, so maybe you can help me make sure I understand: if one thread panics, the panic hook will access the `LocalKey` of the panicking thread, rather than the `LocalKey` of a different thread that did not panic? Or am I misunderstanding this?

kunalmohan:
When accessing a `LocalKey`, the `LocalKey` of the currently executing thread is used.

Since we will be accessing the `LocalKey` from the panic hook, it will access the `LocalKey` of the panicking thread (just like `thread::current()` in panic hook returns the information of panicking thread).

imsnif:
@kunalmohan - sounds like a great solution. Looking forward to see the implementation :)

kunalmohan:
An example of how the context is displayed-
```
thread 'screen' panicked at 'removal index (is 0) should be < len (is 0)': library/alloc/src/vec.rs:1066
ErrorContext {
    calls: [
        "stdin_handler_thread(AcceptInput)",
        "pty_thread(SpawnTerminalVertically)",
        "stream_terminal_bytes(AsyncTask)",
        "screen_thread(HandlePtyEvent)",
    ],
}
   0: mosaic::errors::handle_panic
             at src/errors.rs:11:21
   1: mosaic::start::{{closure}}
             at src/main.rs:146:13
   2: std::panicking::rust_panic_with_hook
             at /rustc/7eac88abb2e57e752f3302f02be5f3ce3d7adfb4/library/std/src/panicking.rs:581:17
   3: std::panicking::begin_panic_handler::{{closure}}
             at /rustc/7eac88abb2e57e752f3302f02be5f3ce3d7adfb4/library/std/src/panicking.rs:484:9
   4: std::sys_common::backtrace::__rust_end_short_backtrace
             at /rustc/7eac88abb2e57e752f3302f02be5f3ce3d7adfb4/library/std/src/sys_common/backtrace.rs:153:18

....
```

kunalmohan:
@imsnif this is ready for the first round of review.

All tests are passing on my local machine. I guess again some race conditions are causing failures in CI.

src/main.rs
```diff
@@ -107,16 +111,16 @@ pub fn start(mut os_input: Box<dyn OsApi>, opts: Opt) {
     let full_screen_ws = os_input.get_terminal_size_using_fd(0);
     os_input.into_raw_mode(0);
     let (send_screen_instructions, receive_screen_instructions): (
-        Sender<ScreenInstruction>,
-        Receiver<ScreenInstruction>,
+        Sender<(ErrorContext, ScreenInstruction)>,
```
imsnif, line 114: Could we have these be the other way around? (eg. `Sender<(ScreenInstruction, ErrorContext)>` I feel it will be clearer)

src/errors.rs
```diff
@@ -14,35 +17,59 @@ pub fn handle_panic(info: &PanicInfo<'_>, send_app_instructions: &SyncSender<App
...
         send_app_instructions
-            .send(AppInstruction::Error(backtrace))
+            .send((err_ctx, AppInstruction::Error(backtrace)))
             .unwrap();
     }
 }
+
+#[derive(Clone, Debug)]
+pub struct ErrorContext {
+    calls: Vec<String>,
```
imsnif, line 63: A few things about this:
1. Can we change `String` here to work with the instruction types we already have? eg. something like `enum { ScreenInstruction, PtyInstruction, AppInstruction }` - I think that will both help performance and prevent overzealous developers from adding weird stuff to this in the future :)
2. Speaking about performance, while I definitely understand why it makes sense for this to be a `Vec` - I think in our case it might be a bit of an issue. Ideally, I would like this type to be `Copy`, because we would be copying it around A LOT (every event we get from the pty, for example, which is every character print, every style change, etc. if we're printing 10000 lines to the terminal, we will really feel this). And of course, we can't do this with a vec... So I was thinking maybe make this a fixed size array. We'll have it as a separate const type (maybe call it something like `MAX_THREAD_CALL_STACK`), we'll set its length to the maximum we have now (I guess it's 3? 4?) and in the future we can make it larger if we need to. What do you think?

imsnif (changes requested):
@kunalmohan - I think your solution with `thread_local` and `OPENCALLS` is very elegant. I could not have imagined anything simpler. Very cool :)

I left some comments about the implementation. We might have to do some back-and-forth about this, because this is a relatively big and very "deep" feature. I hope you can be a little patient with me. :)

Other than that, about displaying:
What do you think of adding some nice colors? Instead of
```
ErrorContext {
    calls: [
        "stdin_handler_thread(AcceptInput)",
        "pty_thread(SpawnTerminalVertically)",
        "stream_terminal_bytes(AsyncTask)",
        "screen_thread(HandlePtyEvent)",
    ],
}
```
something like:
```
Originating threads:
  1. stdin_handler_thread(purple): AcceptInput(green)
  2. pty_thread(purple): SpawnTerminalVertically(green)
  3. stream_terminal_bytes(purple): AsyncTask(green)
  4. screen_thread(purple): HandlePtyEvent(green)

Error:(red)
  <stacktrace>(reset color)
```

src/errors.rs
kunalmohan: I was in fact thinking about the second point myself and was looking at the [arrayvec](https://docs.rs/arrayvec/0.5.2/arrayvec/) crate for the same.

If I understand this correctly, you're suggesting-
```rust
pub enum ContextType {
    ScreenInst(ScreenInstruction),
    PtyInst(PtyInstruction),
    AppInst(AppInstruction),
    IPCServer,
    StdinHandler,
    AsyncTask,
}
```
and have a fixed size array inside `ErrorContext` so that the elements are `Copy` (or basically stored on the stack itself).
I see 2 things ahead of us-
1. We'll have to custom implement Debug for the `***Instruction` enums to prevent enum variants with `Vec` types (such as `ScreenInstruction::ApplyLayout((Layout, Vec<RawFd>))` from filling up the terminal.
2. For `ErrorContext` to implement `Copy` trait, we'll need to have `ContextType` and all the `***Instruction` enums to implement `Copy` trait. These enums have variants with `Vec` type data which won't allow us to have them stored on stack. I maybe misunderstanding your idea of the enum variants of `ContextType`, so please correct me if I'm wrong.

imsnif: You understood me perfectly, and are actually one step ahead of me. :)
I forgot about the `Vec`s in the events, and actually now noticed an old TODO of mine to fix them: https://github.com/mosaic-org/mosaic/blob/main/src/pty_bus.rs#L65 :)

I have a suggestion until we fix those instructions (because they should definitely be `Copy` too):
Let's make `ContextType` like this:
```rust
pub enum ContextType {
    Screen(ScreenContext),
    Pty(PtyContext),
    App(AppContext),
    IPCServer,
    StdinHandler,
    AsyncTask,
}
```
And then we can have for example:
```rust
pub enum ScreenContext {
    Pty,
    Render,
    NewPane,
    // ...
}

impl From<ScreenInstruction> for ScreenContext {
    fn from(screen_instruction: ScreenInstruction) -> Self {
        match ScreenInstruction {
            ScreenInstruction::Pty(_) => ScreenContext::Pty,
            // ...
        }
    }
}
```

What do you think?

kunalmohan: Cool. Sounds good to me. What do you think about the arrayvec crate?

imsnif: I trust your judgment.

kunalmohan: So, I found out that `ArrayVec` doesn't implement `Copy` (for reasons mentioned [here](https://github.com/bluss/arrayvec/issues/32)) so we need to explicitly call `.clone()` but it definitely facilitates stack allocation ensuring better performance.

imsnif: Wouldn't that be defeating the purpose though? Why not use an array with a fixed length and then we can also implement Copy?

kunalmohan: With arrays, we'll have to initialize it. So should we have another enum variant `ContextType::Empty` for that? So `add_call()` will replace the first `Empty` entry it encounters with the current thread instruction.

imsnif: Either that or make it an `Option<ContextType>`, whatever feels better for you.

kunalmohan: `Option` will add 8 bytes to the current size of `ContextType`. So, I'll go with `Empty`.

src/errors.rs
```diff
@@ -14,35 +23,194 @@ pub fn handle_panic(info: &PanicInfo<'_>, send_app_instructions: &SyncSender<App
...
+        }
+    }
+
+    pub fn add_call(&mut self, call: ContextType) {
+        self.calls.push(call);
+        OPENCALLS.with(|ctx| *ctx.borrow_mut() = self.clone());
+    }
+}
+
+impl Debug for ErrorContext {
```
imsnif, line 88: Just curious, why `Debug` and not `Display`?

src/errors.rs
```diff
@@ -14,35 +23,194 @@ pub fn handle_panic(info: &PanicInfo<'_>, send_app_instructions: &SyncSender<App
...
+            }
+            ContextType::AsyncTask => {
+                write!(f, "{}stream_terminal_bytes: {}AsyncTask", purple, green)
+            }
+        }
+    }
+}
+
+#[derive(Debug, Clone, Copy)]
+pub enum ScreenContext {
```
imsnif, line 133: Nice. I initially thought to put these in screen.rs, but I like this better.

src/errors.rs
kunalmohan: Done!

src/input.rs
```diff
@@ -63,10 +67,10 @@ impl InputHandler {
                 }
                 _ => {
                     self.send_screen_instructions
-                        .send(ScreenInstruction::ClearScroll)
+                        .send((ScreenInstruction::ClearScroll, err_ctx))
```
imsnif, line 70: So, this is true to other places as well, but what do you think about making a wrapper struct around our various `Sender`s so that we won't have to keep specifying the `err_ctx` everywhere?

Something like:
```rust
struct SenderWithContext<T> {
    err_ctx: ErrorContext,
    sender: Sender<T>
}
impl<T> SenderWithContext <T>{
    pub fn send(event: T) {
        self.sender.send(event, self.err_ctx);
    }
    pub fn update_error_ctx (new_error_ctx: ErrorContext) {
        self.error_ctx = new_error_ctx;
    }
}
```

imsnif (changes requested):
Looking good! Added some more comments

src/errors.rs
kunalmohan: No particular reason. I just never used `Display` before :sweat_smile: . I can switch to it if you want.

src/input.rs
kunalmohan: This is a good idea. What about the `SyncSender`?

imsnif: Hmm, good question. Either have a `SyncSenderWithContext` version or maybe have `self.sender` be an `enum` somehow? Whatever you feel will look/work better. I think this is one of those cases where developer ergonomics is important.

src/errors.rs
imsnif: Both work, but IMO in this case `Display` is more correct.

src/input.rs
kunalmohan: I have updated this. The only trivial drawback here was that the sender in panic hook cannot be updated because that requires mutable borrow of the sender, but the closure cannot capture the sender mutably. But I don't think it matters much because that info will be displayed only if another panic occurred following an invocation of panic hook, which is rare. The info of the primary panic is still displayed correctly.

Cargo.lock
```diff
@@ -32,9 +32,9 @@ checksum = "4d25d88fd6b8041580a654f9d0c581a047baee2b3efee13275f2fc392fc75034"
 
 [[package]]
 name = "arrayvec"
-version = "0.5.1"
+version = "0.5.2"
```
imsnif, line 35: I think we didn't add this after all, right? Better to stick with what's in the lockfile already in that case.

kunalmohan: Hmn..I removed arrayvec from Cargo.toml but the version inCargo.lock didn't revert back. How can that be done?

imsnif: Personally I'd copy over the file from main and then just do a quick `cargo run`+`cargo test` to make sure all is good.

imsnif (approved):
Everything looks great. Thanks for being so patient (and quick!) with all my requests and suggestions. This is top notch work! I would be very happy to keep seeing you contributing and taking part in our project. :)

PR #86, opened 2020-12-08
It 'aint much, but it's honest work (else-if ladder -> match)

TheLostLambda:


src/terminal_pane/terminal_character.rs
```diff
@@ -300,105 +300,77 @@ impl CharacterStyles {
     }
     pub fn add_style_from_ansi_params(&mut self, ansi_params: &[i64]) {
         let mut params_used = 1; // if there's a parameter, it is always used
-        if ansi_params.is_empty() || ansi_params[0] == 0 {
```
imsnif, line 303: So `ansi_params.is_empty()` happens often. Text styles can be reset with either `[0m` (and then it will be `0`) or `[m` in which case it will be empty. Arcane I know, but I don't make the rules ;P

I thought about just putting an `if ansi_params.is_empty()` and then moving the match into the else, but that might be a little icky. Do you have any rust-magic ideas of doing this nicely? :)

TheLostLambda: @imsnif There seem to be a lot of ways to skin this cat... How do you feel about this one? (In the new commit)

imsnif (approved):
Looks cool, ship it!

PR #90, opened 2020-12-09
fix(compatibility): support rgb color notation

imsnif:
So, apparently terminals with more than 256 colors are a thing ;P

This addresses https://github.com/mosaic-org/mosaic/issues/87 (I hope)


henil:
This fixes my issue :smile:

PR #96, opened 2020-12-11
fix(style): make match arms consistent

henil:
Brought to you by vim macros :smile:


imsnif:
> Brought to you by vim macros smile

My favorite vim feature :) Thanks!

PR #97, opened 2020-12-13
Fix reset causes an unhandled error

5c077m4n:


roee-shapira-tikal:
https://github.com/mosaic-org/mosaic/issues/43

roee-shapira-tikal:
@imsnif ready :)

src/terminal_pane/terminal_pane.rs
```diff
@@ -456,12 +456,7 @@ impl vte::Perform for TerminalPane {
             };
             self.scroll.move_cursor_back(move_back_count);
         } else if c == 'l' {
-            let first_intermediate_is_questionmark = match _intermediates.get(0) {
-                Some(b'?') => true,
-                None => false,
-                _ => false,
-            };
-            if first_intermediate_is_questionmark {
+            if _intermediates.get(0) == Some(&b'?') {
```
imsnif, line 459: I think if we're using this we can remove the leading `_`, right?

src/terminal_pane/terminal_pane.rs
```diff
@@ -471,15 +466,12 @@ impl vte::Perform for TerminalPane {
...
+                self.scroll.clear_all();
             }
         } else if c == 'h' {
-            let first_intermediate_is_questionmark = match _intermediates.get(0) {
-                Some(b'?') => true,
-                None => false,
-                _ => false,
-            };
-            if first_intermediate_is_questionmark {
+            if _intermediates.get(0) == Some(&b'?') {
```
imsnif, line 474: Also - thanks for this. This is of course much better. I should not be coding in my sleep ;P

src/terminal_pane/terminal_pane.rs
```diff
@@ -584,6 +576,19 @@ impl vte::Perform for TerminalPane {
...
+             */
+            if let Some(move_cursor_back) = params.get(1) {
+                let move_cursor_back = *move_cursor_back;
+                if move_cursor_back > 0 {
+                    self.move_cursor_backwards(move_cursor_back as usize);
+                }
+            }
+        } else if c == 'p' {
+            self.clear_scroll();
+            self.reset_size_and_position_override();
```
imsnif, line 591: I'm not sure I understand... we use `reset_size_and_position_override` when the pane was full_screen (had a size and position override) and we want to bring it back. Why do we need it in this case?

imsnif (changes requested):
Looking good. Left a few small comments.
Also: how about writing a test for this?
You can check out the compatibility tests if you like. Essentially what they do is run a fixture (which we get from running mosaic with the `--debug` flag) and comparing it against a snapshot to make sure it is as expected. Want to try and write one? I can provide a more detailed explanation if you like.

src/terminal_pane/terminal_pane.rs
roee-shapira-tikal: Sorry, I saw `reset` and thought: reset!

roee-shapira-tikal:
@imsnif I'll try on my own to add a test and when I'll need help I'll tell you

imsnif:
This has been already been fixed.

PR #116, opened 2020-12-19
Refactor terminal to panes

### LINKED ISSUE #76: refactor Terminals to be panes
In the screen module, there is a terminal BTreeMap attribute which is a collection of all terminal panes, it should be refactored to panes where the id is either RawFd or BuiltInPaneId (connected to #71 ), in the future, this enum will have another option called plugin id.

qballer:
resolves #76


imsnif:
Awesome :)

PR #125, opened 2020-12-29
Fully Working WebAssembly Loading!

TheLostLambda:
We're on a semi-stable version of wasmer now! This is still very WIP, but should be mergable without any harm (all behind a feature flag).


imsnif (approved):
Very cool work on all the upstream fixes!!

PR #127, opened 2020-12-31
Introduce Persistent sessions

kunalmohan:
This is the start of the implementation of persistent sessions.
I'm opening this draft PR for discussion of the changes.

### DIFF
--- a/src/daemon.rs
+++ b/src/daemon.rs
@@ -1,10 +1,28 @@
+use crate::errors::{ContextType, ErrorContext, PtyContext};
 use crate::os_input_output::{daemonize, OsApi};
+use crate::pty_bus::{PtyBus, PtyInstruction};
+use crate::screen::ScreenInstruction;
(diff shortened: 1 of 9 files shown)


src/daemon.rs
```diff
@@ -0,0 +1,67 @@
...
+                            .unwrap();
+                    }*/
+                    ApiCommand::Error(_backtrace) => {
+                        // Send backtrace to all clients and stop
+                    }
+                    ApiCommand::Exit => {
+                        // Server should not exit when a client closes. It should ideally just remove the client.
+                        // Server should close only when there are no active clients and no sessions to store.
+                        // But we should break until we add the check for above.
+                        break;
```
kunalmohan, line 57: We successfully spawn a daemon process now, but it would exit if the client exits because of this break. This is intentional because the flow of this process is not yet defined entirely. You can remove it and check, but you'll have to kill the process manually once you're done.

PR #128, opened 2021-01-01
feat(completions): Add shell completions files

henil:


imsnif:
Hey @henil - sorry for taking a short while to get to this. This looks good - only comment I have is that my personal preference is to:
1. Name files by the name of the struct they expose for easy navigation
2. Give Opt a more descriptive name now that it's out of main.rs - eg. something like CliArgs?

Otherwise, as far as I'm concerned feel free to merge this when you feel it's ready :)

henil:
> Name files by the name of the struct they expose for easy navigation

I have changed it to `cli.rs` rather than `CliArgs.rs` because later on, we could have other functions related to argument parsing. And also it keeps filenames consistent.

PR #132, opened 2021-01-05
Hotkeys

khs26:
Still a WIP, but this should be a no-harm commit (with some dead code!) defining input keys that can be used for hotkey definitions.

As discussed in our weekly catch up, I'm going to raise an issue to track the key mapping task.


khs26:
@imsnif This is the one that I'm struggling with :)

src/input/keybinds.rs
```diff
@@ -0,0 +1,98 @@
...
+            // Scroll
+            defaults.insert(Key::PageUp, Action::ScrollUp);
+            defaults.insert(Key::PageDown, Action::ScrollDown);
+            // Tab controls
+            defaults.insert(Key::Char('1'), Action::NewTab);
+            defaults.insert(Key::Char('2'), Action::GoToNextTab);
+            defaults.insert(Key::Char('3'), Action::GoToPreviousTab);
+            defaults.insert(Key::Char('4'), Action::CloseTab);
+            // New pane
+            defaults.insert(Key::Char('z'), Action::NewPane(Direction::Down));
```
imsnif, line 73: @khs26 - I think the bug in the tests is here. ctrl-z (actually ctrl-m, I think the comment was wrong but I don't remember why it was changed, so I guess either one will do :) ) should not have a direction. This should send `PtyInstruction::NewPane` which should find the largest free space and spawn a pane there (horizontally or vertically, it makes the decision on its own)

PR #147, opened 2021-01-19
feat(plugin): Search in root folder for plugins and layout files

henil:


henil:
(merging this as brooks is busy atm, and this is trivial enough change and is required to package .deb)

PR #154, opened 2021-01-25
fix(compatibility): Use alternate buffer like tmux

henil:


henil:
@imsnif should I merge this?

PR #160, opened 2021-02-09
feat(ui): new keybindings and status bar

imsnif:
This implements the discussion here: https://github.com/mosaic-org/mosaic/discussions/146 and adds a status bar with a new design to match. Also, various plugin APIs and such that were needed for this :)


TheLostLambda:
This looks good to me! Happy to have it merged!! A couple of quick questions:

1. Is there a reason we are using a vector of tuples as opposed to a `HashMap` for the keybinds?
2. What is the invisible boarder thing all about? Just curious :smile:

imsnif:
Thanks for the review! Sorry for being so quick on the merge button :)

>     1. Is there a reason we are using a vector of tuples as opposed to a `HashMap` for the keybinds?
A HashMap doesn't have a predictable entry order and I didn't want to start mucking about with moving a BTreeMap around :/

>     2. What is the invisible boarder thing all about? Just curious smile
It instructs Mosaic not to draw a border around the plugin. I thought the status bar looked nicer that way, but we can of course change it if people don't like it

Issue #178, opened 2021-02-10
[Feature] Add a setting that would set Command mode to be persistent by default


categorille:
### Current behaviour
Right now, pressing Ctrl-g always rotates between Normal mode, Command mode, and Persistent Command mode. There is no way to change this behaviour.

### What should change?
A command-line setting (`--always-locked`?) should be added which makes it so Ctrl-g rotates between Normal mode and Persistent command mode. This could eventually be changed to a config-file setting (when we'll have support for that) or even the default.

### Other considerations and tips
Currently, there exists a `InputMode` enum variant for each input mode, regardless if they are "main" modes (i.e. Normal and Command modes) or submodes of Command mode (i.e. Pane, Resize, Tab, ...). It may be interesting to refactor the code so that it reflects the actual structure of input modes, with two main `InputMode`s and another enum for Command mode submodes. This would both make the code clearer and make it easier to work with command mode persistency.

### Relevant source files
- the [input](https://github.com/zellij-org/zellij/tree/main/src/common/input) module

categorille:
I guess I am assigning myself to both this and #161, given that they are highly related. May even be worth it to work on this before #161 as the refactoring that I am suggesting here will make it easier for that issue to be fixed as well.

imsnif:
The refactor idea sounds good. I have a feeling that starting with it and then making the functional changes here and in #161 would be the longer path, but I definitely might be wrong on this one. In any case, do what you think is right and best. :)
Github unfortunately would not let me assign you to these issues, so if you could write a short comment on #161 as well that you're on it so no one else would do double work on it, that would be great.
Looking forward to the pull requests.

categorille:
Is this still needed, or do we prefer going with the current keybinds (command mode always persistent) without an option?

imsnif:
If it existed, I know I would use it :)

categorille:
Alright, so it's still relevant :) I won't take care of it right now so if anyone else does, plase give it a try (and you can contact me w.r.t. the input module, which is going to see some changes)

imsnif:
So, re-reading this I think I made a mistake and thought this was a different issue. This is actually how things are right now (we only have persistent mode for all modes). So closing this issue (please correct me if I misunderstood @categorille and I'll re-open).

categorille:
[...]

Issue #169, opened 2021-02-10
Feature: custom keybindings


imsnif:
With the default keybindings of Zellij we have a limitation: we can't define a lot of keybindings that are not behind some sort of super/meta key because we would quickly collide with various operating system and terminal keybindings.

It would be nice to let users define these for themselves though (that way they can control what won't collide with their operating system and other applications).

This issue would involve thinking about a configuration format (probably `yaml` since we're using it in other places) and a "language" to allow users to define custom keybindings.

Once the above has been decided and worked out, we should be able to load this file from a default location on startup, set up Zellij to respond to those keybindings and ideally also display them in the status bar.

a-kenji:
I will try to get a rough draft of a custom configuration for keybindings.

imsnif:
That would be really awesome, @a-kenji ! I would very warmly recommend first making a plan and getting feedback on it before writing any code. Makes sense?

a-kenji:
Yes, that is a good idea!

a-kenji:
Configuration Draft:

A first Draft of a possible Configuration file.

- get an ordered list of configuration locations

- use one of the files

- Use yaml .

- Keybindings:

Ultimately have three Options that can be triggered by a keybinding:

Action: Remapping Integrated Functionality (New Pane, etc.)
Commands: Map a command to a keybinding
Macro: Map a sequence of commands, or Actions to a Keybinding

Be able to either map a single key, or a List of Keys.
Use a list of actions associated with each mode:

```
keybindings:
    normal:
        {action: "SwitchToMode(Pane)", keys: "^G"}
    pane:
        {action: "SwitchFocus(Right)", keys: "["L", "Right"]"}
```

In order to possibly not clutter the keybindings too much optionally allow for creation of a separate definition of commands as a list of named commands and macros as a named list of lists of either commands/actions.

```
commands:
{name: "show_all", program: "ls", args: ["-a"]}

keybindings:
    command:
         { command: "show_all", keys: "^L" }
```
is the same as:
```
keybindings:
    command:
        { command: {program: "ls", args: ["-a"]}, keys: "^L"}
```

The same goes for Macros:
```
macros:
 new_show_all:
    {action: "NewPane(Right)"}
    {command "show_all"}

keybindings:
    command:
        {macro: "new_show_all", keys: "^L"}
```
or
```
keybindings:
    command:
        {macro: [
        {action: "NewPane(Right)"}
        {command: "show_all"}
        ], keys: "^L"}

```

[...]

Issue #161, opened 2021-02-09
Feature: ESC should only bring the user one level up, rather than always to Normal mode


imsnif:
Eg. when the user is in "resize input mode" (ctrl-g + r), right now pressing ESC leads back to the initial "normal input mode". Instead, it should lead back to "command input mode". Then, when the user is in "command input mode" - it should lead back to "normal input mode".
The locked state (mode_is_consistent in the codebase) should be maintained as-is until the initial normal state is reached.

A good place to start in the code base (at the time of writing): https://github.com/mosaic-org/mosaic/blob/main/src/common/input/keybinds.rs#L29

This will also require some changes in the tests, for example here (at the time of writing): https://github.com/mosaic-org/mosaic/blob/main/src/tests/integration/close_pane.rs#L359

categorille:
short comment so that i can be assigned :)

Issue #174, opened 2021-02-10
Feature: enhanced Scroll input mode


imsnif:
Right now we have a `Scroll` input mode (`<ctrl-g> + s`) which allows the user to scroll up and down.
It would be great if this can be enhanced to:
1. Behave like a "text box" that the user can copy/paste from (with the keyboard at first and with the mouse when it's supported)
2. Allow scrolling further up/down with pgup/pgdown
3. Allow searching for text in the pane's scrollbuffer

raphCode:
Update:

1. Partially solved:
-mouse selection implemented
-keyboard selection missing (tmux-like functionality)
2. Implemented
3. Still missing

PR #181, opened 2021-02-11
Keybinds categorille

categorille:
### Goal of the PR
Generally improving the input system. Moving towards fixing #178 and fixing #161.

### What changed
The [input](https://github.com/zellij-org/zellij/tree/main/src/common/input) system has seen the start of a reform.
- Command mode is now permanent by default, and can be exited either by pressing `Esc` while in command mode, or pressing `Ctrl-g` while in command mode or any submode (Pane, Tab, Scroll, Resize). This goes in the direction of fixing #178, by making the setting really simple to implement, though it hasn't been implemented in this PR.
- Pressing `Esc` while in one of the Command mode submodes now brings the user "one level back up" into command mode, allowing them to enter another submode right away. Previously, pressing escape would always bring the user back to Normal/Input mode.
- The status bar has been edited slightly to reflect the changes, and some simple improvements (shortcut descriptions, capitalization, etc.) have been added on the way.

### Technical changes
- Keys are now bound to a `Vec<Action>` instead of a single action to allow for more flexibility
- `InputMode`s aren't associated to a "persistency" flag anymore. Instead, whether they are exited after a single command or they should instead remain active is dictated by the `Action` sequences bound to keystrokes (i.e. `Esc`)
- This means that the `InputState` (=`mode: InputMode, persistent: bool`) struct is not needed anymore. The `AppState` struct has been updated accordingly

### What still needs to change
- the feature requested in #178 still has to be implemented (which should be quick) if it is deemed relevant to do so
- the input system should undergo more revision in order to eventually allow customization by the user in an elegant way, as suggested by #169


default-tiles/status-bar/src/main.rs
```diff
@@ -24,18 +24,13 @@ impl Display for LinePart {
```
khs26, line 24: Comment on the whole of this file's changes that this looks way clearer to me!

src/common/input/handler.rs
```diff
@@ -289,46 +268,48 @@ impl Default for InputMode {
...
-            keybinds.push((format!("←↓↑→"), format!("move tab focus")));
-            keybinds.push((format!("n"), format!("new tab")));
-            keybinds.push((format!("x"), format!("exit tab")));
+            keybinds.push((format!("←↓↑→"), format!("Move focus")));
+            keybinds.push((format!("n"), format!("New")));
+            keybinds.push((format!("x"), format!("Close")));
         }
         InputMode::Scroll => {
-            keybinds.push((format!("↓↑"), format!("scroll up/down")));
+            keybinds.push((format!("↓↑"), format!("Scroll")));
```
khs26, line 300: Not in the scope of this change, but I also feel like scroll mode ought to support page up and page down type ops - I haven't seen the previous changes, but guess it's currently just scrolling one line at a time?

khs26:
I think this PR is a vast improvement, and as you say hopefully opens the way for a proper overhaul

categorille:
@khs26 thanks for the feedback! I'm glad it's helpful. I'm currently struggling to get the tests to run correctly (I don't think I broke anything but the tests have hard-coded action sequences which are now outdated). This PR probably shouldn't be merged until the tests run again.

In the current state of things, tests just take very long and ultimately fail, so if you want to run them *you shouldn't wait for them to complete*. If they work, they work fast. If they take a very long time, it means something's wrong.

categorille:
### [NEED HELP]
As I said in my previous comment, I am in the tedious process of fixing tests so they pass (ironic), and I'm getting something really weird. I am correcting all of the keystroke sequences that each test function assigns to its `FakeInputOutput`'s. So far, correcting the keystroke sequences makes all of the tests pass, except that they all run *very* slowly and absolutely hog all of my CPUs.

I will add commits to this PR in a little bit so that you can test out the fixed tests yourself (ha), and so that someone more knowledgeable about Zellij's test framewok can maybe help me out by pointing to a change I made that could have introduced this test performance issue.

categorille:
Some more info because there's a lot of new commits. Commits [7e31fb9](https://github.com/zellij-org/zellij/pull/181/commits/7e31fb9ae5601601a16cb29bcb9bc8bef57fc5cb) and [f2d9421](https://github.com/zellij-org/zellij/pull/181/commits/f2d9421ff0169feb83dbd9b246e59b9244cafc16) are (I'm pretty sure) exactly the same in terms of the modifications I applied to the various test files, but I simply needed to undo all the modifications ([3961dbe](https://github.com/zellij-org/zellij/pull/181/commits/3961dbe6223745654e05037d45665ed73fc13ae0)) and redo them by little batches of files, disabling most of the irrelevant tests all the time (by deleting `#[test]` attributes on tests that weren't modified in that commit), in order to run only a few tests each time so that I could check that they actually individually pass (which they do). I had to do this because running all of the tests in one go leads to extremely long times and ultimately a bunch of failures, when running tests individually / in small groups works. This really needs investigating.

Overall, the test framework probably needs a big overhaul. The idea and main logic is clearly there, with the snapshot system, but it needs refactoring, as there is a ton of copy-pasted code, and it relies on hard-coded values (which are what I changed within the last few commits, trying to fix the test. It would have been an even worse nightmare to change without `vim`'s recording feature).

imsnif:
*points to self* person to blame for the test framework ;P

First: good on you for figuring it out. It is by far the worst and least understandable part of the codebase IMO (and there's quite some competition ;P ).

### About the issue at hand
The tests are timing out in this case because the input loop never ends (this loop here: https://github.com/zellij-org/zellij/pull/181/files#diff-8638c74348fa3aa2afe67bbc7ec64a73f8531b577afb651f04db73c9b2148ec6R60).
Since the tests use the `start` function and not the whole of `main`, their process never actually ends until all the test suite ends. This means that any thread left open in the app will be left open until the test suite ends. Since the input loop never ended, and since we don't join the stdin thread (for various reasons that aren't super substantial, and we should *definitely* change this), this caused one extra thread (with an unbound infinite loop) to hang around for each test.

To fix this, I would bring back the `'input_loop` tag and do a `break `input_loop` if this function tells us to do so: https://github.com/zellij-org/zellij/pull/181/files#diff-8638c74348fa3aa2afe67bbc7ec64a73f8531b577afb651f04db73c9b2148ec6R72
This solved the issue for me locally.

### About the test framework
I agree with everything you said. The testing framework actually predates the input module. When it was written, instead of the input module we had one giant if/else ladder over the stdin input. @khs26 did a great job turning that into what it is today. So I very much agree that right now the testing framework is unwieldy by comparison. And I also use vim macros to change it when necessary ;P
A lot of the repetitions in the tests are there because I generally have this approach that tests should have repetition. Mostly so that if a person needs to change just one test they shouldn't have to understand the whole testing framework. I'm open to re-examining that though, especially since it has given you so much grief!
Maybe we can make a plan about how to change it exactly in a separate issue?

### About this PR
Once you merge from main (right now I think the only conflicting files are the compiled binary wasm plugins - I'd say the right thing to do would be to keep your version of the status bar and take main's version of strider (I added vim keys to it)), our CI will run the tests and when it's green we can go forth and merge this.
The code changes look great, and the functionality is SO MUCH BETTER! I groan to go back to my normal environment without this change. Let's get this into main asap :)

Great job!!

categorille:
Thanks for the comments and the help! It's true that I sort of got rid of the breakpoint in the input loop just because it seemed to work, and there was a comment that said it was only needed until we had bound multiple actions to a single keystroke (which I did), so I assumed it was a relic from a past time. I'll push a final commit to bring it back.

Don't worry though, I didn't get discouraged by all the work that needs to be done on the codebase. Rather, it's quite exciting :)

src/common/input/handler.rs
```diff
@@ -276,10 +263,12 @@ pub enum InputMode {
     Exiting,
 }
 
+/// Represents the help message that is printed in the status bar, indicating
+/// the current [`InputMode`], whether that mode is persistent, and what the
```
imsnif, line 267: "whether that mode is persistent" - not anymore :) (but no biggie, we can remove this on the next pass)

src/common/input/handler.rs
imsnif: Yes! There's even an issue for this: https://github.com/zellij-org/zellij/issues/174

imsnif (approved):
Great job on this - especially all the detailed comments!
To fix formatting, you can run `cargo fmt` (if you have rustfmt installed).
Once that passes, I'll merge.

categorille:
> To fix formatting, you can run `cargo fmt` (if you have rustfmt installed).

right, forgot about `fmt`. I also fixed the comments and references to `InputState` or mode persistency in general on the way

we should (finally) be good! :) omw to document this entire codebase now 🚶🏻‍♀️

PR #185, opened 2021-02-15
First documentation step

categorille:
### Goal(s) of this PR

Adding some documentation wherever I felt confident enough to do so in a clear and concise manner.

### What changed

Mostly just added documentation (`///`) comments throughout the project. These comments aren't limited to public items, as the documentation is meant for developers. `cargo doc` should therefore be passed the `--document-private-items` flag.

I have also reworked some (not many) of the `use` directives, as I think we should try to not `use` functions from external crates directly, but rather `use` the module they live in, as to not confuse them with locally-defined functions (true of `struct`s to a certain extent as well, see the official Rust book's [comment on idiomatic `use` paths](https://doc.rust-lang.org/stable/book/ch07-04-bringing-paths-into-scope-with-the-use-keyword.html#creating-idiomatic-use-paths).

### Future perspectives

I plan to suggest changes to `CONTRIBUTING.md` regarding documentation and `use` directive requirements. The rest of the existing codebase still needs to be documented.

### Other comments

The individual commits are big because I didn't go through the codebase linearly, and didn't know what lines to split my commits along.

### DIFF
--- a/src/client/tab.rs
+++ b/src/client/tab.rs
@@ -1,4 +1,4 @@
-//! `Tab`s holds multiple panes (currently terminal panes). It tracks their coordinates (x/y) and size, as well as how they should be resized
+//! `Tab`s holds multiple panes. It tracks their coordinates (x/y) and size, as well as how they should be resized
 
 use crate::common::{AppInstruction, SenderWithContext};
 use crate::panes::{PaneId, PositionAndSize, TerminalPane};
--- a/src/client/tab.rs
+++ b/src/client/tab.rs
@@ -1,3 +1,5 @@
+//! `Tab`s holds multiple panes (currently terminal panes). It tracks their coordinates (x/y) and size, as well as how they should be resized
+
 use crate::common::{AppInstruction, SenderWithContext};
 use crate::panes::{PaneId, PositionAndSize, TerminalPane};
 use crate::pty_bus::{PtyInstruction, VteEvent};
@@ -12,15 +14,6 @@ use std::{
 use std::{io::Write, sync::mpsc::channel};
 
 use crate::utils::logging::debug_log_to_file;
-
-/*
- * Tab
- *
- * this holds multiple panes (currently terminal panes) which are currently displayed
- * when this tab is active.
- * it tracks their coordinates (x/y) and size, as well as how they should be resized
- *
- */
 
 const CURSOR_HEIGHT_WIDTH_RATIO: usize = 4; // this is not accurate and kind of a magic number, TODO: look into this
 const MIN_TERMINAL_HEIGHT: usize = 2;
--- a/src/common/errors.rs
+++ b/src/common/errors.rs
@@ -1,4 +1,5 @@
-//! All things related to errors and error contexts.
+//! Error context system based on a thread-local representation of the call stack, itself based on
+//! the instructions that are sent between threads.
 
 use super::{AppInstruction, OPENCALLS};
 use crate::pty_bus::PtyInstruction;
@@ -6,12 +7,15 @@ use crate::screen::ScreenInstruction;
 
 use std::fmt::{Display, Error, Formatter};
 
+/// The maximum amount of calls an [`ErrorContext`] will keep track
+/// of in its stack representation. This is a per-thread maximum.
 const MAX_THREAD_CALL_STACK: usize = 6;
 
 #[cfg(not(test))]
 use super::SenderWithContext;
 #[cfg(not(test))]
 use std::panic::PanicInfo;
+/// Custom panic handler/hook. Prints the [`ErrorContext`].
 #[cfg(not(test))]
 pub fn handle_panic(
     info: &PanicInfo<'_>,
@@ -68,19 +72,22 @@ pub fn handle_panic(
     }
 }
 
-/// An [`ErrorContext`] struct contains a representation of the call stack
+/// A representation of the call stack.
 #[derive(Clone, Copy)]
 pub struct ErrorContext {
     calls: [ContextType; MAX_THREAD_CALL_STACK],
 }
 
 impl ErrorContext {
+    /// Returns a new, blank [`ErrorContext`] containing only [`Empty`](ContextType::Empty)
+    /// calls.
     pub fn new() -> Self {
         Self {
             calls: [ContextType::Empty; MAX_THREAD_CALL_STACK],
         }
     }
 
+    /// Adds a call to this [`ErrorContext`]'s call stack representation.
     pub fn add_call(&mut self, call: ContextType) {
         for ctx in self.calls.iter_mut() {
             if *ctx == ContextType::Empty {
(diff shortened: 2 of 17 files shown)


src/client/tab.rs
```diff
@@ -1,3 +1,5 @@
+//! `Tab`s holds multiple panes (currently terminal panes). It tracks their coordinates (x/y) and size, as well as how they should be resized
```
imsnif, line 1: This is outdated, as now a Pane can also be a plugin pane.

src/common/errors.rs
```diff
@@ -108,19 +118,30 @@ impl Display for ErrorContext {
...
     App(AppContext),
     IPCServer,
     StdinHandler,
     AsyncTask,
+    /// An empty, placeholder call. This should be thought of as representing no call at all.
+    /// A call stack representation filled with these is the representation of an empty call stack.
     Empty,
 }
 
+// TODO use the `colored` crate for color formatting
```
imsnif, line 144: +1

src/common/mod.rs
```diff
@@ -54,23 +54,36 @@ pub fn update_state(
...
     drop(app_tx.send(AppInstruction::GetState(state_tx)));
     let state = state_rx.recv().unwrap();
 
     drop(app_tx.send(AppInstruction::SetState(update_fn(state))))
 }
 
-pub type ChannelWithContext<T> = (Sender<(T, ErrorContext)>, Receiver<(T, ErrorContext)>);
-pub type SyncChannelWithContext<T> = (SyncSender<(T, ErrorContext)>, Receiver<(T, ErrorContext)>);
-
+/// An [MPSC](mpsc) asynchronous channel with added context.
```
imsnif, line 65: Maybe specify that this is with added error context?

src/common/mod.rs
```diff
@@ -97,6 +117,7 @@ impl<T: Clone> SenderWithContext<T> {
         }
     }
 
+    /// Updates this [`SenderWithContext`]'s [`ErrorContext`].
```
imsnif, line 120: Maybe add something like "this is the way one adds a call to the error context and works this way in order not to allocate on the heap"?

src/common/mod.rs
```diff
@@ -105,16 +126,23 @@ impl<T: Clone> SenderWithContext<T> {
...
 pub enum AppInstruction {
-    GetState(Sender<AppState>),
+    GetState(mpsc::Sender<AppState>),
     SetState(AppState),
     Exit,
     Error(String),
 }
 
+/// Start Zellij with the specified [`OsApi`] and command-line arguments.
+// FIXME this should definitely be modularized and split into different functions.
```
imsnif, line 145: +1

src/common/os_input_output.rs
```diff
@@ -88,7 +95,15 @@ fn handle_command_exit(mut child: Child) {
...
 
-fn spawn_terminal(file_to_open: Option<PathBuf>, orig_termios: Termios) -> (RawFd, RawFd) {
+/// Spawns a new terminal from the parent terminal with [`termios`](termios::Termios)
+/// `orig_termios`.
+///
+/// If a `file_to_open` is given, the text editor specified by environment variable `EDITOR`
+/// will be started in the new terminal, with the given file open. If no file is given, the
+/// shell specified by environment variable `SHELL` will be started in the new terminal.
+// FIXME this should probably be split into different functions, or at least have less levels
+// of indentation in some way
```
imsnif, line 105: :D

src/common/os_input_output.rs
```diff
@@ -88,7 +95,15 @@ fn handle_command_exit(mut child: Child) {
     }
 }
 
-fn spawn_terminal(file_to_open: Option<PathBuf>, orig_termios: Termios) -> (RawFd, RawFd) {
+/// Spawns a new terminal from the parent terminal with [`termios`](termios::Termios)
+/// `orig_termios`.
+///
+/// If a `file_to_open` is given, the text editor specified by environment variable `EDITOR`
```
imsnif, line 101: + or the environment variable 'VISUAL' if 'EDITOR' is not set

src/common/os_input_output.rs
```diff
@@ -136,64 +151,80 @@ fn spawn_terminal(file_to_open: Option<PathBuf>, orig_termios: Termios) -> (RawF
...
-    fn kill(&mut self, pid: RawFd) -> Result<(), nix::Error>;
+    /// Read bytes from the standard output of the virtual terminal referred to by `fd`.
+    fn read_from_tty_stdout(&mut self, fd: RawFd, buf: &mut [u8]) -> Result<usize, nix::Error>;
+    /// Write bytes to the standard input of the virtual terminal referred to by `fd`.
+    fn write_to_tty_stdin(&mut self, fd: RawFd, buf: &mut [u8]) -> Result<usize, nix::Error>;
+    /// Wait until all output written to the object referred to by `fd` has been transmitted.
+    fn tcdrain(&mut self, fd: RawFd) -> Result<(), nix::Error>;
+    /// Terminate the
+    // FIXME `RawFd` is semantically the wrong type here. It should either be a raw libc::pid_t,
+    // or a nix::unistd::Pid.
```
imsnif, line 180: I'm not sure I understand why?

src/common/screen.rs
```diff
@@ -166,24 +188,31 @@ impl Screen {
...
         }
     }
-    pub fn get_tabs_mut(&mut self) -> &mut BTreeMap<usize, Tab> {
-        &mut self.tabs
-    }
+
+    /// Returns a mutable reference to this [`Screen`]'s active [`Tab`].
     pub fn get_active_tab_mut(&mut self) -> Option<&mut Tab> {
-        let tab = match self.active_tab_index {
+        match self.active_tab_index {
```
imsnif, line 206: oops, thanks! I guess it was left behind in some debugging session

imsnif (changes requested):
This is fantastic. Very thorough work and great attention to details. My only major concern here is that I hope we'll be able to keep up this level of documentation as the code base changes. That being said, I don't think it's a good reason not to try. Do you think there are any clippy rules or other tools we can use to give us developers polite reminders about documenting various code entities?

The generated documentation also looks great. Until we publish to crates.io and get this in docs.rs, do you think you can add brief instructions to `CONTRIBUTING` about how to generate these and view them?

Otherwise, I've left some minor comments and a question about something I didn't fully understand. Let's address these and merge.

src/common/os_input_output.rs
categorille: see `man kill.2`, `man kill.3` and the signature of `nix::sys::signal::kill`. All three take an argument called `pid` of type `pid_t` (just a `typedef`ed C `int` in reality), rather than a file descriptor. [`RawFd`](https://doc.rust-lang.org/std/os/unix/io/type.RawFd.html) happens to also be a `typedef`ed C `int`, so it works technically, but semantically it makes no sense: signals aren't sent to entities associated to file descriptors, they are sent to processes, which are associated to `pid`s.

categorille:
> This is fantastic. Very thorough work and great attention to details.

Thank you! It's definitely been a bit of work getting an understanding of these components sufficient to clearly summarize their role and inner workings.

> I hope we'll be able to keep up this level of documentation as the code base changes. 

it's a worry of mine as well. However, I think it is a worthwhile investment when it comes to attracting new contributors. I happened to have enough time to dig down into (let's be real, quite messy) codebases to create my own idea of its structure, but I think that people who would like to contribute only a small change and not invest themselves all that much on the long term might enjoy being able to get a grasp of the structure at a glance.

We could think of setting a relatively permissive standard regarding the quality of documentation, i.e. just enough to understand the role of the component by looking at its documentation, without all the details about individual arguments (if clear enough from their name and type), the panic cases etc. etc. I can see myself spending some time on making the documentation prettier if the basis is already there.

> do you think you can add brief instructions to `CONTRIBUTING` about how to generate these and view them?

I suggest we already merge this, and I'll keep thinking of what kind standard we should set ourselves (maybe discuss it a bit more with core members). `CONTRIBUTING` probably won't be a source of merge conflicts, so we don't have to worry about it too much right now.

src/common/os_input_output.rs
imsnif: Good catch - that means there's an error in the `spawn_terminal` method as well. It took quite some time to get that one to work properly and I went through quite some iterations of the return types while doing so. I assume the erroneous return value was from one of them.

imsnif:
I definitely agree it's a worthwhile investment. My experience with large code bases being touched by multiple developers (most of whom aren't native English speakers) tells me that unless this is somehow enforced technically (eg. through a linting rule) it won't hold over time.

I agree that the code base is quite messy. It's a painful point for me as well. I very much welcome all work in the direction of making it more comprehensible and friendly to new contributors - but things do change rather quickly because these are the early stages of the project. That causes a mess that (IMO at least) is inevitable. Thank you for your patience and hard work. :) I'm looking forward to more improvements in this and other areas from you.

imsnif (approved):

Issue #187, opened 2021-02-16
Add rust-toolchain.toml file.


a-kenji:
I think it would be valuable to add a rust-toolchain file,
if it is powerful enough.
That way our rust versions and components could be in sync.
```
[toolchain]
channel = "Strings"
components = [ "Strings" ]
targets = [ "Strings" ]
```
https://rust-lang.github.io/rustup/overrides.html#the-toolchain-file

imsnif:
Sounds good! @TheLostLambda - what do you think? As our resident build/cargo/rustup expert? :D

TheLostLambda:
I think I commented on this in the Discord, but I think this is a great idea! I didn't know about it until @a-kenji  mentioned it!

a-kenji:
Awesome! I suggested a small basic toolchain file.
I was not sure especially which targets are supported. 
So would be happy about some input.

TheLostLambda:
So far as I'm aware, we should build on macOS and Linux, x86, x86_64, and aarch64. wasm32-wasi is also what the plugins are built for.

a-kenji:
Cheers. 
Maybe it makes sense to not define the targets in the file then if we decide to use it at all -
they could be a lot and hard to maintain? 

I can build the project with just the file, but maybe I have some stuff thats still cached left over.

PR #189, opened 2021-02-17
Add rust-toolchain.toml file

a-kenji:
https://github.com/zellij-org/zellij/issues/187

I added a simple rust-toolchain file for now. 
Please suggest appropriate targets and or 
components for it.

The rust-toolchain file makes
it possible to share a basic
rustup configuration across the
project.

https://rust-lang.github.io/rustup/overrides.html#the-toolchain-file

Important to note, this is the
current rustup hierarchy:

1. An explicit toolchain, e.g. cargo +beta,
2. The RUSTUP_TOOLCHAIN environment variable,
3. A directory override, ala rustup override set beta,
4. The rust-toolchain file,
5. The default toolchain,

source:
https://github.com/ebroto/rustup/blob/c2db7dac6b38c99538eec472db9d23d18f918409/README.md#toolchain-override-shorthand

This can especially be helpful if in the future we depend on certain rust versions.

### DIFF
--- a/rust-toolchain.toml
+++ b/rust-toolchain.toml
@@ -5,4 +5,4 @@
 # version: "nightly-1.0.0"
 channel = "stable"
 components = [ "rustfmt", "rust-src", "clippy", "rust-analysis"]
-targets = [ "wasm32-wasi" ]
+# targets = [ ]
--- a/rust-toolchain.toml
+++ b/rust-toolchain.toml
@@ -1,0 +1,8 @@
+# https://rust-lang.github.io/rustup/overrides.html#the-toolchain-file
+[toolchain]
+# can be further pinned eg:
+# date: "stable-2020-07-10"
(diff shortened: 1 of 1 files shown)


TheLostLambda (approved):
I don't see any harm in this, but it doesn't do what I thought it would. It doesn't seem to install any missing targets (even before it was commented out).

I'm still fine adding it, but maybe we'll need to figure out the target thing at some point in the future!

a-kenji:
Oh yeah, that would be good to figure out!
I hope it gets more powerful in the future.
It seemed to work for me did you maybe have other
toolchains overriding it, sadly it should only work when you 
have no 1. - 3. installed/setup that will take precedence.

TheLostLambda:
I suppose I'll merge this! It looks like a good change! Thanks for your work on this!

PR #200, opened 2021-02-20
Clippy fixes.

a-kenji:
Low hanging Clippy fixes.

I opted to allow the single match in the signal handler for now,
since there is already a todo item, and as soon as the WINCH
feature comes that should handle more than one case probably
anyway.

### DIFF
--- a/src/client/tab.rs
+++ b/src/client/tab.rs
@@ -1433,7 +1433,7 @@ impl Tab {
                 p.columns() > increase_by && p.columns() - increase_by >= p.min_width()
             });
         } else {
-            return false;
+            false
         }
     }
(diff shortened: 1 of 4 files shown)


denismaxim0v (commented):
So the tests pass, but for some reason I can't run the branch with release flag, can you please confirm if this is the case for you as well?

a-kenji:
I can confirm that the release flag takes a long time on my side, I assumed it was my
questionable hardware. But I can run it.

denismaxim0v:
Can confirm, after some cleanups, I am able to run it too. lgtm

denismaxim0v (approved):

PR #203, opened 2021-02-21
feature(strider): Allow to Hide Hidden Files

### LINKED ISSUE #197: Feature: allow Strider to hide hidden files
With some sort of shortcut (h? something else?) this should be toggled.

a-kenji:
closes #197

default keybinding = '.'
default view = shows hidden files

Alternative:
Hide hidden files by default.


default-tiles/strider/src/state.rs
```diff
@@ -6,6 +6,7 @@ pub struct State {
     pub path: PathBuf,
     pub files: Vec<FsEntry>,
     pub cursor_hist: HashMap<PathBuf, (usize, usize)>,
+    pub not_hidden: bool,
```
imsnif, line 9: Hum - what do you think about changing the name here to `show_hidden_files`? or `hide_hidden_files` if you prefer the bool to be the other way around?

imsnif (changes requested):
Very cool!! I left a minor nitpick about naming and then we can merge :)

default-tiles/strider/src/state.rs
a-kenji: Yes, that does make much more sense.
Should the default stay on showing the hidden files?

imsnif: My preference would be to not show hidden files by default, and ideally to also provide some sort of hint to the user that they can press `.` to show them. That may be out of scope here though. So whatever you think is best, I'm good with all options for now :)

a-kenji: Awesome! Yeah with a help-page - or something similar - it would make sense to hide by default.

TheLostLambda: I personally prefer viewing hidden files by default (it seems to be what most IDEs I've used do), but if others would rather hide them, I'm sure I could live with it too!

a-kenji: I guess the best way would be something configurable down the line!

PR #205, opened 2021-02-22
Input normal mode

### LINKED ISSUE #184: [Bug/Feature] In all non-Normal modes, user input should never be written to the focused pane
### Current behaviour
In all modes (including Normal mode, technically), user input that isn't bound to any sequence of action is written to the focused pane.

### Desired behaviour
When in command mode, no user input should ever be written to the terminal, as this is highly confusing (some characters get written but others are bound to actions, like splitting the focused pane, resizing panes, etc).

### Technical considerations
The current behaviour is due to [this line of code](https://github.com/zellij-org/zellij/blob/c74e2ef27356257c45c6a7bdb24c51a7ae390d8f/src/common/input/keybinds.rs#L176). I have considered creating a `NoOp` [`Action`](https://github.com/zellij-org/zellij/blob/c74e2ef27356257c45c6a7bdb24c51a7ae390d8f/src/common/input/actions.rs#L15), which would replace `Action::Write(input)` as the default action. Sadly, this doesn't work, as the input logic is the same in Normal mode as any other mode: Normal mode has very few keybinds (only [one](https://github.com/zellij-org/zellij/blob/c74e2ef27356257c45c6a7bdb24c51a7ae390d8f/src/common/input/keybinds.rs#L31), in fact, which allows switching to Command mode), and relies on the rest of the unrecognized inputs to be written to the focused pane.

This can probably be considered a part of the general revision of the input module.

a-kenji:
closes #184

Added the NoOp from:
https://github.com/zellij-org/zellij/issues/184

Now text is only written in normal Mode.

### DIFF
--- a/rust-toolchain.toml
+++ b/rust-toolchain.toml
@@ -5,4 +5,4 @@
 # version: "nightly-1.0.0"
 channel = "stable"
 components = [ "rustfmt", "rust-src", "clippy", "rust-analysis"]
-targets = [ "wasm32-wasi" ]
+# targets = [ ]
--- a/rust-toolchain.toml
+++ b/rust-toolchain.toml
@@ -1,0 +1,8 @@
+# https://rust-lang.github.io/rustup/overrides.html#the-toolchain-file
+[toolchain]
+# can be further pinned eg:
+# date: "stable-2020-07-10"
+# version: "nightly-1.0.0"
+channel = "stable"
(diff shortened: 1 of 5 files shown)


categorille:
thanks for implementing this! it was a much needed change

TheLostLambda:
This looks good to me! I'll need to get rid of the old toolchain file, but I'll do that after merging!

a-kenji:
Ah, yeah sorry. I need to get more savvy at git.

PR #207, opened 2021-02-23
Doc build

a-kenji:
To inform people about the `binaryen` dependency.


imsnif:
I think this is no longer the case, right @TheLostLambda ?

TheLostLambda:
It might still be worth mentioning, as wasm-opt is used if available, but you're right that it's not *needed* anymore!

a-kenji:
Oh, I thought I pushed the revised version already.

imsnif (approved):
Thanks!

Issue #166, opened 2021-02-10
Feature: top tab bar


imsnif:
This would add a top status bar to the default layout of zellij (similar to the bottom controls status bar that it has right now). This bar will show the currently open tabs and indicate which one of them is focused.

### How will this work?
The bottom status bar is a webassembly plugin. Its code (at the time of this writing) can be found here: https://github.com/zellij-org/zellij/tree/main/default-tiles/status-bar
Because it's a default plugin, its code is part of the main zellij repository.
Inspiration can be drawn from it in order to create the top bar, including adding it to the default layout (at the time of this writing) here: https://github.com/zellij-org/zellij/blob/main/assets/layouts/default.yaml

imsnif:
This has been implemented

PR #212, opened 2021-02-28
Add ability to set tab name

jcaplan:
#166 
Ctrl-G/t/r to rename the active tab.

This is a fairly moderate change as I needed a way to selectively funnel input streams to a specific plugin. I got the idea that the  `PluginInstruction::Input` needs to be divided into two cases:

1. the input needs to be forwarded to the active pane, which happens to be a plugin, identified by pid (existing case)
2. the input needs to be forwarded to one or more plugins, based on whether or not they are interested in that information

I implemented this by adding  `events` to the layout. When an input comes in that is destined for a plugin (decided based on the InputMode) it is forwarded to the plugin:

In layout.yml:
```
parts:
  - direction: Vertical
    split_size:
      Fixed: 1
    plugin: tab-bar
    events:
      - Tab
```
When a tab loads a plugin it now sends the layout.events:
```
fn apply_layout{
                ...
                self.send_plugin_instructions
                    .send(PluginInstruction::Load(
                        pid_tx,
                        plugin.clone(),
                        layout.events.clone(),
                    ))
```

and then when sending the instruction to the wasm thread:
```
                self.send_plugin_instructions
                    .send(PluginInstruction::Input(
                        PluginInputType::Event(EventType::Tab),
                        c.clone(),
                    ))
```
In the wasm thread I initialize a hashmap that tells which handler is associated with this event type:
```
            let handler_map: HashMap<EventType, String> = [(EventType::Tab, "handle_tab_event".to_string())].iter().cloned().collect();
```

When handling this instruction there are two checks:
```
                               for (instance, plugin_env) in plugin_map.values() {
                                    if !plugin_env.events.contains(&event) {
                                        continue;
                                    }
                                    let handle_key = instance.exports.get_function(
                                        handler_map.get(&event).unwrap()
                                    ).unwrap();
```

First the `plugin_env` is extended to include the event types that the plugin wants to subscribe to (decided at load time), so we skip over plugins that don't care about this type of event. Then we get the handler associated with this event type, and forward it.

The same trick is used for `update_tabs` now. Before, the default handler consumed the stream to keep the plugins stdin clean, now the plugin can be specified as a subscriber for the event.

### DIFF
--- a/assets/layouts/default.yaml
+++ b/assets/layouts/default.yaml
@@ -5,6 +5,8 @@ parts:
     split_size:
       Fixed: 1
     plugin: tab-bar
+    events:
+      - Tab
   - direction: Vertical
     expansion_boundary: true
   - direction: Vertical
--- a/default-tiles/status-bar/src/main.rs
+++ b/default-tiles/status-bar/src/main.rs
@@ -104,7 +104,7 @@ fn key_path(help: &Help) -> LinePart {
                 len,
             )
         }
-        InputMode::Tab => {
+        InputMode::Tab | InputMode::RenameTab => {
             let mode_shortcut_text = "t ";
             let superkey = superkey_text.bold().on_magenta();
             let first_superkey_separator = ARROW_SEPARATOR.magenta().on_black();
--- a/default-tiles/tab-bar/src/line.rs
+++ b/default-tiles/tab-bar/src/line.rs
@@ -115,8 +115,7 @@ fn add_next_tabs_msg(
     title_bar: &mut Vec<LinePart>,
     cols: usize,
 ) {
-    while get_current_title_len(&title_bar) +
-        right_more_message(tabs_after_active.len()).len
+    while get_current_title_len(&title_bar) + right_more_message(tabs_after_active.len()).len
         >= cols
     {
         tabs_after_active.insert(0, title_bar.pop().unwrap());
--- a/default-tiles/tab-bar/src/line.rs
+++ b/default-tiles/tab-bar/src/line.rs
@@ -116,7 +116,6 @@ fn add_next_tabs_msg(
     cols: usize,
 ) {
     while get_current_title_len(&title_bar) +
-        // get_tabs_after_len(tabs_after_active.len()) >= cols {
         right_more_message(tabs_after_active.len()).len
         >= cols
     {
--- a/default-tiles/tab-bar/src/main.rs
+++ b/default-tiles/tab-bar/src/main.rs
@@ -63,6 +63,8 @@ impl ZellijTile for State {
                     tabname = self.new_name.clone();
                 }
                 active_tab_index = t.position;
+            } else if t.active {
+                active_tab_index = t.position;
             }
             let tab = tab_style(tabname, t.active, t.position);
             all_tabs.push(tab);
--- a/default-tiles/tab-bar/src/main.rs
+++ b/default-tiles/tab-bar/src/main.rs
@@ -57,7 +57,11 @@ impl ZellijTile for State {
         for t in self.tabs.iter_mut() {
             let mut tabname = t.name.clone();
             if t.active && self.mode == BarMode::Rename {
-                tabname = self.new_name.clone();
+                if self.new_name.is_empty() {
+                    tabname = String::from("Enter name...");
+                } else {
+                    tabname = self.new_name.clone();
+                }
                 active_tab_index = t.position;
             }
             let tab = tab_style(tabname, t.active, t.position);
--- a/default-tiles/tab-bar/src/main.rs
+++ b/default-tiles/tab-bar/src/main.rs
@@ -12,11 +12,24 @@ pub struct LinePart {
     len: usize,
 }
 
+enum Mode {
+    Normal,
+    Rename,
+}
+
+impl Default for Mode {
+    fn default() -> Self {
+        Mode::Normal
+    }
+}
+
 #[derive(Default)]
 struct State {
     active_tab_index: usize,
     num_tabs: usize,
     tabs: Vec<TabData>,
+    mode: Mode,
+    new_name: String,
 }
 
 static ARROW_SEPARATOR: &str = "";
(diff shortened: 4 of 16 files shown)


TheLostLambda:
Heya! Today is supposed to be my "off day", so I should be able to give things a bit of a closer look in the next couple of days, but I think I've had a bit of a different vision for the data-flow to plugins overall (which isn't really documented anywhere, which is my fault for not getting around to) and I wanted to leave a couple comments for consideration. Some of this commentary also applies to the last plugin PR you've done as well that I didn't quite get a chance to see before merging!

First of all, thank you for putting in these PRs! They are adding some excellent features that will make Zellij quite nice to use!

With that being said, I've got a common thread of concern when it comes to API design / extension: The plugin system was originally conceived to make it as easy as possible to write plugins in *different* languages. As such, one really vital requirement has been keeping `zellij-tile` as small as humanly possible, as every bit of code written there will need to be duplicated to every other language we wish to support. Introducing call-backs to the plugin mandates this sort of change and will break compatibility with every plugin compiled for an older version of the API (Zellij will ask for a function that the WASM module simply doesn't know about).

The alternative to this that I think will be more maintainable in the long-run is the addition of a "host function" like `host_get_help()` in `wasm_vm.rs`. This way the plugin asks for the information from Zellij. This *won't* break plugins using older versions of `zellij-tile` and we don't need to add any new events to the WASM thread (no need for things like `PluginInstruction::UpdateTabs` or the like).

On the point of the event subscriptions, this is something we'll probably need to think about a little more. I'm still not 100% convinced we'll need them (I have some other ideas like adding a single `update(delta_time)` function for timed events), but even if we do end up with event subscriptions, they should ideally be things the plugins themselves can subscribe to (via host-functions) – not something stuck in the layouts which have been designed as a bit of a stop-gap (eventually we want to load plugins dynamically, without needing to add them to any layouts).

I know that's a lot and is, perhaps, quite abstract, but I'm looking at refactoring your last PR a little on Tuesday. If it comes out a bit simpler (which I'm optimistic about), then implementing tab names should be easier in this PR as well.

Additionally, if you totally disagree with my outlook on the plugin system, that's fair! We have meetings Tuesday evenings (18:00 GMT) if you would like to come along and discuss things!

For the short-term, it would be ideal if we could wait to merge this until after I give refactoring a go on Tuesday, but I'm open to dissenting opinions on this :slightly_smiling_face: 

Finally, thank you for all of your hard work on the tab-bar! It's excellent to have it as a plugin and it's very impressive that you've managed to implement it without any proper documentation from Zellij!

jcaplan:
@TheLostLambda 

Thanks for the feedback! I figured it would need some changes.

> As such, one really vital requirement has been keeping zellij-tile as small as humanly possible, as every bit of code written there will need to be duplicated to every other language we wish to support.

In this case, I think that what's missing is basically a more formalized REST-ish API. The problem with `handle_key` is that the data is not tagged in any way, if you're receiving it then you assume you want it. If the answer isn't to add lots of handlers in zellij-tile/lib.rs, then I think there needs to be a more formalized serializable data schema used to pass arbitrary data to a single plugin handle. Then the ball is entirely in the plugins court whether or not to discard the event. I don't think having a handle that passes raw characters is good for much once you accept that non-active panes may still have work to do and expect data to pass from the main app.

I wanted to try to filter on the sender side, but it's easily avoidable. What does seem unavoidable is tagging all data flowing to plugins. The simple use case is strider vs. tab bar. With strider it was assumed that the data need only flow to the active pane, but this assumption doesn't hold for the tab bar. So something needs to be done to let plugins filter out the data they want.

> The alternative to this that I think will be more maintainable in the long-run is the addition of a "host function" like host_get_help() in wasm_vm.rs. This way the plugin asks for the information from Zellij. This won't break plugins using older versions of zellij-tile and we don't need to add any new events to the WASM thread (no need for things like PluginInstruction::UpdateTabs or the like).

At first I tried this way but I hit a problem. There is no global data in common/mod.rs right now. All the data gets moved into each thread. So the screen thread owns the screen data and there's no other way to get some except to ask the screen thread about it.

What happens when a screen tries to render a tab, and the tab tries to render the plugin, and the plugin tries to get data from the screen object, but the screen object won't reply until it finishes rendering and loops around... deadlock.

So that's a fundamental problem assuming that plugins will be able to poll for information that they need the way things are currently designed, if they ever need information from the screen thread. That's why I think the really only viable solution without making pretty moderate architectural changes is to push data out to plugins and let them decide if they need it.

>On the point of the event subscriptions, this is something we'll probably need to think about a little more. I'm still not 100% convinced we'll need them (I have some other ideas like adding a single update(delta_time) function for timed events), but even if we do end up with event subscriptions, they should ideally be things the plugins themselves can subscribe to (via host-functions) – not something stuck in the layouts which have been designed as a bit of a stop-gap (eventually we want to load plugins dynamically, without needing to add them to any layouts).

Yeah that's definitely my least favorite part. I just wanted to make sure that it was possible to attach more metadata about plugins into the plugin_env, but I figured the way I did it would need to change. I just wanted to get it minimally working and get some feedback on the events stuff.

imsnif:
Hey @jcaplan - @TheLostLambda isn't going with the API change for now and will do another different one next week.
Meanwhile, I added some styling to the tab bar in main, so this will require a bit of a merge (but it shouldn't be too difficult I think). Are you comfortable with merging from main and incorporating it with your changes? I think it's mostly about using the methods I added in the new `tab.rs` file in order to render tabs "prettily", and maybe adding some styling to the `Enter name` line, if you're feeling creative.
Sorry for the back-and-forth on this, and thanks for your patience :)

jcaplan:
ok rebase is done. just an fyi the color scheme is quite hard to read on my terminal and some of the characters don't render:

image

imsnif:
Thanks for the rebase @jcaplan. I'm going over everything now.

Oi, that's too bad about the legibility and the unicode characters not being rendered. I would like us to open a separate issue for it regardless, but so that we are able to give that issue the right title, I'll ask a few questions (might have more depending on the answers):
1. What terminal emulator are you using?
2. From the screenshot it seems to me you're running in side tmux, is that the case? If so, does it look fine outside of tmux?

imsnif (changes requested):
Hey @jcaplan - great work on this. I think this one was especially tricky because it required some infrastructure we did not have in our api, and that's why it required some back and forth. I won't comment on the whole API and data flow, because @TheLostLambda is going to do a rewrite of those sections next week anyway (nothing wrong with the way things are here, but we were meaning to take a different approach on this). So we can merge those parts as is.
As for the rest, I left a few minor comments about naming in the code. Otherwise, there are a couple of functionality issues we need to take care of:

1. When naming a new tab, we should pad the name in the end and the beginning with 1 space (this is done with the nameless tabs already).
2. When naming a new tab after naming a previous one, the string from the previous name stays in the `Enter new name` field. I guess we should probably delete it from the state once a new tab name is set?
3. What do you think about styling the rename differently? Instead of it being clear text, maybe the user experience can be that the name you enter is interactively updated in the new tab?

Thanks! Let's do some more work on this and get this merged.

jcaplan:
@imsnif I think these new changes address your suggestions

imsnif:
Hey @jcaplan - good work on this. I fixed two minor bugs:
1. The active tab name wasn't being padded
2. The active tab index wasn't being passed to the `tab_line` component, so it wasn't truncating the right tabs out of the line when it overflowed.

This PR introduces a new issue: when a tab has a name (or when entering a new name) that exceeds the size of the screen, the app crashes. This can be fixed by truncating the name itself (eg. `I am a very long name` ==> `I am a[...]g name`). I'll open a separate issue for it, feel free to pick it up if you want to.

Thanks for your work on this!

PR #213, opened 2021-02-28
fix(build): This is getting uglier :/

henil:
When the user has `CARGO_TARGET_DIR` set, it messes up the build system. This PR fixes it.


TheLostLambda:
This looks good to me! This is all getting a bit gross though, so hopefully a better solution rolls around soon...

TheLostLambda (approved):

PR #204, opened 2021-02-21
add tab toolbar

jcaplan:
#166 

This is adds a toolbar showing the current tab. Tabs are only identified by number. You can select the tab by number key in addition to j and k key bindings.

PR #215, opened 2021-03-02
Tab Bar / Plugin System Refactor

TheLostLambda:
https://user-images.githubusercontent.com/6251883/109726767-2d072880-7bab-11eb-9373-583e46b35e08.mp4

This is related to PR #212 and is a refactor of #204 

This introduces an update-loop to plugins (enabling timed events), gives plugins a way to request a rerender (if their state has been updated), and introduces the `get_tab_info()` host function, ready to pass named tabs as soon as PR #212 is adapted!


imsnif:
Hey - so first: great work. I'm happy to see the plugin API getting into better and better shape.
I think this is a good change, it makes sense to me and we can definitely go forward with it.

That being said, if you feel like putting more work into this - I think the general architecture of the update loop can probably be simplified and made safer. Right now what happens (please correct me if I misunderstood), is that we have an update function which is called every 50ms when the wasm thread is idle. The update method of each plugin can request state from Zellij and update itself, and can also add a re-render.

I see several issues here:
1. The duration between update calls is unpredictable. Because they work on a timeout, they will happen less often when the app is busy. With internal stuff this is not a big issue, but since we're talking about external plugins - this is something that is not super clear to plugin authors.
2. I don't think it's a good idea for plugins to be able to request a re-render. This can create a whole host of deadlocks, infinite loops and can generally create difficult situations to debug (eg. nothing's stopping a plugin from requesting a re-render from the draw function, right?). IMO the better way would be to have plugins subscribe to the state they want on load (eg. TabInfo) and have that state provided to them as part of the draw function (or a separate update function that is called before the draw).
3. Building on point 2, I think it would be simpler not to request information from eg. the Screen thread, but rather have that thread update a central state (I think we already have something like that on the app thread, right?) and then have (parts of) that state sent to the plugins who requested them. Generally this would create an architecture where we stream changes only in one direction (to the plugins) and not have a lot of back-and-forth which we'll have little control over.

Again, these are just my thoughts and what I would do if I were implementing it. I'm sharing because you asked for them. I would have no issue with you merging this as it is - I think it's for sure better than what we have. Up to you.

TheLostLambda:
Thanks for taking the time to give things a look!

I think you understand things correctly! I'm kinda modelling the plugins [after this](https://love2d.org/wiki/Tutorial:Callback_Functions) since that library is so ridiculously easy to use.

Addressing your points:
1. I'm honestly not too bothered by this and I think that updates will always be a little unpredictable – that's why you pass the `dt` for time-sensitive stuff! But we could improve this by creating a new sub-thread for updating things, I tried that briefly yesterday. I think though it's much more important for the user experience to prioritise input processing responsiveness.
2. I think there is no way around it. How would you design things like a clock widget if you aren't doing anything at the same time? Terminals can request re-renders, I think plugins should be capable of the same :slightly_smiling_face: Eventually though, those requests should just rerender that one plugin pane, not the whole screen. Additionally, it's not a blocking call, just queues a render, so it works just fine in the tab-bar `draw()` function even if things get a bit flickery and slow
3. I've come to 100% agree with this. That's something I'd like to do next week. Just push state to the App thread so plugins can query it non-blockingly. That's what I meant by [the comment here](https://github.com/zellij-org/zellij/blob/e1f6eebafe620816a4214e18712567fcba66eb2a/src/common/wasm_vm.rs#L109)!

Let me know what you think! This PR is certainly far from perfect, but I'd rather the plugin / zellij boundary be clear and coherent, even if it means working around some tricky things in Zellij itself.

Also, on the event-subscriptions thing: I'd like to see if we can get things to be decent and performant without them, just from an plugin developer perspective. I'm certainly open to dissenting opinions though!

imsnif:
I think the main issue I have collectively with 1 and 2 is the classic "if you give the developer a gun, be certain they will use it".
If we allow developers to request a re-render with an API call, they will likely at least *try* to create an animation loop. I know that's the first thing I would do.
Right now it won't work because of performance, but I'm kind of afraid of it actually-starting-to-kind-of-work at some point in the near future. Then, we have plugins in the wild which rely on this functionality, we might have libraries that perform this functionality, and then we inevitably have to support it and make sure our performance is always of a level that will support plugins doing complex animations (rendering video in the terminal anyone? :) ). And that's just the example I can think of.

As an aside, the terminal panes aren't able to request a re-render (as far as I'm aware at least), they're just able to mark themselves to be re-rendered so the Screen can optimize who needs to be rerendered or not (even though that mechanism happens to be broken right now and they're all re-rendered every time a render is called).

I definitely understand what you're saying about eg. the clock widget. My preference here would be to have an API called on startup (or ideally have this in the manifest of the plugin somehow) which would ask Zellij to re-render the pane every X interval. Kind of "subscribe to periodic updates" sort of thing.

But really - since we're pre-beta, I think it's totally legit for the API to change quite a bit still. For sure this is an interesting discussion that I'm happy to have, but I wouldn't want it to hold up anyone's work.

TheLostLambda:
Well, I get where you are coming from, but I think a complete plugin system allows for an animation loop eventually. I'm now thinking that I would like to change the way this is done, but I'm still thinking that functions like `request_rerender` have a place in it. At the moment I could (try, even if it doesn't perform well) to play a video in a Zellij terminal pane – I really don't see why plugins shouldn't be capable of something that the terminal already is, especially if we want to eventually make them a plugin too.

I still don't quite get then how something like htop is rerendered if the terminal doesn't have a way to request that. I'm looking to make `request_rerender` behave in the same way as "marking them for rerender" eventually, but as you've pointed out, that's not really functional yet.

Unfortunately though, I've come to the realisation that any sort of polling system (including a re-render every N seconds) is going to be bad for the plugin system. It's simply too slow and too chatty. I think I'm going to try and build a proper event system next week.

What I'll likely do is replace `update(dt)` with `update(event)`, where `event` will be a new enum representing all possible plugin events. In the `load` function, the plugin can pick and choose which events to subscribe to. I'll delete both the `handle_key()` and `handle_global_key()` callbacks, as they'll be replaced by the new `update(event)` system. That leaves every plugin with exactly 3 methods to implement: `load()`, `update(event)`, and `draw(rows, cols)`.

Let me know what you think about that proposal. It should mean that 100% of the information that plugins get sent is relevant and that Zellij never wastes time on needless updates.

imsnif:
Everything sounds good in principle. I feel we're getting out of scope here though. Maybe if you want to discuss the design further we can open an issue/discussion for it? I just don't want this to go on here more than it needs to - this PR blocks other things, I think?

TheLostLambda:
Yeah, I suppose we could merge this now then? The only mess is dealing with the new theme. Should one of us update that then wave this through to be improved again soon?

TheLostLambda:
With that being said actually, if I'm going to be redesigning things radically in a week's time, I might actually just abandon this PR. What do you think? Obviously `main` still works at the moment.

imsnif:
But then what about https://github.com/zellij-org/zellij/pull/212?

TheLostLambda:
Well, if #212 builds on this PR, it will need to be totally rewritten in a week's time, and if it builds on `main`, it will also need to be rewritten in a week's time. We might as well merge #212 so we have that functionality in the short term. The alternative is to just push it back and wait for a better API design, but that's next Tuesday at the earliest.

I think the only thing blocking #212 was if this API was better, but I'm not sure I can say it is.

imsnif:
My vote goes to merging #212 now then, in order not to keep its author waiting. :)

TheLostLambda:
Okay, there is a merge conflict from your recent update to the tab styling. Could I bother you to help resolve that? I'm unfortunately quite busy today :cry: In the meantime I'll close this PR as a bit of a failed experiment!

henil:
I am doing unused branch cleanup so before deleting this, @TheLostLambda just wanted to confirm is it safe to delete this stale branch.

TheLostLambda:
Yeah, should be good to delete! A better system was merged

PR #219, opened 2021-03-09
Rewriting the plugin API to be more general

TheLostLambda:
This will take a little while to finish, but we can keep track of things here:

- [x] Change `init()` to `load()`
- [x] Deduplicate structs that are in both `zellij` and `zellij-tile`
- [x] WASM thread keeps a mapping of plugins and event subscriptions
- [x] Add host-functions for subscribing and unsubscribing from events
- [x] Introduce the `update(event)` callback into the API (which is called to handle plugin events)
- [x] Make mode changes send an event to the WASM thread
- [x] Murder the `AppInstruction::Get/SetState` calls (and any global state)
- [x] Murder the `get_help()` host function
- [x] Add an event system for key-presses
- [x] Add an event for updating tabs
- [x] Add support for renaming tabs
- [x] Murder every callback other than `load()`, `update(event)`, and `draw(rows, cols)`
- [x] Murder every WASM thread event other than Load, Draw, and Update
- [x] Call render after a plugin is updated
- [x] Switch back to upstream `termion`
- [x] Probably more... (loads of lint fixing)
- [x] Don't break everything?

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -27,7 +27,7 @@ signal-hook = "0.1.10"
 strip-ansi-escapes = "0.1.0"
 structopt = "0.3"
 # termion = { git = "https://gitlab.com/TheLostLambda/termion.git", version = "1.6.0", features = ["serde"] }
-termion = { package = "termion_temporary_zellij_fork", version = "1.6.0", features = ["serde"]}
+termion = "1.5.0"
 termios = "0.3"
 unicode-truncate = "0.2.0"
 unicode-width = "0.1.8"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -27,7 +27,6 @@ serde_yaml = "0.8"
 signal-hook = "0.1.10"
 strip-ansi-escapes = "0.1.0"
 structopt = "0.3"
-# termion = { git = "https://gitlab.com/TheLostLambda/termion.git", version = "1.6.0", features = ["serde"] }
 termion = "1.5.0"
 termios = "0.3"
 unicode-truncate = "0.2.0"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -27,17 +27,17 @@ signal-hook = "0.1.10"
 strip-ansi-escapes = "0.1.0"
 structopt = "0.3"
 # termion = { git = "https://gitlab.com/TheLostLambda/termion.git", version = "1.6.0", features = ["serde"] }
-termion = { package = "termion_temporary_zellij_fork" , version = "1.6.0", features = ["serde"]}
+termion = { package = "termion_temporary_zellij_fork", version = "1.6.0", features = ["serde"]}
 termios = "0.3"
 unicode-truncate = "0.2.0"
 unicode-width = "0.1.8"
 vte = "0.8.0"
 strum = "0.20.0"
-strum_macros = "0.20.0"
 lazy_static = "1.4.0"
 wasmer = "1.0.0"
 wasmer-wasi = "1.0.0"
 interprocess = "1.0.1"
+zellij-tile = { path = "zellij-tile/", version = "0.5.0" }
 
 [dependencies.async-std]
 version = "1.3.0"
--- a/default-tiles/status-bar/src/first_line.rs
+++ b/default-tiles/status-bar/src/first_line.rs
@@ -83,7 +83,11 @@ fn unselected_mode_shortcut(letter: char, text: &str) -> LinePart {
         .on(BRIGHT_GRAY)
         .bold()
         .paint(">");
-    let styled_text = Style::new().fg(BLACK).on(BRIGHT_GRAY).bold().paint(text);
+    let styled_text = Style::new()
+        .fg(BLACK)
+        .on(BRIGHT_GRAY)
+        .bold()
+        .paint(format!("{} ", text));
     let suffix_separator = Style::new().fg(BRIGHT_GRAY).on(GRAY).paint(ARROW_SEPARATOR);
     LinePart {
         part: ANSIStrings(&[
@@ -109,7 +113,11 @@ fn selected_mode_shortcut(letter: char, text: &str) -> LinePart {
         .bold()
         .paint(letter.to_string());
     let char_right_separator = Style::new().bold().fg(BLACK).on(GREEN).bold().paint(">");
-    let styled_text = Style::new().fg(BLACK).on(GREEN).bold().paint(text);
+    let styled_text = Style::new()
+        .fg(BLACK)
+        .on(GREEN)
+        .bold()
+        .paint(format!("{} ", text));
     let suffix_separator = Style::new().fg(GREEN).on(GRAY).paint(ARROW_SEPARATOR);
     LinePart {
         part: ANSIStrings(&[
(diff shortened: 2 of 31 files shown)


Cargo.toml
```diff
@@ -28,17 +28,17 @@ signal-hook = "0.1.10"
 strip-ansi-escapes = "0.1.0"
 structopt = "0.3"
 # termion = { git = "https://gitlab.com/TheLostLambda/termion.git", version = "1.6.0", features = ["serde"] }
-termion = { package = "termion_temporary_zellij_fork" , version = "1.6.0", features = ["serde"]}
+termion = "1.5.0"
```
imsnif, line 31: Woohoo! Why do we need the comment still?

default-tiles/status-bar/src/first_line.rs
```diff
@@ -70,90 +70,64 @@ fn unselected_mode_shortcut(letter: char, text: &str) -> LinePart {
...
         .on(BRIGHT_GRAY)
         .bold()
-        .paint(format!(">"));
-    let styled_text = Style::new()
-        .fg(BLACK)
-        .on(BRIGHT_GRAY)
-        .bold()
-        .paint(format!("{} ", text));
+        .paint(">");
+    let styled_text = Style::new().fg(BLACK).on(BRIGHT_GRAY).bold().paint(text);
```
imsnif, line 86: Didn't we drop a padding character here?

default-tiles/status-bar/src/first_line.rs
```diff
@@ -162,7 +136,7 @@ fn disabled_mode_shortcut(text: &str) -> LinePart {
 }
 
 fn selected_mode_shortcut_single_letter(letter: char) -> LinePart {
-    let char_shortcut_text = format!(" {} ", letter);
+    let char_shortcut_text = letter.to_string();
```
imsnif, line 139: Didn't we drop a padding character here too?

default-tiles/status-bar/src/first_line.rs
```diff
@@ -173,16 +147,13 @@ fn selected_mode_shortcut_single_letter(letter: char) -> LinePart {
...
-            ANSIStrings(&[prefix_separator, char_shortcut, suffix_separator])
-        ),
+        part: ANSIStrings(&[prefix_separator, char_shortcut, suffix_separator]).to_string(),
         len,
     }
 }
 
 fn unselected_mode_shortcut_single_letter(letter: char) -> LinePart {
-    let char_shortcut_text = format!(" {} ", letter);
+    let char_shortcut_text = letter.to_string();
```
imsnif, line 156: And here?

default-tiles/status-bar/src/main.rs
```diff
@@ -2,7 +2,7 @@ mod first_line;
 mod second_line;
 
 use std::fmt::{Display, Error, Formatter};
-use zellij_tile::*;
+use zellij_tile::prelude::*;
```
imsnif, line 5: Personally I'm not a fan of wildcard imports. It makes it harder to figure out "where this thing is coming from", but that's just my opinion. :)

default-tiles/status-bar/src/main.rs
```diff
@@ -40,19 +42,25 @@ impl Display for LinePart {
...
         set_selectable(false);
         set_invisible_borders(true);
         set_max_height(2);
+        subscribe(&[EventType::ModeUpdate]);
+    }
+
+    fn update(&mut self, event: Event) {
+        if let Event::ModeUpdate(mode_info) = event {
+            self.mode_info = mode_info;
+        }
```
imsnif, line 55: Very cool :)

src/client/tab.rs
```diff
@@ -268,11 +258,7 @@ impl Tab {
             if let Some(plugin) = &layout.plugin {
                 let (pid_tx, pid_rx) = channel();
                 self.send_plugin_instructions
-                    .send(PluginInstruction::Load(
-                        pid_tx,
-                        plugin.clone(),
-                        layout.events.clone(),
-                    ))
+                    .send(PluginInstruction::Load(pid_tx, plugin.clone()))
```
imsnif, line 261: Tbh, I don't like this approach.
It seems like we're trying to force a shared-state paradigm onto a distributed architecture. Instead of sending a channel over another channel and blocking until we receive a message back on it, can't we do something similar to what we're doing with terminals? We have the same problem there. There we load all of them beforehand. Do you think something like this can work with plugins too?

src/client/tab.rs
```diff
@@ -424,47 +406,44 @@ impl Tab {
...
-                    return;
-                }
-                let terminal_ws = PositionAndSize {
-                    x: active_pane.x(),
-                    y: active_pane.y(),
-                    rows: active_pane.rows(),
-                    columns: active_pane.columns(),
-                };
-                let (top_winsize, bottom_winsize) = split_horizontally_with_gap(&terminal_ws);
+        } else if let PaneId::Terminal(term_pid) = pid {
```
imsnif, line 409: Nice catch :)

imsnif (commented):
Really cool change! Things look so much cleaner now. Stellar work :)
I left some comments in the source. As far as I'm concerned there's no need to do more back-and-forth on this (unless you have questions or concerns, or want to discuss specific comments) - as soon as you feel this is ready, go ahead and merge.

Cargo.toml
TheLostLambda: Whoops! We definitely don't! I just missed that! Gone in the next commit :)

default-tiles/status-bar/src/first_line.rs
TheLostLambda: Good catch on these! Sorry! I was getting a bit overzealous here :sweat_smile:

imsnif: Cool :) I forgot to mention that I didn't run this, I just commented on the code itself. So if you could run it to make sure there's no visual change (just to be sure I caught everything) that would be great :)

default-tiles/status-bar/src/main.rs
TheLostLambda: I agree as a general rule and am using specific imports within Zellij, but I figured that it would be nice to have a single import for plugins. I changed the wild-card import to a `prelude` to follow the "convention" for wildcard imports (as opposed to just having the earlier `zellij_tile::*` wildcard).

I'll probably keep this as is for now, but when you are writing things, feel free to import from specific modules! :)

src/client/tab.rs
TheLostLambda: I think the only tricky bit about this is knowing what "pid" to allocate. At the moment, only the plugin system is aware of how many plugins are loaded. I'm slightly hesitant to duplicate that state, but I'm open to thinking that through some more?

I've been using this system since the initial implementation of the plugin system without too many issues, but I certainly agree that it's not ideal. Are you suggesting that I allocate a PID in `tab.rs` then send that in the loading message?

src/client/tab.rs
TheLostLambda: I didn't know this was even possible! Mixing `if` with `if let`! I'm just following the orders of commander Clippy :paperclip: :heart:

default-tiles/status-bar/src/first_line.rs
TheLostLambda: Strangely, I tested it quite a bit without noticing! I've done a side-by-side with `main` now though and it's looking that same!

TheLostLambda:
Thanks for the review @imsnif ! Good catch on those `format!()` removals!

I'll probably merge this once CI finishes and maybe we could continue chatting about the `.send(PluginInstruction::Load(pid_tx, plugin.clone()))` line going forwards? That might also need to get more complex once we introduce "global" plugins (so that the tab and status bars only exist once and not for every tab).

src/client/tab.rs
imsnif: Not exactly. I'm suggesting we do the same thing we do with terminals. We spawn all of them before we get to this function, and then pass their pids. What happens (roughly) is: when we want to open a new tab, we send a `NewTab` instruction to the pty thread. The pty thread counts all the terminals that need to be spawned, spawns them and then sends the `ApplyLayout` instruction to screen already including the terminal pids. I'm suggesting we add another link to this chain (either before or after the pty thread) where we go through the plugin thread, spawn the necessary plugins, tack on their pids to the instruction and forward everything to screen.
So that in the end, this function would get `new_terminal_pids` and `new_plugin_pids` parameters. Makes sense?

TheLostLambda: Ah, okay, I see what you mean now! I'll take a look at how I might be able to make that happen!

PR #222, opened 2021-03-12
fix(performance): smaller allocations when rendering

imsnif:
This fixes a performance bottleneck that we had when rendering terminals. It reduces the size and amount of allocations performed when converting the state of the terminal panes to stringified ANSI. For me this improves the app's performance significantly.


a-kenji:
This is amazing, significantly is definitely the right word!
Even new tabs are being created fairly quickly on my end, even with the current limitations.

PR #223, opened 2021-03-12
Psuedo Client-Server model

kunalmohan:
Isolate pty-thread to imitate an actual client-server multiprocess model.


imsnif (changes requested):
Hey @kunalmohan - first off: this is fantastic. It's very clear you've put a lot of work and thought into this. It's not an easy task by any means!

Because this is a very big change, I think we'll have to iterate on it a little bit. So I'm going to start with a few major points:

1. The messaging model: I think the client doesn't need to have to know about which parts the server includes. I would much rather the messages look something like: `ServerInstruction::KeyInput` than `ServerInstruction::ToPty(PtyInstruction::KeyInput)`. I think if we flatten the messages in this way and deal with their routing in the router thread, things will become simpler.
2. I really like your solution with the `ClientOsApi` methods being sent to the server when needed, but I think it would be clearer if we extract those messages to a more generic ServerApi. Instead of things like "set_terminal_size_using_fd", we can have "ServerInstruction::ResizeTerminal" (or something of the sort) and then have the server interact with the OS as needed. The client doesn't need to know this call has to interact with the OS, that's the server's job. Makes sense?
3. On the subject of naming, I would rename "notify_server" to "connect_to_server" or "establish_server_connection" or something of that sort
4. I found a small bug: in most cases, ctrl-c doesn't seem to work. I found this when I ran "cargo test" inside Zellij and then did ctrl-c. I think we need to listen to signals on the client side and then send them to the server.
5. As I understand things, plugins now run on the client. Is this correct? My intuition tells me that they need to run on the server, but I'm not 100% sure about this... @TheLostLambda - do you have any thoughts on this?
6. Do you think it would be possible to move everything in the `start` function either to `start_client` or `start_server`? I think it will make things clearer.
7. Finally, while I was expecting this to happen, I think performance here was hurt quite a bit. This makes sense, because now we're passing through IPC rather than passing messages in memory, but we have to address this (or at least improve it a little bit). We need to do some profiling to understand where this is coming from. The way I saw this is just to run a `dmesg` inside a pane with this change and before it. We might have some luck here if we batch some messages on the ipc channel if there's a lot of data flowing. Maybe something similar to what we're doing here: https://github.com/zellij-org/zellij/blob/main/src/common/pty_bus.rs#L206

Looking forward to hearing your thoughts.

TheLostLambda:
Heya! Hopefully I'll have the time to give this a proper look soon, but to address the direct question, yeah, I think that plugins will need to be run on the server-side (so that they can survive a detach).

It would be unfortunate if something like Strider reset its path every time you detached the session. Additionally, at the moment, loading plugins is massively expensive, so it's good to avoid reloading them whenever possible.

imsnif:
> I think that plugins will need to be run on the server-side (so that they can survive a detach).

That makes sense. How about things like the status bar though? Do you think we can find a solution to somehow render plugins differently depending on the client requesting them? Or some other solution? (not for now, just something to think about in this critical architecture stage :) ).

src/common/mod.rs
```diff
@@ -170,36 +199,18 @@ pub fn start(mut os_input: Box<dyn OsApi>, opts: CliArgs) {
...
     let (send_plugin_instructions, receive_plugin_instructions): ChannelWithContext<
         PluginInstruction,
     > = mpsc::channel();
     let send_plugin_instructions =
         SenderWithContext::new(err_ctx, SenderType::Sender(send_plugin_instructions));
 
     let (send_app_instructions, receive_app_instructions): SyncChannelWithContext<AppInstruction> =
-        mpsc::sync_channel(0);
-    let send_app_instructions =
+        mpsc::sync_channel(500);
```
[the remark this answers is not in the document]
kunalmohan: A 0-sized buffer means that this is a rendezvous channel. Earlier, the number of messages and thus the traffic on this channel was quite low, so things worked. But now that every communication between server and client goes through this channel, we might want this buffer size to be non-zero. The 500 value is completely random so we can definitely reduce it.

src/common/os_input_output.rs
```diff
@@ -156,24 +165,53 @@ fn spawn_terminal(file_to_open: Option<PathBuf>, orig_termios: termios::Termios)
...
+    fn send(&mut self, msg: T) -> ipmpsc::Result<()> {
+        self.sender.send(&(msg, self.err_ctx))
+    }
+}
+
+#[derive(Clone)]
+pub struct ServerOsInputOutput {
     orig_termios: Arc<Mutex<termios::Termios>>,
+    server_sender: IpcSenderWithContext<ServerInstruction>,
+    server_receiver: Arc<Mutex<IpcReceiver>>,
```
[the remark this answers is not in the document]
kunalmohan: I was in quite a dilemma on how to store this. We require the clones of `os_input` across threads. I had two choices-
1. Have an `Arc<Mutex<IpcReceiver>>`. This enables `os_input` to be cloned and since the receiver is ideally used by a single thread (`server_router`), the lock resolves immediately every time.
2. Have an `Option<Receiver>`. Here the original `os_input` returned by `get_server_os_input()` would contain `Some(receiver)` and a custom `Clone` implementation would make all the clones created store `None` there. The original `os_input` would then be owned by the `server_router` thread.

I don't have a strong opinion about either. If 2 sounds like a better solution, I can change it.

kunalmohan:
@imsnif Thanks for the detailed review! Regarding your concerns:

1. I had thought about it and the problem I see with flattening is that the `ServerInstruction` enum will become too big and so will match statement in the router thread. Also if one wants to add a new variant in say `PtyInstruction`, they will have to add it in `ServerInstruction` separately and then in the respective `XxxContext` enums as well.
Also, it was foolish of me to not do actual routing inside the router thread itself and simply forward all messages to the main thread. I'll make the change.
2. Makes sense. Will make the changes.
3. Agreed.
4. I not sure I completely understand the bug. Could you explain a bit more? What were you expecting and what did you observe?
5. Yep plugins run on the client-side for now. Earlier, I wasn't sure about which side they should run on. I'll shift them to the server-side now.
6. It should definitely be possible. I'll make the change.
7. Yeah I was expecting performance to get worse and we definitely need to find the bottleneck here. Profiling is a good idea, but I'll need some help with that. I'm not sure how we would batch messages in this case, maybe you could give an idea that?

I have another idea that just came to my mind. Since the Ipc stuff is handled by `os_input` now and we have an `os_input` in each thread, how about we send messages from a thread on the client-side directly to the server_router thread. Messages won't have to go through the client main thread and we will still be using a single Ipc channel. Similar would apply to the server-side threads as well.

imsnif:
1. I understand the trade-off now. Do you think there's maybe a way we can make these more readable with helper functions or macros? `send_resize_message_to_server!()` or something? (ideally a nicer scheme that won't be so verbose :) ).
2. Do you mean remove `ClientOsApiInstruction` completely? If so, I don't think so because the client (at least this terminal client) still has to interact with the OS in some places (eg. `set_raw_mode`). I'm just suggesting to remove the things out of it where it needs to tell the server to interact with the OS, and instead just send API calls to the server. Right now it's `ClientOsApi` => IPC => `ServerOsApi`. I'm suggesting it would be clearer if it's `ServerInstruction::<something>` (eg. resize pty) => IPC => `ServerOsApi`. The client doesn't need to know that this is part of the OS api, it just tells the server to do its thing. Makes sense?
4. The scenario: run `cargo test`, before it's done do `Ctrl-c`. Outside of Zellij, this stops the tests before they are complete and brings you back to the prompt. Inside Zellij, this doesn't work. The `Ctrl-c` is ignored. I haven't tried, but I think this is also the case with other things you can break (eg. doing `sleep 10` and then doing `Ctrl-c` to stop it before the 10 seconds are over).
6. I like your idea of placing both `start_client` and `start_server` inside the `start` function. Then later we can move them depending on what we want to do with them.
7. That's a little tricky. One has to follow the messages from the pty thread until the client and see where the bottleneck is. We might have to get a little creative in certain places. Before that though, my guess is that the IPC messages seem to be going through a synchronous channel - maybe that's what's causing the delay? I know the `AppInstructions` channel needs to be synchronous because of the panic_hook, but maybe we don't have to pass all the other messages through it? I think a good place to start would be to see if we can pass messages between client/server asynchronously (even by just commenting out the panic hook) and seeing if it improves performance.

kunalmohan:
1. I think we can introduce helper methods to `ServerInstruction` enum for building nested messages.
    Eg: We can have 
    ```rust 
    os_input.send_to_server(ServerInstruction::spawn_terminal(None));
    ```
    where spawn_terminal function would be
    ```rust
    impl ServerInstruction {
        fn spawn_terminal(path: Option<PathBuf>) -> Self {
            Self::ToPty(PtyInstruction::SpawnTerminal(path))
        }
    }
    ```
2. I was talking about https://github.com/zellij-org/zellij/blob/23e194cb311b7ff3e58b07e480e7ab3030839bd9/src/common/os_input_output.rs#L310-L315
All the methods inside `ClientOsApi` are executed by the client only. Any os interaction that client wants to do but is unable to do so is sent to server via 
    ```rust 
    os_input.send_to_server(ServerInstruction::OsApi(ServerOsApiInstruction::xxx(..)));
    ``` 
    So here we can either have helper methods under `ServerInstruction` (like I mentioned above) or I can remove the `ServerOsApiInstruction` enum and add those instructions directly under `ServerInstruction` enum.
3. That sounds odd. `ctrl+c` seems to be working fine for me both inside and outside zellij. And its surprising that you're facing that. I've barely touched the input handler in this PR.
4. Cool. I'll make the `start_server` and `start_client` changes.
5. That looks like a good idea. I'll do the profiling once other points are addressed satisfactorily.

src/server/mod.rs
```diff
@@ -1,5 +1,387 @@
...
+    DoneClosingPane,
+    ClosePluginPane(u32),
+    ClientExit,
+    Exit,
+}
+impl ServerInstruction {
+    // ToPty
+    pub fn spawn_terminal(path: Option<PathBuf>) -> Self {
+        Self::ToPty(PtyInstruction::SpawnTerminal(path))
+    }
```
kunalmohan, line 37: @khs26, as discussed in the last meeting, this is the `ServerInstruction` enum which the client uses to send messages to the server. The nested enums (`PtyInstruction` and `ScreenInstruction`) made the `send()` method calls too verbose, so we have defined these helper methods for now.

We have an option to flatten the `ServerInstruction` enum so every `ScreenInstruction` and `PtyInstruction` would be directly defined used the `ScreenInstruction` enum as well. IMO this will be a better option once we have macros for generating contexts from the Instructions and the screen and plugin threads are moved to the server-side (I'm working on this right now).

So if you have any other ideas, let us know.

khs26: Oh, and maybe also implement `From<ServerInstruction>` and `From<PtyInstruction>` for `ServerInstruction` actually does the job when we're having to manipulate these before serializing/sending. :) I should have maybe thought of that first...

khs26: I pushed a change with that in - feel free to discard it if that doesn't help in any way

khs26:
When you're happy to have changes, I'd also like to modularise the huge `start` function, that you moved over to become `start_server`

src/server/mod.rs
khs26: Actually, reading the thread a bit more, I think I got that the problem is just that we don't want to have `ScreenInstruction` and `PtyInstruction` and friends on the client side, is that right? (also, oops...)

If so, I think the main issue we have to solve is defining what we want the client-side instructions to be.

We'd then similarly serialize those messages and not worry about what the server has to do on its side to handle them. My instinct would be to have `input_handler` on the client side just serialize `Action` and ship it over to the server. @imsnif, do you think that makes sense?

The implication of all this would be that the client side "owns":
  - Hotkey definitions
  - Input mode
  - A copy of whatever data they need from the server side to render a screen

kunalmohan:
> Oh, and maybe also implement `From<ServerInstruction>` and `From<PtyInstruction>` for `ServerInstruction` actually does the job when we're having to manipulate these before serializing/sending. :) I should have maybe thought of that first...

I don't think this is gonna help much because, in places where a message originates, we'll still have to write the entire `send()` method something like this:
```rust
os_input.send_to_server(ServerInstruction::from(PtyInstruction::SpawnTerminal(None)));
```
which is not very different from writing the complete message. So, I'll probably have to discard your commit.

I really liked your idea of serializing the `Action` enum and shipping it over to the server. I really think that could work.


> When you're happy to have changes, I'd also like to modularise the huge `start` function, that you moved over to become `start_server`

Sure! Actually more things have moved to the `start-server()` function now (screen and wasm thread as well). I just need to resolve some conflicts before I push those changes. I'll try to push them by tomorrow.

khs26:
> > Oh, and maybe also implement `From<ServerInstruction>` and `From<PtyInstruction>` for `ServerInstruction` actually does the job when we're having to manipulate these before serializing/sending. :) I should have maybe thought of that first...
> 
> I don't think this is gonna help much because, in places where a message originates, we'll still have to write the entire `send()` method something like this:
> 
> ```rust
> os_input.send_to_server(ServerInstruction::from(PtyInstruction::SpawnTerminal(None)));
> ```
> 
> which is not very different from writing the complete message. So, I'll probably have to discard your commit.
Yeah, that's sensible. I wrote that stuff before I understood what the problem was properly. :)

> I really liked your idea of serializing the `Action` enum and shipping it over to the server. I really think that could work.
On this front, I think we might want to still have `handle_action` be on the client side. Since we might have client-only actions (like connecting and disconnecting from the server). However, we could still implement `Serialize` for them and just define which subset get sent to/from the server.

@imsnif, @qballer, do you guys have any thoughts on whether we should have a dedicated enum for messages to the server, or would you just reuse the `Action` enum we already have?

> > When you're happy to have changes, I'd also like to modularise the huge `start` function, that you moved over to become `start_server`
> 
> Sure! Actually more things have moved to the `start-server()` function now (screen and wasm thread as well). I just need to resolve some conflicts before I push those changes. I'll try to push them by tomorrow.
Cool. No rush - I think you're doing much more important things than my addressing this minor gripe! :)

imsnif:
> @imsnif, @qballer, do you guys have any thoughts on whether we should have a dedicated enum for messages to the server, or would you just reuse the `Action` enum we already have?

Personally my vote goes to having a dedicated enum for clarity - but I'm a little out of context on this change so maybe I'm missing something.

kunalmohan:
How about we have a `ServerInstruction` enum that looks like
```rust
pub enum ServerInstruction {
    OpenFile(PathBuf),
    SplitHorizontally,
    SplitVertically,
    MoveFocus,
    // The above 4 will be removed when the respective options are removed from CliArgs
    NewClient(String, PositionAndSize),
    Action(Action),
    Render(Option<String>),    // If None, client should exit. Possible when user closes the last tab or pane.
    PluginUpdate(Option<u32>, Event),
    DoneClosingPane,
    DoneOpeningNewPane,
    DoneUpdatingTabs,
    ClientExit,
    // notify router thread to exit
    Exit,
}
```
The `dispatch_action` method in handler.rs would simply forward the action to the server and optionally block the input thread for atomicity. 
I also think that it may be better to have a single `CondVar` in `command_is_executing` for now, which will be used for blocking and unblocking the input thread. We currently have 3 `CondVar`s, each does the same thing, just with a different name. This would also reduce the three `Donexxx` variants in `ServerInstruction` to just one `UnblockInputThread`.

@imsnif @khs26 What do you guys think?

imsnif:
> @imsnif @khs26 What do you guys think?

@kunalmohan - this looks good to me. I think for simplicity's sake we can go with your idea of the 1 `CondVar`. I'm not sure why I resisted it before, it makes total sense when I think about it. :)

I want to make this a priority (I think this is the number 1 feature people request). So if there's anything I can do to make this easier or provide info (or anything) let me know!

imsnif:
@kunalmohan - I went over this again and this looks great! Very thorough job and top-notch work as always :)

The performance also looks quite good! I didn't even notice a difference from main. Do you think this will change once we move to a 2 process model? As far as I can tell already at this point there is absolutely no shared memory or state between client/server even now, right?

I didn't go over all of the code again, but here are a few things that came to mind when running through it:
1. Is the ipc-pipe secure? Are the messages there encrypted somehow? This is very important because we don't want malicious users to be able to listen to the ipc pipe and deserialize passwords and the like.
2. Continuing from point 1, it might be a good idea to move the ipc pipe files out of the temporary folder and into the user's home folder for extra protection. Later, when we separate to different processes we can give user's a cli option to place the ipc pipe wherever they want.
3. I was a little confused by the naming of the senders/receivers. To make it explicit, do you think we can change them to be: `send_instructions_to_server` / `send_instructions_to_client` / `receive_instructions_from_server` / `receive_instructions_from_client`?
4. Likewise, the server thread should probably be named `server_thread` and not `ipc_thread`.

After these points, how would you like to continue with this?

kunalmohan:
> Do you think this will change once we move to a 2 process model? As far as I can tell already at this point there is absolutely no shared memory or state between client/server even now, right?

I don't think the performance would degrade in an actual 2 process model. There is no shared state between the client and the server now and we are already using IPC here which will remain as is when moving to an actual 2 process model so message delivery speed should remain the same.

> 1. Is the ipc-pipe secure? Are the messages there encrypted somehow? This is very important because we don't want malicious users to be able to listen to the ipc pipe and deserialize passwords and the like.

The ipc-pipe uses Unix Domain sockets under the hood which limits access to sockets via file system permissions. There is no additional security there. And the messages are simply serialized (from Rust types to byte vector) and deserialized (from byte vector to Rust types) by bincode. So IMO there is a possibility that if the intruder has access to the bytes vector, they may be able to deserialize it using bincode if they know the type of message (the code is open source so they definitely know the type). We should maybe consider using a separate encryption library ([openssl](https://docs.rs/openssl/0.10.34/openssl/index.html) maybe). I should point out that I have zero experience in the domain of security of applications, so maybe I'm missing something. Any advice/suggestions are highly appreciated.

> 2. Continuing from point 1, it might be a good idea to move the ipc pipe files out of the temporary folder and into the user's home folder for extra protection. Later, when we separate to different processes we can give user's a cli option to place the ipc pipe wherever they want.

Sure, I'll do that. Just to be clear, exactly under which directory would it reside `XDG_CACHE_HOME` or `XDG_RUNTIME_DIR` or `XDG_DATA_DIR`?

> 3. I was a little confused by the naming of the senders/receivers. To make it explicit, do you think we can change them to be: send_instructions_to_server / send_instructions_to_client / receive_instructions_from_server / receive_instructions_from_client?
> 4. Likewise, the server thread should probably be named server_thread and not ipc_thread.

I'll most definitely change the names to be more clear.

> After these points, how would you like to continue with this?

Once you think that this PR is in a usable state, I'll certainly want to merge it first because the changes are quite big and with all the stuff moving around, merge conflicts are easily introduced and are difficult to address in a big PR. Once this is done/merged, the next task would be to actually have two processes. It will also be beneficial to separate the client, the server and the common stuff into 3 separate crates and have a cargo workspace for it. Then we can move to handle multiple clients by a single server. But I don't have a strong opinion on this. If you have a better approach or anticipate any other challenges that I may have overlooked, please feel free to bring that up so that the transition to the client-server model is as smooth as possible.

imsnif:
> The ipc-pipe uses Unix Domain sockets under the hood which limits access to sockets via file system permissions. There is no additional security there. And the messages are simply serialized (from Rust types to byte vector) and deserialized (from byte vector to Rust types) by bincode. So IMO there is a possibility that if the intruder has access to the bytes vector, they may be able to deserialize it using bincode if they know the type of message (the code is open source so they definitely know the type). We should maybe consider using a separate encryption library ([openssl](https://docs.rs/openssl/0.10.34/openssl/index.html) maybe). I should point out that I have zero experience in the domain of security of applications, so maybe I'm missing something. Any advice/suggestions are highly appreciated.

I think we have to encrypt them then :) It's important that this happens before we merge. Hopefully this will not hurt performance too much...

> Sure, I'll do that. Just to be clear, exactly under which directory would it reside `XDG_CACHE_HOME` or `XDG_RUNTIME_DIR` or `XDG_DATA_DIR`?

I'm not sure what's the best one tbh... would you like to do some research on it and decide?

> Once you think that this PR is in a usable state, I'll certainly want to merge it first because the changes are quite big and with all the stuff moving around, merge conflicts are easily introduced and are difficult to address in a big PR. Once this is done/merged, the next task would be to actually have two processes. It will also be beneficial to separate the client, the server and the common stuff into 3 separate crates and have a cargo workspace for it. Then we can move to handle multiple clients by a single server. But I don't have a strong opinion on this. If you have a better approach or anticipate any other challenges that I may have overlooked, please feel free to bring that up so that the transition to the client-server model is as smooth as possible.

Sounds good! I would recommend that the next task after merging would be to allow users to detach and re-attach to the session - even without having multiple users in one session. We can simply have the existing user disconnected if another user joins for now.

kunalmohan:
> I think we have to encrypt them then :) It's important that this happens before we merge. Hopefully this will not hurt performance too much...

Cool. I'll look into it.

> I'm not sure what's the best one tbh... would you like to do some research on it and decide?

Sure :)

> I would recommend that the next task after merging would be to allow users to detach and re-attach to the session - even without having multiple users in one session. We can simply have the existing user disconnected if another user joins for now.

The only problem with that is we would not have a central registry to see which sessions are running. We'll have to traverse the zellij runtime directories (that contain the ipc socket files) to see which sessions are running and then connect to the specified session. The sessions will have to be named and the socket file names will have to be something like `zellij_<session_name>` to be able to identify the socket file corresponding to a particular session and connect to it.

imsnif:
> The only problem with that is we would not have a central registry to see which sessions are running. We'll have to traverse the zellij runtime directories (that contain the ipc socket files) to see which sessions are running and then connect to the specified session. The sessions will have to be named and the socket file names will have to be something like `zellij_<session_name>` to be able to identify the socket file corresponding to a particular session and connect to it.

How about we use https://crates.io/crates/names to generate session names (I think this is kind of similar to what docker does for its container names). Then when you attach a session you can choose one of those names? I think it'll be friendlier than uuids. We can either do that in a cool terminal menu or by doing a --list-sessions and copy/pasting? Zellij can as you say discover them by looking through the folder. What say you?

kunalmohan:
Sounds good! The generated names will be a lot better than Uuids. I think having a --list-sessions flag should be enough to start with. We can consider terminal menus later. Also, we can add the functionality of renaming sessions later when we move to a single server model (because renaming won't be possible with this model). We can certainly add an option to name the session when starting a new one. If a name is not supplied, we can generate a name using the `names` crate.

imsnif:
Awesome. Let's do this :)

TheLostLambda:
> Sure, I'll do that. Just to be clear, exactly under which directory would it reside XDG_CACHE_HOME or XDG_RUNTIME_DIR or XDG_DATA_DIR?

I think `XDG_RUNTIME_DIR` is the place for this! :)

kunalmohan:
Umm...`runtime_dir` for MacOs and Windows is not defined: https://docs.rs/directories-next/2.0.0/directories_next/struct.ProjectDirs.html#method.runtime_dir

floriankisser:
> > The ipc-pipe uses Unix Domain sockets under the hood which limits access to sockets via file system permissions. There is no additional security there. And the messages are simply serialized (from Rust types to byte vector) and deserialized (from byte vector to Rust types) by bincode. So IMO there is a possibility that if the intruder has access to the bytes vector, they may be able to deserialize it using bincode if they know the type of message (the code is open source so they definitely know the type). We should maybe consider using a separate encryption library ([openssl](https://docs.rs/openssl/0.10.34/openssl/index.html) maybe). I should point out that I have zero experience in the domain of security of applications, so maybe I'm missing something. Any advice/suggestions are highly appreciated.
> 
> I think we have to encrypt them then :) It's important that this happens before we merge. Hopefully this will not hurt performance too much...

Hi there! This thread somehow caught my attention. If access to the socket is restricted to the user exclusively (should be the case under `XDG_RUNTIME_DIR`), then security is en par with tmux and screen already. I'm curious about which threats you want to protect against and how to achieve it?

I can think of two attack vectors here:
* root would be able to gain access to the socket, but he's also able to eavesdrop on the user on several other levels
* there might be rouge processes of the same user gaining access, if they do not run in a sandbox like e.g. flatpak apps do

Some protection against these own unsandboxed processes can be achieved by encrypting connections to the socket by:
* adding TLS, but this also needs a CA out of reach of the user, to prevent MITM attacks (one could just move the socket and create a new one in place, forwarding connections; without proper certificates you won't be able to detect this)
* adding symmetric password encryption, but when enabling this by default the user either has to always set a password when creating a session, or be presented with a generated one, that then needs to be saved somewhere else by the user to be able to reattach to the session

Please let me know if I've completely missed something ^^

imsnif:
Thanks for weighing in, @floriankisser ! This made me look a little deeper into it, specifically into how tmux is doing it, and I think you were right and I was being too cautious. Tmux places the sockets in `/tmp/tmux-1000` and afaict does not use encryption.

My thoughts were indeed a rogue process from the user, but as you say - that process could also follow the user's actions much easily through their shell history for example. So I'm good with not encrypting these as long as place them in the user's home folder somewhere, @kunalmohan.

imsnif:
> Umm...`runtime_dir` for MacOs and Windows is not defined: https://docs.rs/directories-next/2.0.0/directories_next/struct.ProjectDirs.html#method.runtime_dir

Can we fall back to something else there? Maybe @qballer has an idea? (quick context for him: where would be a good place to put socket files to communicate between client/server on a mac somewhere in the user's home folder?)

vikigenius:
Placing the socket somewhere in XDG_RUNTIME_DIR typically /run/user/$uid or /tmp/ seems like the ideal solution. A lot of other programs also place their sockets there and their permissions are basically 700 so that only the user and root can access those sockets.

kunalmohan:
> Placing the socket somewhere in XDG_RUNTIME_DIR typically /run/user/$uid or /tmp/ if it is not set, seems like the ideal solution.

The issue here is with MacOs since there is no `$XDG_RUNTIME_DIR` ([directories-next](https://docs.rs/directories-next/2.0.0/directories_next/struct.ProjectDirs.html#method.runtime_dir) crate). We don't want to keep it in `/tmp/` because it isn't accessible by multiple users (https://github.com/zellij-org/zellij/issues/417). 

> Can we fall back to something else there? Maybe @qballer has an idea?

I was thinking of falling back to the [cache_dir](https://docs.rs/directories-next/2.0.0/directories_next/struct.ProjectDirs.html#method.cache_dir), but I'll wait for the approval.

jovandeginste:
> [...] We don't want to keep it in `/tmp/` because it isn't accessible by multiple users (#417).

What do you mean exactly? I believe zellij should create a specific directory in `/tmp`, for that user. One way is to use the user's `uid` (which should be unique). Best practice (imo.) is to create a directory, eg. `/tmp/zellij-$(id -u)/` with modes `0700` and put the socket file in there. This gives the option to add more (tmp) files if need be, and otherwise keeps `/tmp` somewhat clean.

kunalmohan:
Ahh..sorry. My bad. I misread the linked issue initially. Falling back to a user-specific zellij folder in `/tmp/` sounds like a good solution to me. Thanks for pointing it out @jovandeginste!!

jovandeginste:
Just look at other similar tools, like `ssh-agent`, `tmux`, `vscode`. Their exact implementations vary, but they usually use the user's `uid`. Also, for fun and profit, make sure you run the exact same thing with 2 different users, to be sure it works :-)

jovandeginste:
Would you also allow overriding this through env/config? Eg. `ZELLIJ_SOCKET_DIR` and/or `ZELLIJ_SOCKET_FILE`. This to allow people in different setups to have more flexibility...

jovandeginste:
When looking at the changed code, I believe the whole `ZELLIJ_TMP_DIR` should be changed to include the user's `uid`. Not sure if you're keeping this for a separate PR?

kunalmohan:
> Would you also allow overriding this through env/config? Eg. ZELLIJ_SOCKET_DIR and/or ZELLIJ_SOCKET_FILE. This to allow people in different setups to have more flexibility...

Good Call!

> When looking at the changed code, I believe the whole ZELLIJ_TMP_DIR should be changed to include the user's uid.

Right, that should be the case.

qballer:
Hey @kunalmohan and @imsnif I can confirm that zellij is working on mac.

PR #226, opened 2021-03-13
delete characters when renaming tab

### LINKED ISSUE #221: Feature: tabbar - remove characters while renaming
Make it possible to remove characters from the input while renaming 
the tab-bar.

jcaplan:
handle delete and backspace keys when renaming tab.

closes #221


imsnif (approved):
Cool! Thanks for the quick fix on this.

PR #228, opened 2021-03-13
alt+key shortcut to navigate panes and autoescape

jcaplan:
switching panes is probably the most common action. I really like having a shortcut with as few key presses as possible.

It's somewhat arbitrary which actions should escape back, but basically stuff like scrolling or resizing where you reasonably want to make several adjustments should not auto escape. on the other side, stuff like creating a new pane or tab you probably only want to do once most of the time, so escape from those automatically.

this might be a matter of taste, so I made it configurable in the settings. depends on #224


a-kenji:
I feel the issues too. If you want to create this functionality for a keybinding that only makes sense to you you could also:

```
pane:
        - action: [ NewPane:, SwitchToMode: Normal,]
          key: [Char: 'n',]
        - action: [ NewPane: , ]
          key: [Char: 'N',]
```
That way the default NewPane creation gets you back into the Normal mode, but you can
stay in the NewPane creation with Shift+N.

jcaplan:
@a-kenji yes I really like how you did the config and think I will wind up spending some time hand crafting my own. These two are just particular ones that I think increase usability a lot without forcing anybody to create a really long config.

a-kenji:
I agree!
I think it might make sense to set an escape for the escape in case one wants to stay in the mode,
but maybe that will get to convoluted as a default.

jcaplan:
> I agree!
> I think it might make sense to set an escape for the escape in case one wants to stay in the mode,
> but maybe that will get too convoluted as a default.

I think this gets into the territory of how with vim or tmux you can enter commands with a colon to update the configuration. not sure a key shortcut is the way to go.

imsnif:
@jcaplan, thanks for this! I've been thinking about the keybinds and usability of zellij forever and it's really great to have other minds weigh in on this as well.
I think the alt+direction keys are a great addition, but I have some concerns about having them as the default (looking through the code that's the way I understand them to be now, right? or did I miss what you mentioned about having them as a config option?)
The issue: I've been trying to keep `normal` mode as lean as possible, because I want that to be where most users would spend most of their time (similar to command mode in vim). The more keybindings we add in this mode, the more we would be forcing users to either change their configuration in other apps (eg. I have alt+direction mapped to scrolling in urxvt) or spend more time in locked mode.
So I like what you said about this being a matter of taste and having these provided as an alternative.
If we go in that direction though (providing these as a default alternative) it would cause some more issues that we need to think about: if we just give eg. a command line flag (something like: `--alternative-keybinds`) we would have to:
1. Explain it somehow (what's alternative about it? why do some things go into it and some things go into other configs?)
2. Maintain it: if in 6 months we decide to change something minor about how the actions work, since the alternative configuration is part of our code-base, we would have to make sure it still works. The best way to do this would be to create integration tests for it (I think unit tests will not be enough because we're talking about user experience rather than code testing).

When I think about keybindings, I think it would be best to maintain everything except the default somewhere else (recommended dotfiles, etc.) 
But I'm not 100% certain about this and am willing to discuss it. What do you think about it? Do you have any thoughts about the concerns I've raised? Do you see it differently, or?

a-kenji:
@jcaplan, 
Probably I worded it wrong.
I meant that, maybe shift should invert the escape.
So if I usually want the New Pane Command to exit the mode - shift + n could stay in the mode.
And if I usually want the New Pane Command to stay in the mode - shift + n could exit the mode.

But I am not sure if that is a good default. I will try it out that way!

imsnif:
@jcaplan - tbh the more I think about this change the more I'm warming up to it. Let's go with the alt+direction stuff as part of the default Normal mode (and add them to the help). If they collide with too much existing shortcuts for users and lock mode isn't enough, we'll reconsider it. What do you think?

jcaplan:
@imsnif I will add the alt+arrows to the status bar. are you asking for anything else?

imsnif:
> @imsnif I will add the alt+arrows to the status bar. are you asking for anything else?

Ideally, it would be good to also support alt+arrow-keys - but I think that's not possible.

imsnif:
Already implemented.

Issue #194, opened 2021-02-19
Meta issue: Beta version


imsnif:
# Goals of the beta version
Upon release of this version, Zellij should be stable and have enough features to be the daily driver of those who wish to contribute to its development. Be they active developers or enthusiasts who believe in the future of this app.

This issue tracks all the work that needs to be done to achieve this goal. 

### Stability level
There should be no crashes except in extreme situations.

### Compatibility level
There should be no bugs that require the pane to be re-opened in order to fix: resizing in order to fix is fine for this version.

### Performance level
The performance should not be annoying.

### Features
- [x] supports linux and macos
- [x] tab bar (https://github.com/zellij-org/zellij/issues/166)
- [x] react to terminal window resizes (https://github.com/zellij-org/zellij/issues/34 and https://github.com/zellij-org/zellij/issues/122)
- [x] plugin system documentation (https://github.com/zellij-org/zellij/issues/193)
- [x] status bar should behave properly (https://github.com/zellij-org/zellij/issues/192)

### Bugs
- [x] https://github.com/zellij-org/zellij/issues/184
- [x] https://github.com/zellij-org/zellij/issues/167
- [x] https://github.com/zellij-org/zellij/issues/43

imsnif:
::tada::

PR #229, opened 2021-03-13
additional performance improvements

jcaplan:
polling all pty fds was causing idle cpu utilization and also increasing the load when stuff was happening.

Instead use `async_std::fs::File`, which handles polling the fd.

the savings are easiest to see when idle. here is zellij release build with one pane displaying htop showing it's own usage:

release build off main:
image

release build from this PR:
image

### DIFF
--- a/src/common/os_input_output.rs
+++ b/src/common/os_input_output.rs
@@ -1,5 +1,4 @@
 use crate::panes::PositionAndSize;
-use nix::fcntl::{fcntl, FcntlArg, OFlag};
 use nix::pty::{forkpty, Winsize};
 use nix::sys::signal::{kill, Signal};
 use nix::sys::termios;
@@ -116,12 +115,7 @@ fn spawn_terminal(file_to_open: Option<PathBuf>, orig_termios: termios::Termios)
             Ok(fork_pty_res) => {
                 let pid_primary = fork_pty_res.master;
                 let pid_secondary = match fork_pty_res.fork_result {
-                    ForkResult::Parent { child } => {
-                        // fcntl(pid_primary, FcntlArg::F_SETFL(OFlag::empty())).expect("could not fcntl");
-                        fcntl(pid_primary, FcntlArg::F_SETFL(OFlag::O_NONBLOCK))
-                            .expect("could not fcntl");
-                        child
-                    }
+                    ForkResult::Parent { child } => child,
                     ForkResult::Child => match file_to_open {
                         Some(file_to_open) => {
                             if env::var("EDITOR").is_err() && env::var("VISUAL").is_err() {
--- a/src/common/os_input_output.rs
+++ b/src/common/os_input_output.rs
@@ -1,4 +1,5 @@
 use crate::panes::PositionAndSize;
+use nix::fcntl::{fcntl, FcntlArg, OFlag};
 use nix::pty::{forkpty, Winsize};
 use nix::sys::signal::{kill, Signal};
 use nix::sys::termios;
@@ -115,7 +116,13 @@ fn spawn_terminal(file_to_open: Option<PathBuf>, orig_termios: termios::Termios)
             Ok(fork_pty_res) => {
                 let pid_primary = fork_pty_res.master;
                 let pid_secondary = match fork_pty_res.fork_result {
-                    ForkResult::Parent { child } => child,
+                    ForkResult::Parent { child } => {
+                        // fcntl(pid_primary, FcntlArg::F_SETFL(OFlag::empty())).expect("could not fcntl");
+                        #[cfg(test)]
+                        fcntl(pid_primary, FcntlArg::F_SETFL(OFlag::O_NONBLOCK))
+                            .expect("could not fcntl");
+                        child
+                    }
                     ForkResult::Child => match file_to_open {
                         Some(file_to_open) => {
                             if env::var("EDITOR").is_err() && env::var("VISUAL").is_err() {
--- a/src/common/pty_bus.rs
+++ b/src/common/pty_bus.rs
@@ -42,10 +42,13 @@ impl ReadFromPid {
         }
     }
 
-    #[cfg(not(test))]
     async fn next(&mut self) -> Option<Vec<u8>> {
         let mut buf = [0; 65535];
-        match self.file.as_ref().unwrap().read(&mut buf).await {
+        #[cfg(not(test))]
+        let data = self.file.as_ref().unwrap().read(&mut buf).await;
+        #[cfg(test)]
+        let data = self.os_input.read_from_tty_stdout(self.pid, &mut buf);
+        match data {
             Ok(bytes) => {
                 if bytes == 0 {
                     Some(vec![])
@@ -54,39 +57,6 @@ impl ReadFromPid {
                 }
             }
             _ => None,
-        }
-    }
-}
-
-#[cfg(test)]
-impl Stream for ReadFromPid {
-    type Item = Vec<u8>;
-    fn poll_next(mut self: Pin<&mut Self>, _cx: &mut Context<'_>) -> Poll<Option<Self::Item>> {
-        let mut read_buffer = [0; 65535];
-        let pid = self.pid;
-        let read_result = &self.os_input.read_from_tty_stdout(pid, &mut read_buffer);
-        match read_result {
-            Ok(res) => {
-                if *res == 0 {
-                    // indicates end of file
-                    Poll::Ready(None)
-                } else {
-                    let res = Some(read_buffer[..=*res].to_vec());
-                    Poll::Ready(res)
-                }
-            }
-            Err(e) => {
-                match e {
-                    nix::Error::Sys(errno) => {
-                        if *errno == nix::errno::Errno::EAGAIN {
-                            Poll::Ready(Some(vec![])) // TODO: better with timeout waker somehow
-                        } else {
-                            Poll::Ready(None)
-                        }
-                    }
-                    _ => Poll::Ready(None),
-                }
-            }
         }
     }
 }
--- a/src/common/pty_bus.rs
+++ b/src/common/pty_bus.rs
@@ -202,8 +202,7 @@ pub struct PtyBus {
  * As a result, we have to add the abstraction here instead
  * of in the test. So ReadFromPid implements stream to support
  * tests, and then this function does user space polling of the stream.
- * The production code just does a read on an async_std::fs::File.r!
- * TODO: can the test path be simplified without breaking any tests?
+ * The production code just does a read on an async_std::fs::File.
  */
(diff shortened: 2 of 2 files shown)


imsnif:
Hey, so first: I'm very happy you're tackling the performance problems! I myself have just begun relooking at the should_render issue yesterday.

About the blocking stuff: the reason I didn't originally go for a blocking call is that it won't really work with the async architecture (as far as I understand things anyway). If we do a blocking read, then the whole pty thread will be blocked until we wait for a read - not just the specific async task.
To be sure I'm right here, I checked out this PR and opened a few panes, and indeed after running some commands in each of them I ran into situations where everything gets stuck and only gradually updates itself step by step as I enter commands in one pane. Maybe there's still work that can be done here to optimize stuff, but I'm afraid this solution isn't good for us.

About the should_render: as I said, I started looking into this myself and would be happy to work on this together. We can share a branch on this maybe?

In any case - let's separate this into 2 different PRs, ok? We're dealing with 2 different optimizations and I think this would make it easier for us. What do you think?

jcaplan:
hmm ok. I was a bit confused about why anything was async anyway. I saw several async threads in gdb and assumed there was 1 per pane. I see what's going on now. guess I need to learn more about how rust implements async-io!

I think there are two options:
- https://tokio.rs/tokio/tutorial/io, there's a fair bit of work to hide performant fd polling behind an async abstraction, and hopefully they've done it
- otherwise not sure async is the way to go, it's creating too much CPU busywork. https://docs.rs/mio/0.7.0/mio/struct.Poll.html, if determined to roll out your own, then I think what's missing is a proper event loop underneath. There ought to be a single thread doing a poll on all FDs, with a periodic timeout to check if there are fds that need to registered/unregistered.

imsnif:
>     * https://tokio.rs/tokio/tutorial/io, there's a fair bit of work to hide performant fd polling behind an async abstraction, and hopefully they've done it

Cool. Right now in the codebase we use async_std, namely by implementing a [Stream](https://docs.rs/async-std/1.5.0/async_std/stream/trait.Stream.html) as you've seen. If you think tokio has a better solution for it (or you'd like to try a different approach with async_std) - by all means. Whip up a draft PR and let's see how it behaves.

>     * otherwise not sure async is the way to go, it's creating too much CPU busywork. https://docs.rs/mio/0.7.0/mio/struct.Poll.html, if determined to roll out your own, then I think what's missing is a proper event loop underneath. There ought to be a single thread doing a poll on all FDs, with a periodic timeout to check if there are fds that need to registered/unregistered.

My concern with such a solution (and the reason I went for async in the first place) is that if I understand things correctly (and for sure I might be missing something), such a solution would mean that if we get continuous data from one pane (eg. doing `cat` on a file that's 1M lines long) we would be blocking the flow of data from the other panes, won't we?

jcaplan:
Ok I’ll try again. I’m optimistic about a 1 thread tokio solution. 

re denial of service that’s why I kept around a 10ms sleep after unblocking based on the concerns left in the original comment. I think all the threads waking up are spinning enough to cause a cpu bottleneck. Only threads with work to do should wake up. Once they’re awake then we throttle them with sleeps and hopefully this reduces congestion. 

It may also be the case that with 2+ tokio threads that you support some amount of cpu hogging with a very busy stream without needing the sleeps. 

Once going this route I’ve been curious if most or all of the other threads can be made async. It makes me a bit nervous how many threads there are. Most of them are never doing anything.

imsnif:
Hey @jcaplan - I'm not 100% sure if you're done and want a second review on this?
(as a personal request: it's a little hard for me to get/remember context on PRs if you edit and force-push changes that rewrite the git and PR-comment history, next time if you could add comments instead of editing the original comment that would be helpful, thanks! :) )

jcaplan:
well I'd like to be done but unfortunately bypassing `os_input` and reading directly from the file breaks all the integration tests. I can't find something the compiler likes that let's me mock out the async call in the tests or cfg/ifdef my way out of trouble in the code.

imsnif:
Alright - let e know if you want a second pair of eyes.

jcaplan:
ok so I don't really see a way around having a completely different code path for test, i've put all the old code back in with `cfg(test)` everywhere. it's not ideal but I don't see another way around. Attempts I made at simplifying the old code ended up in lots of test failures.

imsnif:
Hey @jcaplan - thanks for your work on this. I'm also not super wild about the idea of having so many test conditionals. Maybe we can find a creative way around it?
How about moving this part into OsApi: https://github.com/zellij-org/zellij/pull/229/files#diff-8e7424a4b9c2837c3bd08c30c8714ed08891d285594dc2253c571a47196da0efR33-R34
And then implementing the async stream around it? We'll have to adjust the fakes in the test to work with this of course.
Do you think that can work?

jcaplan:
I don't think that parts the problem, that's relatively clean. There's two implementations of next, the test through the Stream trait and this function to stub it out for API equivalence. Turning it into a stream when it already is one underneath will just add layers and cost without making `spawn_terminal` any cleaner, which is the real problem in my opinion.

I get a lot of test failures on my laptop so it's a little hard to change. I'll try to push one thing and see if they all still pass in the github job, if not I'll revert.

imsnif:
The test infrastructure is somewhat convoluted and very much overdue for a refactor and simplification. Until then: if you can take all the elements which interact with ptys and file descriptors and place them as methods inside the `OsApi` trait, I think I can adjust the testing framework to work with it until all tests pass. What do you think?

jcaplan:
I wasn't able to figure out how do that without resorting to an `async fn` in `OsApi which is not allowed. I just pushed a simplified version and the tests all passed with all the optimization stuff removed from `spawn_terminal`, so on the whole it's a lot less clunky. 

another problem I hit when exploring solutions was that replacing the `while` loop with a `loop` with assignments inside the loop doesn't work because you hit all sorts of issues around scope and async closures. 

I also just tried removing the stream trait in fa0b550 but it caused a lot of test failures. I think this is as good as it gets without reworking the tests, which I'm not signing up to include in this PR.

imsnif:
@jcaplan - I took this for a spin and while I understand the benefits in resource usage, a test I did found this degraded performance.
To test, I created a large file (100K lines) and then did a `cat` on it inside Zellij (with just 1 pane). I measured it with the `time` command.

From main:
```
________________________________________________________
Executed in    1.81 secs      fish           external
   usr time    0.59 millis  586.00 micros    0.00 millis
   sys time  235.43 millis  103.00 micros  235.33 millis
```

From this branch:
```
________________________________________________________
Executed in   15.54 secs      fish           external
   usr time    0.35 millis  352.00 micros    0.00 millis
   sys time  316.96 millis  436.00 micros  316.52 millis
```

My guess is that this is because the previous version would cache large amounts of data and only send a render instruction every 30ms. There might be other optimizations here. Would you like to try and work on this further? Or did I miss other situations in which this performs much better than main?

jcaplan:
I would argue this is a good thing. not because the performance in this scenario is acceptable, but because the way the old method was polling large amounts of data periodically is covering up serious inefficiencies in the render function. there's no good reason that rendering should ever consume as much cpu time as it does or take so long to complete a single iteration.

I think you should accept this change and your scenario is a good performance regression benchmark to use to start reducing the cost of renders. there are lots of things to address, my guess at the main culprits are: too many clones/copies instead of global buffer state, lots of overwriting entire buffers with mostly identical data that we can determine is mostly identical, further reduction to the number of bytes being sent to stdout (possible to do more targeted writes with cursor moving?)

but it's up to you.

imsnif:
I very much agree that the render function at the moment is inefficient and could use an overhaul. It was not written with performance in mind, mostly because as I was writing it I was still trying to figure out how terminals work. :)

For sure the caching of data in the `PtyBus` (or more accurately the debounced renders) is masking this, but that's the best solution we have right now and I believe it's definitely in the "good enough" category considering what stage the project is in (pre-beta). I would very much like to move from "good enough" to "excellent", but that would require a medium to big chunk of work in order to make all the changes you mentioned to the render function.

I very much appreciate all the work you put into this, but accepting this change before doing the above would degrade performance. Maybe it will still be in the "good enough" box, but it will surely be worse than it is now. Add to that the fact that this doesn't play very nice with our current testing infrastructure and the picture is clear.

The way I see it, we have 3 options moving forward:
1. Keep this PR open and wait until we optimize the render function (I'm not sure when this will happen, what little development resources the team has are currently directed elsewhere: https://github.com/zellij-org/zellij/issues/194)
2. You can start working on optimizing the render function if this is something that interests you - I would be happy to guide you through it and/or be available to answer any questions you might have.
3. You can adjust this PR to include the 30ms caching mechanism. I think there are some pretty cool things one could do with async streams to achieve this - but maybe you have other ideas.

jcaplan:
ok but you already asked me to split out modifications related to optimizing rendering into a different pr.....

imsnif:
> ok but you already asked me to split out modifications related to optimizing rendering into a different pr.....

Cool - so now the order is clear.

jcaplan:
another big problem with this test case is memory utilization. I'm not really sure how because I thought this was the whole point of rust, but when running cat on a large file over and over again memory utilization goes up several hundred MB every time without getting released. I'll open a new issue for this. I don't agree with deferring performance issues as low priority, since this isn't tuning but suggesting possible deep structural issues that only get harder to change over time. But hey I'm just trying to learn rust and this is your project. However, the memory utilization should definitely be something that sets off red alert. as it happens it will probably be the same sort of fixes to both problems as I suspect they are very related...

imsnif:
> another big problem with this test case is memory utilization. I'm not really sure how because I thought this was the whole point of rust, but when running cat on a large file over and over again memory utilization goes up several hundred MB every time without getting released. I'll open a new issue for this. 

That's definitely an issue. This is not rust's fault - it's ours! We don't currently limit the scroll buffer, so these lines stick around forever. The fix for this should be pretty trivial. We can move the discussion about this there. Thanks for bringing this up!

> I don't agree with deferring performance issues as low priority, since this isn't tuning but suggesting possible deep structural issues that only get harder to change over time. But hey I'm just trying to learn rust and this is your project. 

Well, just to emphasize: this is not my project. I'm part of the team working on it. I don't think it's fair to simplify it to "deferring performance issues as low priority". At the moment my personal goal is to try and make Zellij usable. Last week, I felt addressing performance issues was important because back then the performance was horrid. After some fixes, the performance became passable. Right now, I feel that (for example) making Zellij react to terminal window size changes is more important than squeezing more performance out of the render function. It's a matter of trade-offs in what me and the rest of the team choose to work on.

If *you* feel these things are more important, you are not only encouraged to work on them yourself - I will also devote time to help you out and guide you through the code base as much as you want and need.

imsnif:
This was implemented in a different way.

Issue #195, opened 2021-02-19
Feature: display current path in strider


imsnif:
Right now, the strider plugin (a filesystem tree explorer) only displays the contents of the current directory. It can be very useful if it also displays the relative/full path of the current directory as a title or some other visible UI element.

yhung119:
Hey :) can I help with this issue?

a-kenji:
Hey @yhung119 
Yes that would be awesome!
Please don't hesitate to ask if you have some questions!

imsnif:
A good place to display it would be the pane title (we didn't have pane titles when the issue was originally opened).

RobWalt:
Hey, I'd like to contribute. Can anyone give me some hints on how the PluginPane and the plugins themselves communicate? I'm particularly interested in how I get information from the plugin to the pane.

knidarkness:
Hey,

I've been playing around with `zelij` for a while now and would like to help on this issue, potentially using RobWalt's initial implementation.

imsnif:
Sure thing @knidarkness - let us know if you need any help.

imsnif:
Implemented in https://github.com/zellij-org/zellij/pull/3200

PR #232, opened 2021-03-13
feature(strider): Show current parent folder

a-kenji:
#195

Shows the current parent folder, if opened inside
of strider eg:

It doesn't display the current folder on the root
pane of strider, but if opened a sub folder it displays
the parent folder.

image
image

### DIFF
--- a/default-tiles/strider/src/main.rs
+++ b/default-tiles/strider/src/main.rs
@@ -13,6 +13,7 @@ impl ZellijTile for State {
     }
 
     fn draw(&mut self, rows: usize, cols: usize) {
+
         for i in 0..rows {
             if self.selected() < self.scroll() {
                 *self.scroll_mut() = self.selected();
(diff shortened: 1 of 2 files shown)


imsnif:
Awesome! Right now this seems a little unclear to me visually though. Even knowing what I was looking for, I had a bit of a hard time finding it. Do you think we can make the path clearer? Maybe make it a different color and/or add some background to it, or add some spacing between it and the folder list, or...? I trust your creativity on this one. :)

a-kenji:
Yes that is true! It already has a little different styling but it does blend in with everything else.
I will try to iterate on it a little and see If I find something I am comfortable with! Thanks for the feedback!

a-kenji:
@imsnif 
20210402_14h50m58s_grim

Does that make more sense?The header is now underlined and thus separated from the files.

imsnif:
Hum... I think it's a good direction but my preference would be to differentiate it by color. We took a lot of inspiration from the ranger tool - maybe check out what they do to get some ideas?

a-kenji:
Ah yeah, a different colour may be a good delineation.
Ranger has a hierarchical paradigm, so it still works a little bit different and can
display the parent folder more naturally:
20210405_11h43m38s_grim

PR #235, opened 2021-03-17
feature: Focused pane border coloring

denismaxim0v:
Will leave as draft for now, as there's a little bit more work left on this

### DIFF
--- a/src/client/boundaries.rs
+++ b/src/client/boundaries.rs
@@ -1,7 +1,6 @@
 use crate::tab::Pane;
+use ansi_term::Colour::{self, Fixed};
 use std::collections::HashMap;
-use ansi_term::Colour::*;
-use super::super::utils::logging::debug_log_to_file;
 
 use std::fmt::{Display, Error, Formatter};
 
@@ -26,7 +25,7 @@ pub struct BoundarySymbol {
     boundary_type: BoundaryType,
     invisible: bool,
     should_be_colored: bool,
-    color: ansi_term::Colour
+    color: Colour,
 }
 
 impl BoundarySymbol {
@@ -35,17 +34,14 @@ impl BoundarySymbol {
             boundary_type,
             invisible: false,
             should_be_colored,
-            color: match should_be_colored {
-                true => ansi_term::Colour::Red,
-                false => ansi_term::Colour::White
-            }
+            color: Fixed(245),
         }
     }
     pub fn invisible(mut self) -> Self {
         self.invisible = true;
         self
     }
-    pub fn color(mut self, color: ansi_term::Colour) -> Self {
+    pub fn color(mut self, color: Colour) -> Self {
         self.color = color;
         self.should_be_colored = true;
         self
@@ -56,12 +52,10 @@ impl Display for BoundarySymbol {
     fn fmt(&self, f: &mut Formatter) -> Result<(), Error> {
         match self.invisible {
             true => write!(f, " "),
-            false => {
-                match self.should_be_colored {
-                    true => write!(f, "{}", self.color.paint(self.boundary_type)),
-                    false => write!(f, "{}", self.boundary_type)
-                }
-            }
+            false => match self.should_be_colored {
+                true => write!(f, "{}", self.color.paint(self.boundary_type)),
+                false => write!(f, "{}", self.boundary_type),
+            },
         }
     }
 }
@@ -74,6 +68,10 @@ fn combine_symbols(
     let should_be_colored = current_symbol.should_be_colored || next_symbol.should_be_colored;
     let current_symbol = current_symbol.boundary_type;
     let next_symbol = next_symbol.boundary_type;
+    let color = match should_be_colored {
+        true => Fixed(154),
+        false => Fixed(245),
+    };
     match (current_symbol, next_symbol) {
         (boundary_type::TOP_RIGHT, boundary_type::TOP_RIGHT) => {
             // (┐, ┐) => Some(┐)
@@ -82,7 +80,7 @@ fn combine_symbols(
                 boundary_type,
                 invisible,
                 should_be_colored,
-                color
+                color,
             })
         }
         (boundary_type::TOP_RIGHT, boundary_type::VERTICAL) => {
@@ -92,7 +90,7 @@ fn combine_symbols(
                 boundary_type,
                 invisible,
                 should_be_colored,
-                color
+                color,
             })
         }
         (boundary_type::TOP_RIGHT, boundary_type::HORIZONTAL) => {
@@ -102,7 +100,7 @@ fn combine_symbols(
                 boundary_type,
                 invisible,
                 should_be_colored,
-                color
+                color,
             })
         }
         (boundary_type::TOP_RIGHT, boundary_type::TOP_LEFT) => {
@@ -112,7 +110,7 @@ fn combine_symbols(
                 boundary_type,
                 invisible,
                 should_be_colored,
-                color
+                color,
             })
         }
         (boundary_type::TOP_RIGHT, boundary_type::BOTTOM_RIGHT) => {
@@ -122,7 +120,7 @@ fn combine_symbols(
                 boundary_type,
                 invisible,
                 should_be_colored,
-                color
+                color,
             })
         }
         (boundary_type::TOP_RIGHT, boundary_type::BOTTOM_LEFT) => {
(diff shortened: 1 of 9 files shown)


src/common/screen.rs
```diff
@@ -303,4 +310,9 @@ impl Screen {
             }
         }
     }
+    pub fn change_input_mode(&mut self, input_mode: InputMode) {
+        self.input_mode = input_mode;
+        self.get_active_tab_mut().unwrap().input_mode = self.input_mode;
```
imsnif, line 315: Shouldn't this be changed in all tabs and not just the active one?

src/client/boundaries.rs
```diff
@@ -677,27 +838,29 @@ impl Boundaries {
...
+    pub fn add_rect(&mut self, rect: &dyn Pane, should_be_colored: bool, input_mode: InputMode) {
         if rect.x() > 0 {
             let boundary_x_coords = rect.x() - 1;
             let first_row_coordinates = self.rect_right_boundary_row_start(rect);
             let last_row_coordinates = self.rect_right_boundary_row_end(rect);
             for row in first_row_coordinates..last_row_coordinates {
                 let coordinates = Coordinates::new(boundary_x_coords, row);
                 let mut symbol_to_add = if row == first_row_coordinates && row != 0 {
-                    BoundarySymbol::new(boundary_type::TOP_LEFT)
+                    BoundarySymbol::new(boundary_type::TOP_LEFT, should_be_colored)
```
imsnif, line 849: If there's already a `colored_borders` method on the `Pane` trait, why not use it here instead of passing `should_be_colored`?

src/client/boundaries.rs
```diff
@@ -677,27 +838,29 @@ impl Boundaries {
...
                 } else if row == last_row_coordinates - 1 && row != self.rows - 1 {
-                    BoundarySymbol::new(boundary_type::BOTTOM_LEFT)
+                    BoundarySymbol::new(boundary_type::BOTTOM_LEFT, should_be_colored)
                 } else {
-                    BoundarySymbol::new(boundary_type::VERTICAL)
+                    BoundarySymbol::new(boundary_type::VERTICAL, should_be_colored)
```
imsnif, line 853: Also, maybe we can make the method assigning color to `BoundarySymbol` a Builder type method? Something like:
```
BoundarySymbol::new(boundary_type::VERTICAL).color(colors::GREEN)
```
What do you think? I just think it will be more idiomatic rust.

src/client/tab.rs
```diff
@@ -640,12 +647,16 @@ impl Tab {
             self.full_screen_ws.rows as u16,
         );
         let hide_cursor = "\u{1b}[?25l";
+        //let input_mode = self.get_state(&self.send_app_instructions);
```
imsnif, line 650: oops :)

src/common/screen.rs
```diff
@@ -303,4 +310,9 @@ impl Screen {
...
+        self.render();
```
imsnif, line 316: In the input handler where we send the `ChangeInputMode` instruction, we already explicitly send a `Render` instruction later... maybe we can remove this extra render then?

src/client/boundaries.rs
```diff
@@ -40,28 +52,46 @@ impl BoundarySymbol {
...
 ) -> Option<BoundarySymbol> {
     let invisible = current_symbol.invisible || next_symbol.invisible;
+    let should_be_colored = current_symbol.should_be_colored || next_symbol.should_be_colored;
     let current_symbol = current_symbol.boundary_type;
     let next_symbol = next_symbol.boundary_type;
+    let color = match should_be_colored {
+        true => match input_mode {
+            InputMode::Normal => colors::GREEN,
+            InputMode::Locked => colors::GREEN,
+            InputMode::Pane => colors::WHITE,
```
imsnif, line 78: How about:
```
match input_mode {
    InputMode::Normal | InputMode::Locked => colors::GREEN,
    _ => colors::WHITE
}
```
?

imsnif (changes requested):
Hey @denismaxim0v - this is very cool! Good work.
I left some comments in the code.
Otherwise - what do you think of making the default border color something other than white? I like the green in normal/locked, but maybe something else to distinguish it in the other modes?

src/common/screen.rs
denismaxim0v: Probably should, but I haven't noticed any weird behaviors for user, because when he is in the tab mode and switches between them it still updates everything, but I can of course change that to update all tabs

src/client/boundaries.rs
denismaxim0v: oops, seems like I forgot to remove colored borders that I added initially, wasn't really needed for me for this implementation. Do you think I should go with it instead?

src/client/boundaries.rs
denismaxim0v: That's definitely something I can do

src/client/tab.rs
denismaxim0v: happens :)

src/client/boundaries.rs
denismaxim0v: I did it this way for now, because I wasn't sure whether I would have to change colors individually for more distinction after your review

src/client/boundaries.rs
imsnif: My intuition says it will be cleaner, but maybe I'm wrong :)

src/common/screen.rs
imsnif: I think that just happens to work right now because the tab mode color is white.

denismaxim0v: hmm, could be

src/client/tab.rs
```diff
@@ -73,6 +77,7 @@ pub struct TabData {
     pub position: usize,
     pub name: String,
     pub active: bool,
+    pub input_mode: InputMode,
```
imsnif, line 80: Why do we need the InputMode on the TabData?

src/client/boundaries.rs
```diff
@@ -17,51 +18,76 @@ pub mod boundary_type {
...
 impl Display for BoundarySymbol {
     fn fmt(&self, f: &mut Formatter) -> Result<(), Error> {
-        if self.invisible {
-            write!(f, " ")
-        } else {
-            write!(f, "{}", self.boundary_type)
+        match self.invisible {
+            true => write!(f, " "),
+            false => match self.color {
+                Some(_) => write!(f, "{}", self.color.unwrap().paint(self.boundary_type)),
```
imsnif, line 60: `Some(color) => write!(f, "{}", color.paint(self.boundary_type))` won't work?

src/client/boundaries.rs
```diff
@@ -17,51 +18,76 @@ pub mod boundary_type {
...
 fn combine_symbols(
     current_symbol: BoundarySymbol,
     next_symbol: BoundarySymbol,
+    input_mode: InputMode,
 ) -> Option<BoundarySymbol> {
     let invisible = current_symbol.invisible || next_symbol.invisible;
+    let should_be_colored = current_symbol.color.is_some() || next_symbol.color.is_some();
     let current_symbol = current_symbol.boundary_type;
     let next_symbol = next_symbol.boundary_type;
+    let color = match should_be_colored {
```
imsnif, line 76: Can't we take this from the `Option<Colour>` on the boundary symbol itself?

src/client/tab.rs
```diff
@@ -189,6 +194,7 @@ impl Tab {
         send_app_instructions: SenderWithContext<AppInstruction>,
         max_panes: Option<usize>,
         pane_id: Option<PaneId>,
+        input_mode: InputMode,
```
imsnif, line 197: We keep InputMode both on Tab and Screen - do you think we can only keep it in one place? Maybe Screen? And then pass it to each tab when we render? I think it will make things more maintainable, what do you think?

imsnif (changes requested):
Hey @denismaxim0v - good progress! I've made some comments.

src/client/tab.rs
denismaxim0v: Yeah, don't see why not

src/client/boundaries.rs
denismaxim0v: I think it should, don't know honestly why I left it this way, some things can't be explained 😄

src/client/tab.rs
denismaxim0v: oops, that should've been a part of a cleanup 😅

src/client/tab.rs
denismaxim0v: Actually, we have render called inside the tab itself in lots places, so that would introduce maybe a bit more complexity?

Issue #102, opened 2020-12-15
Feature: Shortcut for opening a new pane in current working directory


kunalmohan:
The new pane should be opened with the current working directory of the active pane instead of the default directory (`/home/user` in case of Linux)

```rust
use std::env;
let path = env::current_dir()?;
```
This should be helpful while implementing this feature.

kunalmohan:
I looked into this, we'll need to get the CWD of the active terminal using https://docs.rs/nix/0.11.0/nix/unistd/fn.getcwd.html instead of `std::env` and provide it to pty_bus while spawning a new terminal.

As I see it we have two implementation paths-
1. `StdinHandler -> Screen(GetCwd) -> PtyBus(SpawnTerminal) -> Screen(SpawnTerminal)`
2. PtyBus also stores the active pane `RawFd` at all times. So would reduce one step here.
`StdinHandler -> PtyBus(SpawnTerminalInCwd) -> Screen(SpawnTerminal)`

imsnif:
Option 2 seems good. Do you want to work on this, or shall I put a `help wanted` on it and someone will pick it up?

kunalmohan:
You can put up the label. I'll prefer working on tabs and persistent sessions first. And since this is relatively easy, this maybe good for a new contributors as well.

qrasmont:
Hi, this seems like a good first issue and I would like to help. Can I go for it :smiley:  ?

imsnif:
> Hi, this seems like a good first issue and I would like to help. Can I go for it smiley ?

You got it. :) Please feel free to reach out if you have any questions.

qrasmont:
I do need a pointer. I found petty quick how create a new shortcut but what I have not yet found is where to call  `unistd::getcwd()` from. Since `unistd::getcwd()` returns the cwd from the calling process, I would need to call it from the current active pane's process. And I saw as mentioned above that `PtyBus` stores the active pane `RawFd`.So I can't put the 2 together yet.

How do I go from knowing the `RawFd` of the active pane to calling `unistd::getcwd()` from whithin its process ? Any tip would be appreciated :+1:.

kunalmohan:
Oops! Looks like I misread the docs for `getcwd()` earlier. I was under the impression that `getcwd()` would accept the `RawFd` as a parameter. I don't have another solution in my mind atm, so I guess we'll have to figure it out. I'll look for a solution and maybe you can help with that? (only if you want to, no pressure)
Sorry for the trouble!

[...]

PR #277, opened 2021-04-18
feat(cwd-pane): Allow to open a new pane from the current working directory

qrasmont:
PR for issue #102 

This allows to open a new pane in the current working directory of the active pane (using: Ctrl-p c).

Tested on both Linux and macOS.


imsnif (changes requested):
Hey @qrasmont - this looks great and I'm sure it was quite a bit of work figuring out all the data paths.
Since this is a bit of a big change I'm going to take a closer look at it in the next few days. Meanwhile, I left one comment. And other than that, I think we don't need a special command for this. Let's make it the default behaviour.

src/common/utils/proc.rs
```diff
@@ -0,0 +1,24 @@
...
+    }
+}
+
+#[cfg(target_os = "linux")]
+pub fn get_cwd(pid: RawFd) -> Option<PathBuf> {
+    match fs::read_link(format!("/proc/{}/cwd", pid)) {
+        Ok(cwd) => Some(cwd),
+        Err(_) => None,
+    }
+}
```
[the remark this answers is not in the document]
qrasmont: Done

qrasmont:
I still need to figure out how I've broken some of the integration tests

src/common/pty_bus.rs
```diff
@@ -169,31 +176,76 @@ impl PtyBus {
...
+            pids: HashMap::new(),
+            active_pane: None,
             debug_to_file,
             task_handles: HashMap::new(),
         }
     }
-    pub fn spawn_terminal(&mut self, file_to_open: Option<PathBuf>) -> RawFd {
-        let (pid_primary, pid_secondary): (RawFd, RawFd) =
-            self.os_input.spawn_terminal(file_to_open);
+    fn terminal_spawner(
```
imsnif, line 185: Do you think we can either give this a more descriptive name or merge it with `spawn_terminal` somehow? I find it a little hard to understand the difference between these two methods without being aware of the whole cwd system...

src/client/tab.rs
```diff
@@ -1889,6 +1904,9 @@ impl Tab {
         } else {
             self.active_terminal = Some(active_terminal.unwrap().pid());
         }
+        self.send_pty_instructions
+            .send(PtyInstruction::UpdateActivePane(self.active_terminal))
+            .unwrap();
```
imsnif, line 1909: I think this is the reason some of the tests are failing. Be sure to only send this instruction if the `active_terminal` has actually changed (eg. here move this inside the `else` block above).

imsnif (changes requested):
Hey, great work once again on the changed. Thanks for being patient with my reviews taking a little bit of time and being somewhat fragmented (I'd ideally have liked to give all of this to you in one go, but times are a little hectic in Zellij right now :) ).

I left some comments in the code, and otherwise here's some issues I found so far:
1. This doesn't seem to work when I open a new tab. The new terminal there opens up in the default folder.
2. If the focused pane is a plugin (eg. `strider`), this crashes. To test this, start zellij with the strider layout: `zellij --layout strider`. Then move focus to the strider file explorer and open a new pane with alt-n.
3. Similarly to 2, if strider tries to open a new file, zellij crashes. (to reproduce: in strider try to press ENTER on a text file - make sure either the `EDITOR` or `VISUAL` env variable on your machine point to a text editor). I imagine this is for the same root cause as 2.

DAlperin:
Hey @qrasmont do you still have time to work on this and get it ready to merge? If not, I can pick this up if that would be helpful.

imsnif:
Hey @DAlperin - since we haven't heard from @qrasmont lately, I think you can definitely pick this up if you want to. Either continuing from this PR or starting over. What do you think?

DAlperin:
Hey @imsnif I got a little busy this month but I think I can work on this a bit next week. I'll try and take this PR and rebase against main and put it in a new PR to start making the requested changes in.

a-kenji:
This is now implemented by #729, 
thanks to everyone working on it!

PR #279, opened 2021-04-19
CI workflow for automatic release

qballer:
This isn't ready but some work went to it, based heavily on stuff from  Bandwhich and online. 
In order to continue this work, we should cache builds and control output file names


TheLostLambda:
As this progresses, I think it would be nice to also add the `workflow_dispatch` trigger so that we have the option to generate binaries on demand between releases (should the need arise).

PR #293, opened 2021-04-21
Reduce crate size by 4.8MB using `cargo diet`, to 77kB

Byron:
Here is my call to action :D: 
Screenshot 2021-04-21 at 09 25 07

On top of that I verified that `cargo package` is still able to build,
adjusting the includes in the process to assure required assets are
packaged, too.

I am happy to adjust the includes to your liking so we find a way to reduce the overall crate size.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -7,7 +7,7 @@ description = "A terminal workspace with batteries included"
 license = "MIT"
 repository = "https://github.com/zellij-org/zellij"
 homepage = "https://zellij.dev"
-include = ["src/**/*", "assets/plugins/*", "assets/layouts/*", "assets/completions/*", "LICENSE.md", "README.md", "build.rs", "!**/*_test.*", "!**/tests/**/*"]
+include = ["src/**/*", "assets/plugins/*", "assets/layouts/*", "LICENSE.md", "README.md", "!**/*_test.*", "!**/tests/**/*"]
 
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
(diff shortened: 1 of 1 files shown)


TheLostLambda:
This is awesome! Thanks for doing this, I'd never used `includes` before!

 We can probably chop a little more if you'd like :slightly_smiling_face: Maybe something like:

```
 include = ["src/**/*", "assets/plugins/*", "assets/layouts/*", "LICENSE.md", "README.md", "!**/*_test.*", "!**/tests/**/*"]
```

The completions aren't really used by `cargo install` right now (only .deb and AUR packaging), so we don't need the `build.rs` for generating them or the completion assets :)

If you chop those out too and make sure `cargo package` still works, I'm happy to merge this!

Byron:
Even better :) - thanks for looking into this!

The fixes work as expected, and I am happy that zellij users will now get even faster installs via `cargo`.

Unsolicited and unrelated idea: "Expert mode" which disables on-screen key combinations and other help text.

TheLostLambda:
@Byron For a poor-man's "Expert Mode", you can modify your `~/.local/share/zellij/layouts/default.yaml` (slightly different location on macOS) and remove the status-bar plugin :wink: 

```yaml
---
direction: Horizontal
parts:
  - direction: Vertical
    split_size:
      Fixed: 1
    plugin: tab-bar
  - direction: Vertical
#  - direction: Vertical
#    split_size:
#      Fixed: 2
#    plugin: status-bar
```

For something a touch less hacky, it'd be great if you could open an issue! :slightly_smiling_face:

Byron:
Thanks for the hint! I think I will do that once I am actually an expert 😅.

PR #294, opened 2021-04-21
Adding field active_at and using it to sort panes

### LINKED ISSUE #230: pane should go to previously active when possible
consider this layout:
image

If I am in the top right pane and then move to the left pane and then back to the right again, zellij chooses the bottom right pane.
I think it is reasonable to expect the top right pane (i.e. the last active pane to the right) to become active when moving back to the right.

The algorithm that determines which pane to move to when multiple panes are on the border of the active pane should give precedence to the last active pane.

dantepippi:
Using last activity time to determine which pane was previously active when moving back to a direction.

Closes #230

### DIFF
--- a/src/client/panes/plugin_pane.rs
+++ b/src/client/panes/plugin_pane.rs
@@ -1,8 +1,9 @@
 use crate::{common::SenderWithContext, pty_bus::VteBytes, tab::Pane, wasm_vm::PluginInstruction};
 
-use std::{sync::mpsc::channel, unimplemented};
-
 use crate::panes::{PaneId, PositionAndSize};
+
+use std::time::Instant;
+use std::{sync::mpsc::channel, unimplemented};
 
 pub struct PluginPane {
     pub pid: u32,
@@ -14,6 +15,7 @@ pub struct PluginPane {
     pub send_plugin_instructions: SenderWithContext<PluginInstruction>,
     pub max_height: Option<usize>,
     pub max_width: Option<usize>,
+    pub active_at: Instant,
 }
 
 impl PluginPane {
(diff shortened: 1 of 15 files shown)


dantepippi:
Just wanted to confirm this. Should I change those failing tests to match the new behavior?
When moving focus to a direction it used to choose the largest overlap but now it will pick the last used pane.

TheLostLambda:
@dantepippi Yeah, we'll probably be fine changing tests, but it would be good if we also had a couple of other people test it out so we can be sure the behavior is "correct" before setting it in stone with snapshots. You'll probably need [`cargo-insta`](https://github.com/mitsuhiko/insta) to update the snapshots once we're sure it's working :)

src/client/panes/plugin_pane.rs
```diff
@@ -13,6 +14,7 @@ pub struct PluginPane {
     pub position_and_size_override: Option<PositionAndSize>,
     pub send_plugin_instructions: SenderWithContext<PluginInstruction>,
     pub max_height: Option<usize>,
+    pub active_at: u64,
```
TheLostLambda, line 17: As a small code tweak though, I'd prefer if we used [`Instant`](https://doc.rust-lang.org/std/time/struct.Instant.html)s here. Those are actually guaranteed to never do anything odd around leap-seconds and are much more idiomatic Rust (just in my opinion)

dantepippi: I'm on it.

a-kenji:
Hey! Thanks! The behavior seems correct to me.

imsnif:
This is a cool solution and I like the behaviour. :)

dantepippi:
> This is a cool solution and I like the behaviour. :)

Thank you! I'm going to need some directions on how to adapt the snapshots for the modified tests.

TheLostLambda:
@dantepippi Stellar work! Nicely done tackling the testing system too, that's no easy beast to tame :sweat_smile: Consider it merged!

PR #298, opened 2021-04-21
Add page scroll

horasal:
In scroll mode, allow 'Ctrl+F' (or PageDown) and 'Ctrl+B' (or PageUp) to scroll one page forward/backward.

This pr add two new action: PageScrollUp, PageScrollDown.
When PageScrollUp/PageScrollDown triggered, current pane will scroll up/down by `pane.rows() - 1` lines.


TheLostLambda:
This is excellent! Adding an action, things to the default config, and tests!

It works like a charm for me and is something I'd been meaning to implement for a while now :slightly_smiling_face: 

One last thing to do though! We need to add some help text to the [`get_mode_info()` function](https://github.com/zellij-org/zellij/blob/138889f49a8fb7a4f0254466e54eca7712cf687f/src/common/input/handler.rs#L272) (used by the plugin system for the help on the status bar). I'd personally put "PGUP" and "PGDN" or some variation that's better capitalized, but you'll still be able to use 'f' and 'b' just fine, even if they aren't in the help :)

Happy to merge after we make that tweak!

horasal:
Added PGUP/PGDN to get_mode_info, new help info looks like:

TheLostLambda:
This looks great! Thanks a ton for adding it!

PR #312, opened 2021-04-21
feat(ci): build release binary for `aarch64-apple-darwin` (#279)

tesaguri:
This adds `aarch64-apple-darwin` (Apple Silicon macOS) target support to #279.

I don't have an Apple Silicon Mac and didn't verify that the binary actually works, but it at least compiles fine.

Cc @qballer.


qballer:
I've pinged a colleague to check it works on his machine (from the fork).

PR #318, opened 2021-04-22
Toward enable `should_render` (fix pane render)

horasal:
This pr is the first step toward enable `should_render`, by fixing pane render when switching/resize tab.

When `should_render` is enabled, pane won't be re-rendered unless its content is changed. When switching tab, panes in target tab may be not changed from its last render. Thus the render requests will be ignored  and those panes' areas will mess up.

This pr add a `set_force_render` to `Tab`, which will clear the `should_render` flag of panes belongs to this tab. Then `tab::rander` call can rerender all panes.

currently, `set_force_render` is called when:

- go to tab
- switch tab next
- switch tab prev
- terminal resize

pane `should_render` will be true if

- reflow_lines happens
- new character is written to terminal
- scroll


horasal:
All tests passed after enable `should_render` and remove `cfg(test)`, results can be found at [actions](https://github.com/zellij-org/zellij/runs/2407952755) for commit https://github.com/zellij-org/zellij/pull/318/commits/4c94f573b867251204faa251335c212b3d7beb30

After that, https://github.com/zellij-org/zellij/pull/318/commits/d949a99cf0115824d44114f01f926b900cfae01b changes `should_render()` to `true` because we still have wide char issue.
https://github.com/zellij-org/zellij/pull/318/commits/8c474016b74087b3c3ab54ed7babce348b4d0a40 and https://github.com/zellij-org/zellij/pull/318/commits/aa25ba26927169f7e3832ce9fdcd627c23e00ac1 do `cargo fmt` and `cargo clippy`.

src/client/tab.rs
```diff
@@ -656,6 +656,11 @@ impl Tab {
     pub fn toggle_fullscreen_is_active(&mut self) {
         self.fullscreen_is_active = !self.fullscreen_is_active;
     }
+    pub fn set_force_render(&mut self) {
```
imsnif, line 659: Nice. I would have gone with something on the state of `Tab` that bypasses the `should_render` of each pane, but the more I think about it, the more I like this solution.

src/client/panes/terminal_pane.rs
```diff
@@ -315,6 +316,7 @@ impl TerminalPane {
         let rows = self.get_rows();
         let columns = self.get_columns();
         self.grid.change_size(rows, columns);
+        self.set_should_render(true);
```
imsnif, line 319: Elegant!

horasal:
> are you interested in continuing the work on this? As I understand it, the main thing we have left is dealing with the implications of wide-characters on this. Do you have a plan to deal with them?

I did some experiments but still need many works and far from becoming a PR. 

https://github.com/horasal/zellij/commit/e8e8639159ec773cf4c35902d9f4f479ce6fbbf5

(In current zellij, `grid` and `cursor` treat number of character = width of string = position. However for widechar, width can be larger than character so this rule doesn't apply. I'm not sure a major revise on grid is the right way or not)

> I have some ideas otherwise

It will be great if you can make zellij works with widechar!

imsnif:
For sure making Zellij support wide characters is on the roadmap, but as you say this requires a bigger change.
I was thinking we can improve on the current functionality (namely, uncomment that `if`), if we do the following:

1. Identify wide characters as part of the `TerminalCharacter` struct
2. Force a tab re-render if there's a wide character anywhere in the viewport.

A little hacky, but I think it's better than what we have now and will probably give us a performance boost until we can handle them properly. What do you think?

horasal:
It sounds good so I implemented it: https://github.com/horasal/zellij/tree/widechar_force_render
but panes still messed up... Did I miss something?

imsnif:
> It sounds good so I implemented it: https://github.com/horasal/zellij/tree/widechar_force_render
> but panes still messed up... Did I miss something?
> 

Unfortunately, you're not missing anything - this happens in `main` too :D 

Since we don't support wide unicode characters yet, we don't properly detect them in order to line wrap inside a pane. So they end up being line-wrapped by the terminal emulator itself.

This issue is super high priority for us to fix, and the only reason I didn't get to it myself yet is that there are other super high priority fixes on my plate before it :)

That being said, I think it should not interfere with this PR... I'm happy you implemented this idea! Is it ready for a review?

horasal:
> That being said, I think it should not interfere with this PR... I'm happy you implemented this idea! Is it ready for a review?

I agree with you. This pr should not be blocked by widechar problem. This is ready for a review.

imsnif:
Hey, this is great! I'm good to go with this, but I would prefer to go with your solution of the wide characters and then uncomment the `if true` line in the render function. What do you think?

horasal:
merge widechar_force_render, and uncomment `if true` in pane render func.

imsnif:
Thank you very much for your work on this!

PR #355, opened 2021-04-24
Document the server-client API

khs26:
This PR is for reviewing the API that's going to be offered by the server to client(s). Comments and contributions are very welcome!


docs/SERVER-API.md
```diff
@@ -0,0 +1,214 @@
...
+### Write input
+
+**Purpose:** Writes a set of bytes to the terminal, not to be interpeted by zellij.
+
+**Parameters:** Vector of bytes to be written.
+
+**Response:** None.
+
+### Switch input mode
+@@@Is this part of the API or a client detail?
```
khs26, line 21: Is it? I think the decision to make is if we want modes to be implemented by all the different clients, or by the server. Could we have both?

If we want it to be implemented just by clients, should we offer a helper library and in what language?

docs/SERVER-API.md
```diff
@@ -0,0 +1,214 @@
...
+
+**Response:** @@@do we have a lock here?
+
+### Open file
+
+**Purpose:** Indicates that a new terminal should be spawned.
+
+**Parameters:** File name (optional) - the name of a file to be opened, using a terminal application determined by first the `EDITOR` and (if `EDITOR` is not set) `VISUAL` environment variables.
+
+**Response:** @@@do we have a lock here?
```
khs26, line 59: This feels like an odd API. Do we actually just want the "new client" API below to optionally take some config to open a layout, with each pane in the layout optionally taking a filename to be opened using the described heuristic?

docs/SERVER-API.md
```diff
@@ -0,0 +1,214 @@
...
+
+## Pane controls
+
+### New pane
+
+**Purpose:** Opens a new [pane](./TERMINOLOGY.md#pane).
+
+**Parameters:** Direction (optional) - the position of the new pane relative to the focus pane.
+
+**Response:** Done opening new pane - indicates that the server has finished opening the pane. This exists for the purpose of synchronization @@@what's blocked?
```
khs26, line 69: @imsnif I don't understand the various locks we have between client and server. Is it something that the client should care about at all? Should a client block in any way? If not, then we probably want to isolate this to the server side.

docs/SERVER-API.md
```diff
@@ -0,0 +1,214 @@
...
+
+**Parameters:** None.
+
+**Response:** Finished opening new pane - indicates that the server has finished closing the pane. This exists for the purpose of synchronization @@@what's blocked?
+
+### Switch focus pane
+
+**Purpose:** Switch the focus [pane](./TERMINOLOGY.md#pane). The focus pane is the target for many other pane-related actions.
+
+**Parameters:** @@@none?
```
khs26, line 83: I think we want to have some option of moving in a particular direction. The `Action` enum does currently have that in the docs, but not the implementation(!) :)

docs/SERVER-API.md
```diff
@@ -0,0 +1,214 @@
...
+
+**Response:** None.
+
+### Move focus pane
+
+**Purpose:**
+
+**Parameters:** 
+
+**Response:**
```
khs26, line 93: I don't think I know what this does clearly enough to write it up. Does it take the focus pane and swap it with whichever pane is its immediate neighbour in the direction specified?

docs/SERVER-API.md
```diff
@@ -0,0 +1,214 @@
...
+
+**Purpose:** Resizes the focus [pane](./TERMINOLOGY.md#pane). 
+
+**Parameters:** Direction in which the pane should be resized. @@@ When does it shrink, when does it grow? Is it the top left that's fixed?
+
+**Response:** None.
+
+### Horizontal split
+
+**Purpose:** Split the current focus [pane](./TERMINOLOGY.md#pane) horizontally, creating a new pane @@@above or below?
```
khs26, line 113: On this and the next one, I don't know in what direction the new panes are spawned. @imsnif ?

docs/SERVER-API.md
```diff
@@ -0,0 +1,214 @@
...
+
+**Parameters:** None.
+
+**Response:** None.
+
+### Go to tab
+
+**Purpose:** Switches to a specific [tab](./TERMINOLOGY.md#tab).
+
+**Parameters:** Index - the index of the tab within the list. @@@ Shouldn't this really be based on name? Index is quite internal.
```
khs26, line 166: There might be a use for switching by index, but I think if there is, we want to also share the indices with the client in the previous APIs.

docs/SERVER-API.md
```diff
@@ -0,0 +1,214 @@
...
+
+### Rename tab
+
+**Purpose:** Renames the current [tab](./TERMINOLOGY.md#tab).
+
+**Parameters:** A vector of bytes containing the new name for the tab.
+
+**Response:** None.
+
+## Session management
```
khs26, line 186: I think we also probably want:
- Detach (other) client
- Attach to session
- Get session name
- Rename session

docs/SERVER-API.md
```diff
@@ -0,0 +1,214 @@
...
+
+### New client
+
+**Purpose:** Indicates that a new client is connecting to the server.
+
+**Parameters:**
```
khs26, line 192: As described above, I think we could do with merging the open file API into this.

docs/SERVER-API.md
denismaxim0v: I don't know if we really need to have this on the server to be honest. But if we really do I think it would be easier to have them on both sides. Or am I missing something?

kunalmohan: The plugins run on the server-side. So server needs to be aware of the current input mode.

docs/SERVER-API.md
kunalmohan: The new pane is spawned below in this case and towards the right in the next one.

docs/SERVER-API.md
kunalmohan: This is actually a good point. We should not try to create atomicity in operations across processes. Instead of blocking the input thread on the client-side, we should consider blocking the router thread (only if we decide to have a separate router thread for each client).

docs/SERVER-API.md
```diff
@@ -0,0 +1,214 @@
...
+- The response that the client can expect to receive, including any possible error messages. N.B. Response does not detail side-effects of the request, in particular changes in what's rendered on screen.
+
+The document does not describe implementation details on the server side and is intended for use by those implementing new zellij clients.
+
+
+## Basic controls
+
+### Write input
+
+**Purpose:** Writes a set of bytes to the terminal, not to be interpeted by zellij.
```
imsnif, line 14: I'd change this (just to be explicit) to: "Writes a set of bytes to the focused pane"

docs/SERVER-API.md
imsnif: I agree with @kunalmohan. I think blocking the router thread on the server side is a good solution, and we can probably get away with just 1 kind of lock. As long as we make sure that one action that would cause a lock doesn't spawn another action that can release it.

eg. If we want the "new tab" action to lock the router thread until the new tab is open, we shouldn't allow the "new pane" action to release the lock, because the "new pane" option can be taken as a result of the "new tab" action. In the current state of affairs this is a non-issue, but if we imagine opening a new tab with a complex 10+ pane layout, this might become one...

docs/SERVER-API.md
imsnif: This currently does work though... am I misunderstanding?

docs/SERVER-API.md
imsnif: Yeah, in the UI we already changed this to the (IMO clearer) "split_right" / "split_down", might be a good idea to change it in the code-base and thus the API too :)

docs/SERVER-API.md
imsnif: So are you suggesting that this should only be possible through a layout?

imsnif:
@khs26 - I'm happy we're having this conversation :) I left some comments directly in the document.

Otherwise: we seem to be giving a flat API and also allowing the user to switch between input modes. I understand wanting to give client developers the flexibility, but this seems a little confusing to me. As in: if I can always send the "split_horizontally" action, why do I need to switch to pane mode?

I'm not sure what I think a good solution to this might be, tbh. Did you have any thoughts on the subject?

a-kenji:
Closing this for now, since this seems to have gone stale.

PR #369, opened 2021-04-25
Add subcommand to generate completions, remove build.rs

matthiasbeyer:
As promised, this is the PR to change the CLI completion generation to not be in a build-time process, but rather in a subcommand for the binary itself.

Tell me what you think!

---

After working the first time with structopt, I must say it is way worse than I expected. Using plain clap just feels so much more powerful than structopts interface. Just as a note, not the concern of the zellij project of course.


a-kenji:
I personally am happy with it in general.
Maybe we want to put the completions in a file called `setup.rs` or put it in `install.rs` in order to not fill up the main too much?

cc @henil  @TheLostLambda @qballer 

Is this something we can merge soon, or do we need to change quite a few things in the build infrastructure for that? I am not sure how we have used the completions in the past.

>After working the first time with structopt, I must say it is way worse than I expected. Using plain clap just feels so much more powerful than structopts interface. Just as a note, not the concern of the zellij project of course.

I have already lost a few hours to structopt weirdness. I think for now it is ok. We might want to or even need to switch to clap at some point, since structopt is limited in what it can derive. I personally am eager to try out v3 once it stabilizes.

a-kenji:
>Maybe we want to put the completions in a file called setup.rs or put it in install.rs in order to not fill up the main too much?

Ah, no I looked at the file again it should be totally fine for now.

matthiasbeyer:
I am using clapv3 beta for a few months now in one project at work and it is awesome.

I think the main won't get filled too much if you'd use the clap interface in a different way. That might be included in a refactor from structopt to clap, I guess.

matthiasbeyer:
Pushed a fixup commit for formatting, will rebase before merge.

Not sure about the test failure tho.

a-kenji:
>I am using clapv3 beta for a few months now in one project at work and it is awesome.

Nice!

>Not sure about the test failure tho.

If I understand it correctly, sometimes the snapshot tests get confused, if they are too close to each other.

matthiasbeyer:
> If I understand it correctly, sometimes the snapshot tests get confused, if they are too close to each other.

ah whut? How?

henil:
> Is this something we can merge soon, or do we need to change quite a few things in the build infrastructure for that? I am not sure how we have used the completions in the past.

I am out of loop for sometime, so cannot say for sure but I think it should not require any big changes. Simply deleting `build.rs` should be fine.

matthiasbeyer:
> Simply deleting build.rs should be fine.

That is done in this PR :smile:

a-kenji:
Awesome!

matthiasbeyer:
Yeah, that rebase was still pending. Either way... Thanks.

a-kenji:
Oh, right. My bad!

PR #391, opened 2021-04-27
Config Loading

a-kenji:
* Configuration is now handled in this order:
- --config flag
- ZELLIJ_CONFIG_FILE env
- config_dir flag
- ZELLIJ_CONFIG_DIR env
-  get_default_config_dir() (directories-next)
- $HOME/.config/zellij (for mac convenience)
- SYSTEM_DEFAULT_CONFIG_DIR (for distributions to specify sensible defaults ontop of zellij)
-  from assets

* Fix default.yaml file

* Move default.yaml file to assets directory

@qballer Could you check, if the config is loaded from `.config/zellij/config.yaml` now please? In order for that to happen the original config directory on mac shouldn't exist.

The default keys are now not anymore specified in the code, but taken from the `default.yaml` file.

I am happy about suggestions or any input on this pr!


TheLostLambda:
From a first glance, this is looking pretty good! Certainly quite a few different options now! One thing I think we should be sure to do is document this well. Your description in this PR is good, but should probably make it's way somewhere more permanent too :slightly_smiling_face:

a-kenji:
@TheLostLambda 
Cheers! Would it be possible to get the documentation back in this repo 
at some point and have the website render everything out?

src/common/install.rs
```diff
@@ -39,3 +42,44 @@ pub fn populate_data_dir(data_dir: &Path) {
         }
     }
 }
+
+pub fn default_config_dir() -> Option<PathBuf> {
+    vec![
```
TheLostLambda, line 47: Could maybe use an `or_else()` chain here if you'd like. Similar to [here](https://github.com/zellij-org/zellij/blob/169e25af66fe534a5091583b4b5f62dd1b7eea9b/src/common/mod.rs#L473)

Also, does just an array work here? Just `[]` instead of `vec![]`? I suppose you wouldn't need either with the `or_else()` chain though :slightly_smiling_face:

a-kenji: Cheers, that is good! I'll try that out.

TheLostLambda:
> Cheers! Would it be possible to get the documentation back in this repo
> at some point and have the website render everything out?

Yes, I think so, and that would be a much much better way to do things! I'm not sure when I'll have the time, but I'll bump that up my todo list. It's just some trickery with GitHub actions I think.

a-kenji:
>Yes, I think so, and that would be a much much better way to do things! I'm not sure when I'll have the time, but I'll bump that up my todo list. It's just some trickery with GitHub actions I think.

Awesome, no rush. Just good to know that its a possibility!


@TheLostLambda 
And also, what would be a good way to set for SYSTEM_DEFAULT_CONFIG_DIR a prefix on compilation?
I just have seen conditional compile flags so far, not one where we can pass stuff in yet, for this here it actually shouldn't matter too much but if we want to do it in the future.

TheLostLambda:
> I just have seen conditional compile flags so far, not one where we can pass stuff in yet, for this here it actually shouldn't matter too much but if we want to do it in the future.

I could be mistaken, but I think things like `env!()` will do that. All of the macros with the `!` run at compile time, as far as I'm aware. There are probably even more macros for doing stuff at compile-time like that.

a-kenji:
@TheLostLambda 
Awesome, thanks! Yeah that is what I was looking for.

src/common/install.rs
a-kenji: I don't see a way to implement that or_else() in a good way here. I want the contents, in this case the path, not the value that is returned.

PR #399, opened 2021-04-28
Added Screen Config File: screen-overview.yaml and screen.nonfunctional,

### LINKED ISSUE #286: Feature(Config): Emulate Screen Default Keybindings.
We want to provide sane defaults for people, 
so having a configuration file that emulates the screen config makes sense.
Inspiration in `/example/config.yaml` and `example/default.yaml`,
as well as the keybindings [documentation]( https://zellij.dev/documentation/keybindings.html).
`/example/screen.yaml` could be a suitable location.

SaintFenix:
closes #286 

This is an early implementation of `screen` -like bindings for Zellij using the [GNU screen Manual](https://www.gnu.org/software/screen/manual/screen.html) and the `<prefix> ?` command in Screen itself from version `4.08.00`

It follows the same formatting as my previous pull request: #362

In that pull Request @a-kenji and I worked though these as possible control schemes, while providing placeholder names for features that Zellij may eventually implement that will copy or exceed the functionality of screen in this case.

Many of the functions of screen are exceptionally archaic or out of place in Zellij and were included in the lists commented out just for completion sake. I have made a similar [screenjustification.txt](https://github.com/zellij-org/zellij/files/6388832/screenjustification.txt) as before to see my reasoning and why some commands work and others. This one doesn't line up as neatly as tmux due to how screen formats its output, but I attempted to keep it roughly the same.


example/screen-overview.yaml
```diff
@@ -0,0 +1,352 @@
+---
+#This file functions as an overview of using screen binds in place of the lock mode, it currently functions fine in Zellij, but doesn't give the user the opportunity to use lock mode.
```
a-kenji, line 2: Can you put that in 2 lines please?

a-kenji (changes requested):

a-kenji:
This is great, thank you!

example/screen-overview.yaml
SaintFenix: Done and done.

SaintFenix:
Got that edited for ya. 👍

a-kenji (approved):

a-kenji:
Cheers!

PR #400, opened 2021-04-28
refactor(handler): refer to self.config.keybinds internally

yykamei:
I believe this is a trivial change, so I created a PR directly.

Because `handle_key` can refer to `self.config.keybinds` by itself, it doesn't have to receive the third argument. Also, this change may slightly improve the performance as it deletes `clone()`.

By the way, I love zellij and started to use it on my laptop.
Thanks,


a-kenji:
Thank you very much! 
This makes a lot of sense.

Issue #313, opened 2021-04-21
More lightline like toolbars [Feature Request]


alexhallam:
I have a very minimal set up.

When I open the multiplexer this is what my toolbar looks like.

![image](https://user-images.githubusercontent.com/9298693/115584999-3e121300-a299-11eb-910b-bec97580e684.png)

I would like to request a simplified version that does not have the font/glyph deps. Something like a light line option (see https://github.com/itchyny/lightline.vim)

![image](https://user-images.githubusercontent.com/9298693/115585527-b11b8980-a299-11eb-8654-09319b39f07e.png)

a-kenji:
Thank you! Currently we have no configuration system for themes and plugins implemented yet, but the Font issue has come up often enough, that I think it is reasonable to try to build the bar with a more general font and include it by default, so people can switch it out in their default.yaml - as an intermediary fix. I am curious what other people think.

cc @imsnif

keeslinp:
Worst case someone could just make a new wasm plugin and load a different layout right? (or just replace the `status-bar.wasm` file I guess)

imsnif:
A relatively quick fix for this (in case users don't want to install new fonts) is:

1. Create versions of the default plugins (status-bar and tab-bar) as @keeslinp suggested, which do not include the arrow font (or include a different separator instead it, maybe a pipe?)
2. Create a "simplified.yaml" layout which is the copy of the default layout and loads them instead
3. Run: `zellij --layout simplified`

If anyone wants to work on this, I'd be happy to merge this PR. I think this would be valuable to include with Zellij.

Bonus points if you can come up with a better name than "simplified" :)

alexhallam:
I can give this a try. But I have a couple questions first:

1. For new versions of `status-bar` and `tab-bar` I am guessing I just cp `default-plugins/status-bar/src/main.rs` and `./default-plugins/tab-bar/src/main.rs` to something like `default-plugins/status-bar-simple/src/main.rs` and `./default-plugins/tab-bar-simple/src/main.rs` and make the edits. From there I will have to read the docs.
2. I see `./example/default.yaml` I am just going to cp that to `./example/simplified.yaml` and make the edits.
3. I will have to dig in to find where command line args are called. So I can add the --layout flag.
4. I will skip the bonus points, but if someone else has a better name than simplified I am all ears. haha

a-kenji:
Hey @alexhallam , that would be awesome!

[...]

PR #416, opened 2021-04-30
Add simplified

alexhallam:
This PR adds a simplified layout. For #313 

# Top tab bar
top_new

# Bottom status bar
bottom

# Design Decision
I originally used the pipe symbol `|`, but found that it was unattractive and repetitive (see below image). 

top

I felt the text and color highlighting was sufficient to delineate selections.


TheLostLambda:
This is a great feature to have, but really duplicates a lot of code.

I wonder if there is a way to detect if the terminal can or can't show these characters at runtime. Is there a way to figure that sort of thing out @imsnif ?

Personally, I would like to add a "capabilities" struct to the plugin `ModeInfo` so that plugins know if they can use these characters or certain colors when they are run by Zellij. That would allow you to pass a flag like `zellij --simplified` to notify all of the plugins to use simpler fonts (even 3rd party ones). Hopefully that would knock the diff down to less than 100 lines without any duplication.

If you're interested in the command-line flag and `ModeInfo` solution, I'm happy to help!

a-kenji:
Thank you @alexhallam !
I am sorry I partly sent you on a false quest here, since after seeing the implementation I am also of @TheLostLambda opinion.
I really like how it turned out though.

>Personally, I would like to add a "capabilities" struct to the plugin ModeInfo so that plugins know if they can use these characters or certain colors when they are run by Zellij. That would allow you to pass a flag like zellij --simplified to notify all of the plugins to use simpler fonts (even 3rd party ones).

That would be a very elegant solution.

imsnif:
> I wonder if there is a way to detect if the terminal can or can't show these characters at runtime. Is there a way to figure that sort of thing out @imsnif ?

I don't know of such a way :/ Maybe when we have a setup wizard we can ask the user if they can see the character properly. Otherwise I also like the --simplified solution

a-kenji:
Until that is implemented would it maybe be good to ship anyway with these extra plugins? But the characters are switched out by compile time options? That way we could generate them from the same file and the overhead of maintaining them wouldn't be so huge, until we find a more permanent solution.

And we could still use this `simplified.yaml` layout file, for people that have font problems, and they could simply just move  that to `default.yaml` if they are so inclined, or start with `--layout simplified`

alexhallam:
If the arrow symbol is tied to a font then `fc-list` can output all fonts and the font of interest could be grepped. If the arrow symbol is a unicode issue then maybe we can use `'\u{003E}'.to_string()` as an alternative. If it is a terminal issue then we can keep a list of compatible terminal names that can be checked. I am not sure if the issue is a font problem or a unicode problem or a terminal problem, something else. If you want I can try to figure that out.

a-kenji:
@alexhallam That would be awesome!
As far as  I know this is a missing font problem, `nerdfonts` most notably include these symbols.
But there could be more cases where that doesn't work.
Still, supporting people that can't show this should be done imo.

alexhallam:
Okay, it would only be a partial solution because `fc-list` is a linux only solution. Also, I am not sure if there is stigma against using system commands in Rust code. What are your thoughts.

TheLostLambda:
> Until that is implemented would it maybe be good to ship anyway with these extra plugins? But the characters are switched out by compile time options?

I was thinking this too and nearly included it in my first comment, but after thinking about it, I think adding that "capabilities" struct would actually be quicker and easier :sweat_smile: 

I reckon that if you get some value from the `--simplified` flag into a struct like this in `zellij-tile/src/data.rs`:
```rust
struct PluginCapabilities {
    arrow_fonts: bool
}
```
then pass it to [this function](https://github.com/zellij-org/zellij/blob/3c431ee6254858e5a46fd0870e5aff47f57f64e3/src/common/input/handler.rs#L287), that's most of the work done. Then each plugin just needs to read that struct and switch to simple chars.

Let me know what you think of that and if you'd like to pursue that path! I'd personally avoid system commands where possible (simply for the sake of portability and having a static binary), but there are some places I'm sure where they are appropriate.

a-kenji:
@alexhallam 
>Okay, it would only be a partial solution because fc-list is a linux only solution. Also, I am not sure if there is stigma against using system commands in Rust code. What are your thoughts.

Thanks! That seems hard to get right though!


@TheLostLambda 
 
>I was thinking this too and nearly included it in my first comment, but after thinking about it, I think adding that "capabilities" struct would actually be quicker and easier sweat_smile

Especially in the long run. This is definitely a more permanent solution!

a-kenji:
@TheLostLambda 

I am looking at implementing this right now and I have trouble coming up with a good way
passing the potential `opts.simplified` to the `ModeInfo` struct. Do you have any potential pointers?

a-kenji:
@TheLostLambda Nvm, I think I figured it out!

TheLostLambda:
@a-kenji nice! Sorry I couldn't get back to you in time, I've been a bit tied up! What solution have you gone with?

a-kenji:
@TheLostLambda No problem! And sorry that I almost always seem to tag you, most of what I do seems to correlate often.

I think I will be able to have a draft rdy today or tomorrow!

a-kenji:
Thanks @alexhallam for starting this, otherwise we probably wouldn't be as far right now!

Im closing this in favor of #479.

alexhallam:
@a-kenji My pleasure.

PR #423, opened 2021-04-30
Load config before start()

### LINKED ISSUE #414: Let a user know why the configuration is invalid
Hi,

I'm curious whether Zellij can load a config file before calling [`start`](https://github.com/zellij-org/zellij/blob/b5d38e95c876f231a1c52ab2bcf805f65958d56c/src/common/mod.rs#L123). When I deliberately write an invalid configuration, Zellij exits with the status code 1 without any output. It would be inconvenient for a Zellij user, at least me, because the user should want to know why Zellij does not launch.

```console
$ cat config.yaml
keybinds: invalid
$ zellij --debug --config config.yaml
(Nothing appears)
```

I guess this is because the config loading starts **after** Zellij enters an alternative screen buffer. The most straightforward way to solve this might be just to disable an alternative screen buffer when Zellij detects the passed configuration is invalid. However, I think Zellij can load configuration before calling `start`, and it's more reasonable as Zellij doesn't have to set up the terminal when the config file is wrong. If it's good, I'm going to open a PR.

FYI, this is output after disabling the alternative screen buffer:

```console
$ zellij --version  --debug --config config.yaml; echo -e '\e[?1049l'
There was an error in the config file:
Deserialisation error: keybinds: invalid type: string "invalid", expected struct KeybindsFromYaml at line 1 column 11

```

By the way, it'll get even better to provide a feature to let users reload a configuration, but it's out of scope so far.

## Information

`zellij --version`: `0.7.0` (commit: b5d38e95c876f231a1c52ab2bcf805f65958d56c)

yykamei:
Close #414

Previously, a config file was loaded within `start()`, and if the config file is invalid, Zellij was supposed to show a user what's wrong with it. However, since `start()` starts setting up its terminal with an alternative screen buffer, neither standard output nor standard error could display such an error.

This change intends to address this issue by making Zellij load a config file before `start()`.

In addition, the patch also includes some refactorings:

* Redefine `from_cli_config` with `TryFrom`, which was introduced in Rust 1.34
* Remove conditional declaration `cfg(not(test))` because `start()` now receive a `Config` as the third argument
* Introduce [`tempfile`](https://crates.io/crates/tempfile) in order to run tests with actual files
* Typo?: "Deserialisation" -> "Deserialization"

### DIFF
--- a/src/common/input/config.rs
+++ b/src/common/input/config.rs
@@ -6,14 +6,17 @@ use std::io::{self, Read};
 use std::path::{Path, PathBuf};
 
 use super::keybinds::{Keybinds, KeybindsFromYaml};
-use crate::cli::ConfigCli;
+use crate::cli::{CliArgs, ConfigCli};
 use crate::common::install;
 
 use serde::Deserialize;
+use std::convert::TryFrom;
+
+const DEFAULT_CONFIG_FILE_NAME: &str = "config.yaml";
 
 type ConfigResult = Result<Config, ConfigError>;
 
-/// Intermediate deserialisation config struct
+/// Intermediate deserialization config struct
 #[derive(Debug, Deserialize)]
 pub struct ConfigFromYaml {
     pub keybinds: Option<KeybindsFromYaml>,
@@ -27,13 +30,13 @@ pub struct Config {
 
 #[derive(Debug)]
 pub enum ConfigError {
-    // Deserialisation error
+    // Deserialization error
     Serde(serde_yaml::Error),
     // Io error
     Io(io::Error),
     // Io error with path context
     IoPath(io::Error, PathBuf),
-    // Internal Deserialisation Error
+    // Internal Deserialization Error
     FromUtf8(std::string::FromUtf8Error),
 }
 
@@ -41,6 +44,35 @@ impl Default for Config {
     fn default() -> Self {
         let keybinds = Keybinds::default();
         Config { keybinds }
+    }
+}
+
+impl TryFrom<&CliArgs> for Config {
+    type Error = ConfigError;
+
+    fn try_from(opts: &CliArgs) -> ConfigResult {
+        if let Some(ref path) = opts.config {
+            return Config::new(&path);
+        }
+
+        if let Some(ConfigCli::Config { clean, .. }) = opts.option {
+            if clean {
+                return Config::from_default_assets();
+            }
+        }
+
+        let config_dir = opts.config_dir.clone().or_else(install::default_config_dir);
+
+        if let Some(ref config) = config_dir {
+            let path = config.join(DEFAULT_CONFIG_FILE_NAME);
+            if path.exists() {
+                Config::new(&path)
+            } else {
+                Config::from_default_assets()
+            }
+        } else {
+            Config::from_default_assets()
+        }
     }
 }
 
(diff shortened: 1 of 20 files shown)


src/common/input/config.rs
```diff
@@ -6,14 +6,17 @@ use std::io::{self, Read};
...
 
 use serde::Deserialize;
+use std::convert::TryFrom;
+
+const DEFAULT_CONFIG_FILE_NAME: &str = "config.yaml";
 
 type ConfigResult = Result<Config, ConfigError>;
 
-/// Intermediate deserialisation config struct
+/// Intermediate deserialization config struct
```
a-kenji, line 19: Good catch!

src/common/input/config.rs
```diff
@@ -44,6 +47,34 @@ impl Default for Config {
     }
 }
 
+impl TryFrom<&CliArgs> for Config {
```
a-kenji, line 50: This is awesome! Thanks for putting `TryFrom` on my radar.

src/common/input/config.rs
```diff
@@ -44,6 +47,34 @@ impl Default for Config {
...
+    type Error = ConfigError;
+
+    fn try_from(opts: &CliArgs) -> Result<Self, Self::Error> {
```
a-kenji, line 53: Does it make sense to use `ConfigResult` here, or do you think that this is clearer?

src/common/input/config.rs
```diff
@@ -6,14 +6,17 @@ use std::io::{self, Read};
...
 
 use super::keybinds::{Keybinds, KeybindsFromYaml};
-use crate::cli::ConfigCli;
+use crate::cli::{CliArgs, ConfigCli};
 use crate::common::install;
 
 use serde::Deserialize;
+use std::convert::TryFrom;
+
+const DEFAULT_CONFIG_FILE_NAME: &str = "config.yaml";
```
a-kenji, line 15: This makes a lot of sense.

a-kenji (approved):

src/common/input/config.rs
yykamei: Thank you. It makes sense :+1:

a-kenji:
Thank you, this is awesome!
This one test shouldn't be a big problem to fix.

>Does it make sense to use ConfigResult here, or do you think that this is clearer?

This is just a small thing.

>Remove conditional declaration `cfg(not(test))` because `start()` now receive a `Config` as the third argument

This is quite an awesome side benefit of this change! Now even parts of changes are possible to test reasonably well in the integration tests. Good job!

>Introduce tempfile in order to run tests with actual files

Also a good idea!

> Typo?: "Deserialisation" -> "Deserialization"

This was a translation error, thanks for catching that.

Just write a message if you get that test fixed and feel its ready for a proper review, then I'll have another look at it.

yykamei:
Thanks for your quick review in spite of such a large diff :pray: 

One of the tests always fails on GitHub Actions. I'm looking into this, but I'm not sure why it happens...

src/common/input/config.rs
```diff
@@ -44,6 +47,35 @@ impl Default for Config {
...
+            return Config::new(&path);
+        }
+
+        if let Some(ConfigCli::Config { clean, .. }) = opts.option {
+            if clean {
+                return Config::from_default_assets();
+            }
+        }
+
+        let config_dir = opts.config_dir.clone().or_else(install::default_config_dir);
```
yykamei, line 64: I misunderstood the behavior of `default_config_dir`, which I thought would create a directory.
`default_config_dir` just indicates a possible configuration directory, so it should be called here as the previous implementation.

By the way, as `default_config_dir` returns `Option<PathBuf>`, I made `opts.config_dir` call `clone()` to match these types. Maybe, I can delete this `clone()` by refactoring `default_config_dir`, but it should be done in another PR.

yykamei:
Please ping me if rebase is required!

src/common/input/config.rs
a-kenji: >I misunderstood the behavior of `default_config_dir`, which I thought would create a directory.
`default_config_dir` just indicates a possible configuration directory, so it should be called here as the previous implementation.

Yeah, I glossed over that too now. The naming does suggest some sort of creation. I'll put it on my list to try to make the intention clearer.

>By the way, as `default_config_dir` returns `Option<PathBuf>`, I made `opts.config_dir` call `clone()` to match these types. Maybe, I can delete this `clone()` by refactoring` default_config_dir`, but it should be done in another PR.

I would love that, as I am also unhappy about the state of `default_config_dir`, so feel free to open a PR on that, if you want to!

a-kenji (approved):
Thank you! This is looking good. If you want you can rebase, then I'll merge it.

yykamei:
Thank you. I rebased and squashed commits into a single commit!

PR #428, opened 2021-05-01
Refactoring sender/receiver stuff

khs26:
## Changes
- Move inter-thread comms into its own module
- Add a `Bus` and `ThreadSenders` struct to wrap the various senders and receivers for our mpsc channels and the OS input
- Move as much thread work as possible out of the main module into the relevant threads
- Changes requested by `cargo clippy` and `cargo fmt`

## Manual testing
- Opened, closed, resized and resized panes
- Wrote some long text (found a bug in main!) and tested scrolling
- Opened, closed and switched tabs

## Deferred changes
- Actually have proper error handling here (lots of `as_ref().unwrap()` everywhere, though I've tried to hide it from the user)
- Think a bit more about ownership rather than just cloning senders/receivers/OS input when I was unsure #RustProblems


src/common/thread_bus.rs
```diff
@@ -0,0 +1,142 @@
...
+/// A container for senders to the different threads in zellij
+#[derive(Clone)]
+pub struct ThreadSenders {
+    pub to_screen: Option<SenderWithContext<ScreenInstruction>>,
+    pub to_pty: Option<SenderWithContext<PtyInstruction>>,
+    pub to_plugin: Option<SenderWithContext<PluginInstruction>>,
+    pub to_app: Option<SenderWithContext<AppInstruction>>,
+}
+
+impl ThreadSenders {
```
khs26, line 81: I thought about doing something clever with dynamic dispatch, but a) I'm not familiar enough with it in Rust, b) not sure it's worth it.

TheLostLambda:
Eek, the failing tests here seem repo-wide. They aren't working in another PR as well. I'll give this a review anyhow and hope that gets fixed up soonish (they seem to only fail in CI, so perhaps more fun race conditions).

src/common/screen.rs
```diff
@@ -330,3 +314,188 @@ impl Screen {
...
+    full_screen_ws: PositionAndSize,
+) {
+    let mut screen = Screen::new(bus, &full_screen_ws, max_panes, ModeInfo::default());
+    loop {
+        let (event, mut err_ctx) = screen
+            .bus
+            .recv()
+            .expect("failed to receive event on channel");
+        err_ctx.add_call(ContextType::Screen(ScreenContext::from(&event)));
+        match event {
```
TheLostLambda, line 331: It seems that here we have the match in the `thread_main` function, but the `Pty` has a separate function for handling the events (the `match`) and the main loop (`thread_main` function). I'm somewhat partial to the having everything together in the same `thread_main` like this and the WASM thread, that seems quite clean, but I don't really mind either so long as we're consistent :)

src/common/thread_bus.rs
TheLostLambda: With only this many threads, I think this is plenty fine :) To get around this way of doing things, we'd probably need another layer of Enums or to add Traits to things for the dynamic dispatch. I think this is a fine tradeoff to make :)

If you are interested in de-duplicating this region, I reckon it could be done with a decently simple macro :)

src/common/thread_bus.rs
```diff
@@ -0,0 +1,142 @@
...
+    pub fn send_to_app(
+        &self,
+        instruction: AppInstruction,
+    ) -> Result<(), mpsc::SendError<(AppInstruction, ErrorContext)>> {
+        self.to_app.as_ref().unwrap().send(instruction)
+    }
+}
+
+/// A container for a receiver, OS input and the senders to a given thread
+pub struct Bus<T> {
```
TheLostLambda, line 112: A great place for some simple generics! Go developers are weeping

src/common/wasm_vm.rs
```diff
@@ -28,14 +36,97 @@ pub enum PluginInstruction {
 pub struct PluginEnv {
     pub plugin_id: u32,
     // FIXME: This should be a big bundle of all of the channels
```
TheLostLambda, line 38: Hey! We can finally get rid of this comment I left here!

src/common/pty.rs
```diff
@@ -77,16 +77,90 @@ pub enum PtyInstruction {
...
-    pub receive_pty_instructions: Receiver<(PtyInstruction, ErrorContext)>,
+pub struct Pty {
+    pub bus: Bus<PtyInstruction>,
     pub id_to_child_pid: HashMap<RawFd, RawFd>,
-    os_input: Box<dyn OsApi>,
     debug_to_file: bool,
     task_handles: HashMap<RawFd, JoinHandle<()>>,
 }
 
+fn handle_event(
```
TheLostLambda, line 87: We could potentially merge this into `thread_main` for consistency with the other threads :)

TheLostLambda (commented):
This looks great! I left a couple of small nitpicky / maybe helpful comments, but I'm overall very happy with this change :)

kunalmohan (commented):
LGTM! Just a couple of small things to consider. Otherwise looks great!

src/common/thread_bus.rs
```diff
@@ -0,0 +1,142 @@
...
+    pub receiver: Option<mpsc::Receiver<(T, ErrorContext)>>,
```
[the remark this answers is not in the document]
khs26: Ahh yes, that's because I originally didn't have a separate `ThreadSenders` struct for the cases when there's no `receiver`. I can unoptionalise this :)

src/common/thread_bus.rs
```diff
@@ -0,0 +1,142 @@
...
+    pub static ASYNCOPENCALLS: RefCell<ErrorContext> = RefCell::default()
+}
+
+/// A container for senders to the different threads in zellij
+#[derive(Clone)]
+pub struct ThreadSenders {
+    pub to_screen: Option<SenderWithContext<ScreenInstruction>>,
+    pub to_pty: Option<SenderWithContext<PtyInstruction>>,
+    pub to_plugin: Option<SenderWithContext<PluginInstruction>>,
+    pub to_app: Option<SenderWithContext<AppInstruction>>,
```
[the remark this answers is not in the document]
khs26: So I did also think this, but my other thought was that we probably want to generally be explicit about when a thread can talk to another thread, e.g. in general, threads don't send messages to themselves, except actually, the plugin stuff does.

What do you and @TheLostLambda think about that?

src/common/wasm_vm.rs
khs26: Yes - there are other instances of these comments I forgot to fix too!

src/common/pty.rs
khs26: So the lack of consistency is because the first thing the pty thread does when we kick it off is to send itself a `PtyInstruction::NewTab`. I'd rather push for consistency the other way - i.e. add `handle_event` functions for the other thread main functions. Do you think that would be an imprveoemtn?

src/common/thread_bus.rs
khs26: > Eek, the failing tests here seem repo-wide. They aren't working in another PR as well. I'll give this a review anyhow and hope that gets fixed up soonish (they seem to only fail in CI, so perhaps more fun race conditions).

So the tests reliably pass locally and seem to be flaky here (I re-ran and they pass). Do you remember what test it was? I'm not sure I can still see...

khs26:
> Eek, the failing tests here seem repo-wide. They aren't working in another PR as well. I'll give this a review anyhow and hope that gets fixed up soonish (they seem to only fail in CI, so perhaps more fun race conditions).

Do you remember what test it was that failed? We should definitely get a fix (or at least an issue) for that.

src/common/thread_bus.rs
khs26: Yeah, I think on balance I agree. I also think it's good to have the type checking and explicitness (i.e. `send_to_screen` takes a `ScreenInstruction` otherwise you've messed up).

src/common/screen.rs
khs26: Replied to this below.

src/common/wasm_vm.rs
khs26: Think I got most of them...

khs26:
I have a bit of work to do to merge with the colors stuff too, which I'm going to pick up later. So more comments are welcome in the meantime!

src/common/thread_bus.rs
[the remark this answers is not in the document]
kunalmohan: That's a fair point. IMO it becomes a personal choice as to what you're willing to trade-off. I personally would prefer to have all the senders in each `ThreadSender` object and not use them when not required rather than keeping track of each `ThreadSender` object separately. But not a big deal, I'll go with whatever the majority of the folks are comfortable with.

src/common/pty.rs
TheLostLambda: Yeah, I do think that consistency is good here, so I'm fine with pushing that consistency the other way too :)

khs26: Kunal's changes for server stuff means that the pty thread doesn't need to self-pulse any more, so I've gone without the `handle_event` function. :)

Issue #434, opened 2021-05-02
Panic in `hex_to_rgb` called from `load_palette`


yykamei:
I got a panic with `cargo make run` on the commit: d6a49f6.
After running the command, my terminal has got broken with the backtrace like this:

```
[cargo-make] INFO - cargo make 0.32.17
[cargo-make] INFO - Project: zellij
[cargo-make] INFO - Build File: Makefile.toml
[cargo-make] INFO - Task: run
[cargo-make] INFO - Profile: development
[cargo-make] INFO - Execute Command: "cargo" "make" "--disable-check-for-updates" "--no-on-error" "--loglevel=info" "--profile=development" "--allow-private" "--skip-init-end-tasks" "--makefile" "/home/kamei/git/Fork/zellij-org/zellij/Makefile.toml" "build"
[cargo-make][1] INFO - Project: zellij
[cargo-make][1] INFO - Build File: /home/kamei/git/Fork/zellij-org/zellij/Makefile.toml
   2: std::panicking::rust_panic_with_hook
             at /rustc/f97769a2b7bc5a6b3d8f7140613cf26352f2b518/library/std/src/panicking.rs:595:17
   3: std::panicking::begin_panic_handler::{{closure}}
             at /rustc/f97769a2b7bc5a6b3d8f7140613cf26352f2b518/library/std/src/panicking.rs:495:13
   4: std::sys_common::backtrace::__rust_end_short_backtrace
             at /rustc/f97769a2b7bc5a6b3d8f7140613cf26352f2b518/library/std/src/sys_common/backtrace.rs:141:18
   5: rust_begin_unwind
             at /rustc/f97769a2b7bc5a6b3d8f7140613cf26352f2b518/library/std/src/panicking.rs:493:5
   6: core::panicking::panic_fmt
             at /rustc/f97769a2b7bc5a6b3d8f7140613cf26352f2b518/library/core/src/panicking.rs:92:14
   7: core::panicking::panic
             at /rustc/f97769a2b7bc5a6b3d8f7140613cf26352f2b518/library/core/src/panicking.rs:50:5
   8: core::option::Option<T>::unwrap
             at /home/kamei/.rustup/toolchains/beta-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/core/src/option.rs:386:21
   9: zellij::common::utils::shared::hex_to_rgb
             at src/common/utils/shared.rs:48:20
[...]

PR #437, opened 2021-05-02
Make Xrdb Loading More Forgiving

a-kenji:
Attempt at fixing  #434.

The `hex_to_rgb` function propagates the Option now.
If it is None, it tries to load the default Palette in most cases.

### DIFF
--- a/src/common/os_input_output.rs
+++ b/src/common/os_input_output.rs
@@ -1,5 +1,5 @@
 use crate::panes::PositionAndSize;
-use crate::utils::shared::{colors, detect_theme, hex_to_rgb};
+use crate::utils::shared::{default_palette, detect_theme, hex_to_rgb};
 use nix::fcntl::{fcntl, FcntlArg, OFlag};
 use nix::pty::{forkpty, Winsize};
 use nix::sys::signal::{kill, Signal};
@@ -14,7 +14,7 @@ use std::path::PathBuf;
 use std::process::{Child, Command};
 use std::sync::{Arc, Mutex};
 use xrdb::Colors;
-use zellij_tile::data::{Palette, PaletteSource, Theme};
+use zellij_tile::data::{Palette, PaletteSource};
 
 use signal_hook::{consts::signal::*, iterator::Signals};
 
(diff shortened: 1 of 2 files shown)


src/common/utils/shared.rs
```diff
@@ -43,12 +43,33 @@ pub mod colors {
     pub const BLACK: (u8, u8, u8) = (0, 0, 0);
 }
 
-pub fn hex_to_rgb(hex: &Option<String>) -> (u8, u8, u8) {
+pub fn hex_to_rgb(hex: &Option<String>) -> Option<(u8, u8, u8)> {
```
yykamei, line 46: How about removing `Option`? Because `hex_to_rgb` seems to be a util function, I wonder it doesn't have to care about the argument is `Some` or not.

By the way, `Rgb::from_hex_str` returns `Result`, but this function looks like ignoring it with `unwrap`. I think it's OK because `hex_str` is usually a valid color hex in many cases. However, using `expect` might be more helpful for developers to detect errors.

To summarize, I came up with this:

```rust
pub fn hex_to_rgb(hex: &str) -> (u8, u8, u8) {
    let rgb = Rgb::from_hex_str(hex)
        .expect("The passed argument should be valid color hex.")
        .as_tuple();
    (rgb.0 as u8, rgb.1 as u8, rgb.2 as u8)
}
```

yykamei (commented):
LGTM. I like this approach. It makes sense to return with `default_palette()` though some users might use their own terminal with a light background color.

src/common/os_input_output.rs
```diff
@@ -268,10 +268,23 @@ impl OsApi for OsInputOutput {
     fn load_palette(&self) -> Palette {
         let palette = match Colors::new("xresources") {
             Some(palette) => {
-                let fg = hex_to_rgb(&palette.fg);
-                let bg = hex_to_rgb(&palette.bg);
-                let colors: Vec<(u8, u8, u8)> =
-                    palette.colors.iter().map(|c| hex_to_rgb(c)).collect();
+                let fg = if let Some(foreground) = hex_to_rgb(&palette.fg) {
```
yykamei, line 271: If `hex_to_rgb` should be like [this](https://github.com/zellij-org/zellij/pull/437/files#r624784691) comment, this line might be fixed with the following change (`palette.bg` is also fixed as the same change :smile:)

```suggestion
                let fg = if let Some(foreground) = palette.fg.as_deref().map(hex_to_rgb) {
```

src/common/os_input_output.rs
```diff
@@ -268,10 +268,23 @@ impl OsApi for OsInputOutput {
...
+                let bg = if let Some(background) = hex_to_rgb(&palette.bg) {
+                    background
+                } else {
+                    return default_palette();
+                };
+
+                let colors: Vec<(u8, u8, u8)> = palette
+                    .colors
+                    .iter()
+                    .map(|c| hex_to_rgb(c).unwrap())
```
yykamei, line 286: :thinking: In my environment, all elements of `palette.colors` are `None`, so it might be better to consider the case `c` is `None`.

src/common/utils/shared.rs
a-kenji: Thank you, yes that makes a lot of sense. Your solution is very elegant!

src/common/os_input_output.rs
a-kenji: Yes, that makes sense :+1: .

src/common/os_input_output.rs
yykamei: It might be ugly but practical to use `for` instead of `iter()`. This will get fallback to the default when any item of `palette.colors` is `None`.

```rust
let mut colors: [(u8, u8, u8); 16] = [(0, 0, 0); 16]; // NOTE: `16` is the same as the length of `palette.colors`.
for (idx, color) in palette.colors.iter().enumerate() {
    if let Some(c) = color {
        colors[idx] = hex_to_rgb(c);
    } else {
        return default_palette();
    }
}
```

denismaxim0v:
LGTM

denismaxim0v (approved):

src/common/os_input_output.rs
a-kenji: Its not as nice as the `iter()`, but much nicer and more concise as I came up with, since we sadly couldn't use a closure here.

a-kenji:
@yykamei 
Thanks for checking it over and leaving so much good advice!
> It makes sense to return with `default_palette()` though some users might use their own terminal with a light background color.

I hope we will soon-ish be able to have theming options in the config.

PR #441, opened 2021-05-03
Fixes Default Keybinds in Resize Mode #440

### LINKED ISSUE #440: Bug: Resize with `ctrl+R` only switches between panes and does not resize with arrow keys (v0.6.0)
With the latest release 0.6.0, hit `alt+n` to create a new pane. Then `ctrl+R` to jump into resize mode. Then hitting the arrow keys only switches between panes and does not actually resize.

Bug on linux x86_64 (Ubuntu 20.04).

a-kenji:
Closes #440

Changes arrow keybinds from switching focus
to resizing.


imsnif:
@a-kenji - I think this is not right. In Pane mode it needs to be normal arrows (not alt). This is also true to Tab and Scroll. Normal arrows should behave just like hjkl in those cases. Alt should remain the same in all modes

imsnif:
(the alt case is a little weird and doesn't work 100% when another non-alt shortcut with the same key is bound... so it's okay if they just work in normal mode actually)

Sorry for being hectic, just wanted to chime in before the merge :)

a-kenji:
Thanks for checking over this!
I think you got mixed up in the keybindings:

This is the current pane mode:
```
 - action: [MoveFocus: Left,]
          key: [ Char: 'h', Left,]
        - action: [MoveFocus: Right,]
          key: [ Char: 'l', Right,]
        - action: [MoveFocus: Down,]
          key: [ Char: 'j', Down,]
        - action: [MoveFocus: Up,]
          key: [ Char: 'k', Up,]
```

This is the current resize mode:

```
      - action: [Resize: Left,]
          key: [Char: 'h', Left,]
        - action: [Resize: Down,]
          key: [Char: 'j', Down,]
        - action: [Resize: Up,]
          key: [Char: 'k', Up, ]
        - action: [Resize: Right,]
          key: [Char: 'l', Right,]
```

a-kenji:
@imsnif 
No problem this is good input, we want the default configuration to be correct!

Ah, now I understand what you mean with the alt.

imsnif:
True, my bad! I only looked at it very quickly :) Sorry!

Additionally though: the arrow keys don't work in tab/scroll mode. Maybe we can also fix them as part of this?

a-kenji:
I like that idea,
In Scroll Mode: Up/Down Lines, and Right/Left Page?

In Tab Mode: Left/Right: Next/Previous Tab?
Do you got any idea for the Up Keys?

Or other ideas?

imsnif:
That sounds cool :) I would just be happy if scroll up/down arrow and tab left/right arrows work - but these also sound great.

a-kenji:
I tried adding the parity to the scroll and tab mode, could you please do a sanity check on that?

imsnif:
LGTM! :+1:

Issue #175, opened 2021-02-10
Feature: mouse support


imsnif:
We should allow switching pane focus as well as resizing panes with the mouse.
We should also allow copying text from within a pane (and not including the text from the surrounding panes) when marking it with a mouse.

Other mouse suggestions are of course very welcome :)

a-kenji:
Trying to get together the possible issues:
- [ ] handle mouse input at all -> use the suggested NoOp 
- [ ] make it possible to pass through mouse input
- [ ] switching pane focus
- [ ] resizing panes
- [ ] copying text, not just from the terminal, but just inside the focused pane

miguelraz:
After poking @imsnif on Discord I would like to also suggest 
- [ ] mouse scrolling

imsnif:
> After poking @imsnif on Discord I would like to also suggest
> 
>     * [ ]  mouse scrolling

Added, thank you!

zareami10:
Note sure if it is discussed but it would be great if we could also switch (and maybe at some point reorder) tabs with mouse.

a-kenji:
@zareami10 
Thank you for the input, yes that is definitely planned,
a prerequisite for that would be #607.

icyJoseph:
What is the mouse support status? 

I have a bad habit of using my touchpad to click on `vim`, and it really prevents usage of zellij to develop, instead I use it to monitor logs in Alacritty. Is there anything I can do already?

I've tried to google and navigate the issues looking for mouse support, and it seems that the information is kinda fragmented, most likely because the feature is not fully supported yet.

imsnif:
Hey @icyJoseph - right now Zellij has mouse support for some of its own UI (eg. changing pane focus, scrolling inside a pane). There's work being done to make it send mouse events to the programs running inside the terminal (eg. to vim) here: https://github.com/zellij-org/zellij/pull/624

benruijl:
I'd love to have mouse support to change tabs. Would that be difficult to add?

imsnif:
That would be cool... @TheLostLambda - where are we on accepting mouse events in plugins? I'm out of the loop on this one I'm afraid.

a-kenji:
@imsnif 
He seems to be saying this could be quite close to being merged #629.
I don't know if qepasa has still some reservations here though.

qepasa:
[...]

PR #448, opened 2021-05-03
Initial mouse support

tlinford:
* enable mouse support (termion MouseTerminal)
* handle scroll up and down events

This is a very initial WIP for #175 now, with quite a few things that need discussing in more detail, but I figure it's a start:
* have mouse support be an option?
* once mouse reporting is turned on it left clicks are captured, but shift+left click allows text selection; right clicks seem to be forwarded normally, not sure how/if mouse input can be passed through


imsnif:
Hey @tlinford - happy to see you taking this up!
I haven't yet tried this out or looked at the code, but to some of your concerns:
We have to have the ability to let users mark text with left-click. Ideally the marking should stop at the boundaries of the pane, but this is very much an advanced feature. :)
Do you think it would be possible to still leave text-marking, even if we implement it ourselves?

tlinford:
Hi @imsnif ,
After some tinkering I've gotten changing the focused pane on left click working decently, but the code is a bit of a mess.

We can implement text-marking ourselves for sure, but as you say it's going to be an advanced thing with quite a lot of work to get it going
My idea at this point is that one needs to send clicks to the correct pane and the pane can figure out what to do in regards to text selection.
Also have to figure out how to have click events sent to plugins, that way we could change tabs in the tab-bar, or turn modes on/off etc.

imsnif:
> We can implement text-marking ourselves for sure, but as you say it's going to be an advanced thing with quite a lot of work to get it going

Right. The issue is that we can't publish a version that doesn't allow text-marking. I think it will kind of render the app unusable (not being able to copy/paste from the terminal). Maybe there's some sort of workaround, or?

> Also have to figure out how to have click events sent to plugins, that way we could change tabs in the tab-bar, or turn modes on/off etc.

That would be very cool!

> Also, are you thinking of having mouse support always enabled, or should it be a toggle?

IMO it should be always enabled.

tlinford:
> Right. The issue is that we can't publish a version that doesn't allow text-marking. I think it will kind of render the app unusable (not being able to copy/paste from the terminal). Maybe there's some sort of workaround, or?

Agreed, the only thing that I've found for now is that shift click bypasses mouse capture so that can be used. Haven't found a way to capture the events and let them pass through yet.

imsnif:
If it's a termion issue, I am very open to looking at other solutions (eg. crossterm) if it helps...

tlinford:
I could easily be wrong about this, as I don't really know much about terminal escapes codes, but from what I found it's not really a termion thing. Once the csi sequence escape codes (termion uses csi!("?1000h\x1b[?1002h\x1b[?1015h\x1b[?1006h")) to enable mouse reporting are written, shift click is the only way to inhibit that behaviour.

imsnif:
> I could easily be wrong about this, as I don't really know much about terminal escapes codes, but from what I found it's not really a termion thing. Once the csi sequence escape codes (termion uses csi!("?1000h\x1b[?1002h\x1b[?1015h\x1b[?1006h")) to enable mouse reporting are written, shift click is the only way to inhibit that behaviour.

Good find! I played a little bit with this sequence and it seems that the one part of it that actually stops mouse reporting tot he terminal is `[?1002h`. The most readable reference I could find to it is in the alacritty source code: https://github.com/alacritty/alacritty/blob/master/alacritty_terminal/src/ansi.rs#L572 which refers to it as `ReportCellMouseMotion`. Whereas `ReportMouseClicks`, which is what we want, is `[?1000h`. So I think this might be a sort of catchall by termion and we could actually do well with just part of it.

So if this is not something that's possible with termion, maybe it's possible with crossterm? Or something else? Or maybe we can somehow hack around this by immediately cancelling the `ReportCellMouseLocation` to the user's terminal by sending a `[?1002l` signal back? What say you?

tlinford:
Hey @imsnif,
things are still a bit rough, but for now we have:

- select active pane with mouse
- scroll pane content with mouse wheel
- select text within a pane, with initial copy to clipboard, for now on wayland only, since that's what I'm using (https://github.com/alacritty/copypasta won't work on wayland as it needs a surface)

The text selection is still buggy when selecting across multiple panes, but I'd appreciate some feedback on the direction this is going

imsnif:
Hey @tlinford - I didn't have a whole lot of time today to take a deep look at this. I plan on doing so tomorrow. Meanwhile my thoughts on copy-paste: I would recommend going in a different direction. The problem with using something that interacts directly with the clipboard is that aside from having to support many different platforms, we would encounter problems when Zellij is run on a remote server.

There is a terminal signal that tells the terminal emulator to place something in the clipboard (OSC 52). Meaning we'd be sending it to stdout and the terminal emulator would do the rest. I only did some brief searching for it, and found this generally related article: https://jdhao.github.io/2021/01/05/nvim_copy_from_remote_via_osc52/

So if you're having trouble finding the exact syntax of the signal let me know and I'll do some more thorough research.

imsnif:
@tlinford - I took a deeper look and also tried this. It looks really good. As you said, still a little rough but I'm very happy to see where this is going.

A behaviour I noticed that needs changing: it's possible to select text in more than one pane at the same time (not sure if this is what you meant with the multiple pane bugginess?)

Other than that, let me know if you want a hand with the OSC stuff. Let me know when you're done and I'll do a thorough go-over of the code (though at a glance things seem fine).

imsnif:
Oh! One more thing: I'm doing some work on the render function to improve its performance. Considering that, it might actually be better to handle the character selection in the grid and not in the render function. I know I said otherwise on discord - sorry about that :)

The reason for this is that the optimization I'm doing would mean that only changed characters will be sent to the terminal render function from the grid - so if you do it that way, the merge will be less painful I think. I hope that doesn't throw a wrench in your plans?

tlinford:
Glad to hear that you are happy with things so far :)
 
- Regarding OSC52, absolutely, in the rush to get the selection copied to the clipboard I kind of forgot some important details, will give that a try as soon as possible
- Multiple panes and selection: yes, that's what I meant with bugginess, there's quite a bit of work missing there, mainly a way to handle selecting text in one pane at a time, and resetting (or maybe it would be cool to recalculate) the selection on pane resizes.
- Rendering: after you suggested doing it in render, I realized that it's much simpler that way, because modifying the character styles in the grid makes it necessary to also be able to undo the changes when the selection changes or is reset. But it totally makes sense from a performance perspective to redraw only changed characters, I'll see what I can come up with.

imsnif:
>     * Rendering: after you suggested doing it in render, I realized that it's much simpler that way, because modifying the character styles in the grid makes it necessary to also be able to undo the changes when the selection changes or is reset. But it totally makes sense from a performance perspective to redraw only changed characters, I'll see what I can come up with.

Don't work too hard on this (or maybe leave it to the end?) - I'm in very initial stages of the performance improvements and things might change.

tlinford:
@imsnif Quick update:

- osc52 seems to be working fine
- selecting with multiple panes should be improved, once a selection is started by clicking in a pane, drag and release events are considered only for that pane. For now non empty selections in other panes are not modified, I think we could go both ways on that.

I realized that the current approach is a bit limited, as the selection is relative to the viewport, and not the whole terminal's text buffer, so for example holding left mouse button, and scrolling up/down won't behave as one would expect at the moment.

imsnif:
Sounds good! Sorry, I just realized maybe you're stuck because of me?

imsnif:
Hey @tlinford - this is really cool! I absolutely love the behaviour and the performance is stellar.
I think it would be best to start with some back-and-forth about some issues I'm experiencing before we delve more deeply into the code (as it might change through these):
1. I had a hard crash when I tried to mark some text with the mouse (and might have moved my mouse outside the terminal window as it was pressed, though I'm not sure this was the reason for it). It happened because of a "subtract with overflow" here:  https://github.com/zellij-org/zellij/pull/448/files#diff-9090eff77858ec02312fbbf524b5ab874d0ac52f89fbefb5501e2e4785113ee4R33
I think it might be a good idea to use `saturating_sub` there and in other places where we do subtraction.
2. I can't get copying to work for me. Is it working for you? I'm marking with the mouse and the text isn't getting into my clipboard, even though doing the same thing works outside of my terminal emulator (alacritty). Am I doing something wrong?

tlinford:
Heya, glad you liking it :)
1. will look into that
2. copying is working for me, not sure if I mentioned that for now it's bound to ctrl+b, have you tried with that or with the usual ctrl+shift+c? I've been testing in alacritty and it's been working reliably.

imsnif:
>     2. copying is working for me, not sure if I mentioned that for now it's bound to ctrl+b, have you tried with that or with the usual ctrl+shift+c? I've been testing in alacritty and it's been working reliably.

Aha! Sorry, I missed that. Yes, that's working :)
In most (all?) terminal emulators I think copying happens whenever the user lifts the left mouse button if there's text selected. Do you think we could also behave that way?

tlinford:
> In most (all?) terminal emulators I think copying happens whenever the user lifts the left mouse button if there's text selected. Do you think we could also behave that way?

Should not be a problem at all, should it be like in tmux, where the selection also gets reset after release?

imsnif:
Yeah, that sounds good!

tlinford:
Have changed the subtractions to use saturating_sub (although I find it a bit strange since Termion mentions coordinates are 1 based).
Also, the selection now gets automatically copied to clipboard on mouse release and reset.

imsnif (changes requested):
So I've gone over all the changes and kicked the tires quite a bit. This is fantastic work. It's very clear you've devoted quite a bit of time and effort here and I'm very excited about getting this merged ASAP.

I left a few comments in the code itself. Other than that, a few things:
1. We took a bit of a performance hit here. It's subtle, but important. One can see this in the following scenario: do a `cat /dev/urandom` and then try to stop it with `ctrl-c`. It never really stops. the only way to stop it is to kill the process from another terminal. I'm not 100% sure this is a performance issue, but that's what it was the last time around. Be sure to use the `--release` flag when building and testing this. Let me know if you want a hand debugging this - I started on my own and so far can still reproduce this after commenting out the `enable_mouse_support` and the lines in the `terminal_pane` render function.
2. We need to restore the terminal mouse support when Zellij crashes as well. I think we could probably do this where we restore the screen buffer (`[?1049l`) in the `handle_error` call in zellij_client/lib.rs.
3. I saw that there's some initial handling of scrolling a pane up while selecting (I also remember we briefly talked about this before). Will it be possible to include the full functionality in this PR, or would you rather keep this for later? It's just that it would be a very very nice user experience to have this from the get-go, but definitely not critical.
4. As mentioned in the linked issue above about neovim, we should at some point support forwarding mouse events to panes if they themselves request a mouse terminal. This can definitely be left for later - just mentioning it here in case you think it's an easy win somehow.

tlinford:
Hi @imsnif, awesome, thanks for the kind words :)
will look into everything over the next days.
Regarding scrolling while selecting, it should be already working, have you tried a noticed it not working?

tlinford:
> So I've gone over all the changes and kicked the tires quite a bit. This is fantastic work. It's very clear you've devoted quite a bit of time and effort here and I'm very excited about getting this merged ASAP.

Also excited about getting this merged!  Has been a great learning experience.
And would love to work on more features after that, like resizing panes and switching tabs with the mouse, and so on. 

> I left a few comments in the code itself. Other than that, a few things:
> 
>     1. We took a bit of a performance hit here. It's subtle, but important. One can see this in the following scenario: do a `cat /dev/urandom` and then try to stop it with `ctrl-c`. It never really stops. the only way to stop it is to kill the process from another terminal. I'm not 100% sure this is a performance issue, but that's what it was the last time around. Be sure to use the `--release` flag when building and testing this. Let me know if you want a hand debugging this - I started on my own and so far can still reproduce this after commenting out the `enable_mouse_support` and the lines in the `terminal_pane` render function.

I tried some times to reproduce this, but I can't see any difference the latest 0.13 binary and one i built with `cargo make build-release`. Not sure if it has something to do with my current setup, but I'll try on a different machine.

>     2. We need to restore the terminal mouse support when Zellij crashes as well. I think we could probably do this where we restore the screen buffer (`[?1049l`) in the `handle_error` call in zellij_client/lib.rs.
> 

Absolutely

>     3. I saw that there's some initial handling of scrolling a pane up while selecting (I also remember we briefly talked about this before). Will it be possible to include the full functionality in this PR, or would you rather keep this for later? It's just that it would be a very very nice user experience to have this from the get-go, but definitely not critical.
> 

This should be working already as I mentioned above

>     4. As mentioned in the linked issue above about neovim, we should at some point support forwarding mouse events to panes if they themselves request a mouse terminal. This can definitely be left for later - just mentioning it here in case you think it's an easy win somehow.

Do you mean this issue? https://github.com/zellij-org/zellij/issues/579
I'm afraid I don't really know for now, need to do some experimenting to see how it can be done.

zellij-server/src/panes/grid.rs
```diff
@@ -488,10 +494,13 @@ impl Grid {
             }
             let line_to_insert_at_viewport_bottom = self.lines_below.remove(0);
             self.viewport.push(line_to_insert_at_viewport_bottom);
+            self.selection.move_up(1);
             self.output_buffer.update_all_lines();
         }
     }
     pub fn change_size(&mut self, new_rows: usize, new_columns: usize) {
+        // TODO: recalculate selection after resize, for now reset selection to empty
```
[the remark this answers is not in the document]
tlinford: yep

zellij-client/src/input_handler.rs
```diff
@@ -54,6 +57,9 @@ impl InputHandler {
         let alt_left_bracket = vec![27, 91];
         let bracketed_paste_start = vec![27, 91, 50, 48, 48, 126]; // \u{1b}[200~
         let bracketed_paste_end = vec![27, 91, 50, 48, 49, 126]; // \u{1b}[201
+
+        // move this somewhere else?
+        self.os_input.enable_mouse();
```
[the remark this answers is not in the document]
tlinford: was fighting a bit with this at the beginning, sounds good.

zellij-server/src/panes/grid.rs
```diff
@@ -1135,6 +1146,128 @@ impl Grid {
...
+        // maybe some of these can be avoided, but the logic is tricky
+        lines_to_update.insert(old_selection.start.line.0);
+        lines_to_update.insert(old_selection.end.line.0);
+        lines_to_update.insert(new_selection.start.line.0);
+        lines_to_update.insert(new_selection.end.line.0);
+
+        let old_lines: HashSet<isize> = self.get_visible_indices(old_selection).collect();
+        let new_lines: HashSet<isize> = self.get_visible_indices(new_selection).collect();
+
+        old_lines.symmetric_difference(&new_lines).for_each(|&l| {
```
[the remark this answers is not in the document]
tlinford: yep, amazing stuff to have in the stdlib :)

zellij-server/src/panes/grid.rs
```diff
@@ -1135,6 +1146,128 @@ impl Grid {
...
+            };
+
+            if start_column == end_column {
+                continue;
+            }
+
+            let empty_row = Row::from_columns(vec![EMPTY_TERMINAL_CHARACTER; self.width]);
+
+            // get the row from lines_above, viewport, or lines below depending on index
+            let row = if l < 0 {
```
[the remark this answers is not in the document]
tlinford: yes, the selection basically can have indices that are negative, when the selected text is in lines_aboves, or greater then grid.height when the text is in lines_below. The selection gets adjusted in when the grid scrolls up or down.

zellij-server/src/panes/plugin_pane.rs
```diff
@@ -39,6 +39,13 @@ impl PluginPane {
 }
 
 impl Pane for PluginPane {
+    fn start_selection(&mut self, _start: &Position) {}
+    fn update_selection(&mut self, _to: &Position) {}
+    fn end_selection(&mut self, _end: Option<&Position>) {}
+    fn reset_selection(&mut self) {}
+    fn get_selected_text(&self) -> Option<String> {
+        None
+    }
```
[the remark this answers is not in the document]
tlinford: done

zellij-server/src/panes/grid.rs
```diff
@@ -1135,6 +1146,128 @@ impl Grid {
     fn set_preceding_character(&mut self, terminal_character: TerminalCharacter) {
         self.preceding_char = Some(terminal_character);
     }
+    pub fn start_selection(&mut self, start: &Position) {
```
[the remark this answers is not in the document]
tlinford: I'm not seeing an easy way for the moment, we need to update the lines in grid's output_buffer to be able to highlight them, or access the selected lines when copying to clipboard.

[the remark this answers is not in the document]
tlinford: as a small improvement, moved the line diffing calculation between to selections to selection

imsnif:
Hey @tlinford - I'm not sure why but it seems that the last bit of our conversation here was deleted :O 

I remember I had to reply about marking and scrolling:
I just tested this again locally and it doesn't work for me. What I'm doing:
* Type `ls` a few times inside a pane to create scrollback
* Mark text inside a pane, starting from the middle
* Keep the left-mouse-button down and go up, beyond the pane to the next pane above it (or above the edge of the window if there's just one pane)
* The pane does not scroll

I'm using Alacritty, but I have a feeling this doesn't have to do with the terminal emulator. Are you doing the exact same thing and it's working for you?

imsnif:
Also, are there any other subjects that were somehow deleted that I had to respond to? Or anything else you'd like my input on?

tlinford:
>     * Keep the left-mouse-button down and go up, beyond the pane to the next pane above it (or above the edge of the window if there's just one pane)
> I'm using Alacritty, but I have a feeling this doesn't have to do with the terminal emulator. Are you doing the exact same thing and it's working for you?

Ah no, haven't gotten around to implementing that for now, what works is scrolling is the mouse wheel while holding left mouse button down.

tlinford:
> Also, are there any other subjects that were somehow deleted that I had to respond to? Or anything else you'd like my input on?

Try scrolling up, I can see everything, but it say 31 hidden items.

imsnif:
Oops, you're right - thanks. My bad :)

> Ah no, haven't gotten around to implementing that for now, what works is scrolling with the mouse wheel while holding left mouse button down.

Aha, yes. Okay, that's working. I think most users would expect the other behaviour... looks like most of the infrastructure is there in order to do it. Do you think it can be implemented easily here (eg. something like: if there is a selection already on mouse hold, treat the coordinates on screen relative to the ones in the pane)? Really no biggie if we want to do this in a subsequent PR of course. You've already done a lot of work here.

> I tried some times to reproduce this, but I can't see any difference the latest 0.13 binary and one i built with `cargo make build-release`. Not sure if it has something to do with my current setup, but I'll try on a different machine.

Let's try to get to the bottom of this, because this is important.

Here's what I'm doing:

How I'm compiling the binaries:
* Check out either `main` or this branch
* `cargo build --release` (no need to go through cargo make here if you already have the plugins and everything compiled)
* run the below test-case on the created binary at `target/release/zellij`

My test-case:
* Run the binary
* `cat /dev/urandom`
* Wait 1-2 seconds
* type `ctrl-c` to stop the output

My results:
* in `main` the output stops immediately and I get a prompt back
* in this branch the output never stops and I have to kill either zellij or the `cat` process from another pane/terminal window

Is this behaviour different for you? How exactly?

> I'm afraid I don't really know for now, need to do some experimenting to see how it can be done. Are you thinking of just handling the event, and then rewriting it to the terminal in the pane?

Sort of. We have a bit of a similar issue and behaviour with [DECCKM](https://vt100.net/docs/vt510-rm/DECCKM.html). Here we need to conditionally send different "keystrokes" depending on what the shell/other-program requested. We do it here: https://github.com/zellij-org/zellij/blob/main/zellij-server/src/panes/terminal_pane.rs#L77
My suggestion is that we do the same with the mouse. When a program requests to be a "mouse terminal" itself, then if the user does any mouse actions inside that pane, we send them directly to the program as is. Otherwise we handle them ourselves. I'm not sure if this is feasible, makes sense or how easy it is to implement. So if you want to keep this for later after this PR that's totally legit. :)

tlinford:
> Aha, yes. Okay, that's working. I think most users would expect the other behaviour... looks like most of the infrastructure is there in order to do it. Do you think it can be implemented easily here (eg. something like: if there is a selection already on mouse hold, treat the coordinates on screen relative to the ones in the pane)? Really no biggie if we want to do this in a subsequent PR of course. You've already done a lot of work here.

In general I'd be happy to try and merge sooner rather than later, can get a bit troublesome to rebase everything if there are some bigger changes in master :D - I'd be happy to work on this in the near future though.
> Let's try to get to the bottom of this, because this is important.
> 
> Here's what I'm doing:
> 
> How I'm compiling the binaries:
> 
>     * Check out either `main` or this branch
> 
>     * `cargo build --release` (no need to go through cargo make here if you already have the plugins and everything compiled)
> 
>     * run the below test-case on the created binary at `target/release/zellij`
> 
> 
> My test-case:
> 
>     * Run the binary
> 
>     * `cat /dev/urandom`
> 
>     * Wait 1-2 seconds
> 
>     * type `ctrl-c` to stop the output
> 
> 
> My results:
> 
>     * in `main` the output stops immediately and I get a prompt back
> 
>     * in this branch the output never stops and I have to kill either zellij or the `cat` process from another pane/terminal window
> 
> 
> Is this behaviour different for you? How exactly?
> 
Yes, I'm having no luck reproducing the problem you describe: in my case on both master and mouse-support I see the same behaviour, after cat /dev/urandom, with ctrl+c the output stops and i get the prompt back.

What I find a bit strange about this is that you mentioned that it happens in this situation:

> I started on my own and so far can still reproduce this after commenting out the enable_mouse_support and the lines in the terminal_pane render function.

Without enable_mouse_support the code in this branch should really not be doing much at all, there is that extra check in `TerminalPane render`, so this is very perplexing. Are you running on linux or macos?

> > I'm afraid I don't really know for now, need to do some experimenting to see how it can be done. Are you thinking of just handling the event, and then rewriting it to the terminal in the pane?
> 
> Sort of. We have a bit of a similar issue and behaviour with [DECCKM](https://vt100.net/docs/vt510-rm/DECCKM.html). Here we need to conditionally send different "keystrokes" depending on what the shell/other-program requested. We do it here: https://github.com/zellij-org/zellij/blob/main/zellij-server/src/panes/terminal_pane.rs#L77
> My suggestion is that we do the same with the mouse. When a program requests to be a "mouse terminal" itself, then if the user does any mouse actions inside that pane, we send them directly to the program as is. Otherwise we handle them ourselves. I'm not sure if this is feasible, makes sense or how easy it is to implement. So if you want to keep this for later after this PR that's totally legit. :)

I think keeping it for later seems preferable at this point.

One other thing that I'm a bit concerned about, is the fact that OSC52 support is not present in some terminal emulators, most notably gnome-terminal and konsole out of the ones I've tried. Do you think it might be a good idea to just add an option to disable mouse-support to avoid weird behaviour in such cases?

imsnif:
As dicussed offline, I stopped being able to reproduce this issue after some computer and docker restarts. I cannot explain why it happened, but it's clearly particular to my local environment. Sorry about that!

tlinford:
Hi @imsnif, scrolling while selecting by moving the cursor above/below the active pane should now be working.
There might be a bit of tuning to be done to make the scrolling speed feel right. Also considering whether to increase the number of scrolled lines, based on how far the cursor moves.

zellij-client/src/input_handler.rs
```diff
@@ -173,6 +208,20 @@ impl InputHandler {
             .send(ClientInstruction::Exit(ExitReason::Normal))
             .unwrap();
     }
+
+    // Repeatedly send fake mousehold events, to allow scrolling up/down with mouse while selecting
+    fn start_hold_repeater(&mut self, position: Position) {
+        let mut poller = StdinPoller::default();
```
imsnif, line 214: Cool solution :) Since this interfaces with the OS though, it needs to go into os_input_output

zellij-client/src/input_handler.rs
```diff
@@ -193,3 +242,45 @@ pub(crate) fn input_loop(
...
+        let stdin = 0;
+        let mut stdin_fd = SourceFd(&stdin);
+        let events = Events::with_capacity(128);
+        let poll = Poll::new().unwrap();
+        poll.registry()
+            .register(&mut stdin_fd, Token(0), Interest::READABLE)
+            .expect("could not create stdin poll");
+
+        // TODO: decide what timeout to use
+        let timeout = Duration::from_millis(50);
```
imsnif, line 278: I think this can even be less, 10ms?

zellij-server/src/panes/terminal_pane.rs
```diff
@@ -285,6 +300,41 @@ impl Pane for TerminalPane {
...
     }
+
+    fn start_selection(&mut self, start: &Position) {
+        self.grid.start_selection(start);
+        self.set_should_render(true);
+    }
+
+    fn update_selection(&mut self, to: &Position) {
+        // TODO: decide how often to scroll
+        let should_scroll = self.selection_scrolled_at.elapsed() >= time::Duration::from_millis(50);
```
imsnif, line 311: Maybe I'm misunderstanding, but if we have the `mio` timeout in the stdin thread, why do we need to keep track of the time here too?

zellij-server/src/panes/unit/grid_tests.rs
```diff
@@ -555,3 +555,67 @@ fn wrap_wide_characters_at_the_end_of_the_line() {
...
+fn copy_selected_text_from_viewport() {
+    let mut vte_parser = vte::Parser::new();
+    let mut grid = Grid::new(27, 125, Palette::default());
+    let fixture_name = "grid_copy";
+    let content = read_fixture(fixture_name);
+    for byte in content {
+        vte_parser.advance(&mut grid, byte);
+    }
+
+    println!("{:?}", grid);
```
imsnif, line 569: oops :)

zellij-server/src/panes/unit/grid_tests.rs
```diff
@@ -555,3 +555,67 @@ fn wrap_wide_characters_at_the_end_of_the_line() {
...
+fn copy_selected_text_from_lines_above() {
+    let mut vte_parser = vte::Parser::new();
+    let mut grid = Grid::new(27, 125, Palette::default());
+    let fixture_name = "grid_copy";
+    let content = read_fixture(fixture_name);
+    for byte in content {
+        vte_parser.advance(&mut grid, byte);
+    }
+
+    println!("{:?}", grid);
```
imsnif, line 590: ^^

zellij-utils/src/input/options.rs
```diff
@@ -24,6 +24,9 @@ pub struct Options {
     /// subdirectory of config dir
     #[structopt(long, parse(from_os_str))]
     pub layout_dir: Option<PathBuf>,
+    #[structopt(long)]
+    #[serde(default)]
+    pub disable_mouse_mode: bool,
```
imsnif, line 29: I couldn't get this to work. I did `zellij options --disable-mouse-mode` and the mouse was still functional. Am I doing something wrong or misunderstanding?

imsnif (changes requested):
Hey @tlinford - I left some comments about the recent changes. The scrolling selection looks great and I honestly could not find any functional issue with it - and not for lack of trying. :)

I left some comments in the code. Otherwise, as we discussed on discord, the integration tests are gone, so it would be best to move the ones in this PR to either be e2e tests or unit tests (preferably one basic e2e test to make sure the mouse flow is working and the rest should be unit tests).

zellij-server/src/panes/terminal_pane.rs
tlinford: a bit hard to explain, but basically if we don't have it scrolling speed becomes faster when moving the mouse, as every change of position generates a new mouse hold event. This way it stays constant both while moving the cursor and when it's still, and the events are generated by the repeater

zellij-client/src/input_handler.rs
```diff
@@ -54,6 +62,10 @@ impl InputHandler {
         let alt_left_bracket = vec![27, 91];
         let bracketed_paste_start = vec![27, 91, 50, 48, 48, 126]; // \u{1b}[200~
         let bracketed_paste_end = vec![27, 91, 50, 48, 49, 126]; // \u{1b}[201
+
+        if !self.config.options.disable_mouse_mode {
```
a-kenji, line 66: Since this is the direct `config` struct it doesn't merge the config options,
that would be the `options` struct you can see in `opts` in `main.rs`.
Maybe it would make sense to also merge the options back into the config
struct? Or maybe it makes more sense to pass the merged `options` struct.

zellij-utils/src/input/options.rs
a-kenji: It should work in the config file already with `disable_mouse_mode: true`, but the merged config is not used yet.

a-kenji (commented):
Awesome! Looks great already.

The `options` config is merged [here](https://github.com/zellij-org/zellij/blob/1b70b6958ca365223bd104fb3ea549d39f8b3d12/src/main.rs#L39) currently and not yet attached back to the config struct, maybe that would be the best way to go for now.

zellij-client/src/input_handler.rs
tlinford: related to the comment just below, so it should be 10ms to have faster scrolling?

zellij-client/src/input_handler.rs
tlinford: `input_handler` is also reading `keybinds` from `config`, so maybe passing both is best?

tlinford:
Hi @imsnif, I migrated the tests to the new e2e setup (very cool!), the action repeater is now in `os_input_output`, and the problem with the disable command line flag should be fixed. Scrolling speed has been updated to 10ms.

imsnif:
Hey @tlinford - this looks great. Thanks for all the changes. I think I'm ready to merge this into `main` - unless there's something I'm forgetting? Some issue still open or that we still need to decide about?

tlinford:
> Hey @tlinford - this looks great. Thanks for all the changes. I think I'm ready to merge this into `main` - unless there's something I'm forgetting? Some issue still open or that we still need to decide about?

Hey @imsnif, awesome,  I think we covered everything we've discussed so far.

imsnif:
Amazing :)

So thanks very much for your incredible hard work and perseverance in implementing this feature. You've done a really great job!

PR #457, opened 2021-05-04
A New Resize Algorithm

### LINKED ISSUE #406: i3wm - proportionate adjustment between panes
zellij version is 0.5.1

I am using i3wm as window manager on Arch Linux. 
When i launch a window of other application, this window is proportionately adjusted with other window (i.e terminal (alacritty) with zellij having two panes). Terminal with zellij becomes unusable **with disproportionate adjustment of panes**.

screenshot_20210429113309_IMGP

TheLostLambda:
Should eventually close #406 

This is still a bit borked. I need to:

- [x] Find out if changing the behavior of the `change_pos_and_size()` function breaks anything
- [ ] ~~Store and internal `f64` size and external `usize` for Pane~~
- [ ] Use proper percent and fixed constraints, translated to real column and row counts on resize

Eventually, things should work like this:

https://user-images.githubusercontent.com/6251883/117079734-43a93780-ad34-11eb-8281-0617045b2c38.mp4


TheLostLambda:
Don't worry, this PR is still quite alive, I've just had to rebase on the wildly different `main` and change direction a little!

PR #465, opened 2021-05-06
 Add Complile Time System Directories

### LINKED ISSUE #380: paths for system wide config and plugin dir
We are working on a package for opensuse and it seems so far it only searches for plugins in the user home dir but not in a system path yet? same for the system wide config. which path should be used there?

a-kenji:
closes #380

* --features enables_automatic_asset_installation

enables automatic installation of the `data_dir`

* a set PREFIX env will use `$PREFIX/share/zellij` as its system `data_dir`
Will now look in `xdg_data_dir`, if it doesn't find anything there falls back to the system `data_dir`


src/common/setup.rs
```diff
@@ -68,9 +72,26 @@ pub fn find_default_config_dir() -> Option<PathBuf> {
...
 
+/// Looks for an existing dir, uses that, else returns a
+/// dir matching the config spec.
+pub fn get_default_data_dir() -> PathBuf {
+    vec![
+        Some(xdg_data_dir()),
+        Some(Path::new(SYSTEM_DEFAULT_DATA_DIR_PREFIX).join("/share/zellij")),
+    ]
+    .into_iter()
+    .filter(|p| p.is_some())
```
TheLostLambda, line 83: Any reason this is needed? (I assume it's a form of future-proofing? When will these values be `None`?

TheLostLambda (approved):
Looks good to me! Left a code question, but this is some good flexibility to have! 

Thanks for putting this together!

src/common/setup.rs
a-kenji: I think think this a remnant of me trying a different approach, thanks! That is a good catch.

a-kenji:
Does the `makefile` need adjustment for this functionality? Or should the packagers invoke `cargo` directly?

TheLostLambda:
I think cargo directly for now should be fine, but perhaps adding a way to specify features in the makefile would be helpful eventually (probably just passing extra arguments through like we are already doing with `cargo make run`). Up to you!

a-kenji:
Yeah, that is a good idea! Maybe we even get some feedback in that direction at some point.

PR #466, opened 2021-05-06
improve bug report template

matu3ba:
Fold instructions, give copy-paste commands and give example of typical
information density. Refer to common sense of the issue author.


imsnif:
Hey, thanks for this! I like all the changes except the folding of the instructions... why do you feel this is important? This is quite a short document...

matu3ba:
@imsnif The idea was to remove clutter from reading the issues, since people tend to only add things. I was expecting the template to increase in size due to more features.

Folding the instructions could also be added at a later point, so I removed it for now.

imsnif:
I definitely understand what you're saying about most people usually just adding stuff. Let's keep an eye on it then.

Issue #426, opened 2021-05-01
Neovim not rendering correctly


badrbouslikhin:
Hello,

Thank you for your efforts!

I have rendering issues when using Neovim (nightly) inside Zellij + kitty. I am also using Fish Shell.

**Neovim without Zellij**

<img width="1792" alt="Screenshot 2021-05-01 at 15 56 55" src="https://user-images.githubusercontent.com/7818591/116784674-dc6e5780-aa95-11eb-9563-fb36427e6b1f.png">

**Neovim with Zellij**

<img width="1792" alt="Screenshot 2021-05-01 at 15 57 49" src="https://user-images.githubusercontent.com/7818591/116784704-fc9e1680-aa95-11eb-9bc3-f880d91e279a.png">

## Logs
[log.txt](https://github.com/zellij-org/zellij/files/6409651/log.txt)
[zellij-3.log](https://github.com/zellij-org/zellij/files/6409652/zellij-3.log)
[zellij-6.log](https://github.com/zellij-org/zellij/files/6409653/zellij-6.log)
[zellij-8.log](https://github.com/zellij-org/zellij/files/6409654/zellij-8.log)

## Information

`zellij --version`: zellij 0.6.0
`tput lines`: 55
`tput cols`: 210
`nvim --version`: NVIM v0.5.0-dev+nightly-2-g66f9dd3c6
`kitty --version`: kitty 0.19.3
`fish --version`: fish, version 3.2.2

imsnif:
Thanks for reporting this. This is now fixed in `main` and will be released in the next version.

PR #469, opened 2021-05-07
fix(compatibility): upgrade vte to support csi subparameters

imsnif:
By supporting csi subparameters (see discussion here: https://github.com/alacritty/vte/issues/22) we fix lots of coloring issues. These were instructions we were simply ignoring before, which is why a lot of colors looked wrong.


imsnif:
This addresses (among other things) https://github.com/zellij-org/zellij/issues/426

PR #475, opened 2021-05-09
Config tweaks

khs26:
Still a WIP, but doing some misc tweaking of the config stuff, along with the beginnings of a schema.

### DIFF
--- a/src/common/input/config_schema.json
+++ b/src/common/input/config_schema.json
@@ -33,9 +33,6 @@
             }
         }
     },
-    "required": [
-        "keybinds"
-    ],
     "$defs": {
         "mode_keybinds": {
             "description": "the set of keybinds for a particular input mode",
--- a/src/common/input/config_schema.json
+++ b/src/common/input/config_schema.json
@@ -1,0 +1,62 @@
+{
+    "$schema": "https://json-schema.org/draft/2020-12/schema",
+    "$id": "zellij/config_schema.json",
+    "description": "Configuration for the zellij terminal multiplexer",
+    "type": "object",
+    "properties": {
+        "keybinds": {
+            "description": "user-defined keybinds",
+            "type": "object",
+            "properties": {
+                "unbind": {},
+                "normal": {
+                    "$ref": "#/$defs/mode_keybinds"
+                },
+                "locked": {
+                    "$ref": "#/$defs/mode_keybinds"
+                },
+                "resize": {
+                    "$ref": "#/$defs/mode_keybinds"
+                },
+                "panes": {
+                    "$ref": "#/$defs/mode_keybinds"
+                },
+                "tab": {
+                    "$ref": "#/$defs/mode_keybinds"
+                },
+                "scroll": {
+                    "$ref": "#/$defs/mode_keybinds"
+                },
+                "renametab": {
+                    "$ref": "#/$defs/mode_keybinds"
+                }
+            }
+        }
+    },
+    "required": [
+        "keybinds"
+    ],
+    "$defs": {
+        "mode_keybinds": {
+            "description": "the set of keybinds for a particular input mode",
+            "type": "array",
+            "items": {
+                "description": "an individual keybind",
+                "properties": {
+                    "action": {
+                        "description": "the actions to be executed, in order",
+                        "type": "array"
+                    },
+                    "key": {
+                        "description": "the key sequence(s) which invoke this keybind",
+                        "type": "array"
+                    }
+                },
+                "required": [
+                    "action",
+                    "key"
+                ]
+            }
+        }
+    }
+}
--- a/src/common/input/config_schema.json
+++ b/src/common/input/config_schema.json
@@ -8,7 +8,9 @@
             "description": "user-defined keybinds",
             "type": "object",
             "properties": {
-                "unbind": {},
+                "unbind": {
+                    "$ref": "#/$defs/unbind"
+                },
                 "normal": {
                     "$ref": "#/$defs/mode_keybinds"
                 },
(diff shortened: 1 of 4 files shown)


src/common/input/keybinds.rs
```diff
@@ -29,7 +29,7 @@ pub struct KeybindsFromYaml {
 #[serde(untagged)]
 enum KeyActionUnbind {
     KeyAction(KeyActionFromYaml),
-    Unbind(UnbindFromYaml),
```
a-kenji, line 32: This breaks current configurations.
You are not able to anymore unbind keys in this way:

```
keybinds:
    normal:
        - unbind : true
        - action: [GoToTab: 1,]
          key: [F: 1,]
```

or 

```
keybinds:
    normal:
        - unbind : [Alt: 'n', Ctrl: 'p']
        - action: [GoToTab: 1,]
          key: [F: 1,]
```

a-kenji (changes requested):
I will try to look at what a Json Schema is more when I get back home later. But I think this should still be there. Since the keyword makes it a lot clearer. Unless you find a way to still have this sort of configuration without the extra struct.

src/common/input/keybinds.rs
khs26: I think I might need your help with understanding the intended functionality here. I removed it on the basis of it looking potentially redundant (since `struct UnbindFromYaml` just contained `enum Unbind`, which I put in as its replacement).

However, when I set the default config to contain both of your examples:
- All the tests pass
- I'm not sure anything is actually being unbound

Both with and without my change.

khs26: `example/default.yaml` is not `assets/config/default.yaml` 😄

khs26:
FWIW, I'm using http://json-schema.org/learn/getting-started-step-by-step.html and http://json-schema.org/understanding-json-schema/reference/ as my references for it

src/common/input/keybinds.rs
a-kenji: The intended functionality there is that you can be granular in the unbinding configuration, since we add a lot of default keybindings and a person might just want to remove all the alt keybindings, or maybe just all the alt keybindings for a specific mode. Or wants to just unbind a specific mode.

https://zellij.dev/documentation/keybindings.html#to-unbind-the-default-keybindings

khs26: Cool - that makes a lot of sense as a goal for what the functionality should be.

I just hadn't noticed that the `default.yaml` I was editing was the one from the example folder

a-kenji:
Thank you for the links! If I understand correctly we write the configuration again. Is that not duplicated effort?
Serde already checks the types and it also has an error message if things are wrong, which is how it caught the naming mismatch.
I was thrown off with the recent bug in the macro, but the error message was quite clear, it said it wasn't a variant of the enum,
which meant it wasn't either in one of those forms:
```
 - action: Vec<Action> 
    key:  Vec<Key>
```
or
```
- unbind: Vec<Key>
```
or 
```
- unbind: bool
```

I have been wanting to improve the error message there for a while, and it is indeed on my list at the moment.
But we can certainly give a more detailed error message there. That said, I probably have overlooked some of the 
benefits of the Schema so I am eager about your input!

src/common/input/keybinds.rs
a-kenji: Ah, no problem! Yes that can be confusing and I think at some point we should remove it. I think though atm it is still a good reference.

Yes, the struct looks redundant. It is a way to tell serde that we want the unbind keyword there and match an untagged enum, if we don't have a struct there we probably would need to write our own deserialiser implementation, at least if I understood everything correctly.

khs26:
> Thank you for the links! If I understand correctly we write the configuration again. Is that not duplicated effort?
> Serde already checks the types and it also has an error message if things are wrong, which is how it caught the naming mismatch.
> I was thrown off with the recent bug in the macro, but the error message was quite clear, it said it wasn't a variant of the enum,
> which meant it wasn't either in one of those forms:
> 
> ```
>  - action: Vec<Action> 
>    key:  Vec<Key>
> ```
> 
> or
> 
> ```
> - unbind: Vec<Key>
> ```
> 
> or
> 
> ```
> - unbind: bool
> ```
> 
> I have been wanting to improve the error message there for a while, and it is indeed on my list at the moment.
> But we can certainly give a more detailed error message there. That said, I probably have overlooked some of the
> benefits of the Schema so I am eager about your input!

You're definitely right that there's duplication of effort here, but I think the advantages might still be worth it.

My motivation was to have a simpler way to validate the schema and as an added bonus, get syntax highlighting for VSCode. Our config model isn't very complex just yet, but I can imagine with layouts and keybinds and suchlike all combined, it could get to the point where someone ha to spend a fair bit of time learning how to write config. The JSON schema takes away the overhead of then verifying that it's structured correctly.

There are a few extensions I'd like to make to the schema, that could make other basic diagnostics a bit easier:
- A programmatic dump of all the actions, to restrict the range of things that can go in the `action` array (this would have caught my bug, with an error along the lines of `SyncActiveTab not a valid element` or some such).
- A definition of the available key types.

There are also a few other benefits that we can't take advantage of right now:
- It's possible to [generate basic serde stuff from a schema](https://docs.rs/schemars/0.8.3/schemars/), and I'd argue that writing a schema is a bit easier to iterate on when adding some new config. I don't *think* this is possible for serde_yaml just yet though.
- When we want to do fuzz testing of config, a schema could help with generating inputs.

a-kenji:
Would it maybe make sense to use a config of the form:
```
keybinds:
    normal:
        unbind: true
        keys:
             - action: []
               keys: []
```
I tried to avoid so much nesting, from a user perspective, but maybe it would be clearer if we look at it from the schema perspective and then also error matching would be easier.

a-kenji (commented):
I just realized this was also still pending ... !

src/common/input/config_schema.json
```diff
@@ -0,0 +1,75 @@
...
+                            "unbind": {
+                                "$ref": "#/$defs/unbind"
+                            }
+                        }
+                    }
+                ]
+            }
+        },
+        "unbind": {
+            "description": "unbind a particular key"
```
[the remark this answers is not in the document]
khs26: I definitely want these kind of things in here, but haven't yet been successful in applying them. I know that YAML has some peculiar habits with types (see [my colleague's blogpost](https://www.lewisgaul.co.uk/blog/coding/2021/04/18/zig-nestedtext/)) but I'm also just not very familiar with it.

Feel free to play around with it yourself, if you get time. Otherwise I'll pick it up when I'm doing zellij stuff and finished with anything Kunal needs.

a-kenji:
This doesn't seem to be relevant currently, closing for now.

PR #504, opened 2021-05-14
Remove unused imports/functions

matthiasbeyer:
This PR removes unused stuff.

Lets see what CI tells us about me deleting code :smile:

### DIFF
--- a/src/server/mod.rs
+++ b/src/server/mod.rs
@@ -19,7 +19,7 @@ use crate::common::{
     screen::{screen_thread_main, ScreenInstruction},
     setup::{get_default_data_dir, install::populate_data_dir},
     thread_bus::{ChannelWithContext, SenderType, SenderWithContext},
-    utils::consts::{ZELLIJ_IPC_PIPE, ZELLIJ_PROJ_DIR},
+    utils::consts::ZELLIJ_IPC_PIPE,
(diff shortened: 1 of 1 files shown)


src/common/setup.rs
```diff
@@ -21,37 +21,6 @@ macro_rules! asset_map {
     }
 }
 
-pub mod install {
```
a-kenji, line 24: This is actually [still used](https://github.com/zellij-org/zellij/blob/a24c7f79f1ac40f1e8554aa1615b0a7e020c7ca5/src/server/mod.rs#L190), but good catch! 
I guess the CI should check with --all_features and maybe also just the default features?

Or we need to address this issue in the makefile.

a-kenji (changes requested):
Thanks for this pr! I think this shows that we need to adjust our tests better for conditional compilation.

src/common/setup.rs
matthiasbeyer: Yep, I guess there's room for improvement with the testsuite then :laughing:

Want me to remove this patch from the patchset?

a-kenji: Yes, please do so.

matthiasbeyer:
After rebasing to latest master, this is what remains! :smile:

a-kenji (approved):
Thank you!

Issue #477, opened 2021-05-10
Panic on main thread when somehting is passed via stdin


bric3:
Thank you for taking the time to file an issue!
You can erase any parts of this template not applicable to your issue.

I was playing with `zellij` and tried to see if it accepted something on stdin, however it fails immediately.

```shell
$ echo "ls" | ~/.cargo/bin/zellij
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: Sys(ENOTTY)', /Users/bric3/.cargo/registry/src/github.com-1ecc6299db9ec823/zellij-0.8.0/src/common/os_input_output.rs:282:49
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

And there's no log in `/tmp/zellij-501/zellij-log`

## Information

`zellij --version`: 0.8.0
macos : big sur 11.3.1

imsnif:
Oi, thanks for finding this!
A good solution would be to not accept any input piped to Zellij at the moment :) Maybe in the future we'll want to add some features there, but for now it would be best if we don't crash.

This shouldn't be a super difficult fix, so if you or anyone else wants to take it up that would be cool.

bric3:
Unfrtunately I can barely read rust code. So I won't know how to fix that in a timely maneer.

a-kenji:
@bric3 
This should be adressed in #511.
Please don't hesitate to reach out again, if that doesn't work for you.

bric3:
👍 

Actually I had to hold back on zellij, as I am facing this small but repeated disruption https://github.com/zellij-org/zellij/issues/265 (On my keyboard I need to use the Alt / Option key to enter special characters, may there's a way to map to left or right alt, this may help).

PR #511, opened 2021-05-15
handle error on termios initialization

tlinford:
fix for #477


khs26:
GH actions are down, so I had to do the CI myself, but it all looks good :) Thanks for the fixes!

PR #513, opened 2021-05-15
Default mode 368

a-kenji:
This aims to make setting a default_mode (the first mode you start with) possible.
This works already, but the plugin doesn't correctly show it yet until the first change of events.


a-kenji:
@TheLostLambda 
Would it make sense to add some sort of `update` function to the
`load` function?

I want the plugin to read the `InputMode` already on load, so it can display
the locked mode.

TheLostLambda:
@a-kenji Is the issue that you aren't getting a properly updated mode when the plugin first loads? If so, we'll have to smush a bug somewhere because I made the `ModeInfo` struct special at some point. Every plugin that is loaded should have a `ModeInfo` event dispatched to it automatically. I've been able to start plugins with different modes on the first load in the past (that's what happens every time you open a new tab, since that reloads the plugins), so it should be easy to start in a different mode.

I suspect a bug more than a missing feature here, but I'll have to take a closer look soon (I'm still a bit busy this week with exams)

a-kenji:
@TheLostLambda 
Thanks for the response - no pressure!

> Is the issue that you aren't getting a properly updated mode when the plugin first loads? If so, we'll have to smush a bug somewhere because I made the `ModeInfo` struct special at some point.

I think so yeah, I suspect the plugin is loaded with defaults on the first creation.

I'll into it a little more!

a-kenji:
@TheLostLambda 

Thank you for the help!
I figured it out now, the plugins are getting an update when
a tab is created, but not on the initial load.

I propagate the information now on creation.
I think that should be sufficient for now. 
As soon as I am able I will merge from main.

a-kenji:
@kunalmohan,

I am having Issues merging this change into the current main,
maybe you have some idea where the mode behaviour change
may come from.

Pre merge, the mode functions properly, even when starting in the
locked mode. But after the merge I need to switch to the normal mode
twice in order for it to have an effect. 

And as always no rush!

kunalmohan:
@a-kenji how do the plugins receive the default_mode from the config options?

a-kenji:
@kunalmohan 
Either through:
`zellij options --default-mode [mode]`
or setting `default_mode: [mode]` in the config file

kunalmohan:
@a-kenji Sorry, I wasn't clear. I meant in the code. Where/when does the wasm thread receive the default_mode at startup?

zellij-server/src/screen.rs
```diff
@@ -415,9 +416,10 @@ pub(crate) fn screen_thread_main(
             capabilities: PluginCapabilities {
                 arrow_fonts: capabilities,
             },
+            mode: default_mode,
```
a-kenji, line 419: @kunalmohan 
Here is the mode passed in for the plugins.

a-kenji:
@kunalmohan 
No problem! 
I left a comment where the changes are picked up for the plugin.

But as I said. It works correctly prior to the merge in 
https://github.com/zellij-org/zellij/pull/513/commits/80fe803ffd49c0aa05c3883d04c0239e5dd5bae1
but after the merge it still picks up the correct mode on startup,
but it stays there until I switch the mode twice back to normal mode.
This is not the plugin that just doesn't show the mode switch,
it is the mode that is not switched.

I will keep on looking into it but I don't understand currently what has
changed here.

PR #515, opened 2021-05-16
Restructure workspace: Create separate crates for client, server and utilities

kunalmohan:
* Separates `zellij` into 4 crates- `zellij`, `zellij-utils`, `zellij-client`, `zellij-server`.
* `zellij-utils` contains all utilities shared by the client and the server. And `zellij` parses the command-line arguments and calls appropriate API methods exposed by `zellij-client` and `zellij-server`.
* Since `cfg(test)` is not propagated to the dependencies, `zellij-client`, `zellij-server` and `zellij-utils` now have a feature `test` which `zellij` enables while testing. So `cfg(test)` has been replaced by `cfg(any(feature="test", test))` in the 3 crates.
* `Makefile.toml` has been modified to skip various workspace crates so that crates are compiled parallelly as a dependency, instead of compiling each crate explicitly in a sequential manner.


khs26:
Does this change how you run zellij at all? It looks like it probably doesn't, but if it does, we should update the docs

kunalmohan:
> Does this change how you run zellij at all?

Nope, the commands are the same. I've made a few changes so that the build is not sequential, but the user interface is same. I haven't tested the install and publish commands, but I think they should work fine. I'm hoping that @TheLostLambda  will be able to point out any potential bugs from a quick review of the `Makefile.toml` changes.

Makefile.toml
```diff
@@ -2,6 +2,7 @@
 [env]
 CARGO_MAKE_EXTEND_WORKSPACE_MAKEFILE = true
 CARGO_TARGET_DIR = "${CARGO_MAKE_WORKSPACE_WORKING_DIRECTORY}/target"
+CARGO_MAKE_WORKSPACE_SKIP_MEMBERS = ["zellij-utils", "zellij-server", "zellij-tile", "zellij-client", "zellij-tile-utils"]
```
TheLostLambda, line 5: I'm not sure how I feel about this exactly, I've been using the `workspace = false` on tasks that should only be done once. I'm a touch worried this will break things like running clippy on all crates.

Additionally, I'm not a massive fan of having a `test` feature, but I also don't know if there is a better way. Do the other crates need to expose different functionality for `zellij` to have tests?

TheLostLambda: Overall though, I like this reorganization! As long as Clippy, running all tests, and the release building still works, I'm happy :)

kunalmohan: > I'm not sure how I feel about this exactly, I've been using the `workspace = false` on tasks that should only be done once. I'm a touch worried this will break things like running clippy on all crates.

Yeah, that's reasonable. I'll revert this change. The compile-time also hasn't reduced much.
 
> Additionally, I'm not a massive fan of having a `test` feature, but I also don't know if there is a better way. Do the other crates need to expose different functionality for `zellij` to have tests?

Yeah, we have used `#[cfg(test)]` and `#[cfg(not(test))]` in a number of places which are now in the 3 different crates. But I think things will improve as the multiprocess model develops and a new testing framework is adopted and hopefully we won't need the `test` feature in the future.

kunalmohan: Although, my one tiny concern is that cargo-make build everything with `--all-features` flag. So when you run `cargo make run/build`, it will first build the 3 `zellij-*` crates with the `test` feature enabled, and later on it builds them as a dependency of `zellij` without the `test` feature. Would it be alright to remove the `--all-features` flag? (Not a big deal though)

TheLostLambda: Yeah, I'm happy to remove the all-features flag! Thanks for testing all of this!

PR #521, opened 2021-05-18
Random fixes

kxt:
Assorted refactors and fixes:
* use Pid instead of RawFd for processes in ServerOsApi
* improve debug performance by writing entire buffers instead of per-byte
* remove muts where not needed

They should have no user-visible impact (except for faster debug output on heavy workloads).


imsnif (approved):
This is great! Tested it locally and everything seems in order. Thanks for this :)

Issue #525, opened 2021-05-19
High memory usage and latency when a program produces output too quickly


kxt:
version `0.12.0`

### Reproduction
Run `time cat 1_000_000_lines.txt` and compare it with the time measured by wall clock. The amount of time `time` reports is significantly less than what is observed. Having a slow computer and using non-optimized (non-`release`) build helps.

### Root cause
A program's output is being processed by the `Pty` thread. This thread reads the output and propagates it to the `Screen` thread for rendering through an unbounded channel called `to_screen`. If the program produces output faster than what `Screen` can render, `to_screen`'s queue starts to grow indefinitely. This results in growing memory usage and input latency.

An alternative explanation (kudos to `@aeinstein`): running nearly at the speed of light, `cat` is experiencing time dilation compared to a stationary observer.

As PR #523 increases the throughput of `Pty` thread, it exacerbates this issue.

kxt:
Currently there is a workaround in `Pty` calling `task::sleep` to slow it down.

However, as there is no right amount of sleep that works for every machine, the proper solution would be to have back pressure between `Pty` and `Screen`, so if `Screen` can't keep up, `Pty` will stop reading from the program until the queue clears up. This'd need  a bounded queue between `Pty` and `Screen`, however, as `Pty` is async, we can't simply replace the queue with an mpsc `SyncSend`, as it is blocking. This means an async queue is needed, like [async-std's bounded channel](https://docs.rs/async-std/1.9.0/async_std/channel/fn.bounded.html).

kxt:
I found a workable way to add back pressure between `pty` and `screen` threads without triggering the mentioned deadlock.

The idea is to have a dedicated bounded channel between `pty` and `screen`, so we gain back pressure, while keeping the unbounded channel between every other component and `screen`, to avoid deadlocks.

This required a change from `mpsc` to `crossbeam`, as mpsc lacks a way to select between multiple Receivers.

@imsnif If this is an acceptable approach ([poc patch here](https://github.com/kxt/zellij/commit/dacdc76fc0b1c8212adaf457064ecce79384817e)) I'll whip up a PR-able implementation.

imsnif:
@kxt - at a brief glance this looks like a good direction (though I haven't run it or looked too deeply at it yet). I'm good with moving forward with a PR so we can look deeper. I might want to bring in @kunalmohan to also take a look (if he has the time) since he's done a lot of work in this area.

kunalmohan:
[...]

PR #534, opened 2021-05-25
refactors for #525

kxt:
These patches are assorted refactors in preparation for fixing #525.


imsnif:
Hey @kxt - this looks good. Thanks for dividing it up into byte-sized commits to make it easier to review.
The only question I have is what you plan on doing with https://github.com/zellij-org/zellij/pull/534/commits/79ac50521028f4a84062e93d525b0b42f1a8ef8f? Why is it necessary?

kxt:
>  79ac505? Why is it necessary?

That is the right question. As this is a refactor patch, it does not change any observable behavior, it just makes `Bus` ready to accept multiple receivers, and it's `recv` method returns incoming data on any of them.

The first actual beneficiary of this is the [back pressure patch](https://github.com/kxt/zellij/pull/1/commits/aa65f897af196fef061fd08b346802a27145c5f9), where both the bounded and unbounded queue receivers will be passed for the `Bus` used by `Screen`. 

I'll update the commit message to make it clear.

imsnif:
Sounds good. I hope you don't mind if I hold off on merging this until tomorrow. I plan on releasing a version tomorrow morning and it's already quite huge. Will merge it afterwards.

zellij-utils/src/channels.rs
```diff
@@ -2,28 +2,18 @@
...
-/// An [MPSC](mpsc) synchronous channel with added error context.
-pub type SyncChannelWithContext<T> = (
-    mpsc::SyncSender<(T, ErrorContext)>,
-    mpsc::Receiver<(T, ErrorContext)>,
-);
+pub type ChannelWithContext<T> = (Sender<(T, ErrorContext)>, Receiver<(T, ErrorContext)>);
 
 /// Wrappers around the two standard [MPSC](mpsc) sender types, [`mpsc::Sender`] and [`mpsc::SyncSender`], with an additional [`ErrorContext`].
 #[derive(Clone)]
 pub enum SenderType<T: Clone> {
```
kunalmohan, line 14: I think we can remove the `SenderType` enum entirely since we only have a single type of sender. We can have the `crossbeam_channel::Sender` directly as a field of `SenderWithContext`.

kxt: That's right, although I liked the idea of abstracting away the queue implementation in the original code, even though it was leaky quite.

I'll remove it in a subsequent patch.

kunalmohan: > even though it was leaky quite.

I'm sorry I didn't quite understand what you meant by that. What's wrong with it?

kxt: That was a typo, I meant "quite leaky". If you are not familiar with the phrase "leaky abstraction", it means that what was being abstracted was still visible by the users of the abstractions, in this case mpsc types.

But it was all right, crossbeam is almost a drop-in replacement, so I barely had to do anything to change it.

kunalmohan: Ohh. TIL. Thanks for clarifying that 🙂. I wasn't familiar with the phrase.

PR #539, opened 2021-05-28
Fix build on platforms with TIOCGWINSZ / ioctl() integer type mismatch.

riggs-:
The conversion is required e.g. on 64 bit FreeBSD to convert `TIOCGWINSZ` (hard coded to `u32`) into an `u64` which is expected by `ioctl()`. It might be an issue on other platforms as well.

### DIFF
--- a/zellij-client/src/os_input_output.rs
+++ b/zellij-client/src/os_input_output.rs
@@ -45,7 +45,7 @@ pub(crate) fn get_terminal_size_using_fd(fd: RawFd) -> PositionAndSize {
         ws_ypixel: 0,
     };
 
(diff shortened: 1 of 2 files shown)


kunalmohan:
@riggs- Thanks for the fix! Looks like clippy doesn't like your changes. Could you also add a small comment in the code mentioning the reason for this change (or a link to this PR itself, so that other developers are not tempted to remove the ignore-lint attribute and the conversion) and the relevant `#[allow(clippy::..)]` attributes so that the check doesn't fail. Then we can go ahead and merge this.

riggs-:
Sure, will do later.

PR #547, opened 2021-05-29
Fix build on platforms with TIOCGWINSZ / ioctl() integer type mismatch.

riggs-:
As discussed in https://github.com/zellij-org/zellij/pull/539#issuecomment-850349322, provide context, comment and clippy lint hint.


kunalmohan:
@riggs- please run `cargo fmt` and push the changes again.

kunalmohan (approved):
Thanks!

PR #550, opened 2021-05-29
Add color theme config

### LINKED ISSUE #390: Color scheme support
Hi,
Are there any plans on adding support for color scheme settings? I'm using a light terminal color scheme and the bright green characters in the Tips e.g. are really hard to read on a white (or other light) terminal background.

a-kenji:
* add option `theme` that allows for setting of a theme,
the default is `default`

* under `themes` the themes can be described as follows:

either
```
themes:
  default:
    fg: [0,0,0]
    bg: [0,0,0]
    black: [0,0,0]
    red: [0,0,0]
    green: [0,0,0]
    yellow: [0,0,0]
    blue: [0,0,0]
    magenta: [0,0,0]
    cyan: [0,0,0]
    white: [0,0,0]
    orange: [0,0,0]
```
or
```
themes:
  default:
    fg: 0
    bg: 0
    black: 0
    red: 0
    green: 0
    yellow: 0
    blue: 0
    magenta: 0
    cyan: 0
    white: 0
    orange: 0
```

If the key is different from default, it needs to either be specified on
start with `options --theme [THEME]`, or in the configuration file under
theme: [THEME].

closes #390

### DIFF
--- a/zellij-client/src/lib.rs
+++ b/zellij-client/src/lib.rs
@@ -91,7 +91,14 @@ pub fn start_client(
     let take_snapshot = "\u{1b}[?1049h";
     let bracketed_paste = "\u{1b}[?2004h";
     os_input.unset_raw_mode(0);
-    let palette = os_input.load_palette();
+    let config_options = Options::from_cli(&config.options, opts.command.clone());
+    let palette = config.themes.clone().map_or_else(
+        || os_input.load_palette(),
+        |t| {
+            t.theme_config(&config_options)
+                .unwrap_or_else(|| os_input.load_palette())
+        },
+    );
     let _ = os_input
         .get_stdout_writer()
         .write(take_snapshot.as_bytes())
@@ -101,8 +108,6 @@ pub fn start_client(
         .write(clear_client_terminal_attributes.as_bytes())
         .unwrap();
     std::env::set_var(&"ZELLIJ", "0");
-
-    let config_options = Options::from_cli(&config.options, opts.command.clone());
 
     let full_screen_ws = os_input.get_terminal_size_using_fd(0);
     let client_attributes = ClientAttributes {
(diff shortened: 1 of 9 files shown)


a-kenji:
This is not quite ready yet, since the theme reverts to its default
once its mode is changed. 

I am not sure where we should attach the client `attrs` in route in order to get the
attributes of the client session.

imsnif:
Just a heads up: some terminals don't support truecolor. That's the reason our default theme uses 256 colors. It might be a good idea to differentiate the themes to 256 and truecolor themes. Or at least provide a configuration option for 256 colored themes.

a-kenji:
Oh yeah, that is also possible to specify currently.
I will update the current post.

kunalmohan:
> I am not sure where we should attach the client `attrs` in route in order to get the
> attributes of the client palette attributes.

The idea was to have screen and plugins to store a copy of the palette and generate the mode_info themselves when the mode changes (instead of having the router thread do that). Original comment: https://discord.com/channels/771367133715628073/773128553524494367/843837258120298577

But it'll, probably, be a while before this is done (Brooks is currently working on the resizing algorithm), so maybe go with plan B and store the palette in `SessionMetaData` for now and use that palette [here](https://github.com/zellij-org/zellij/blob/fe299325eb54e6642d27a417a3922a757b4390e4/zellij-server/src/route.rs#L37) instead of the default one?

What do you think @imsnif ?

zellij-client/src/lib.rs
```diff
@@ -91,7 +91,15 @@ pub fn start_client(
...
-    let palette = os_input.load_palette();
+    let config_options = Options::from_cli(&config.options, opts.command.clone());
+    let palette = config
+        .themes
+        .clone()
+        .map(|t| {
+            t.theme_config(&config_options)
+                .unwrap_or_else(|| os_input.load_palette())
+        })
+        .unwrap_or_else(|| os_input.load_palette());
```
kunalmohan, line 102: We can probably use `map_or_else` here.

a-kenji: Awesome, yes!

a-kenji:
I think that is a good idea for now then.
Thank you for showing me the discussion @kunalmohan!

Issue #474, opened 2021-05-09
Don't change the names of "unnamed" tabs when closing/opening tabs


khs26:
Currently, if I have e.g. four tabs open:
- `foo`
- `Tab #2` (unnamed)
- `Tab #3` (unnamed)
- `bar`

If I close tab `foo`, then the following changes occur:
- `foo` -> `<dead>`
- `Tab #2` -> `Tab #1`
- `Tab #3` -> `Tab #2`
- `bar` -> `bar`

The renaming of the numbered tabs is quite confusing. We should probably treat the numbers as automatically generated names, that are retained for the lifetime of the tab.

AdaShoelace:
I'm currently working on implementing this but it has me asking some questions.
Lets say that we have three tabs open e.g `Tab #1`, `Tab #2` and `Tab #3`. We then decide to close `Tab #2`, which will leave us with `Tab #1` and `Tab #3`. In the case that we open a new tab will it fill the "hole" or just get appended as `Tab #4` and leave the list with a "hole" at index 1?

imsnif:
> Lets say that we have three tabs open e.g `Tab #1`, `Tab #2` and `Tab #3`. We then decide to close `Tab #2`, which will leave us with `Tab #1` and `Tab #3`. In the case that we open a new tab will it fill the "hole" or just get appended as `Tab #4` and leave the list with a "hole" at index 1?

My vote goes for leaving a hole and opening `Tab #4`. I find the other behaviour a little surprising as a user "What? I just opened a tab, where is it?" - but maybe that's just me :)

khs26:
Yeah, I'd agree with just appending to the list of tabs and leaving a hole.

AdaShoelace:
Great, I agree with leaving a "hole" aswell.
Know I just need to understand the relationship between the name and rendering the text. From what I can gather the tab itself will have an empty string for a name if it is unnamed but still get rendered as `Tab #<num>`. 
Judging from this discussion this is not the way we want it to behave. But rather that it has a generated name.

a-kenji:
This should be adressed by #554.

PR #554, opened 2021-05-31
Don't rename 'unnamed' tabs upon deletion of other tabs

AdaShoelace:
This PR aims to solve issue #474.
I decided to put the logic in `Tab::new`. This approach felt the "cleanest" to me.
If this breaks any style guides or should be done differently for other reasons please let me know and I'll change it.

### DIFF
--- a/zellij-server/src/tab.rs
+++ b/zellij-server/src/tab.rs
@@ -259,7 +259,11 @@ impl Tab {
             index,
             position,
             panes,
-            name,
+            name: if name.is_empty() {
+                format!("Tab #{} (unnamed)", position + 1)
+            } else {
+                name
+            },
             max_panes,
             panes_to_hide: HashSet::new(),
             active_terminal: pane_id,
--- a/zellij-server/src/tab.rs
+++ b/zellij-server/src/tab.rs
@@ -257,7 +257,7 @@ impl Tab {
         };
 
         let name = if name.is_empty() {
-            format!("Tab #{} (unnamed)", position + 1)
+            format!("Tab #{}", position + 1)
         } else {
             name
         };
--- a/zellij-server/src/tab.rs
+++ b/zellij-server/src/tab.rs
@@ -255,15 +255,18 @@ impl Tab {
         } else {
             BTreeMap::new()
         };
+
+        let name = if name.is_empty() {
+            format!("Tab #{} (unnamed)", position + 1)
+        } else {
+            name
+        };
+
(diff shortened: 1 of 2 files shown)


zellij-server/src/tab.rs
```diff
@@ -259,7 +259,11 @@ impl Tab {
             index,
             position,
             panes,
-            name,
+            name: if name.is_empty() {
```
a-kenji, line 262: I would prefer the initialization here just like the `panes` are initialized.

`let name =` prior and then keep the name here.

a-kenji (changes requested):
Thank you, this is already pretty great and
I think including it in `tab` does make a lot of sense!

There is still one thing that I would like to see a little bit different:

When opening a tab and not really caring about the name,
I think I know what you want to convey with the (unnamed), but I think it
does take up too much space. And should be fine with being omitted.

AdaShoelace:
> Thank you, this is already pretty great and
> I think including it in `tab` does make a lot of sense!
> 
> There is still one thing that I would like to see a little bit different:
> 
> When opening a tab and not really caring about the name,
> I think I know what you want to convey with the `(unnamed)`, but I think it
> does take up too much space. And should be fine with being omitted.

"(unnamed)" has been removed.

a-kenji (approved):
Awesome, thank you!

ammkrn:
@a-kenji 

It seems like some underlying map of tabs to numbers still exhibits the old behavior. If I have `tab1, tab2, tab3, tab4` and close`tab2`, the remaining tabs are `tab1, tab3, tab4`, but in tab mode, pressing the `2` key jumps to `tab3` and the `3` key jumps to `tab4`.

If other people are observing this I can take a look at it. Hints for where to look would be appreciated.

a-kenji:
@ammkrn,
This is correct, but sadly still confusing behaviour.
The tabs are now named that way by default, but the keys
map to the position in the tab. I hope with #557 we can show 
the number of the tab and then the name of the tab with a 
running program/shell behind.

I believe we want (am happy about more input):
To go to a specific tab-name: #379. (An ongoing pr atm).
To go to a specific tab#. (The current 2 behaviour).

I imagined this here to be in scope of #557:
To show the number of the current tab.
To show the name of the current tab. (possibly with name of currently running process)

ammkrn:
Thanks for clearing that up; I look forward to seeing how those solutions work out.

Issue #533, opened 2021-05-23
How to search in the scroll buffer?


gaydenko:
Have not found the way to search the scroll buffer. Attempt to use konsole or alacritty search facilities  just searches in the current screen.

imsnif:
We unfortunately do not have this feature yet. PRs very much welcome!

msirringhaus:
I would really like to have that feature as well. PR #566 seems to be dead in the water currently, possibly because it is unclear at the moment how to integrate the feature in the UI. 
Therefore I would like to rekindle the discussion here about that.
(Note that I have _not_ yet looked in the code much, so I'm arguing from an abstract point here)

I see 2 more or less obvious choices where the search could be located:
1. Have it's own mode next to "Move" and "Scroll".
2. Have it as a sub-function of "Scroll".

## 1. Own mode
### Pro
* Conceptually, it **is** sort of a separate mode. When searching, the pane behaves differently to all other modes, accepting different commands, etc.
* Using it's own mode, modifiers ("case sensitive", "whole word only", "wrapping search", etc.) and commands ("Search next", "Search prev", ..) can easily be displayed
* The very obvious shortcut `Ctrl+F` is still free and not clashing with bash-commands
* Less keys to press to start searching

### Con
* As mentioned above, it is only sort of a separate mode, as we would like to be able to use all the scrolling-commands inside a search. So we would have to copy the scroll-help over, or accept them 'silently' (hindering discoverability)
* There are already a lot of modes, so the menu might get a bit crowded

## 2. Sub-functionality of "Scroll"
### Pro
* As above, it can also be viewed as an additional function on top of scrolling, so putting it there would also make sense.
* Menu would not be cluttered with ever more modes
* Scrolling commands are already displayed, helping discoverability
* Ability to modify the search, _before_ starting to search

### Con
* Needs some visual way to distinguish if a pane is in search mode or not
* The sub-menu of "Scroll" might get very crowded, if all search-commands and modifiers are added there as well.
* _Maybe_ a bit less discoverable, because it is hidden behind the Scroll-mode
* More key presses needed to get there


I'm myself undecided on which variation I would prefer as a user.
The choice might be obvious, when looking at the code, though.

msirringhaus:
Relabeling the `Scroll Mode` to `Search Mode` and putting everything there sounds like a good idea (or mayhaps even `Search/Scroll Mode`).

[...]

PR #566, opened 2021-06-02
wip: Search in scrollback history

Stoope:
Implementation of  #533
I'm planning to add support for search in scrollback buffer, something similar to what you can do in tmux.


badrbouslikhin:
Looking forward to this!

PR #568, opened 2021-06-03
Overhauling Zellij's Resize & Layout System

### LINKED ISSUE #619: Default to a proportionate `split_size` in `layouts`
If no `split_size` is specified, then split proportionate to the number of 
`parts` in a split.

Eg:
```
parts:
    - direction: Horizontal
    - direction: Horizontal
    - direction: Horizontal
```
Should each have the same split size by default.

TheLostLambda:
Should close #619 and #406 when merged :)


TheLostLambda:
It's been a very, very, very long journey (and there is still a ways to go) but I'm pinging @a-kenji for some testing of the layout system which should be fully ported over to the new parametric resize.

I'm looking primarily for feedback on bugs and behavior at this point – the code itself needs a lot of cleaning up still.

Feel free to smoke-test the layout system using these test layouts and / or any of your own!
[Test Layouts.zip](https://github.com/zellij-org/zellij/files/6913546/Test.Layouts.zip)

Manual pane resizing (i.e. all of resize mode) is currently broken, and trying to use it may cause an explosive memory leak? No idea why, but maybe be careful around that for now...

Things still to do:

- [ ] Port manual resizing over to the parametric system
- [ ] Add back all of the tests that I just deleted from the branch (and make them pass)
- [ ] Clean up the code and delete legacy cruft

Also pinging @imsnif if he wants a part of this chaos :heart: 

A demo video, if you're a curious passer-by:
https://user-images.githubusercontent.com/6251883/127787597-24a671cc-743e-4d80-bc06-e86e16eb3619.mp4

a-kenji:
@TheLostLambda
This already looks great! I did come across some issues.
I am using i3 while testing this.

It seems that on every resize the memory does increase?
And I also had the memory leakage bug that you showed offline.
Also on a very small window it can't satisfy the constraint anymore and does panic.

There seems also an issue with this layout  `implicit.yaml`:
image
Then closing one pane:
image
Closing the next pane does close the session. 
Resizing in this state does give me an unsatisfiable constraint.
Log of this session with resize: 
[zellij.log](https://github.com/zellij-org/zellij/files/6917467/zellij.log)

TheLostLambda:
@a-kenji Heya! Sounds like things are mostly what I expected there, but memory leaking on resize is something I certainly need to look into...

The reason you had issues closing a pane (I assume) is because I've not really implemented any of the logic for that :sweat_smile: I don't think that creating, resizing, or closing panes works yet.

The crash when things get too small is also to be expected. Currently I'm calling unwrap on that Result, but I have a commented-out line that nicely propagates the error – aborting the resize but without crashing the program. I have it intentionally crashing now so that I can find the bugs more easily :slightly_smiling_face: 

Perhaps a silly question, but do you know if the memory use on each resize is from the Zellij process or the large log file being created in `/tmp`?

Thanks for giving things a look!

a-kenji:
@TheLostLambda 

> Perhaps a silly question, but do you know if the memory use on each resize is from the Zellij process or the large log file being created in /tmp?

That could be! I'll try to look more into it.

> The crash when things get too small is also to be expected. Currently I'm calling unwrap on that Result, but I have a commented-out line that nicely propagates the error – aborting the resize but without crashing the program. I have it intentionally crashing now so that I can find the bugs more easily.

The problematic memory problems I think come from not crashing in certain situations currently?
If that makes sense. When I sometimes make the window small and then resize again,
if it doesn't crash it freezes my whole session for a while.

TheLostLambda:
> The problematic memory problems I think come from not crashing in certain situations currently?

Well, I didn't quite believe this at first, but then I tested things even with my non-crashing code and you are quite correct :sweat_smile: 30GB of RAM gone in the blink of an eye...

I'll look into this deeper! Thanks for the good find!

TheLostLambda:
@a-kenji Haha, well, I figured by reckless casting of types would lead to this... I present to you `-1` as a `usize`:
image
Just a hunch, but `18446744073709551612` rows on the screen could allocate a quite large buffer... :sweat_smile: :wink:

TheLostLambda:
@a-kenji , if you'd like to test commit `955a337`, the "Make things a bit less crashy" commit, then that should be hard (impossible?) to crash to get the memory leaks on? Let me know if it gives you any issues!

imsnif:
Hey @TheLostLambda - following your request on Discord I took a look at the e2e tests. We're using an instance of `TerminalPane` in the e2e test runner in order to interpret the output of the tests and create the snapshots. Looks like when we were constructing it, we were creating it with columns and rows of 1 (`PaneGeom::Default()`). Which caused some issues :) I fixed that and pushed it. Do note that the rest of the e2e tests need some attention. In some places it's about changing the expected cursor position (since we're now drawing an outer frame in all cases) but I think in other cases they indicate some real bugs. I think they'll help fixing some functionality.

A few other things:
1. We weren't using the default title (eg. `Pane #1`) in the terminal panes. I fixed that.
2. In the terminal pane rendering function it seems like we're creating and rendering the pane frame on every render. This will probably cost us dearly in performance. I'd recommend going back to the old way of only rendering it if the title/scroll/dimensions/position have changed.
3. I tried to test 2, but when I did a `cargo make build-release` I could not run the binary. I got this error:
```
Error: thread 'wasm' panicked at 'called `Result::unwrap()` on an `Err` value: Link(Import("zellij", "host_set_fixed_height", UnknownImport(Function(FunctionType { params: [I32], results: [] }))))': zellij-server/src/wasm_vm.rs:111
   0: zellij_utils::errors::handle_panic
   1: std::panicking::rust_panic_with_hook
             at /rustc/a178d0322ce20e33eac124758e837cbd80a6f633/library/std/src/panicking.rs:626:17
   2: std::panicking::begin_panic_handler::{{closure}}
             at /rustc/a178d0322ce20e33eac124758e837cbd80a6f633/library/std/src/panicking.rs:519:13
   3: std::sys_common::backtrace::__rust_end_short_backtrace
             at /rustc/a178d0322ce20e33eac124758e837cbd80a6f633/library/std/src/sys_common/backtrace.rs:141:18
   4: rust_begin_unwind
             at /rustc/a178d0322ce20e33eac124758e837cbd80a6f633/library/std/src/panicking.rs:515:5
   5: core::panicking::panic_fmt
             at /rustc/a178d0322ce20e33eac124758e837cbd80a6f633/library/core/src/panicking.rs:92:14
   6: core::result::unwrap_failed
             at /rustc/a178d0322ce20e33eac124758e837cbd80a6f633/library/core/src/result.rs:1355:5
   7: zellij_server::wasm_vm::wasm_thread_main
   8: std::sys_common::backtrace::__rust_begin_short_backtrace
   9: core::ops::function::FnOnce::call_once{{vtable.shim}}
  10: <alloc::boxed::Box<F,A> as core::ops::function::FnOnce<Args>>::call_once
             at /rustc/a178d0322ce20e33eac124758e837cbd80a6f633/library/alloc/src/boxed.rs:1575:9
      <alloc::boxed::Box<F,A> as core::ops::function::FnOnce<Args>>::call_once
             at /rustc/a178d0322ce20e33eac124758e837cbd80a6f633/library/alloc/src/boxed.rs:1575:9
      std::sys::unix::thread::Thread::new::thread_start
             at /rustc/a178d0322ce20e33eac124758e837cbd80a6f633/library/std/src/sys/unix/thread.rs:71:17
  11: start_thread
  12: __GI___clone
⋊> ~/c/zellij on resize-overhaul ⨯  
```

I would like to stress that I have not given this a thorough go-over. These are just things I happened to find when fixing the e2e tests. When you feel this is in good enough shape, I very much encourage you to give it a good shake (this is a big change and I feel you're in the best position to know what's changed and what might be affected) - and once you've done that, I'd be happy to give it another one myself just to be sure.

Good work so far! Looking forward to seeing this in main :)

TheLostLambda:
Hey @imsnif ! Thanks for giving things a look! I'm sure there are still a set of bugs that need to be trampled, but we are getting a lot closer :)

I'd really like to make sure that rendering performance is truly affected before adding back complexity though. I think you just need to run with some updated plugins. You could probably run `rm ~/.local/share/zellij -r` if you don't have anything important there. Alternatively, how have you been benchmarking render performance?

TheLostLambda:
@imsnif I'm still struggling with the tests I think. I'm getting timeouts on most of them. Things like `Timed out waiting for data on the SSH channel for test close_pane. Was waiting for step: Split pane to the right`. I don't quite understand what's going on there,  given splitting certainly works when tested manually.

imsnif:
@TheLostLambda - I would be incredibly surprised if rendering performance is not affected. The way things are here, we render the whole pane frame for each letter we type in the terminal. When I added in the frames, I saw it being affected even only by rendering the title/scroll every time rather than the whole frame. But if you'd like to check it, by all means:

I check things by creating a file of 2M lines and then `cat`ing it inside a pane. I measure the performance with `hyperfine`:
`hyperfine --show-output "cat /tmp/bigfile"`

If I need to remove things from `.local`, won't that mess things up for me when going back to the former version of Zellij? What's the upgrade path for this once we release?

TheLostLambda:
@imsnif I agree, I'd assume performance is affected but it's good for my peace of mind to test it. Any particular way you've been creating your 2M line file? And no worries about removing things from `.local`, I can test things. It's something automatically handled by the build-script I believe – updating the assets on things like cargo install. It will certainly need testing to be sure though.

imsnif:
> @imsnif I'm still struggling with the tests I think. I'm getting timeouts on most of them. Things like `Timed out waiting for data on the SSH channel for test close_pane. Was waiting for step: Split pane to the right`. I don't quite understand what's going on there, given splitting certainly works when tested manually.

I'd recommend trying to debug the test itself. The tests send key sequences through the ssh channel. Do the key sequences get through? If you send something else, will that get through? If it got to this step successfully, what happened in the previous step? How is that different?

imsnif:
> @imsnif I agree, I'd assume performance is affected but it's good for my peace of mind to test it. Any particular way you've been creating your 2M line file? And no worries about removing things from `.local`, I can test things. It's something automatically handled by the build-script I believe – updating the assets on things like cargo install. It will certainly need testing to be sure though.

Cool. I do a fish loop: `for i in (seq 500000); echo sldkfjsldkjflskdjflskdjflksjdflksjdflksjdlfkjs$i >> /tmp/bigfile; end` and repeat it 4 times because fish doesn't like too many expansions :)

TheLostLambda:
@imsnif I've done some benchmarking and while `main` (I think?) is usually a bit faster, it's bordering on statistically insignificant in some comparisons. Using `Ctrl-p z` to toggle off pane frames speeds up both branches significantly.

I'll try to change my branch to render frames only when needed so I can test that change in isolation.

```
File Creation:
dd if=/dev/random bs=1M count=100 | base64 > /tmp/100MB

Benchmarking:
hyperfine --warmup 5 --min-runs 25 --show-output "cat /tmp/100MB"

Alacritty:
  Time (mean ± σ):      2.412 s ±  0.021 s    [User: 1.0 ms, System: 1175.3 ms]
  Range (min … max):    2.378 s …  2.443 s    10 runs
  
Tmux:
  Time (mean ± σ):      3.127 s ±  0.049 s    [User: 1.7 ms, System: 1265.2 ms]
  Range (min … max):    3.055 s …  3.190 s    10 runs
  
Zellij (main):
  Time (mean ± σ):      3.430 s ±  0.021 s    [User: 2.4 ms, System: 1454.5 ms]
  Range (min … max):    3.392 s …  3.459 s    10 runs
  
  Time (mean ± σ):      3.454 s ±  0.024 s    [User: 1.9 ms, System: 1514.4 ms]
  Range (min … max):    3.396 s …  3.512 s    25 runs
  
  Time (mean ± σ):      3.482 s ±  0.035 s    [User: 1.4 ms, System: 1532.7 ms]
  Range (min … max):    3.400 s …  3.556 s    25 runs
  
Zellij (main – no borders):
  Time (mean ± σ):      3.182 s ±  0.042 s    [User: 1.8 ms, System: 1486.6 ms]
  Range (min … max):    3.124 s …  3.239 s    10 runs
  
Zellij (resize-overhaul):
  Time (mean ± σ):      3.489 s ±  0.029 s    [User: 1.3 ms, System: 1503.6 ms]
  Range (min … max):    3.441 s …  3.527 s    10 runs

  Time (mean ± σ):      3.557 s ±  0.030 s    [User: 1.8 ms, System: 1583.1 ms]
  Range (min … max):    3.515 s …  3.616 s    25 runs
  
  Time (mean ± σ):      3.480 s ±  0.028 s    [User: 1.2 ms, System: 1568.8 ms]
  Range (min … max):    3.425 s …  3.554 s    25 runs
    
Zellij (resize-overhaul – no borders):
  Time (mean ± σ):      3.218 s ±  0.030 s    [User: 1.9 ms, System: 1481.8 ms]
  Range (min … max):    3.192 s …  3.275 s    10 runs
```

imsnif:
Just a note: statistically insignificant with very strong hardware becomes rather significant with weaker harder at times. I also assume that the difference with the borders is due to updating the scroll (but that's just a guess)

TheLostLambda:
@imsnif I've continued benchmarking and even wrote a version of my branch that only renders when the new PaneFrame is different from the last rendered one, and I tested things in the much slower-to-render Gnome terminal. I cannot find a performance difference between any of the three and it's become a bit of a time-sink. I might carry on with bug fixing and leave this performance matter for later to see if anyone can produce a significant difference. I can't test the effectiveness of my rendering changes when bench-marking reveals no difference to begin with.

```
Gnome Updated Resize Branch:
  3.272 ± 0.034
  
Gnome Old Resize Branch:
  3.221 ± 0.037
  
Gnome Main Branch:
  3.280 ± 0.032
```

PR #576, opened 2021-06-14
Move `layout.rs` from `server` to `utils`

a-kenji:
* give feedback on most errors in the layout-file


a-kenji:
@TheLostLambda 
Thank you for checking the errors!
Do you think it can lead to problems omitting the
```
#[serde(skip_serializing_if = "Option::is_none")]
```
attributes, or is the struct just slightly larger now?

TheLostLambda:
> @TheLostLambda
> Thank you for checking the errors!
> Do you think it can lead to problems omitting the
> 
> ```
> #[serde(skip_serializing_if = "Option::is_none")]
> ```
> 
> attributes, or is the struct just slightly larger now?

I think that should be fine. I think this is actually the only place the struct is serialised at the moment? If I recall correctly it may have once been intended to produce cleaner outputs when saving a layout from Zellij, but having an empty YAML field really isn't too bad, I don't think :)

a-kenji:
Thank you!
Yes it is the only location that serializes the struct currently.

Issue #541, opened 2021-05-28
Feature request: zellij attach --create


srid:
A command that automatically creates (if it doesn't exist) and joins the given named-session would be useful.

It can come in handy in creating quick project-specific sessions, for instance:

```
zellij attach --create $(basename `pwd`)
```

srid:
Tmux analogue

```
tmux new-session -A -s Foo
```

a-kenji:
Thank you for taking the time creating this issue!
This would be a welcome addition.

pedrohjordao:
It seems like it should be easy enough. I would like to take a stab at it.

Any traps I should be aware of?

imsnif:
@pedrohjordao - tbh I'm personally not familiar with this part of the code very well, the person who wrote it is a little busy in the coming weeks (which is why I don't want to tag him :) ). Which is to say that I'm not aware of any specific traps - but I'm sure we can figure this out. Feel free to reach out here or on the Discord #development channel if you have any questions or would like some guidance.

kunalmohan:
Hi @pedrohjordao! I don't think there are any traps in this part of the code. The logic for implementing this will go [here](https://github.com/zellij-org/zellij/blob/7f67e6d97f158a00d1f2c21326fe320388ec8446/src/main.rs#L57-L76). This block of code can be a bit tricky to understand, so if you need any help, feel free to ping me (here or on discord) 🙂

kunalmohan:
Implemented in #731

PR #581, opened 2021-06-18
Detach session when forced close detected (handling SIGTERM and other signals)

### LINKED ISSUE #572: Feature request: Autodetach
Would be nice to autodetach if closed like byobu/tmux

dantepippi:
Auto detach the session when receiving a forced close signal (SIGWINCH, SIGTERM, SIGINT, SIGQUIT, SIGHUP).

The current behavior when the user manually closes Zellij (ctrl-q) remains the same and does not auto detach.

Closes #572

### DIFF
--- a/zellij-client/src/lib.rs
+++ b/zellij-client/src/lib.rs
@@ -210,11 +210,9 @@ pub fn start_client(
                         }
                     }),
                     Box::new({
-                        let send_client_instructions = send_client_instructions.clone();
+                        let os_api = os_input.clone();
                         move || {
-                            send_client_instructions
-                                .send(ClientInstruction::Exit(ExitReason::ForceDetached))
-                                .unwrap()
+                            os_api.send_to_server(ClientToServerMsg::Action(Action::Detach));
                         }
                     }),
                 );
--- a/zellij-client/src/lib.rs
+++ b/zellij-client/src/lib.rs
@@ -213,7 +213,7 @@ pub fn start_client(
                         let send_client_instructions = send_client_instructions.clone();
                         move || {
                             send_client_instructions
-                                .send(ClientInstruction::Exit(ExitReason::Normal))
+                                .send(ClientInstruction::Exit(ExitReason::ForceDetached))
                                 .unwrap()
                         }
                     }),


zellij-client/src/lib.rs
```diff
@@ -213,7 +213,7 @@ pub fn start_client(
                         let send_client_instructions = send_client_instructions.clone();
                         move || {
                             send_client_instructions
-                                .send(ClientInstruction::Exit(ExitReason::Normal))
+                                .send(ClientInstruction::Exit(ExitReason::ForceDetached))
```
kunalmohan, line 216: Does this behave as expected?
I think we should rather be sending `ClientToServerMsg::Action(Action::Detach)` to the server for the actual detach to happen.
(Even the existing logic seems suspicious to me. We should have been sending `ClientToServerMsg::Action(Action::Quit)` for a clean exit. Must have missed changing this part when the refactor happened.)

dantepippi: Hey @kunalmohan! In my tests with this code I was able to force close zellij (with interrupt signals) and see the session available with list-sessions - that didn't happen with main. But I'm ok with doing some refactoring.

On a related note, what do you think about changing the behavior for `zellij attach` so instead of an error if there are no detached sessions, it would start a new one. What I'm thinking is to always start zellij with the attach option so I recover some eventually detached session when available or create a new one otherwise. Maybe recovering the previous session  should even be the default.

kunalmohan: I believe it is being worked on in https://github.com/zellij-org/zellij/issues/541. The idea is to have a --create (-c) flag with the attach command to create a new session if one does not already exist. I'm not sure about making it default. Attaching to an existing session and creating a new one are two separate operations. Attach command having a default behaviour of "attach or create" sounds a bit weird and confusing. IMO having a config option to change the default behaviour of this command can be a good option.

kunalmohan (approved):
LGTM! Thanks!!

zellij-client/src/lib.rs
dantepippi: But that would involve me knowing the name of a session to use with --create (session-name). My suggestion was more in the line that if a user has a detached session (because it was previously interrupted) when he opens Zellij again he probably wants to continue from where he left, would be a nice UX not having to know that session name. If you have just one session  detached  "recover" it instead of creating new sessions.
So the proposal would be "zellij attach" attaches if there is one session available or creates a new one otherwise.

kunalmohan: How about `zellij a -c`? The session name is already optional for attach command, so in case of `zellij a -c`:
1. If only single session exists, recover it.
2. If more than one sessions exists, exit with a message (as we do now).
3. If no session exists, create a new one with a random name.

kunalmohan: Re: Whether `zellij attach` should have a separate `--create` flag or a default behaviour of "recover-or-create". I'm inclined towards the former, but don't have a strong opinion about it. I'm guessing that other folks might also have an opinion about this, so maybe we can move this discussion to discord or a new issue? We can then decide accordingly.

dantepippi: Thanks @kunalmohan ! I honestly would be happy with both scenarios (default or not) as long as we have the option to "recover or create" without requiring a name to recover.
Yes, lets continue this on Discord.

PR #591, opened 2021-06-29
Add new dependencies to `CONTRIBUTING.md`

a-kenji:
I hit this while trying to test the new system, 
are these dependencies correct?

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -44,6 +44,8 @@ cargo make manpage
 To run `install` or `publish`, you'll need the package `binaryen` in the
 version `wasm-opt --version` > 97, for it's command `wasm-opt`.
 
+To run `test`, you will need the package `pkg-config` and a version of `openssl`.
+
 ## Running the end-to-end tests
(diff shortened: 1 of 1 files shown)


imsnif:
The new test system is just about running `e2e-test`... did it change the normal test for you too? (I personally haven't encountered a need for these packages in either `test` or `e2e-test`, but that doesn't mean it's not true to other systems)

a-kenji:
They seem to have been needed by `openssl-sys`.
It seems to need it even for the `cargo make test` case.

`ssh2` -> `libssh2-sys` -> `openssl-sys` -> `openssl` & `pkg-config`.

Maybe you have already openssl and pkg-config installed?

imsnif:
Ah yeah - ok. That's the `ssh2` package that we use in order to connect to the docker container in the e2e tests. It's a dependency of our tests, which is why you're getting this I suppose.
Good find, thanks!

a-kenji:
Ok, then I will merge this, thanks for the sanity check!

Issue #561, opened 2021-06-01
Zellij Needs a Proper Logging System


TheLostLambda:
Zellij needs a central logging bus that captures the `stderr` from plugins, the client, and the server. Information coming over these channels should be tagged with it's location of origin (e.g. thread / plugin name) and timestamped.

Ideally, using `dbg!()` or something extraordinary similar with "just work", with the values being `dbg!()`ed appearing in the logs. 

This is a mentored issue, so if you'd like to take this on, you can count on some extra help and guidance from some of our maintainers.

qepasa:
Looks like a cool issue. I'd like to give this one a go.

TheLostLambda:
Awesome! Do you have any particular approach in mind or do you need any help understanding the codebase? If you need anything, just let us know!

qepasa:
Some ideas for followup:
- change logging level to more verbose when zellij is started with --debug parameter
- custom log writer that would write to file with colors

imsnif:
This has already implemented. Thanks @qepasa! if you want to make the followups into new issues that can be great.

PR #592, opened 2021-06-30
feat: Add logger to Zellij

qepasa:
We add log4rs create for logging across Zellij. Additionally, we capture
`stderr` output from plugins and log it the same log file as other
Zellij logs.

We set up the logger in `zellij-utils/src/logging.rs`. The logger parses
configuration from `zellij-utils/assets/config/log4rs.yml` file. After calling
`configure_logger` from `logging.rs` the logger should be available in the
calling thread and all children threads.
Second part of this PR is the mechanism that captures `stderr` output from
plugins. We setup a customized drain (`DecoratingPipe`) for plugin 
processes in `zellij-server/src/wasm_vm.rs`. The `DecoratingPipe` logs the
`stderr` output from plugins using the same logger as we will use everywhere
else but with custom format.

Implements: #561 

Todo:
- [x] (kind of, see comment in `DecoratingPipe::flush`) make sure we detect byte boundaries correctly in `DecoratingPipe::write`
- [x] work on log format i.e.: add colors, pad certain parts (looks like there is no way to add colors to file for now)
- [x] cleanup

Note:
I didn't go with `slog` due to it not being fully compatible with standard rust `log::info` (warn, err etc.), instead that project prefers to use it's own `slog::info` (warn, err etc.). I tried to use adapter provided by slog project but it was difficult to share the logger between packages. Slog basically expects you to pass log instance across your codebase, which seems like a lot of effort in case of a project that wasn't initially created with slog usage in mind. Also, there is no official support for adding thread name to log, emitting colored log to file or filtering modules.

Sample from log file:
```
INFO   |zellij_utils::logging    | 2021-07-13 23:15:16.650 [main      ] [zellij-utils/src/logging.rs:56]: Zellij logger initialized 
INFO   |zellij                   | 2021-07-13 23:15:16.650 [main      ] [src/main.rs:26]: Starting Zellij! 
INFO   |zellij_utils::logging    | 2021-07-13 23:15:16.698 [main      ] [zellij-utils/src/logging.rs:56]: Zellij logger initialized 
INFO   |zellij                   | 2021-07-13 23:15:16.698 [main      ] [src/main.rs:26]: Starting Zellij! 
INFO   |zellij_server            | 2021-07-13 23:15:16.709 [main      ] [zellij-server/src/lib.rs:113]: starts server 
INFO   |zellij_server::wasm_vm   | 2021-07-13 23:15:16.762 [wasm      ] [zellij-server/src/wasm_vm.rs:61]: Wasm main thread starts 
INFO   |tab-bar                  | 2021-07-13 23:15:21.357 [id: 0     ] Creating decorating pipe for plugin: tab-bar! 
DEBUG  |tab-bar                  | 2021-07-13 23:15:21.396 [id: 0     ] [default-plugins/tab-bar/src/main.rs:23] "hello from plugin main :)" = "hello from plugin main :)" 
INFO   |strider                  | 2021-07-13 23:15:27.802 [id: 1     ] Creating decorating pipe for plugin: strider! 
DEBUG  |strider                  | 2021-07-13 23:15:27.843 [id: 1     ] [default-plugins/strider/src/main.rs:8] "hello from plugin main :)" = "hello from plugin main :)" 
DEBUG  |strider                  | 2021-07-13 23:15:27.843 [id: 1     ] [default-plugins/strider/src/main.rs:12] "hello from load2?" = "hello from load2?" 
INFO   |status-bar               | 2021-07-13 23:15:32.170 [id: 2     ] Creating decorating pipe for plugin: status-bar! 
DEBUG  |status-bar               | 2021-07-13 23:15:32.198 [id: 2     ] [default-plugins/status-bar/src/main.rs:22] "hello from plugin main :)" = "hello from plugin main :)" 
DEBUG  |status-bar               | 2021-07-13 23:15:32.199 [id: 2     ] [default-plugins/status-bar/src/main.rs:136] "hello from load" = "hello from load" 
INFO   |zellij_client            | 2021-07-13 23:24:45.110 [main      ] [zellij-client/src/lib.rs:304]: Bye from Zellij! 
INFO   |zellij_utils::logging    | 2021-07-13 23:27:27.729 [main      ] [zellij-utils/src/logging.rs:63]: Zellij logger initialized 
INFO   |zellij                   | 2021-07-13 23:27:27.729 [main      ] [src/main.rs:24]: Starting Zellij! 
```


TheLostLambda:
This looks super cool! I'm still a bit homeless at the moment, as I'm between places and moving, but I'll try to give this a look either today or tomorrow! thanks so much for doing this!

Is there anything in particular you'd like me to look at or think about?

qepasa:
Hey, thanks! For now this is still a bit of WIP, but I think that the general mechanism for plugin logs will not change. Neither will the logging crate (unless you or other contributors suggest otherwise). 

As for the initial review, feel free to have a look at it but please remember that it's a bit crude at the moment and might change a bit. I will try to finish it as fast as possible and ping on discord when it's ready for a proper, detailed review. :smiley:. I added a small guide in the initial comment to make the review as easy as possible.

zellij-server/src/decorating_pipe.rs
```diff
@@ -0,0 +1,94 @@
...
+                    "{}: {}",
+                    self.plugin_name,
+                    std::str::from_utf8(&self.buffer.make_contiguous().split_last().unwrap().1)
+                        .unwrap()
+                );
+                self.buffer.clear();
+            }
+        }
+        Ok(buf.len())
+    }
```
qepasa, line 55: I think we need to add code here to ensure that we detect the 'log boundaries' correctly. For example, the plugin might call `write` on `stderr` in the middle of a log message. In such case we shouldn't call `info!` but buffer the bytes and wait until we the buffer on our side gets a `\n`. Also, I am not sure if current way of calling `from_utf` is safe. The `buf` might end on incorrect utf8 symbol boundary. Lastly, we probably should add some unit tests for this to make sure we're handling all those cases described above.

zellij-utils/assets/config/log4rs.yml
```diff
@@ -0,0 +1,42 @@
...
+  stderr:
+    kind: console
+    target: stderr
+
+  logFile:
+    kind: file
+    path: "zellij.log"
+    append: false
+    encoder:
+      pattern: "{h(|{M}| {d} {l} [{T}] [{f}:{L}]: {m})} {n}"
```
qepasa, line 15: I should probably add comments explaining the patterns so everyone can see what the expected format is without going into log4rs documentation.

zellij-server/src/decorating_pipe.rs
qepasa: We don't do it. I added comment in `DecoratingPipe::flush`.

zellij-utils/assets/config/log4rs.yml
qepasa: Done, made the format more verbose so it should be clear how it should look like

zellij-utils/assets/config/log4rs.yml
```diff
@@ -0,0 +1,46 @@
...
+      # module is padded to *at least* 25 bytes and thread is padded to be between 10 and 15 bytes.
+      pattern: "|{module:<25}| {date} {highlight({level})} [{thread:<10.15}] [{file}:{line}]: {message} {n}"
+
+  # plugin appender. To be used in decorating_pipe to forward stderr output from plugins.
+  logPlugin:
+    kind: file
+    path: "zellij.log"
+    encoder:
+      # {n} means platform dependent newline
+      pattern: "{message} {n}"
```
TheLostLambda, line 26: I think we should make this as similar as possible to the `logFile` pattern, then use [`log!`](https://docs.rs/log/0.4.14/log/macro.log.html) in the `DecoratedPipe`.

We should at least include the `date` and maybe even `level` if possible.

Additionally, are there different (shorter) ways to format the `date`?

zellij-utils/assets/config/log4rs.yml
```diff
@@ -0,0 +1,46 @@
...
+appenders:
+  # An appender named "stderr" that writes to stderr
+  stderr:
+    kind: console
+    target: stderr
+
+  # default zellij appender, should be used across most of the codebase.
+  logFile:
+    kind: file
+    path: "zellij.log"
```
TheLostLambda, line 13: We should look into setting this at runtime in Zellij. We already have a logging directory, so it would be nice if these logs were always written to wherever that happens to be.

zellij-utils/assets/config/log4rs.yml
```diff
@@ -0,0 +1,46 @@
...
+    append: false
```
TheLostLambda, line 14: Eventually, this should probably also be true?

TheLostLambda: Having a `date` on all of the messages resolves some ambiguity and we could have some startup / shutdown messages that make it clear when Zellij is restarted.

TheLostLambda (changes requested):
Thank you for all of your hard work! This is looking great! I left a couple of comments as a first pass.

The main things I think we should look into are making the plugin log-lines fit with the rest of the log lines (adding a date at least), and writing to the proper Zellij log directory in `/tmp`. 

Overall, I'm very pleased with this!

zellij-server/src/decorating_pipe.rs
```diff
@@ -0,0 +1,223 @@
...
+        let amt = std::cmp::min(buf.len(), self.buffer.len());
+        for (i, byte) in self.buffer.drain(..amt).enumerate() {
+            buf[i] = byte;
+        }
+        Ok(amt)
+    }
+}
+
+impl Write for DecoratingPipe {
+    fn write(&mut self, buf: &[u8]) -> std::io::Result<usize> {
```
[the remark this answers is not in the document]
qepasa: It's kind of tricky to tag it here. The calls to `write()` might be chunked and spanned across multiple log lines. If we want to add time, probably we should to it in `flush()`. Alternatively, my initial idea was to offload all log formatting to plugin itself. On zellij side, we just do `|plugin_name|: {message}` and that's it. Although, I understand that for convenience, we mainly intend using `dbg!` on plugin side so maybe adding timestamp (in `flush()`) is a better idea.

zellij-utils/assets/config/log4rs.yml
qepasa: We definitely can format date differently! What is the smallest unit we want to include? Milliseconds?

zellij-utils/assets/config/log4rs.yml
qepasa: Sure, we can move logger initialization to runtime and get rid of this file entirely.

zellij-utils/assets/config/log4rs.yml
qepasa: Sounds good!

zellij-utils/assets/config/log4rs.yml
qepasa: I went with hybrid approach. I moved `level` to the beginning of log message but still do some formatting in `logging_pipe`. This way logs from plugin have `plugin_name` instead of `module` and `plugin_id` instead of `thread`.

zellij-server/src/wasm_vm.rs
```diff
@@ -73,6 +76,8 @@ pub(crate) fn wasm_thread_main(bus: Bus<PluginInstruction>, store: Store, data_d
 
                 let output = Pipe::new();
                 let input = Pipe::new();
+                let stderr =
```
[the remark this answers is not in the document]
qepasa: Added plugin id. Plugin name is capped at 25 characters.

zellij-server/src/lib.rs
```diff
@@ -114,6 +116,7 @@ pub(crate) enum SessionState {
 }
 
 pub fn start_server(os_input: Box<dyn ServerOsApi>, socket_path: PathBuf) {
+    info!("starts server");
```
qepasa, line 119: No idea why this (and `Zellij logger initialized`) are printed twice :|

TheLostLambda: Hmm, I may ping @kunalmohan since he did a lot of the daemonization stuff. This could be because that process requires forking twice, but that's just a shot in the dark!

kunalmohan: @qepasa did you mean `Starting Zellij!` (instead of `starts server`)?

I can see why this happens. When the user starts zellij, the control enters the `main()` function which calls `start_client()`. `start_client()` then optionally spawns a new server (if a new sessions is to be started). The way this is done is- we spawn a new zellij process with `zellij --server <socket-path>` ([here](https://github.com/zellij-org/zellij/blob/3df0210647e65170576f1ea97328cce26c211a43/zellij-client/src/lib.rs#L62-L65)). So the main function is called again, but this time it further calls `start_server()`.

We can see that the control goes to the `main()` function twice (in two separate processes). Since both the above logs are printed in main, we see them twice.

qepasa: Makes sense, thanks for explanation😀

zellij-server/src/logging_pipe.rs
```diff
@@ -0,0 +1,258 @@
...
+            Local::now().format("%Y-%m-%d %H:%M:%S.%3f"),
+            format!("id: {}", self.plugin_id),
+            message
+        );
+    }
+}
+
+impl Read for LoggingPipe {
+    fn read(&mut self, buf: &mut [u8]) -> std::io::Result<usize> {
+        let amt = std::cmp::min(buf.len(), self.buffer.len());
```
TheLostLambda, line 52: Usually when you run into manual indexing like this in Rust there is another way to go about it. Sometimes that other way can be super elegant, sometimes less so... I think this is one of the less elegant times, but here is a slightly more "functional" (as in paradigm) approach if you are interested:

image

The same for copy-paste:
```rs
impl Read for LoggingPipe {
    fn read(&mut self, mut buf: &mut [u8]) -> std::io::Result<usize> {
        let data: Vec<_> = self.buffer.drain(..buf.len()).collect();
        buf.write(&data)
    }
}
```

zellij-server/src/logging_pipe.rs
```diff
@@ -0,0 +1,258 @@
...
+                endline \\n symbol.";
+            error!("{}: {}", self.plugin_name, error_msg);
+            self.buffer.clear();
+            return Err(std::io::Error::new(
+                std::io::ErrorKind::InvalidData,
+                error_msg,
+            ));
+        }
+
+        self.buffer.extend(buf);
```
TheLostLambda, line 75: This function looks great now!

TheLostLambda (approved):
I love it! I left a couple of small comments for you to address if you'd like, but I'm happy with merging this! Feel free to do any final cleanup (I'm not sure if you want to keep all of the test messages in?) and merge when ready!

zellij-server/src/lib.rs
TheLostLambda: As for this, if you do want to keep some logging messages in for when Zellij itself is starting (I think that could be a nice way to see where sessions begin and end in the log file), maybe we could try to find a way to print separate messages for client and server startup?

zellij-server/src/logging_pipe.rs
```diff
@@ -0,0 +1,258 @@
...
+pub struct LoggingPipe {
+    buffer: VecDeque<u8>,
+    plugin_name: String,
+    plugin_id: u32,
+}
+
+impl LoggingPipe {
+    pub fn new(plugin_name: &str, plugin_id: u32) -> LoggingPipe {
+        info!(
+            "|{:<25.25}| {} [{:<10.15}] Creating decorating pipe for plugin: {}!",
```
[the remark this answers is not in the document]
qepasa: I am going to remove this log

zellij-server/src/logging_pipe.rs
```diff
@@ -0,0 +1,258 @@
...
+            plugin_name
+        );
+        LoggingPipe {
+            buffer: VecDeque::new(),
+            plugin_name: String::from(plugin_name),
+            plugin_id,
+        }
+    }
+
+    fn log_message(&self, message: &str) {
```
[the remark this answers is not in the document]
qepasa: Removed the `info!` call above.

zellij-server/src/logging_pipe.rs
```diff
@@ -0,0 +1,258 @@
...
+}
+
+impl LoggingPipe {
+    pub fn new(plugin_name: &str, plugin_id: u32) -> LoggingPipe {
+        info!(
+            "|{:<25.25}| {} [{:<10.15}] Creating decorating pipe for plugin: {}!",
+            plugin_name,
+            Local::now().format("%Y-%m-%d %H:%M:%S.%3f"),
+            format!("id: {}", plugin_id),
+            plugin_name
```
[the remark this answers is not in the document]
qepasa: Removed the this log altogether

zellij-server/src/logging_pipe.rs
qepasa: Thanks, I like your code more. To be honest, this was straight copied from wasmer-wasi `Pipe`.
Thinking about it, I think we should remove this code altogether. We're draining our buffer in `flush()`, so I guess if someone were to chain this `LoggingPipe` with something that would read from it, it would see very weird behavior and also logs would be screwed up.

I changed to your version, added note that it should probably be removed.

One caveat:
I had to do:
`buf.as_mut().write(&data)`
without `as_mut()` I was getting:
```

error[E0596]: cannot borrow `buf` as mutable, as it is not declared as mutable
  --> zellij-server/src/logging_pipe.rs:47:9
   |
47 |         buf.write(&data)
   |         ^^^
   |         |
   |         cannot borrow as mutable
   |         try removing `&mut` here

error: aborting due to previous error; 1 warning emitted

```

I have no idea that's the case. I thought `buf` is already mut?

TheLostLambda: I think Rust is being a tad strange here. When calling `.write()` it's attempting to take a `&mut` of the `buf`. Even if the `buf` *stores* a mutable reference, the `buf` variable itself is non mutable, so Rust doesn't let you take a mutable reference of it. If you add `mut buf: ...` instead of just `buf: ...` in the arguments of `read()`, that should fix things (I think)!

With that being said, you seem write about us not using this! I think that, since people really shouldn't be calling `.read()`, we shouldn't let them! How do you feel about mirroring the `Seek` implementation with something like:
image

Copy paste:
```rs
fn read(&mut self, _buf: &mut [u8]) -> std::io::Result<usize> {
    Err(std::io::Error::new(
        std::io::ErrorKind::Other,
        "can not read a LoggingPipe",
    ))
}
```

TheLostLambda:
@qepasa Let me know if you need any help with merge conflicts! I think it's just been a bump in the crate versions though, so it shouldn't be too difficult :)

zellij-server/src/logging_pipe.rs
qepasa: Yeah, it's good idea to just return error for reads. Obviously, it can be changed in future by not draining buffer but keeping a pointer instead. Changed it to return error, should be fine for now.

qepasa:
@TheLostLambda Done :smiley: Also, changed the `LoggingPipe::Read` to return error as discussed above

TheLostLambda:
@qepasa Looks stellar! Thank you for all of your hard work on this! It's been lovely working with you and I'm excited to see you around more!

I'll merge it now and update the changelog!

PR #598, opened 2021-07-04
Remove test feature and hacks

kunalmohan:
This PR aims to remove the "test" feature and the little hacks we had for our former testing framework

### DIFF
--- a/zellij-server/src/panes/grid.rs
+++ b/zellij-server/src/panes/grid.rs
@@ -1766,9 +1766,10 @@ impl Perform for Grid {
                 _ => {}
             }
         } else {
-            let result = debug_log_to_file(format!("Unhandled csi: {}->{:?}", c, params));
-            #[cfg(not(any(feature = "test", test)))]
-            result.unwrap();
+            drop(debug_log_to_file(format!(
+                "Unhandled csi: {}->{:?}",
+                c, params
+            )));
         }
     }
 
--- a/zellij-server/src/panes/terminal_pane.rs
+++ b/zellij-server/src/panes/terminal_pane.rs
@@ -387,7 +387,6 @@ impl TerminalPane {
     pub fn read_buffer_as_lines(&self) -> Vec<Vec<TerminalCharacter>> {
         self.grid.as_character_lines()
     }
-    #[cfg(any(feature = "test", test))]
     pub fn cursor_coordinates(&self) -> Option<(usize, usize)> {
         // (x, y)
         self.grid.cursor_coordinates()
--- a/zellij-utils/src/setup.rs
+++ b/zellij-utils/src/setup.rs
@@ -12,7 +12,6 @@ const CONFIG_LOCATION: &str = ".config/zellij";
 const CONFIG_NAME: &str = "config.yaml";
 static ARROW_SEPARATOR: &str = "";
 
-#[cfg(not(any(feature = "test", test)))]
 /// Goes through a predefined list and checks for an already
 /// existing config directory, returns the first match
 pub fn find_default_config_dir() -> Option<PathBuf> {
@@ -21,11 +20,6 @@ pub fn find_default_config_dir() -> Option<PathBuf> {
         .filter(|p| p.is_some())
         .find(|p| p.clone().unwrap().exists())
         .flatten()
-}
-
-#[cfg(any(feature = "test", test))]
-pub fn find_default_config_dir() -> Option<PathBuf> {
-    None
 }
 
 /// Order in which config directories are checked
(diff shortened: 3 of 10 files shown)


zellij-utils/src/setup.rs
```diff
@@ -23,11 +22,6 @@ pub fn find_default_config_dir() -> Option<PathBuf> {
         .flatten()
 }
 
-#[cfg(any(feature = "test", test))]
```
a-kenji, line 26: This could potentially cause some confusion,
when I run the tests locally now I have one failing test,
since it sources the local configuration.

A possible solution could be starting with the `setup --clean`
in the makefile?

a-kenji (commented):
Hey! Good idea. 
One little thing I am a little concerned about,
but that could be solved differently too.

zellij-utils/src/setup.rs
kunalmohan: Thanks for pointing this out!
Which test is failing for you? `cargo test` does not accept the arguments the same as a normal zellij binary, so I'm not sure how `setup --clean` would help.
For e2e tests, we can add it somewhere in https://github.com/zellij-org/zellij/blob/main/src/tests/e2e/remote_runner.rs but for unit tests we'll have to see.

zellij-server/src/panes/grid.rs
```diff
@@ -1766,9 +1766,10 @@ impl Perform for Grid {
                 _ => {}
             }
         } else {
-            let result = debug_log_to_file(format!("Unhandled csi: {}->{:?}", c, params));
-            #[cfg(not(any(feature = "test", test)))]
-            result.unwrap();
+            drop(debug_log_to_file(format!(
```
kunalmohan, line 1769: Should I unwrap the result from `debug_log_to_file()`? To me, it looks like an error here should not be the reason a user is unable to zellij.

zellij-server/src/panes/terminal_pane.rs
```diff
@@ -387,7 +387,6 @@ impl TerminalPane {
     pub fn read_buffer_as_lines(&self) -> Vec<Vec<TerminalCharacter>> {
         self.grid.as_character_lines()
     }
-    #[cfg(any(feature = "test", test))]
     pub fn cursor_coordinates(&self) -> Option<(usize, usize)> {
```
kunalmohan, line 390: This function is only used in e2e testing, is it okay to have it in normal builds as well `#[cfg(any(feature = "test", test))]`? Maybe add a comment stating that it's required only for e2e tests?

kunalmohan (commented):
@imsnif I need your comments on a couple of changes here before I merge this. I've mentioned them below. Please take a look whenever you're free 🙂

zellij-utils/src/setup.rs
a-kenji: It is `try_from_cli_args_default()`, but now that I think more
about it that probably can get removed now.
I think It mattered more in the old test system.

Ah, I thought we could pass the flags through somehow,
since I think we do use the `--data-dir` flag in the tests atm?

kunalmohan: Are you sure we can remove it? Since it is used in a unit test, we can simply have it behind `#[cfg(test)]` attribute. We won't need a separate "test" feature.
The `--data-dir` is used in `build` command. For tests, `fn main()` is not the entry point and hence we cannot use zellij flags there.

a-kenji: Ah, yes then it probably is still better to keep using
the distinction with `#[cfg(test)]`, if not for now.
I can see it being a point of confusion in the future
if personal configuration changes could affect the
outcome of tests.

zellij-server/src/panes/grid.rs
imsnif: No no, no need to unwrap here at all! If it fails, let it fail.

zellij-server/src/panes/terminal_pane.rs
imsnif: Sure. I think it's not a problem to have this in normal builds as well (it's just a few lines of code). I can also imagine occasions where we might need it in the app itself at some point. So maybe we can leave it in and suppress the compiler warning?

imsnif:
> @imsnif I need your comments on a couple of changes here before I merge this. I've mentioned them below. Please take a look whenever you're free slightly_smiling_face

Commented inline... did I miss anything?

Also - thanks for doing this! If you can find a way to get rid of the `ignore` stuff and still have a separation between the e2e and the normal tests flows that would be amazing (but no sweat, I tried a bit and gave up ;P )

kunalmohan:
> Commented inline... did I miss anything?

Nope, that's it. Thanks!

> If you can find a way to get rid of the `ignore` stuff and still have a separation between the e2e and the normal tests flows that would be amazing

Sure, I'll take a look at it :)

PR #600, opened 2021-07-05
Add commands to layout

a-kenji:
Add the ability to run commands on loading a layout:
```
  - direction: Horizontal
    split_size:
      Percent: 50
    run:
      command: {cmd: htop}
```
or respectively:
```
  - direction: Horizontal
    split_size:
      Percent: 50
    run:
      command: {cmd: htop, args: ["-C"]}
```

In order to specify the difference of commands and
plugins now the plugins need to be under the `run` section:
```
  - direction: Vertical
    split_size:
      Fixed: 2
    run:
      plugin: status-bar
```
This also means that this is a breaking change for people
that already have a custom layout.

Example layouts:
```
---
direction: Vertical
parts:
  - direction: Horizontal
    split_size:
      Percent: 50
    parts:
      - direction: Vertical
        split_size:
          Percent: 50
      - direction: Vertical
        split_size:
          Percent: 50
        run:
          command: {cmd: htop}
  - direction: Horizontal
    split_size:
      Percent: 50
    run:
      command: {cmd: htop}
```
and:
```
---
direction: Horizontal
parts:
  - direction: Vertical
    split_size:
      Fixed: 1
    run:
      plugin: tab-bar
  - direction: Vertical
    parts:
    - direction: Vertical
      parts:
      - direction: Vertical
        split_size:
          Percent: 50
        run:
          command: {cmd: htop}
      - direction: Vertical
        split_size:
          Percent: 50
        run:
          command: {cmd: htop, args: ["-C"]}
  - direction: Vertical
    split_size:
      Fixed: 5
  - direction: Vertical
    split_size:
      Fixed: 2
    run:
      plugin: status-bar
```

### DIFF
--- a/zellij-server/src/pty.rs
+++ b/zellij-server/src/pty.rs
@@ -1,11 +1,3 @@
-use zellij_utils::async_std;
-
-use async_std::future::timeout as async_timeout;
-use async_std::task::{self, JoinHandle};
-use std::collections::HashMap;
-use std::os::unix::io::RawFd;
-use std::time::{Duration, Instant};
-
 use crate::{
     os_input_output::{AsyncReader, Pid, ServerOsApi},
     panes::PaneId,
@@ -14,9 +6,22 @@ use crate::{
     wasm_vm::PluginInstruction,
     ServerInstruction,
 };
+use async_std::{
+    future::timeout as async_timeout,
+    task::{self, JoinHandle},
+};
+use std::{
+    collections::HashMap,
+    os::unix::io::RawFd,
+    time::{Duration, Instant},
+};
 use zellij_utils::{
+    async_std,
     errors::{get_current_ctx, ContextType, PtyContext},
-    input::{command::TerminalAction, layout::Layout},
+    input::{
+        command::TerminalAction,
+        layout::{Layout, Run},
+    },
     logging::debug_to_file,
 };
 
@@ -237,19 +242,36 @@ impl Pty {
     pub fn spawn_terminals_for_layout(
         &mut self,
         layout: Layout,
-        terminal_action: Option<TerminalAction>,
+        default_shell: Option<TerminalAction>,
     ) {
-        let total_panes = layout.total_terminal_panes();
+        let extracted_run_instructions = layout.extract_run_instructions();
         let mut new_pane_pids = vec![];
-        for _ in 0..total_panes {
-            let (pid_primary, pid_secondary): (RawFd, Pid) = self
-                .bus
-                .os_input
-                .as_mut()
-                .unwrap()
-                .spawn_terminal(terminal_action.clone());
-            self.id_to_child_pid.insert(pid_primary, pid_secondary);
-            new_pane_pids.push(pid_primary);
+        for run_instruction in extracted_run_instructions {
+            match run_instruction {
+                Some(Run::Command(command)) => {
+                    let cmd = TerminalAction::RunCommand(command);
+                    let (pid_primary, pid_secondary): (RawFd, Pid) = self
+                        .bus
+                        .os_input
+                        .as_mut()
+                        .unwrap()
+                        .spawn_terminal(Some(cmd));
+                    self.id_to_child_pid.insert(pid_primary, pid_secondary);
+                    new_pane_pids.push(pid_primary);
+                }
+                None => {
+                    let (pid_primary, pid_secondary): (RawFd, Pid) = self
+                        .bus
+                        .os_input
+                        .as_mut()
+                        .unwrap()
+                        .spawn_terminal(default_shell.clone());
+                    self.id_to_child_pid.insert(pid_primary, pid_secondary);
+                    new_pane_pids.push(pid_primary);
+                }
+                // Investigate moving plugin loading to here.
+                Some(Run::Plugin(_)) => {}
+            }
         }
         self.bus
             .senders
--- a/zellij-server/src/tab.rs
+++ b/zellij-server/src/tab.rs
@@ -26,7 +26,10 @@ use std::{
 };
 use zellij_tile::data::{Event, InputMode, ModeInfo, Palette};
 use zellij_utils::{
-    input::{layout::Layout, parse_keys},
+    input::{
+        layout::{Layout, Run},
+        parse_keys,
+    },
     pane_size::PositionAndSize,
     shared::adjust_to_size,
 };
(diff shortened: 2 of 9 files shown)


zellij-server/src/tab.rs
```diff
@@ -318,26 +322,50 @@ impl Tab {
...
+                            .send_to_plugin(PluginInstruction::Update(
+                                Some(pid),
+                                Event::ModeUpdate(self.mode_info.clone()),
+                            ))
+                            .unwrap();
+                    }
+                    // A command pane
+                    Run::Command(command) => {
+                        let cmd = TerminalAction::RunCommand(command.clone());
+                        self.senders
```
a-kenji, line 351: This is for current testing and most likely not the way to go.
If I understand correctly we need the `pid`, which the `ScreenInstruction` doesn't
provide to us.

zellij-server/src/tab.rs
```diff
@@ -318,26 +322,50 @@ impl Tab {
...
+                            .send_to_pty(PtyInstruction::SpawnTerminal(Some(cmd)))
+                            .unwrap();
+
+                        // I would expect something like this to work,
+                        // but I must still be overlooking something.
+                        //let (_, pid) = self.os_api.spawn_terminal(Some(cmd));
```
a-kenji, line 357: This was my most recent attempt.
As far as I understand we want to:
- create the terminal -> get the pid
- use the pid to create a new terminal struct and pass the correct size
- then insert the terminal into the tab struct

a-kenji (commented):
This is still a rought draft and is definitely not working as planned yet.
I would appreciate some pointers here.

It is already possible to just spawn a command in a new pane, but not
according to the constraints specified in the `layout.rs` file.

I left comments on relevant places.

There is also an extra boundary that i can't make sense of.

 
image

imsnif:
Hey @a-kenji, I'd be happy to take a look. It would be helpful to me if we at least start with guided questions rather than code. Want to let me know what you're having trouble with? If it's more than one thing maybe we can start with a list and work our way through it?

a-kenji:
Hey @imsnif, thank you and no problem!

The aim is to start a new terminal in `tab.rs`, resize it accordingly
and add it to the total tabs.

The solution I would assume works is:
- self.os_api.spawn_terminal function to spawn a new terminal and get the relevant pid
- use the pid to create a new terminal struct and resize accordingly
- use the pid to add the terminal struct to the pane

This errors out from the screen thread, so I am hoping you 
might have more insight into this problem.

imsnif:
> This errors out from the screen thread, so I am hoping you
> might have more insight into this problem.

Hey, so I'm not sure why this is erroring - but in any case I don't think this is the way to go. When we create a layout, there is a path that we follow through the various parts of the application. The screen thread is the last step of that path. If you notice, the `apply_layout` method in the screen gets a `new_pids` vector with the pids of terminals that have already been opened in the `pty` thread.

IMO the best thing to do would be to go through that same path. That would probably mean adjusting this method: https://github.com/zellij-org/zellij/blob/47206866b68a00222a95b02abbfdbf0a6f38c781/zellij-server/src/pty.rs#L237 so that it knows to scan the layout to find all the command panes, open them like it does the terminal panes, and then send a less naive structure to the screen with all the "normal" terminal ids on the one hand and the ids of the commands on the other.

How does that sound? Did I manage to explain myself properly?

a-kenji:
Thank you, yes I think that makes sense!

a-kenji:
@imsnif 
This is now ready for a review!

zellij-server/src/pty.rs
```diff
@@ -237,19 +242,36 @@ impl Pty {
...
+                    let (pid_primary, pid_secondary): (RawFd, Pid) = self
+                        .bus
+                        .os_input
+                        .as_mut()
+                        .unwrap()
+                        .spawn_terminal(default_shell.clone());
+                    self.id_to_child_pid.insert(pid_primary, pid_secondary);
+                    new_pane_pids.push(pid_primary);
+                }
+                // Investigate moving plugin loading to here.
```
imsnif, line 272: +1 !!!!!

zellij-utils/src/input/layout.rs
```diff
@@ -29,14 +33,23 @@ pub enum SplitSize {
...
 }
 
+#[derive(Debug, Serialize, Deserialize, Clone)]
+#[serde(crate = "self::serde")]
+pub enum Run {
+    #[serde(rename = "plugin")]
+    Plugin(Option<PathBuf>), // 1 to 100
+    #[serde(rename = "command")]
+    Command(RunCommand), // An absolute number of columns or rows
+}
```
imsnif, line 43: I think maybe the comments here are not relevant?

zellij-server/src/pty.rs
```diff
@@ -237,19 +242,36 @@ impl Pty {
     pub fn spawn_terminals_for_layout(
         &mut self,
         layout: Layout,
-        terminal_action: Option<TerminalAction>,
+        default_shell: Option<TerminalAction>,
     ) {
-        let total_panes = layout.total_terminal_panes();
+        let total_runnables = layout.total_runnables();
```
imsnif, line 247: I think calling this method "total_runnables" is a little confusing. "total" implies we're talking about an amount, and here we're getting back a run instruction... maybe we can call it `extract_run_instructions` and then in the loop call the singular version `run_instruction`? What do you think?

imsnif (changes requested):
This is great! I haven't run it because I trust you made sure it works properly :)
I left a few minor comments in the code. Otherwise, can you maybe include an example with a layout that has a run command in it? Just to make it easier for people to understand how to do this.

imsnif:
I just thought of something else: how do we know to open the command in the pane specified in the layout and not in a different pane? Is there a handling here for it that I missed?

zellij-server/src/pty.rs
a-kenji: Thank you that is very helpful. I think that is a good idea.

zellij-utils/src/input/layout.rs
a-kenji: Oh,good catch!

a-kenji:
> I just thought of something else: how do we know to open the command in the pane specified in the layout and not in a different pane? Is there a handling here for it that I missed?

It opened them already in the correct order.
I tried to test it with various configurations and it just worked.
In `tab.rs` the pane and `position_and_size` are iterated over,
I actually thought it wasn't possible this way since the plugins 
are handled differently, but it turns out that the plugins take their
respective `position_and_size` out of the list, so it is fine again.

So the `runnable` vector is filled in exactly the same order the
`position_and_size` is and as long as they remain the same,
the layout should be correct.

As a side note, I think previously it might have spawned too many
`terminals` and needed to close some again, I think that might
not be necessary anymore but I am not too sure.

>Otherwise, can you maybe include an example with a layout that has a run command in it? Just to make it easier for people to understand how to do this.

Sure, I have one in the commit message, should I just take that one and but it under `examples/run-htop-layout.yaml` ?

imsnif:
Ahhh, that makes sense about the iteration order! That's actually quite an elegant solution. I wonder if it happened by accident ;P

And yeah, placing the one in the commit message in `examples` sounds good.

PR #608, opened 2021-07-08
Display session name within `zellij`

### LINKED ISSUE #573: [Request] Display session name in Zellij
I usually have at least two sessions open for Zellij; 1 for work (Work), and 1 for personal things (Personal). Sometimes I open other sessions for specific projects.

I'd like to be able to view the name of the session somewhere in Zellij, instead of having to do `env | grep -i zellij`.
Perhaps in the top-left, which currently reads "Zellij", or the top right.
Or perhaps somewhere in the bottom bar, just like in Tmux.

I know I can have a different theme for each session, and I'm looking into that, but I like to be able to read the name as well.

Would this be possible?

LovecraftianHorror:
Closes #573 

This displays the session name next to `Zellij` in the top-left corner by passing the information in `ModeInfo` as described in #573. Example:

image

The session name is stored as an `Option<String>` so if, for whatever reason, `None` gets passed to the plugins then the name will fall back to the old style of just displaying `Zellij`. I did this because the `State` for the plugin has to implement `Default`, but I haven't been able to see a situation where it receives `None`.

This currently gets the session name by reading the `ZELLIJ_SESSION_NAME` env var, but I would prefer to have the name passed in as a `String` instead and am planning on making this change before switching this off a draft.

I did also move `get_mode_info()` into `zellij_tile` as `ModeInfo::new()` since it seemed like the function was just serving as a constructor for `ModeInfo`, but let me know if there was any compelling reason to keep it as a separate function in a different crate.

Also just to note from a new contributor I had to:

- Remove `-C target-cpu=native` from my `rustflags` to get the wasm portions to build without erroring
- Install the Linux musl target from `rustup` along with the musl C standard library from my package manager
    - `rustup target add x86_64-unknown-linux-musl`
    - `sudo pacman -S musl` for Arch at least

### DIFF
--- a/default-plugins/tab-bar/src/line.rs
+++ b/default-plugins/tab-bar/src/line.rs
@@ -144,8 +144,12 @@ fn add_next_tabs_msg(
     title_bar.push(right_more_message);
 }
 
-fn tab_line_prefix(palette: Palette) -> LinePart {
-    let prefix_text = " Zellij ".to_string();
+fn tab_line_prefix(session_name: Option<&str>, palette: Palette) -> LinePart {
+    let mut prefix_text = " Zellij ".to_string();
+    if let Some(name) = session_name {
+        prefix_text.push_str(&format!("({}) ", name));
+    }
+
     let prefix_text_len = prefix_text.chars().count();
     let prefix_styled_text = style!(palette.white, palette.cyan)
         .bold()
@@ -165,6 +169,7 @@ pub fn tab_separator(capabilities: PluginCapabilities) -> &'static str {
 }
 
 pub fn tab_line(
+    session_name: Option<&str>,
     mut all_tabs: Vec<LinePart>,
     active_tab_index: usize,
     cols: usize,
@@ -181,7 +186,7 @@ pub fn tab_line(
     };
     tabs_to_render.push(active_tab);
 
-    let prefix = tab_line_prefix(palette);
+    let prefix = tab_line_prefix(session_name, palette);
     populate_tabs_in_tab_line(
         &mut tabs_before_active,
         &mut tabs_after_active,
--- a/default-plugins/tab-bar/src/line.rs
+++ b/default-plugins/tab-bar/src/line.rs
@@ -145,10 +145,8 @@ fn add_next_tabs_msg(
 }
 
 fn tab_line_prefix(session_name: Option<&str>, palette: Palette) -> LinePart {
-    let mut prefix_text = " Zellij ".to_string();
-    if let Some(name) = session_name {
-        prefix_text.push_str(&format!("({}) ", name));
-    }
+    let session_name = session_name.unwrap_or("Zellij");
+    let prefix_text = format!(" {} ", session_name);
 
     let prefix_text_len = prefix_text.chars().count();
     let prefix_styled_text = style!(palette.white, palette.cyan)
--- a/default-plugins/tab-bar/src/line.rs
+++ b/default-plugins/tab-bar/src/line.rs
@@ -145,8 +145,10 @@ fn add_next_tabs_msg(
 }
 
 fn tab_line_prefix(session_name: Option<&str>, palette: Palette) -> LinePart {
-    let session_name = session_name.unwrap_or("Zellij");
-    let prefix_text = format!(" {} ", session_name);
+    let mut prefix_text = " Zellij ".to_string();
+    if let Some(name) = session_name {
+        prefix_text.push_str(&format!("({}) ", name));
+    }
 
     let prefix_text_len = prefix_text.chars().count();
     let prefix_styled_text = style!(palette.white, palette.cyan)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,8 +5,28 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 
 ## [Unreleased]
+* Plugins don't crash zellij anymore on receiving mouse events (https://github.com/zellij-org/zellij/pull/620)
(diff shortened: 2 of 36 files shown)


a-kenji:
Thank you for this pr!

> I did also move `get_mode_info()` into `zellij_tile` as `ModeInfo::new()` since it seemed like the function was just serving as a constructor for `ModeInfo`, but let me know if there was any compelling reason to keep it as a separate function in a different crate.

I like it much more after your change. I think we thought already about moving it.
Maybe someone else has more input on that.

> Also just to note from a new contributor I had to:
> 
>     * Remove `-C target-cpu=native` from my `rustflags` to get the wasm portions to build without erroring

How did you set up your toolchain? Does the `rust-toolchain` file inherit
global rustflags? Maybe we need to investigate more what happened here.

>     * Install the Linux musl target from `rustup` along with the musl C standard library from my package manager
>       
>       * `rustup target add x86_64-unknown-linux-musl`
>       * `sudo pacman -S musl` for Arch at least

The `musl` target is only used for the e2e tests right now,
maybe we need to look a little bit more on splitting the dependencies up.


And a little bit general feedback:

For me personally the string is a bit long.
I would like to see the session name without the zellij prefix.
But maybe that is too confusing for especially new users,
maybe it would make sense to just show the session name by default
when it is set manually rather than generated?

This is all still a little bit complicated since we haven't fleshed out
a system for configuring the plugins yet.

TheLostLambda:
This is looking quite cool in general! Just a note about this bit:

> I did also move get_mode_info() into zellij_tile as ModeInfo::new() since it seemed like the function was just serving as a constructor for ModeInfo, but let me know if there was any compelling reason to keep it as a separate function in a different crate

The compelling reason here is that *everything* in `zellij_tile` gets compiled into *every* plugin. In my opinion, there is really no case in which `ModeInfo::new()` would be needed by a plugin. It seems (to me) to be exclusively used by Zellij itself. Even if a plugin did have a reason to call this constructor, it's likely that it would fail to capture the correct session-name because plugins can have different environment variables from the main Zellij process.

I'd personally go back to the `get_mode_info()` approach so that it's impossible for a plugin developer to get confused and attempt to call this function. Let me know what you think though!

LovecraftianHorror:
@a-kenji 

> How did you set up your toolchain? Does the rust-toolchain file inherit
> global rustflags? Maybe we need to investigate more what happened here.

Checking over things again, it came from setting `lld` as my linker in my cargo config

```toml
[build]
rustflags = "-C link-arg=-fuse-ld=lld
```

Building the plugins just had a lot of text about ignoring settings caused by having `target-cpu=native` set which is what threw me off, but the builderror was actually from trying to use `lld` as my linker

> For me personally the string is a bit long.
> I would like to see the session name without the zellij prefix.
> But maybe that is too confusing for especially new users,
> maybe it would make sense to just show the session name by default
> when it is set manually rather than generated?

I went ahead and changed it to just display the session name (falling back to `Zellij` if it's somehow not set). It would take a bit more work to only display it when the session name isn't generated since it requires somehow passing and possibly storing that information.

@TheLostLambda 

> The compelling reason here is that everything in `zellij_tile` gets compiled into every plugin. In my opinion, there is really no case in which `ModeInfo::new()` would be needed by a plugin. It seems (to me) to be exclusively used by Zellij itself. Even if a plugin did have a reason to call this constructor, it's likely that it would fail to capture the correct session-name because plugins can have different environment variables from the main Zellij process.

I think that's a very good reasoning. I went ahead and moved it back into `zellij-utils/input/mod.rs` as `get_mode_info()`

a-kenji:
@TheLostLambda thank you for the insight - that makes a lot of sense.

> I went ahead and changed it to just display the session name (falling back to Zellij if it's somehow not set). It would take a bit more work to only display it when the session name isn't generated since it requires somehow passing and possibly storing that information.

Thank you for changing that, but I think if it is not configurable yet, then the
prior state that you already had was a better default.

default-plugins/tab-bar/src/line.rs
```diff
@@ -144,8 +144,10 @@ fn add_next_tabs_msg(
     title_bar.push(right_more_message);
 }
 
-fn tab_line_prefix(palette: Palette) -> LinePart {
-    let prefix_text = " Zellij ".to_string();
+fn tab_line_prefix(session_name: Option<&str>, palette: Palette) -> LinePart {
```
a-kenji, line 147: Please do change this part back the way you had it in the prior commit, 
we now have a more clear idea of how to potentially
configure plugins, but it probably still will take some more time.

Until then we would like to show as much information as possible.

a-kenji (changes requested):
Thank you for this awesome pr! With this little change I think it is ready to be merged.

default-plugins/tab-bar/src/line.rs
LovecraftianHorror: I've gone ahead and reverted that commit, so it's back to the old behavior

LovecraftianHorror:
Sorry life got busy (moving and all), so I left this PR out to dry on accident. I went ahead and reverted the commit so it's back to displaying `Zellij (<session-name>)` now. Let me know if there are any other changes that you all want.

I think I'll leave the current method for getting the session name as well instead of trying to weave that information through everything (although if you all want that changed then feel free to do so! I'm sure it would be much easier for people that are more familiar with the codebase)

a-kenji (approved):
No problem, life has priority! I wasn't sure if I was clear enough that's why I tried to explain it in a different way.
Thank you for this awesome work.
I think this looks good for now.

a-kenji (approved):
Thank you very much! I went ahead and merged it now.

techhazard:
@LovecraftianHorror Thank you so much!

PR #622, opened 2021-07-21
feat(tab): add keybind to go to last tab visited

### LINKED ISSUE #398: Feature: Keybind for Returning to Previous Tab:
As part of #376 I'll be creating issues for missing tmux features for discussion for inclusion in Zellij, or overall improvement if Zellij's equivalent doesn't have as much functionality as Tmux.

This one is fairly simple.

A keybind in which will select the most recent tab you were on before the current one.

Basically Zellij could keep a history of what tabs you were on, and return you to the previous one.

It could be in Normal Mode, or Tab Mode, though I'd imagine it most useful in Normal Mode.

sagittarius-a:
Fixes #398.

Tab key is used as default for the `GoToLastTab` action.

A test have been added, All other tests still pass.

### DIFF
--- a/zellij-client/src/input_handler.rs
+++ b/zellij-client/src/input_handler.rs
@@ -188,7 +188,7 @@ impl InputHandler {
             | Action::GoToPreviousTab
             | Action::CloseTab
             | Action::GoToTab(_)
-            | Action::GoToLastTab
+            | Action::ToggleTab
             | Action::MoveFocusOrTab(_) => {
                 self.command_is_executing.blocking_input_thread();
                 self.os_input
--- a/zellij-client/src/input_handler.rs
+++ b/zellij-client/src/input_handler.rs
@@ -188,6 +188,7 @@ impl InputHandler {
             | Action::GoToPreviousTab
             | Action::CloseTab
             | Action::GoToTab(_)
+            | Action::GoToLastTab
             | Action::MoveFocusOrTab(_) => {
                 self.command_is_executing.blocking_input_thread();
                 self.os_input
--- a/zellij-server/src/screen.rs
+++ b/zellij-server/src/screen.rs
@@ -342,9 +342,7 @@ impl Screen {
     /// Consumes the last entry in tab history.
     pub fn get_previous_tab(&mut self) -> Option<&Tab> {
         let last = self.tab_history.pop();
-        if last.is_none() {
-            return None;
-        }
+        last?;
         match last.unwrap() {
             Some(tab) => self.tabs.get(&tab),
             None => None,
--- a/zellij-server/src/screen.rs
+++ b/zellij-server/src/screen.rs
@@ -68,6 +68,7 @@ pub(crate) enum ScreenInstruction {
     ToggleActiveSyncTab,
     CloseTab,
     GoToTab(u32),
+    GoToLastTab,
     UpdateTabName(Vec<u8>),
     TerminalResize(PositionAndSize),
     ChangeMode(ModeInfo),
@@ -133,6 +134,7 @@ impl From<&ScreenInstruction> for ScreenContext {
             ScreenInstruction::MouseRelease(_) => ScreenContext::MouseRelease,
             ScreenInstruction::MouseHold(_) => ScreenContext::MouseHold,
             ScreenInstruction::Copy => ScreenContext::Copy,
+            ScreenInstruction::GoToLastTab => ScreenContext::GoToLastTab,
         }
     }
 }
@@ -150,6 +152,7 @@ pub(crate) struct Screen {
     position_and_size: PositionAndSize,
     /// The index of this [`Screen`]'s active [`Tab`].
     active_tab_index: Option<usize>,
+    last_active_tab_index: Option<usize>,
     mode_info: ModeInfo,
     colors: Palette,
     session_state: Arc<RwLock<SessionState>>,
@@ -170,6 +173,7 @@ impl Screen {
             position_and_size: client_attributes.position_and_size,
             colors: client_attributes.palette,
             active_tab_index: None,
+            last_active_tab_index: None,
             tabs: BTreeMap::new(),
             mode_info,
             session_state,
@@ -194,6 +198,7 @@ impl Screen {
             self.colors,
             self.session_state.clone(),
         );
+        self.last_active_tab_index = self.active_tab_index;
         self.active_tab_index = Some(tab_index);
         self.tabs.insert(tab_index, tab);
         self.update_tabs();
@@ -219,6 +224,7 @@ impl Screen {
         for tab in self.tabs.values_mut() {
             if tab.position == new_tab_pos {
                 tab.set_force_render();
+                self.last_active_tab_index = self.active_tab_index;
                 self.active_tab_index = Some(tab.index);
                 break;
             }
@@ -238,6 +244,7 @@ impl Screen {
         for tab in self.tabs.values_mut() {
             if tab.position == new_tab_pos {
                 tab.set_force_render();
+                self.last_active_tab_index = self.active_tab_index;
                 self.active_tab_index = Some(tab.index);
                 break;
             }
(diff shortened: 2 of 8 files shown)


zellij-client/src/input_handler.rs
```diff
@@ -188,6 +188,7 @@ impl InputHandler {
             | Action::GoToPreviousTab
             | Action::CloseTab
             | Action::GoToTab(_)
+            | Action::GoToLastTab
```
a-kenji, line 191: I believe `ToggleTab` might be clearer.

zellij-server/src/screen.rs
```diff
@@ -67,6 +67,7 @@ pub(crate) enum ScreenInstruction {
     ToggleActiveSyncTab,
     CloseTab,
     GoToTab(u32),
+    GoToLastTab,
```
a-kenji, line 70: Same as above.

zellij-utils/src/input/mod.rs
```diff
@@ -41,6 +41,7 @@ pub fn get_mode_info(
             keybinds.push(("x".to_string(), "Close".to_string()));
             keybinds.push(("r".to_string(), "Rename".to_string()));
             keybinds.push(("s".to_string(), "Sync".to_string()));
+            keybinds.push(("Tab".to_string(), "Last".to_string()));
```
a-kenji, line 44: Same as above, I think `Toggle` might carry the intention better than `Last`.

zellij-server/src/screen.rs
```diff
@@ -148,6 +150,7 @@ pub(crate) struct Screen {
     position_and_size: PositionAndSize,
     /// The index of this [`Screen`]'s active [`Tab`].
     active_tab_index: Option<usize>,
+    last_active_tab_index: Option<usize>,
```
a-kenji, line 153: I think `previous_active_tab_index` would be clearer here, last could be a little ambiguous,
what do you think?

a-kenji (changes requested):
Thank you! This is already looking great.

I personally would prefer if it also kept track of the last tab
it visited while going to the last tab, so that you could easily
toggle between the two last used tabs. Do you think that might
be in scope of this pr?

If that is the case I do have these small considerations about naming.

sagittarius-a:
Regarding the naming I'll apply your suggestions.

Regarding your comment on the "two last used tabs", I'm not sure to have properly identified your request. But to be more explicit about the current behavior:

- visit tab number 1. Let's call it dev.
- visit tab number 2. Let's call it make.
- use `ToggleTab` action will go to `dev`
- use `ToggleTab` again will bring you back to `make`

With this behavior I aimed an easy to switch between 2 tabs, hopefully staying in `Tab mode` would require a single `Tab` key press to switch between them.

Is this what you described ? If not would you mind giving an illustrated example ? :)

I'll implement it later this day if everything goes as planned.

Thank you for your time

sagittarius-a:
Regarding the naming I'll apply your suggestions.

Regarding your comment on the "two last used tabs", I'm not sure to have properly identified your request. But to be more explicit about the current behavior:

- visit tab number 1. Let's call it dev.
- visit tab number 2. Let's call it make.
- use `ToggleTab` action will go to `dev`
- use `ToggleTab` again will bring you back to `make`

With this behavior I aimed an easy to switch between 2 tabs, hopefully staying in `Tab mode` would require a single `Tab` key press to switch between them.

Is this what you described ? If not would you mind giving an illustrated example ? :)

I'll implement it later this day if everything goes as planned.

Thank you for your time

sagittarius-a:
Regarding the naming I'll apply your suggestions.

Regarding your comment on the "two last used tabs", I'm not sure to have properly identified your request. But to be more explicit about the current behavior:

- visit tab number 1. Let's call it dev.
- visit tab number 2. Let's call it make.
- use `ToggleTab` action will go to `dev`
- use `ToggleTab` again will bring you back to `make`

With this behavior I aimed an easy to switch between 2 tabs, hopefully staying in `Tab mode` would require a single `Tab` key press to switch between them.

Is this what you described ? If not would you mind giving an illustrated example ? :)

I'll implement it later this day if everything goes as planned.

Thank you for your time

a-kenji:
Oh yeah, thanks that is indeed already largely how I thought it would work, i guess I was in a weird state while testing it.

My example was on creating a New Tab it would also know what the previous tab was.

Currently when opening a tab one would already need to switch once manually before being able to toggle between both of them:

- start with tab 1
- open tab 2
- toggle tab -> doesn't toggle (but I would assume it would toggle already)

Thank you and no rush!

zellij-client/src/input_handler.rs
sagittarius-a: Done.

zellij-server/src/screen.rs
sagittarius-a: Done.

zellij-utils/src/input/mod.rs
sagittarius-a: Done.

zellij-server/src/screen.rs
sagittarius-a: Done.

sagittarius-a:
I updated the PR according to your comments.

I spent some time struggling with the feature when deleting tabs between other tabs. During this process I found something quite inconsistent:

`screen.go_to_tab` function is declared as follows:
```
    pub fn go_to_tab(&mut self, mut tab_index: usize)
```

The issue can be illustrated with the following scenario:

1. Create 3 tabs (#1, #2, #3 respectively)
2. Go to tab #2
3. Delete it
4. Toggle previous tab (initially using it's index)

After step 2, here is the status of tabs:

```
Tab: index: 0 -- position: 0 -- name: Tab #1 -- active: false -- previous: false
Tab: index: 1 -- position: 1 -- name: Tab #2 -- active: true  -- previous: false
Tab: index: 2 -- position: 2 -- name: Tab #3 -- active: false -- previous: true
```

After step 3, here is the status of tabs:
```
Tab: index: 0 -- position: 0 -- name: Tab #1 -- active: true  -- previous: false
Tab: index: 2 -- position: 1 -- name: Tab #3 -- active: false -- previous: true
```

As shown, for the `Tab #3`, `index` and `position` are now different. But if I
strictly following semantic used by the `screen.go_to_tab` function, I use the
index to toggle to the previous active tab. By doing so, it fails to find the
`Tab #3` since the search is not performed with the `index` field but with
`position` (see last line of snippet below):

```
    pub fn go_to_tab(&mut self, mut tab_index: usize) {
        tab_index -= 1;
        let active_tab_index = self.get_active_tab().unwrap().index;
        if let Some(t) = self.tabs.values_mut().find(|t| t.position == tab_index) {
```

Maybe the semantic for the `screen.go_to_tab` could be updated to use `position`
instead of `index` ?

I'm not sure it fits in the current PR but I wanted to address it here since it
is closely related to the changes performed.

Thanks !

a-kenji (changes requested):
Thank you for the changes, this is looking much better to me.
There is still a situation where closing a Tab will render the toggle useless,
and I think this is currently the best we are going to get while saving a single number.

I think it could be beneficial here to use a `Vec` as a stack and `push` on creation,
`pop` on delete and `pop & push` on toggle, what do you think? Maybe you will find
a more elegant way.

This is the situation that currently fails:
- open tab 1
- open tab 2
- open tab 3
- GoToTab 1
- CloseTab
- ToggleTab

Good find on that `index` inconsistency! If I understand you correctly
I think I agree that changing it is the best course of action for now, I think the index
is there to have a unique number associated with each tab that stays the same regardless
of reordering. But using that number for a `go_to_tab` action might not be the best choice.

Please do so In a separate commit and put the awesome diagram in your message,
if you put it here, or in a separate pr is up to you.

sagittarius-a:
I finally took the time to work on this !

I will address a new PR for the variable naming issue, I find it more relevant.

Let me know if you find some improvements that can be made :)

a-kenji:
@sagittarius-a,
Thank you for this awesome work! This looks great to me.

PR #626, opened 2021-07-24
feat(scroll): Restore scroll position on SIGINT

### LINKED ISSUE #606: Restore scroll position on SIGINT
Currently when scrolling the text, the only way to restore
the scroll position is by scrolling down, or using the `[PgDn]` key.

I believe it would make sense to restore the position when receiving
a SIGINT, maybe even go out of the scrolling mode as well.

sagittarius-a:
Fixes #606

Currently scroll down to bottom then exits scroll mode.


a-kenji (approved):
This is looking good, thank you!

PR #641, opened 2021-08-10
support monochrome mode

### LINKED ISSUE #614: Add option to start zellij in monochrome mode
Strip out all color from zellij and programs that are running in zellij.

Similar to `htop -C`

tw4452852:
When in monochrome mode, all color related escape sequences are stripped.

Fixes #614

Signed-off-by: Tw <wei.tan@intel.com>

### DIFF
--- a/default-plugins/status-bar/src/main.rs
+++ b/default-plugins/status-bar/src/main.rs
@@ -70,64 +70,92 @@ pub struct ColoredElements {
 // we need different colors from palette for the default theme
 // plus here we can add new sources in the future, like Theme
 // that can be defined in the config perhaps
-fn color_elements(palette: Palette) -> ColoredElements {
-    match palette.source {
-        // "cyan" here is used as a background as a dirty hack
-        // this is because the Palette struct doesn't have a "gray" section
-        // and we can't use its "bg" because that is now dynamically taken from the terminal
-        // and might often not actually fit the rest of the colorscheme
-        //
-        // to fix this, we need to restructure the Palette struct
-        PaletteSource::Default => ColoredElements {
-            selected_prefix_separator: style!(palette.cyan, palette.green),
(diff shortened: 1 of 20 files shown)


imsnif:
Hey @tw4452852 - this looks really cool! I only took a brief look so far, but it seems to me that this strips the colors only from the interface and not from the programs running inside it, right?
I'm not sure what the use case for the latter is, so I don't know if it's important. Maybe @a-kenji can shed more light on it?

tw4452852:
Hi @imsnif 

Yes, it only with the default plugins currently. I need further investigation to find out how to strip out color escape sequence from the server side.

tw4452852:
sure:)

imsnif:
@tw4452852, apologies for the delay:

I think ignoring the styles here would do the trick for terminals: https://github.com/zellij-org/zellij/blob/main/zellij-server/src/panes/terminal_pane.rs#L239

tw4452852:
Hi @imsnif

Thank you very much, it saved a lot of time for me.

After rethinking the monochrome mode, I think it makes more sense to let the palette be optional instead of introducing a `Mono` in the `Color` enum.  In this way, there will be no palette if we're in the monochrome mode.

BTW, a dedicated unit test is also added for the monochrome mode.

imsnif:
Hey @tw4452852 - my apologies for taking so long to review this!

This looks good to me, but I think my main issue is placing the `Palette` behind an `Option`. It seems like it creates a lot of extra code. Do you think we can:
1. for the plugins, create a monochrome palette and work with it?
2. for the TerminalPane, pass the Palette as normal but have a `strip_colors: bool` attribute on it that we'd control with the builder pattern? Something like `let terminal_pane = TerminalPane::new(...).strip_colors()`?

Or do you think it'll be too complex, or cause other issues that I'm not thinking of?

imsnif:
Closing as stale

Issue #640, opened 2021-08-09
Feature: indicate to the user when text is copied to the clipboard


imsnif:
Right now, whenever you mark text with the mouse, Zellij copies it to the system clipboard immediately. This behaviour is great, but it's not very clear to the user that this is happening.

We should:
1. Change the "tip" line in the status bar to "Text copied to clipboard." whenever this happens
2. Any input from the user after the above message appears should reset the message to the usual "tip" line.

tlinford:
Have a wip version of this, will hopefully finish it up soon.

PR #642, opened 2021-08-12
Indicate to the user when text is copied to the clipboard

tlinford:
Resolves zellij-org/zellij#640

- broadcast CopyToClipboard event to plugins after selection has been
  copied, and InputReceived event after any input has been received.
- add new ClientToServerMsg InputReceived
- subscribe status-bar plugin to new events, modify second line after
  CopyToClipboard, reset it on InputReceived.


imsnif (changes requested):
Hey, looking good! Do you think we can tie this to existing events rather than sending a new one though?
Also, can we maybe make the text stand out a little more? Maybe be bold and orange/green?

tlinford:
> Hey, looking good! Do you think we can tie this to existing events rather than sending a new one though?

Not sure to be honest, out of the existing `Event` variants only `Key` seems to be close but imo doesn't really apply, specially for the event that signals that text was copied.
Do you think something along these lines this would be cleaner?
```
pub enum Event {
    ...,
    Action,
}

pub enum Action {
    InputReceived,
    SelectionCopiedToClipboard,
}
```

Action is maybe a little too generic/vague?

> Also, can we maybe make the text stand out a little more? Maybe be bold and orange/green?
No problem.

imsnif:
> Not sure to be honest, out of the existing `Event` variants only `Key` seems to be close but imo doesn't really apply, specially for the event that signals that text was copied.

What I was thinking is that instead of using `InputReceived` we could add the relevant line here: https://github.com/zellij-org/zellij/blob/main/zellij-server/src/route.rs#L26

Like - any action we get is essentially an "InputReceived", right? So before the match we could clear the "copied" line from the status bar. What do you think?

imsnif:
P.S. - sorry for delaying so long on replying :)

tlinford:
Hi @imsnif, I've removed the InputReceived ClientToServerMsg, using existing Actions to know when to send the Event to the plugins.
Also made the hint bold green, much better.

imsnif:
Looks great! Thanks for this cool feature :)

PR #655, opened 2021-08-21
Fix(scroll): Prevent scroll buffer overwrite

### LINKED ISSUE #306: scrolling when text updates
running mpv in audio-only mode led me to discover that if I scroll, the text update happens in the last >visual< line, not the actual last line as one expects it to happen

I imagine this would happen with any application that updates the terminal while one scrolls

so I tested it with a bash for loop printing and I discovered a different? bug, it doesn't allow me to scroll while it prints

kunalmohan:
Fix #306 
Currently, if a command is producing output (eg `tail`), then scrolling up causes the command to overwrite the scroll buffer.
This PR fixes it. Now when a pane is scrolled up, new vte events are buffered and when we exit scroll, the events are played at once.

### DIFF
--- a/zellij-server/src/panes/grid.rs
+++ b/zellij-server/src/panes/grid.rs
@@ -361,6 +361,7 @@ pub struct Grid {
     pub pending_messages_to_pty: Vec<Vec<u8>>,
     pub selection: Selection,
     pub title: Option<String>,
+    pub is_scrolled: bool,
 }
 
 impl Debug for Grid {
@@ -405,6 +406,7 @@ impl Grid {
             title_stack: vec![],
             title: None,
             changed_colors: None,
+            is_scrolled: false,
         }
     }
     pub fn render_full_viewport(&mut self) {
@@ -537,6 +539,7 @@ impl Grid {
     }
     pub fn scroll_up_one_line(&mut self) {
         if !self.lines_above.is_empty() && self.viewport.len() == self.height {
+            self.is_scrolled = true;
             let line_to_push_down = self.viewport.pop().unwrap();
             self.lines_below.insert(0, line_to_push_down);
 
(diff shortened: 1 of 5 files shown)


imsnif (changes requested):
Very cool - happy to see you getting into these parts of Zellij as well. :)

A few things:
1. I'm a little afraid of the buffer being uncapped. If I leave a pane scrolled up with input running overnight I might build up quite the memory utilization by morning... do you think we can cap it somehow and then if it exceeds that cap we reset the scroll? not the best ux, but I think it's better than now, no? Or do you maybe have a different idea?
2. When we play the pending vte events, at least on my machine, there's quite a long period where things are stuck. I assume it's when the events are interpreted. Do you think we might be able to add some renders in the middle of this somehow? So the user sees that something is happening?

kunalmohan:
@imsnif I've incorporated your suggestions and updates the PR. For now, I've kept the buffer size to 7000 events per pane and the tab will render after processing every 100 pending events. I can update the numbers as per your (or anyone else's) suggestions.

Also, this works at all times- inside scroll mode or outside (when scrolling with the mouse).

imsnif:
@kunalmohan - I think this isn't working for me, but maybe I'm doing something wrong.

What I did:
1. Started an infinite loop that prints the date every 0.1 seconds (while true; echo (date); sleep 0.1; end)
2. Scrolled up (ctrl+s and k a few times)
3. Pressed ENTER to leave scroll mode
4. I'm still scrolled up and not at the bottom of the pane
5. When I scroll down it updates but I think there's a gap there? I'm not sure about this one

kunalmohan:
I've addressed the review comments and made the required updates. Things work for me as expected, therefore, merging this now.
Any further changes can be made later based on user experience and suggestions.

PR #666, opened 2021-08-27
feat(plugin): add exec_cmd helper for executing command in host

tw4452852:
Signed-off-by: Tw <wei.tan@intel.com>


tw4452852:
Hi @imsnif ,

I've added a flag to allow plugin run host command (forbidden by default). Can you give it another look?

zellij-server/src/wasm_vm.rs
```diff
@@ -248,6 +256,20 @@ fn host_set_timeout(plugin_env: &PluginEnv, secs: f64) {
     });
 }
 
+fn host_exec_cmd(plugin_env: &PluginEnv) {
+    let mut cmdline: Vec<String> = wasi_read_object(&plugin_env.wasi_env);
+    let command = cmdline.remove(0);
+
+    process::Command::new(command)
+        .args(cmdline)
+        .spawn()
```
tw4452852, line 265: Just leave a note here, to avoid block plugin thread, we don't wait for the command to exit.

imsnif:
Hey @tw4452852 - sorry for the misunderstanding. I didn't mean to add a flag that would allow all plugins to execute commands. That might be quite dangerous I find. I was thinking right now to do something like add this to the layout where we load the plugin. Maybe under "run"?

imsnif:
@tw4452852 - sounds great!

tw4452852:
@imsnif 
Already done.

zellij-server/src/wasm_vm.rs
```diff
@@ -54,6 +54,7 @@ pub(crate) struct PluginEnv {
     pub senders: ThreadSenders,
     pub wasi_env: WasiEnv,
     pub subscriptions: Arc<Mutex<HashSet<EventType>>>,
+    pub _allow_exec_host_cmd: bool,
```
TheLostLambda, line 57: The leading `_` is good here! If you want to be even more explicit about this being temporary, if might be nice to leave a comment here :) Maybe starting like: `// FIXME: ...` so it's easy to find later :)

zellij-utils/src/input/unit/layout_test.rs
```diff
@@ -45,7 +45,10 @@ fn default_layout_merged_correctly() {
                 borderless: true,
                 parts: vec![],
                 split_size: Some(SplitSize::Fixed(1)),
-                run: Some(Run::Plugin(Some("tab-bar".into()))),
+                run: Some(Run::Plugin(Some(RunPlugin {
+                    path: "tab-bar".into(),
+                    _allow_exec_host_cmd: false,
```
TheLostLambda, line 50: If you want, you could `#[derive(Default)]` on `RunPlugin` and do:

```rs
RunPlugin {
    path: "tab-bar".into(),
    ..Default::default()
}
```
I get the feeling that the `RunPlugin` struct will be growing and this should prevent us from needing to update this file every time that happens.

zellij-server/src/wasm_vm.rs
```diff
@@ -248,6 +256,20 @@ fn host_set_timeout(plugin_env: &PluginEnv, secs: f64) {
     });
 }
 
+fn host_exec_cmd(plugin_env: &PluginEnv) {
```
TheLostLambda, line 259: This function, here on the host-side, needs to check `plugin_env._allow_exec_host_cmd` before running anything!

TheLostLambda: If permission is denied, I'd just use `log::warn()` to tell the user about it and return immediately. We probably don't need to crash and logging a warning + ignoring the request should do for now.

zellij-server/src/wasm_vm.rs
```diff
@@ -248,6 +256,20 @@ fn host_set_timeout(plugin_env: &PluginEnv, secs: f64) {
...
+    let mut cmdline: Vec<String> = wasi_read_object(&plugin_env.wasi_env);
+    let command = cmdline.remove(0);
+
+    process::Command::new(command)
+        .args(cmdline)
+        .spawn()
+        .unwrap();
+}
+
+fn host_get_exec_permission(plugin_env: &PluginEnv) {
```
TheLostLambda, line 269: Once we move the permission check to the host-side, we can delete this function :)

zellij-tile/src/shim.rs
```diff
@@ -37,6 +45,14 @@ pub fn open_file(path: &Path) {
 pub fn set_timeout(secs: f64) {
     unsafe { host_set_timeout(secs) };
 }
+pub fn exec_cmd(cmd: &[String]) -> Result<(), PermissionDenied> {
```
TheLostLambda, line 48: I don't think we need the `Result` here and I'd be fine with things failing silently (we don't currently have a way to see if the command was run successfully, even if we did have permissions).

I'd personally just return nothing here, but if you disagree, just let me know!

zellij-tile/src/shim.rs
```diff
@@ -3,6 +3,9 @@ use std::{io, path::Path};
 
 use crate::data::*;
 
+#[derive(Debug)]
+pub struct PermissionDenied;
```
TheLostLambda, line 7: Probably don't need this if we change `exec_cmd` to return nothing.

zellij-tile/src/shim.rs
```diff
@@ -37,6 +45,14 @@ pub fn open_file(path: &Path) {
 pub fn set_timeout(secs: f64) {
     unsafe { host_set_timeout(secs) };
 }
+pub fn exec_cmd(cmd: &[String]) -> Result<(), PermissionDenied> {
+    if !get_exec_permission() {
```
TheLostLambda, line 49: No need to have this check here once the check has been moved inside of `host_exec_cmd()` in `wasm_vm.rs`!

zellij-utils/src/input/layout.rs
```diff
@@ -53,11 +53,19 @@ pub enum SplitSize {
 #[serde(crate = "self::serde")]
 pub enum Run {
     #[serde(rename = "plugin")]
-    Plugin(Option<PathBuf>),
+    Plugin(Option<RunPlugin>),
     #[serde(rename = "command")]
     Command(RunCommand),
 }
 
+#[derive(Debug, Serialize, Deserialize, Clone, PartialEq, Eq)]
```
TheLostLambda, line 61: I'd toss in `Default` here :)

TheLostLambda (changes requested):
Hey @tw4452852 ! Thank you for all of your excellent work here! I left a couple of small comments, but there is really just one thing we need to address before finally merging this!

Right now there is actually a pretty major security hole: everything in `shim.rs` is just convenience code for the plugins and isn't actually needed to write a plugin (it wouldn't be used at all in non-Rust plugins, for example). That means that any plugin can just directly call the `host_exec_cmd` without checking permissions.

I'll leave some comments with more details, but we essentially need to move the permissions check to the host side of things.

zellij-utils/src/input/unit/layout_test.rs
tw4452852: Really good advice, thanks.

zellij-tile/src/shim.rs
tw4452852: I'd personally prefer to return something if we wait for the command exit, users may know if it's successful or not.
But right now, we don't wait for the command to finish, so I agree with your opinion for simplicity.

tw4452852:
Hi @TheLostLambda , Thanks a lot for your review, they're really good advice, I adopt them all.

zellij-tile/src/shim.rs
TheLostLambda: @tw4452852 I totally agree :) I'd like to monitor the running of the command eventually, but I'm not sure how we'll do that just yet, so maybe it's best for a second PR :)

zellij-server/src/wasm_vm.rs
```diff
@@ -248,6 +256,23 @@ fn host_set_timeout(plugin_env: &PluginEnv, secs: f64) {
     });
 }
 
+fn host_exec_cmd(plugin_env: &PluginEnv) {
+    let mut cmdline: Vec<String> = wasi_read_object(&plugin_env.wasi_env);
+    let command = cmdline.remove(0);
+
+    // Bail out if we're forbidden to run command
+    if plugin_env._allow_exec_host_cmd {
```
TheLostLambda, line 264: Almost perfect, but this line needs to be `if !plugin_env._allow_exec_host_cmd {` with the `!`. Right now everything that shouldn't have permissions does and vice versa

TheLostLambda (changes requested):

zellij-tile/src/shim.rs
```diff
@@ -37,6 +37,10 @@ pub fn open_file(path: &Path) {
 pub fn set_timeout(secs: f64) {
     unsafe { host_set_timeout(secs) };
 }
+pub fn exec_cmd(cmd: &[String]) {
```
TheLostLambda, line 40: Additionally, I might change this to take either `&[&str]` or maybe `&[impl AsRef<str> + Serialize]` if you really want to pass strings in directly.

Both solutions allow very simple calls like: 
```rs
exec_cmd(&["playerctl", "pause"]);
```

And the more complex type allows this as well, which you can decide whether you'll need or not:
```rs
let cmds: Vec<String> = vec!["playerctl".into(), "play".into()];
exec_cmd(&cmds);
```

zellij-server/src/wasm_vm.rs
tw4452852: Ah, sorry, my bad...

zellij-tile/src/shim.rs
tw4452852: I prefer `&[&str]` because it's more friendly to the plugin as most of the time `&str` is sufficient.

tw4452852:
Hi @TheLostLambda 

Thanks for your review comments. In the latest version, I also add more context to the warning log (the command to run).

zellij-tile/src/shim.rs
tw4452852: I originally wanted to return the exit code which indicates the success or not. But I'm not sure if this is enough or general. Let's hold on to this for this moment now.

TheLostLambda (approved):
This looks great! Thank you for all of your hard work on this and feel free to merge things whenever you're ready!

TheLostLambda:
Additionally, @tw4452852 if you can't actually press the merge button, even after approval, just let me know and I'll merge for you!

tw4452852:
Hi @TheLostLambda 
It seems I don't have the permission, feel free to merge, I'm done.

TheLostLambda:
No worries, all merged!

PR #669, opened 2021-08-29
feat(ui): round frame corners

TheLostLambda:
I think rounded corners look nice! This PR adds them by default:
Screenshot from 2021-08-29 21-19-34

For our font-deficient friends, we fall back to square corners with `--simplified-ui`:
Screenshot from 2021-08-29 21-21-09

Let me know what you think and if things display correctly with your font!


a-kenji:
I agree, this looks very nice, I personally would vote for a separate option and to not make that the default, because of the font issue. 

To tie it to simplified_ui is not a bad idea, but it could be an even worse first experience for people that don't just have extended fonts installed.

imsnif:
Fully agree with @a-kenji

TheLostLambda:
Okay, then I don't think it's really worth a command-line flag – should I just add an option in the config that defaults to false? A rounded-corners flag?

TheLostLambda:
On second thought @a-kenji  and @imsnif , wouldn't anywhere the corners are broken also have broken status and tab bars? We have the fancy status bars by default, is this that much different? Our defaults kinda already require fonts with these nice characters, don't they?

imsnif:
I don't know if every font that has the arrow characters also has the rounded corners. :/ When I made the arrow characters I was sure all the fonts have them. I don't know how to check the validity of this.

TheLostLambda:
> I don't know if every font that has the arrow characters also has the rounded corners. :/ When I made the arrow characters I was sure all the fonts have them. I don't know how to check the validity of this.

That's a valid concern. I'll do some testing of different fonts and Linux distributions if you think that would be helpful? How were you sure that all fonts had arrow characters?

imsnif:
I just didn't think about it to be honest. If you want to take responsibility for this, then go for it. Just be aware that:
1. If some fonts don't have this, we'll likely lose users (this is a very bad first experience)
2. I personally will not know how to check if your tests were conclusive enough.

a-kenji:
@TheLostLambda,
No I think you are right, I personally think this is a little different here though.
While it is a default we ship currently, it is in the default plugins and not the default application,
while that is currently a very arbitrary difference maybe that will not be in the future.
I am not sure if that warrants a different look at things.

But starting zellij without any plugins by default currently does not require certain fonts to be installed.

TheLostLambda:
For this, are people thinking that we should not make it the default and have a separate option for it? I'd personally be comfortable making it the default and would probably prefer that (I never ran into missing fonts accept on the TTY after testing half a dozen distros), but my resolve has softened enough to keep the square corners the default if others are particularly set on that.

imsnif:
For me: as I mentioned offline, I'd feel confident enough to make this the default if I understood more. Why doesn't the TTY font have it, for example? Where's the difference?

Even if it were just the TTY though - the benefit seems so small to me in making this the default (I personally would not be able to tell the difference at a glance, but maybe it's my lack of aesthetic sense :) ), versus the risk of losing even one user to it not working out of the box as they would expect it to, that my vote is no.

That's really just my opinion though.

a-kenji:
@TheLostLambda 
I think adding this functionality as an optional value would be great.
Are you still interested in that?

TheLostLambda:
Hi @a-kenji ! Sorry this has sat around for ages! I'm happy to add it as something optional!

Should we make it a config option that allows the user to specify an arbitrary character to use for the corners or just have a `rounded_pane_corners` boolean toggle?

a-kenji:
@TheLostLambda,

Id say have a boolean toggle for now would be the best option.

a-kenji:
@TheLostLambda,
Just in case this pr is not on your mind anymore a slight ping.

TheLostLambda:
Closing in favour of #1227 !

PR #676, opened 2021-08-30
Fix E2E Tests With Plugins

TheLostLambda:
This is a dirty spin-off of a previous branch. The changes to the `Makefile.toml` and `remote-runner.rs` work on my local machine, but not in Docker CI.


TheLostLambda:
Should have been fixed by #1047 !

PR #686, opened 2021-09-03
fix(tabs): use tab index for tab name

prscoelho:
This PR fixes a small visual bug that causes tab names to have the same name if we close any tab that is not the last and create a new tab.

Previously, if we had tabs 2, 3, and created a new tab, we would end up with tabs 2, 3, 3. Now we get 2, 3, 4.


imsnif:
Good find!! Thank you :)

PR #693, opened 2021-09-04
fix(example) Update NewTab action in example/default.yaml to new syntax

synul:
The default keybind file in the examples folder has a syntax error for the NewTab action. The file is being referenced in the docs so when copy pasting it (like I did) to customize off of it, zellij won't be able to read the config on start up.


a-kenji:
Thank you @synul, that is a good catch!

a-kenji (approved):
Cheers!

Issue #638, opened 2021-08-05
Feature: Scroll mode should be persistent in a scrolled pane


imsnif:
When I:

1. Enter scroll mode and scroll up inside a pane
2. Move the focus to another pane, exit scroll mode and do something else
3. Go back to the scrolled up pane

I should be placed in scroll mode again.

This is so that I don't try to scroll and then lose my position because the pane resets back to its bottom most position.

imsnif:
Hey @v0idx - could you link to the parts of the code you mean? It would be easier to discuss them that way. :)

v0idx:
Hey @imsnif of course I can! The mode_info attribute can be found here: https://github.com/zellij-org/zellij/blob/e54dfab40f73bbd0c7e27b0caf525c7cc5af607b/zellij-server/src/screen.rs#L138-L154
My thought was that as there is already an attribute that stores the current mode for a screen, by moving this value to the pane object, you would be able to keep track of the current mode, incl. scrolling, on a pane by pane basis

imsnif:
Hey @v0idx - thanks for the link. So to make sure I understand what you're suggesting: you'd like to also store a specific `pane_mode` in each pane (be it TerminalPane or PluginPane) and then when we focus that pane, we'll do a quick query to see if the pane has a specific mode and then "override" the current mode with it?

v0idx:
Yeah @imsnif that was the general jist of what I was suggesting, does this sound applicable to your issue?

a-kenji:
From looking over it sporadically, this sounds like a good solution to me.

@imsnif, @v0idx

imsnif:
Yeah, that sounds good. My only suggestion is to have it as something like `mode_override` in the Pane object and keeping the original in the Screen. I feel this is the shorter path.

v0idx:
That sounds good, I'll get started on an implementation tonight on my fork of zellij

arefem-project:
Hey @v0idx, I looked into this one as well and never finished it. It was actually a little trickier then it seemed at first. Hopefully I can save you some time here.

[...]

PR #696, opened 2021-09-06
Add support for scroll mode persistence

spacemaison:
This PR implements persisting the scroll mode for a pane after focusing a different pane and then moving back into it.

This resolves issue #638.

### DIFF
--- a/src/tests/e2e/remote_runner.rs
+++ b/src/tests/e2e/remote_runner.rs
@@ -2,7 +2,10 @@ use zellij_tile::data::Palette;
 
 use zellij_server::panes::TerminalPane;
 use zellij_utils::pane_size::{Dimension, PaneGeom, Size};
-use zellij_utils::{vte, zellij_tile};
+use zellij_utils::{
+    vte,
+    zellij_tile::{self, data::ModeInfo},
+};
 
 use ssh2::Session;
 use std::io::prelude::*;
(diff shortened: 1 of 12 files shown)


imsnif:
Hi @spacemaison - I'm going to continue the discussion here because I feel it will be easier to keep track as we're starting to discuss the implementation.

> I've been driving myself crazy with this for half a day now. Maybe you can compile my branch and futz around with it to see if it feels normal to you?

I totally get the feeling. These UX problems are the kind of problems you really want to get "right", and it's not always easy :)

I checked out the branch and think the behavior is really cool (I haven't looked at the code yet). I do get the caveat you mentioned and agree that it's a little odd and unexpected.

The way I see it, the behavior needs to be like this:
1. If a pane is in scroll mode, it should remain in scroll mode until the user explicitly exits it
2. If a user is in non-normal mode (eg. Pane mode) and moves into a pane with scroll mode, the user should remain in their original mode (eg. Pane mode) - this was your addition and I really liked it.
3. If the user exits scroll mode in another pane, only that other pane should exit scroll mode

1 + 2 are already implemented here. IMO if we implement 3, it will be really predictable UX-wise. What do you think?

a-kenji:
>  
    1. If a pane is in scroll mode, it should remain in scroll mode until the user explicitly exits it
    2. If a user is in non-normal mode (eg. Pane mode) and moves into a pane with scroll mode, the user should remain in their original mode (eg. Pane mode) - this was your addition and I really liked it.
    3. If the user exits scroll mode in another pane, only that other pane should exit scroll mode



I have to agree with @imsnif, great addition on number 2. 
And I think the whole interaction at least seems to be very intuitive to me.

imsnif:
Closing as stale

PR #701, opened 2021-09-07
feat(keybinding): Introduce a pseudo `All` mode to define keybindings for all real input modes

tw4452852:
With this patch you could define some sort of global key bindings, they'll be used no matter which mode you're in currently.

### DIFF
--- a/default-plugins/status-bar/src/first_line.rs
+++ b/default-plugins/status-bar/src/first_line.rs
@@ -246,6 +246,7 @@ pub fn superkey(palette: ColoredElements, separator: &str) -> LinePart {
 pub fn ctrl_keys(help: &ModeInfo, max_len: usize, separator: &str) -> LinePart {
     let colored_elements = color_elements(help.palette);
     match &help.mode {
+        InputMode::All => Default::default(),
         InputMode::Locked => key_indicators(
             max_len,
             &[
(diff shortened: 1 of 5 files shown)


a-kenji:
Hey! Thanks for this awesome pr.
I think this can be valuable, but to make it more applicable I would like to be able to specify
in what modes this `all` mode should be present, what do you think?

For example currently I want most modes to be able to share configuration options
through a shared mode, but I want to exclude the `locked` mode from that configuration,
as well as the `renametab` mode.

Maybe call it a `Shared` mode and configure through the config options with what modes
the configuration is actually shared? And it defaulting to `all`, or `all` except for the locked mode.

tw4452852:
Thanks @a-kenji 

`Shared` is a more proper name in this case.

Instead of introducing a new option to exclude modes which `Shared` applies to, how about specifying them in keybinding scope, like this:

```
keybinds:
  shared:
    ...
  locked:
    ...
    include_share: false
  normal:
    ...
```

If you don't specify `include_share: false` explicitly, keybinds in `shared` will be included by default.
I think this is more applicable personally. What's your opinion?

a-kenji:
I am not sure if putting more attributes in the modes currently is a good idea.
That being said I agree that introducing a new option might not be too good as well.
But I didn't mean to specify which modes would be excluded, rather which ones would be included.

The problem here is:
```
keybinds:
  shared:
    ...
  locked:
    ...
    include_share: false
  normal:
    ...
```

Currently that wouldn't be possible, it would need to be like this:
```
keybinds:
  shared:
    ...
  locked:
    ...
    - include_share: false
  normal:
    ...
```
This seems weird, at least to me and would be not as easy to parse.
The configuration for modes and the keybindings is sadly not as 
robust as I would like it to be, maybe we should open a general issue for that?
The problem I see here is that too much behavior here would be implicit rather than explicit.

tw4452852:
I agree. Let's hold on to this, it needs a double consideration. Or someone else has a better idea?

Issue #699, opened 2021-09-07
Tab bar width not correctly calculated with simplified-ui option


prscoelho:
Tabs are not taking up all possible space with simplified-ui on.

![tab_bar_spacing](https://user-images.githubusercontent.com/9609090/132270341-e7b53103-e097-41bf-8702-7797fc5c8c7a.png)

imsnif:
Fixed in main, thanks for the fix @prscoelho !!

PR #703, opened 2021-09-07
fix(tab-bar): prevent active tab from being hidden

### LINKED ISSUE #668: Inconsistent Tab Bar Behaviour at small window sizes
These sorts of UI bugs seem best explained with a video, but the gist is that, as the terminal width is reduced, the tab bar exhibits unexpected (and arguably buggy) behavior.

https://user-images.githubusercontent.com/6251883/131230830-0828ce2f-bc40-47b9-ab31-0a7a20bd1f64.mp4

prscoelho:
This PR fixes #668 (somewhat) and #699. The active bar won't disappear anymore, but which tabs render is still inconsistent. I think the only way to make it more consistent would be to calculate space required for the collapsed tabs beforehand and remove that amount before calling tabs_to_render.

### DIFF
--- a/default-plugins/tab-bar/src/line.rs
+++ b/default-plugins/tab-bar/src/line.rs
@@ -8,39 +8,82 @@ fn get_current_title_len(current_title: &[LinePart]) -> usize {
     current_title.iter().map(|p| p.len).sum()
 }
 
+// move elements from before_active and after_active into tabs_to_render while they fit in cols
+// adds collapsed_tabs to the left and right if there's left over tabs that don't fit
 fn populate_tabs_in_tab_line(
     tabs_before_active: &mut Vec<LinePart>,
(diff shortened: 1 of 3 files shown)


prscoelho:
Oops. Forgot clippy...

prscoelho:
Sorry about another force push.. I wasn't fully happy with the solution since it was still not consistent and in certain widths not using up all possible space.

The issue was that we can't really populate tabs and then collapse after. I thought about it and we should start from a state of collapsed left, middle, collapsed right and expand tabs while they fit. This current version is completely consistent from what I can tell! Now it should fully close #668 and #699. But it does not attempt to keep active bar in the middle.

prscoelho:
Okay, now the active tab should be kept in the middle.

imsnif:
@prscoelho - this looks great! Thanks for your meticulous work on this :)

PR #708, opened 2021-09-10
Add prompt functionality to zellij

a-kenji:
* add action to interact with the prompt:

 Confirm: Confirms the action
 Deny: Denies the action

* add a default prompt mode

* add action to automatically confirm any potential prompt:
  SkipConfirm: Action

Eg:
zellij-prompt


a-kenji:
@imsnif the `input_handler::grid_tests` fail here, I have played around with it a little,
I thought the reason was because quit doesn't break the input loop here, but the error seems somewhere else.

```
---- input_handler::grid_tests::bracketed_paste stdout ----
thread 'input_handler::grid_tests::bracketed_paste' panicked at 'ran out of stdin events!', zellij-client/src/./unit/input_handler_tests.rs:111:13
```

I have not touched the tests in this pr, as I couldn't make them work yet, but I have tried 
out quite a few iterations already.

imsnif:
Hey @a-kenji - sorry for taking a few days to get to this.

Your intuition is right. The error comes from the tests, specifically here: https://github.com/zellij-org/zellij/blob/main/zellij-client/src/unit/input_handler_tests.rs#L111

Since QUIT now no longer breaks the input loop, the loop continues and the test does not have anymore input to give it. I haven't looked into this PR deeply, but my recommendation would be to try and fix these tests by faking an event from the server somehow that brings the input thread into prompt mode in the very next iteration of the loop after sending the QUIT, and then sending the 'Y' key.

Let me know if you'd like a hand or another pair of eyes on this.

a-kenji:
@imsnif,
Thank you that is helpful, I will try to work if that.
Otherwise I will get in touch again!

dwvisser:
I would **love** to see this feature in a release. This PR seems to be inactive for some months at this point. Is there something blocking it?

Issue #706, opened 2021-09-10
Ability to name tabs from the config file


stratosgear:
After looking around (and consulting in the discord channel) it seems there is no way to name your tabs from the config file (you can only name them manually, from a running instance).

I think there **ought** to be a way to do that from the config file, too!

a-kenji:
Thank you for taking the time writing down this issue.

For anyone that wants to pick this up: 
I think it would only make sense to have them be named in the layout file, as opposed to the config file for now.

stratosgear:
Yes, I meant the layout file (sorry for messing up the terminology)

tw4452852:
I can help on this

a-kenji:
That would be awesome! @tw4452852

a-kenji:
@stratosgear 
No problem, this is very valuable!

prscoelho:
Hey guys, since I recently made a PR for tab-bar, my mind is still fresh on this and I have some questions.

Will control characters be allowed in the tab name from layout file? And can we prevent that easily? Currently an unescaped \n or \t will completely mess up the spacing count and rendering, but I don't know if we care at all, it's deliberate bad input and it doesn't result in a panic, just a bad result from bad input.

And since any kind of characters will be allowed into tab name I guess we should probably change the width count from chars().count() to unicode-width. I can make a PR for that.

a-kenji:
@prscoelho 
Thanks for raising these issues! 
I am all for the tab width pr.
I am not too sure about the escaping one. Maybe open an issue? At some point templating for configured strings make sense maybe. So we can keep that for collecting ideas.

PR #709, opened 2021-09-11
Calculate width with unicode-width in tab-bar and utils

prscoelho:
In preparation for tab names from layout files #706, we need to account for wide characters.

After changing tab-bar, this was the result (notice the line below tab bar).

screenshot-2021-09-11 16:49:13+01:00

The culprit was in zellij-utils, using unicode-width in ansi_len fixed it.


a-kenji (approved):
Awesome, thank you!

PR #715, opened 2021-09-13
feat(screen): support specifying tab's name in layout

### LINKED ISSUE #706: Ability to name tabs from the config file

tw4452852:
fix #706 


stratosgear (commented):
I would assume that the docs will have to somehow make a mention of this new feature.

a-kenji (approved):
This is awesome, thank you!

PR #717, opened 2021-09-13
add visiblity event for plugin

tw4452852:

### DIFF
--- a/zellij-server/src/screen.rs
+++ b/zellij-server/src/screen.rs
@@ -188,22 +188,37 @@ impl Screen {
         }
     }
 
+    /// A helper function to switch to a new tab at specified position.
+    fn switch_active_tab(&mut self, new_tab_pos: usize) {
+        if let Some(new_tab) = self.tabs.values().find(|t| t.position == new_tab_pos) {
+            let current_tab = self.get_active_tab().unwrap();
+
+            // If new active tab is same as the current one, do nothing.
+            if current_tab.position == new_tab_pos {
+                return;
+            }
+
+            current_tab.visible(false);
+            let new_tab_index = new_tab.index;
+            let new_tab = self.get_indexed_tab_mut(new_tab_index).unwrap();
+            new_tab.set_force_render();
+            new_tab.visible(true);
+
+            let old_active_index = self.active_tab_index.replace(new_tab_index);
+            self.tab_history.retain(|&e| e != Some(new_tab_pos));
+            self.tab_history.push(old_active_index);
+
+            self.update_tabs();
+            self.render();
+        }
+    }
+
     /// Sets this [`Screen`]'s active [`Tab`] to the next tab.
     pub fn switch_tab_next(&mut self) {
         let active_tab_pos = self.get_active_tab().unwrap().position;
         let new_tab_pos = (active_tab_pos + 1) % self.tabs.len();
 
-        for tab in self.tabs.values_mut() {
-            if tab.position == new_tab_pos {
-                tab.set_force_render();
-                self.tab_history.retain(|&e| e != Some(tab.index));
-                self.tab_history.push(self.active_tab_index);
-                self.active_tab_index = Some(tab.index);
-                break;
-            }
-        }
-        self.update_tabs();
-        self.render();
+        self.switch_active_tab(new_tab_pos);
     }
 
     /// Sets this [`Screen`]'s active [`Tab`] to the previous tab.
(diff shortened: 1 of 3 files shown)


zellij-server/src/tab.rs
```diff
@@ -2400,6 +2400,29 @@ impl Tab {
             }
         }
     }
+
```
TheLostLambda, line 2403: Hey! Thanks for submitting this PR! I like the functionality that this enables, but I think we could clean things up a little by having something like `Visible(bool)`, with `Visible(true)` replacing `Active` and `Visible(false)` replacing `InActive`.

That way we could clean up these duplicated functions and keep the number of new Events minimal! :)

Hopefully you won't need to use this hack for your timer in the near future, but this will be useful even after that I think :)

TheLostLambda (changes requested):

zellij-server/src/tab.rs
tw4452852: This is a nice idea for merging two events into one. 
Currently, I use this mechanism to save and restore the state across multiple instances of a plugin.

zellij-server/src/screen.rs
```diff
@@ -188,22 +188,40 @@ impl Screen {
...
+    fn switch_active_tab(&mut self, new_tab_pos: usize) {
+        let current_tab = self.get_active_tab().unwrap();
+        let current_tab_pos = current_tab.position;
+
+        // If new active tab is same as the current one, do nothing.
+        if current_tab_pos == new_tab_pos {
+            return;
+        }
+
+        // FIXME: old active tab should only be inactive once the new active tab is determined,
```
tw4452852, line 201: @TheLostLambda Do you have any idea about this ？

zellij-server/src/tab.rs
TheLostLambda: Nice! That's what I thought :)

zellij-server/src/screen.rs
TheLostLambda: Took a while, but what do you think about this? Does this still work?

```rs
    /// A helper function to switch to a new tab at specified position.
    fn switch_active_tab(&mut self, new_tab_pos: usize) {
        if let Some(new_tab_idx) = self.tabs.values().position(|t| t.position == new_tab_pos) {
            let current_tab = self.get_active_tab().unwrap();

            // If new active tab is same as the current one, do nothing.
            if current_tab.position == new_tab_idx {
                return;
            }

            current_tab.inactive();

            let new_tab = self.get_indexed_tab_mut(new_tab_idx).unwrap();
            new_tab.set_force_render();
            new_tab.active();

            let old_active_index = self.active_tab_index.replace(new_tab_idx);
            self.tab_history.retain(|&e| e != Some(new_tab_idx));
            self.tab_history.push(old_active_index);

            self.update_tabs();
            self.render();
        }
    }
```

tw4452852: Wow, looks very nice! Thanks a lot!

tw4452852: @TheLostLambda 
Sorry, I'm happy too fast.

This is also forbidden:

```
warning: cannot borrow `*self` as mutable because it is also borrowed as immutable
   --> zellij-server/src/screen.rs:202:27
    |
194 |         if let Some(new_tab) = self.tabs.values().find(|t| t.position == new_tab_pos) {
    |                                --------- immutable borrow occurs here
...
202 |             let new_tab = self.get_indexed_tab_mut(new_tab.index).unwrap();
    |                           ^^^^                     ------------- immutable borrow later used here
    |                           |
    |                           mutable borrow occurs here
    |
    = note: `#[warn(mutable_borrow_reservation_conflict)]` on by default
    = warning: this borrowing pattern was not meant to be accepted, and may become a hard error in the future
    = note: for more information, see issue #59159 <https://github.com/rust-lang/rust/issues/59159>

error[E0499]: cannot borrow `self.active_tab_index` as mutable more than once at a time
   --> zellij-server/src/screen.rs:206:36
    |
202 |             let new_tab = self.get_indexed_tab_mut(new_tab.index).unwrap();
    |                           ---- first mutable borrow occurs here
...
206 |             let old_active_index = self.active_tab_index.replace(new_tab.index);
    |                                    ^^^^^^^^^^^^^^^^^^^^^         ------------- first borrow later used here
    |                                    |
    |                                    second mutable borrow occurs here
```

tw4452852:
Hi @TheLostLambda 

I did some changes with your excellent help and suggestions. Also, I rebased onto the latest main

PR #723, opened 2021-09-14
fix(plugin): Set preopen directories properly

tw4452852:
Every plugin will have following two directories for its use:

`./`: Plugin's own data should be saved here, every plugin will have its own directory.
`/global/`: All plugins have access to this directory, Some shared data could be saved here.

### DIFF
--- a/zellij-server/src/wasm_vm.rs
+++ b/zellij-server/src/wasm_vm.rs
@@ -56,6 +56,7 @@ pub(crate) struct PluginEnv {
     pub subscriptions: Arc<Mutex<HashSet<EventType>>>,
     // FIXME: Once permission system is ready, this could be removed
     pub _allow_exec_host_cmd: bool,
+    plugin_own_data_dir: PathBuf,
 }
 
 // Thread main --------------------------------------------------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


TheLostLambda (changes requested):
Another promising feature! I think we'll need to do a little more thinking about this though (it totally breaks `strider` for example, which relies on being able to access the current directory tree).

Additionally, while I think that having a private data directory for each plugin is good, I'm not really sure if we'd need / want a `/global` directory. It feels like that would primarily be for communicating between different plugins, but I think that's something that I think would be better implemented in another way.

I'd be happy with changing the `plugin_own_data_dir` to `.join(path)` so that all of your timer plugin instances can still share data like you need them to!

To summarize, how would you feel about:

1. Changing `.` to `/host` and mounting the current directory there (this will also require Strider to be updated, but is better than just changing `.` back to the current directory, since that would conflict with `/global` in some cases)
2. Changing `/global` to `/data` and naming that directory after the path of the loaded plugin somehow instead of the `plugin_id` (this way you can still share files between several instances of the same plugin)

tw4452852:
Hi @TheLostLambda 

Very appreciate your good suggestions.

> 1. Changing . to /host and mounting the current directory there (this will also require Strider to be updated, but is better than just changing . back to the current directory, since that would conflict with /global in some cases)

I'm fine with this.

> 2. Changing /global to /data and naming that directory after the path of the loaded plugin somehow instead of the plugin_id (this way you can still share files between several instances of the same plugin)

Per my understanding, here `join(path)` means `join(plugin_name)`, right?
Currently, we have two ways to specify a plugin, either by name or by absolute path of the `wasm` binary. 
For the second way, shall we extract its from the loading path? Like `/path/to/test.wasm => test`

TheLostLambda:
@tw4452852 

> For the second way, shall we extract its from the loading path? Like /path/to/test.wasm => test

I'd actually be fine with either that or the full paths, you decide what you think works best! We can improve it to take into account versions and things in the future if we need :)

But yeah, plugin_name overall so plugins of the same type can share the folder like you need :)

tw4452852:
@TheLostLambda 

I'm done.

tw4452852:
Hold on please, there's an issue when unloading the plugin.

TheLostLambda:
@tw4452852 Good catch! Let me know once things are working!

tw4452852:
@TheLostLambda I'm done, but I leave a TODO for cleanup when unloading.

TheLostLambda:
> @TheLostLambda I'm done, but I leave a TODO for cleanup when unloading.

Ah! To be honest, I was kinda thinking that the folders could just stick around. That way plugins can actually store information between Zellij sessions!

tw4452852:
Ah, yes.

TheLostLambda:
@tw4452852 If you happy with that, I can merge this and then make the Strider fixes? Do you reckon this is ready to merge?

tw4452852:
Yes, it's ready. Feel free to merge. Thanks.

PR #724, opened 2021-09-14
fix: don't pad empty pane title

### LINKED ISSUE #713: On empty pane frame title it shows a gap
On empty pane frame title it shows a gap.
Setting an empty title still has the following behavior:
image

prscoelho:
Closes #713


imsnif:
Very nice! Thank you :)

imsnif:
oops, wrong button ;P

Issue #509, opened 2021-05-15
CPU usage when idle


kxt:
version: 0.10.0

I've noticed that an idle zellij process (eg. running only a barebones shell) uses quite a lot of CPU and reduces battery life too. Having multiple panes/tabs makes this more severe.

I found that in `pty.rs`, each fd is being actively polled to check if new data is available from a client. This can be verified by running zellij with strace, it reports thousands of syscalls per second during steady state (ie. idle).

I've created a [proof-of-concept patch](https://github.com/kxt/zellij/commit/39cfb77320d3ce0e9b5bb8a1bbe8cbc108b94014) that uses async-std to manage io and timeout (for batching updates to reduce render time). With this patch, activity during idle drops significantly, the only activity reported by strace is `os_input_output.rs`'s `handle_command_exit` looping.

I need someone to review it and discuss whether an approach like this is suitable for zellij and if so, how to move forward (eg. `ServerOsApi` would need to become and async trait if an approach like this would be used).

imsnif:
@kxt - this definitely looks like a good direction! I haven't tested this out yet, but how about you whip up a PR and we'll see how it behaves?

Key things to check: performance when dumping lots of lines to the screen (eg. cat a 100000 lines file and time it).

kxt:
Thank you for your input! I'm going to create 2 PRs:
* one is some assorted cleanup patches that I created while working on the async code, these are probably not very controversial and can be merged quickly
* the other is the actual fix for this issue

kxt:
I've created the first PR with the refactor patches, https://github.com/zellij-org/zellij/pull/521.
For the actual fix, I have some problems figuring out the fakes to investigate some failing testcases.

imsnif:
First PR merged!
Do feel free to reach out if you want a hand with the fakes. The whole testing framework is a little iffy and we're very eager to replace it :) You can reach out here or on discord, as you like.

PR #728, opened 2021-09-16
fix(server): fix leakage of SessionMetaData

tw4452852:
Currently, when the server exits, Session data is leaked because 'server_listener' thread still holds a reference to it.
Due to this thread blocking on the socket accepting, to fix this problem, we change it to polling mode (the timer interval is 10 milliseconds) and add a flag to indicate exit.

### DIFF
--- a/zellij-server/src/lib.rs
+++ b/zellij-server/src/lib.rs
@@ -170,19 +170,13 @@ pub fn start_server(os_input: Box<dyn ServerOsApi>, socket_path: PathBuf) {
                             thread_handles.lock().unwrap().push(
                                 thread::Builder::new()
                                     .name("server_router".to_string())
-                                    .spawn({
-                                        let session_data = session_data.clone();
-                                        let os_input = os_input.clone();
-                                        let to_server = to_server.clone();
-
-                                        move || {
-                                            route_thread_main(
-                                                session_data,
-                                                session_state,
-                                                os_input,
-                                                to_server,
-                                            )
-                                        }
+                                    .spawn(move || {
+                                        route_thread_main(
+                                            session_data,
+                                            session_state,
+                                            os_input,
+                                            to_server,
+                                        )
                                     })
                                     .unwrap(),
                             );
@@ -267,7 +261,6 @@ pub fn start_server(os_input: Box<dyn ServerOsApi>, socket_path: PathBuf) {
                 }
             }
             ServerInstruction::ClientExit => {
-                *session_data.write().unwrap() = None;
                 os_input.send_to_client(ServerToClientMsg::Exit(ExitReason::Normal));
                 break;
             }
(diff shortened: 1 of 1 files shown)


kunalmohan:
@tw4452852 Thanks for catching this! I'm a bit curious about how big of an issue this is and whether there is an alternate approach to solve this. This only happens when zellij is exiting and IIUC the OS releases any memory allocated to a process after it exits.

The reason I'm concerned is that the new change adds polling to the thread so the thread will wake up every 10ms. We had a similar polling approach in some other section of zellij in some of the earlier releases due to which users complained of higher CPU and power consumption (Linked Issue: https://github.com/zellij-org/zellij/issues/509).

tw4452852:
Hi @kunalmohan 

Indeed, any allocated memory will be released by the OS when the process exits. But as I did [some temporary files cleanup](https://github.com/zellij-org/zellij/pull/723/files#diff-72baa051cbfabfc51139227f2a7a27671782c29e1e29537326fef9614e945ef9R171) in PR #723 when `wasm` thread exits. And this depends on the notification sent when `SessionMetaData` drops:

https://github.com/zellij-org/zellij/blob/72b0474d024abc799a2f5e90a1078f68b150502a/zellij-server/src/lib.rs#L106

Regarding the CPU usage when polling, it's almost ignorable compared to the blocking mode. As I use this during my daily work.

imsnif:
I hope it's okay if I chime in here: if this is in order to facilitate something else that hasn't been merged yet, maybe we can look into that first and then see if we still need this one? I guess @TheLostLambda will probably have some useful input regarding cached plugin disk metadata.

kunalmohan:
FWIW we drop the `SessionMetaData` object on exit here: https://github.com/zellij-org/zellij/blob/72b0474d024abc799a2f5e90a1078f68b150502a/zellij-server/src/lib.rs#L270
But there are other cases where we don't drop it on exit ([this](https://github.com/zellij-org/zellij/blob/72b0474d024abc799a2f5e90a1078f68b150502a/zellij-server/src/lib.rs#L288) and [this](https://github.com/zellij-org/zellij/blob/72b0474d024abc799a2f5e90a1078f68b150502a/zellij-server/src/lib.rs#L296)). So maybe we can move the above statement outside the loop, before:
https://github.com/zellij-org/zellij/blob/72b0474d024abc799a2f5e90a1078f68b150502a/zellij-server/src/lib.rs#L300-L304

tw4452852:
Ah, I see!

tw4452852:
BTW, after re-reading the code, I think the second clone is unnecessary:

https://github.com/zellij-org/zellij/blob/72b0474d024abc799a2f5e90a1078f68b150502a/zellij-server/src/lib.rs#L167-L187

Correct me if I'm wrong.

kunalmohan:
@tw4452852 yes, you are right!

tw4452852:
Thanks @kunalmohan 

I have updated the patch according to your suggestions.

kunalmohan (approved):
LGTM! Thanks!!

PR #729, opened 2021-09-16
fix(cwd-pane): Fix for cwd not being inherited when switching tabs

### LINKED ISSUE #727: Working directory inheritence isn't quite working
I just noticed that when switching between tabs and then splitting into a new pane the current working directory is inherited from the previous tabs pane and not the pane that's focused after switching tabs. This is likely a because we're not updating the pty's active pane when switching between tabs. I will look into why this is happening later tonight and file a PR to fix it.

spacemaison:
fixes #727

Inheriting the current working directory didn't work when switching between tabs. This happened because the event to notify the pty of an pane change was triggered when setting the active pane inside of the current tab, but switching between tabs will only cause a re-render of the newly selected tab and it's panes without setting the active pane. This was fixed by moving the event to notify the pty of the pane change into the tabs render method.

### DIFF
--- a/zellij-server/src/tab.rs
+++ b/zellij-server/src/tab.rs
@@ -295,13 +295,6 @@ impl Tab {
         }
     }
 
-    fn set_active_terminal(&mut self, pane_id: Option<PaneId>) {
-        self.active_terminal = pane_id;
-        self.senders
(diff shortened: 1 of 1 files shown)


imsnif:
Thanks for the quick fix!

PR #731, opened 2021-09-16
 Implement `attach --create` subcommand flag to create session if one does not exist

GPery:
Hey!
Addressing #541 to help my automated zellij remote workspace :D

Thanks!

### DIFF
--- a/src/sessions.rs
+++ b/src/sessions.rs
@@ -102,7 +102,7 @@ pub(crate) fn list_sessions() {
     process::exit(exit_code);
 }
 
-pub(crate) fn session_exists(name: &str) -> Result<bool, ErrorKind> {
+pub(crate) fn session_exists(name: &str) -> Result<bool, io::ErrorKind> {
     return match get_sessions() {
         Ok(sessions) => {
             if sessions.iter().any(|s| s == name) {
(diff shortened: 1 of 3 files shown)


zellij-utils/src/cli.rs
```diff
@@ -85,6 +85,11 @@ pub enum Sessions {
         /// zellij client (if any) and attach to this.
         #[structopt(long, short)]
         force: bool,
+
+        /// Create a session if one does not exist.
+        #[structopt(long)]
```
kunalmohan, line 90: Let's have the `short` version as well.

src/sessions.rs
```diff
@@ -1,12 +1,13 @@
 use std::os::unix::fs::FileTypeExt;
 use std::{fs, io, process};
 use zellij_utils::{
+    async_std::io::ErrorKind,
```
kunalmohan, line 4: Why can't we use `std::io::ErrorKind` instead of this?

kunalmohan (changes requested):
@GPery Thanks for taking this up! This looks very good. Just a couple of small changes and it should be ready to merge.

src/sessions.rs
GPery: No reason, fixed

zellij-utils/src/cli.rs
GPery: Of course
I initially removed the short option because it was for `zellij` and not for `attach`
Added

kunalmohan (approved):
LGTM! Thanks for working on this! It is a much-needed feature.

PR #740, opened 2021-09-24
Mirrored sessions

imsnif:
This change introduces "mirrored sessions". With this change, when one tries to attach to a session that is already attached, instead of getting an error, the session will be mirrored in the new terminal window.

This change also introduces some fixes and stabalizations for the e2e tests.

@kunalmohan - if you could take a look mainly at the parts involving the server os_input_output to make sure I didn't make any fatal mistakes there, that would be great. :)

I know this change is a little big, so I'd be happy to walk you through the code as needed. Thanks!

EDIT:
The main changes to the client involve moving the input handler to its own thread and creating a new thread that would just read events from stdin (and interpret them with termion).

The reason for this is that the mirrored sessions need to be in the same mode. Since we kept the input mode on the client side, if one client changed the input mode, we would need to let the other clients know somehow. So I introduced a new ServerToClient::SwitchToMode variant that would let the client know a different client (or it itself) switched a mode and it needs to update its own state. This is handled by the new "input thread" that also handles events from the stdin thread.


zellij-server/src/route.rs
```diff
@@ -282,47 +289,75 @@ pub(crate) fn route_thread_main(
...
+                to_server.send(new_client_instruction).unwrap();
             }
-            ClientToServerMsg::AttachClient(_, force, _) => {
-                if *session_state.read().unwrap() == SessionState::Attached && !force {
-                    os_input.send_to_temp_client(ServerToClientMsg::Exit(ExitReason::CannotAttach));
-                } else {
-                    os_input.add_client_sender();
-                    to_server.send(instruction.into()).unwrap();
-                }
+            ClientToServerMsg::AttachClient(client_attributes, force, opts) => {
```
kunalmohan, line 350: The `force` option seems to be obsolete now (It is not being used here or in `ServerInstruction::AttachClient(..)`. Can we remove that?

kunalmohan (approved):
This looks great! Just a small concern with the `--force` flag in the attach command. I haven't looked at the tests though, you know it better than me.
I'm very happy with the changes you made 😄, especially the ones in the `ServerOsApi` (It looks very neat now).
Merge when you're ready 🙂

zellij-server/src/route.rs
imsnif: Good catch - fixed. Thanks!

imsnif:
Thanks for the review!

PR #745, opened 2021-09-28
feat: add `kill-session` command

### LINKED ISSUE #644: Feature: add command `kill-session` like tmux
Hi, zellij 👋 

Maybe I missed it, but currently there doesn't seem to be a feature to close **individual** or **all** sessions via terminal commands.

Are there any plans to add that feature? If there is a plan, I would like to contribute.

jaeheonji:
Fix #644 

Hi, everyone 👋 
I simply added a command to close the session. 

The purpose of this feature is to close a session via a command **inside** or **outside** the session.

In order to give the same experience as `tmux` as much as possible, I implemented it referring to the commands of `tmux` like below.

* `kill-session [-a] [-t target-session]`

By the way, I'm a Rust **newbie** and I need a lot of feedback from community!

Anyone feel free to give feedback 😸


imsnif:
Hey, just wanted to apologize for not getting to this yet! I plan on taking a look tomorrow.

imsnif:
Hey @jaeheonji - first of all, I'm really sorry for taking so long to get to this (and not getting to this when I promised). It's been a week for me.

Second - I'm happy to see this being picked up. We just had a request for this on Discord.

So, before I look at the code - I tried this out and it isn't working for me. Though maybe I'm doing something wrong?

What I'm doing:

1. `cargo make run`
2. In a different terminal window, `target/debug/zellij kill-session -t <session_name>`
3. Nothing happens

I also tried with `target/debug/zellij kill-session -a`... am I missing something?

Also, if I just do cargo make run -- kill-session I get a hard-crash (panicked on unwrapping a None).

jaeheonji:
Thanks for checking @imsnif 

Hm.. interesting.. Actually, when I tested it before merging upstream, it worked fine. However, as you pointed out, now that I check it, there seems to be something wrong.

It is dawn where I am now. I'll get up tomorrow morning and check it out 👍

imsnif:
Aha! I think I know why this happened. I recently added a "mirrored sessions" feature to allow users to connect to the same session from different terminal windows. With this session the behaviour of the `QUIT` action changed (it only quits if you're the last user connected). So you might want to add a `KILL_SESSION` action and use it instead (or solve it in some other way that you think is better).

Have a good sleep meanwhile :)

jaeheonji:
There was such a story! thank you :)

jaeheonji:
There is one concern I have added during development.

If the user **unintentionally** uses `kill-all-sessions`, the state of all sessions is lost. so I added the **interactive question** when executing it.

The code can be weird because it's simple to implement without using additional libraries 😭 

If you think it's too functionally **overkill**, please leave a comment.

HKalbasi:
Just used this feature. Thanks!

PR #754, opened 2021-09-30
feat(plugins-options): Add ability to configure plugins with user supplied options

spacemaison:
This PR allows for users to add an options fields in either the `plugins` list in the config file or the `run` field in layout files. Those options are then passed into plugins and deserialized as structs that are passed into the load method of the plugin.


spacemaison:
Hey @TheLostLambda, sorry for the reciprocal delay in responding to your comments! I do like you're second suggestion to the plugin API better than any of the others, and I changed the PR to do it that way. I also left a comment above to clarify why `PluginOptions` may be necessary. Let me know what you think.

TheLostLambda:
Sorry for the delay @imsnif and @spacemaison ! I'm reviewing this now!

In the future it would be ideal if we could avoid force pushing (it makes things a bit harder to review and all of the PR gets squashed on merge anyways), but no worries if just starting from main to solve merge conflicts is easier – avoiding force pushing is just a preference, certainly not some rule :)

default-plugins/status-bar/src/main.rs
```diff
@@ -136,7 +136,9 @@ fn color_elements(palette: Palette) -> ColoredElements {
 }
 
 impl ZellijPlugin for State {
-    fn load(&mut self) {
+    type Options = ();
+
+    fn load(&mut self, _: ()) {
```
TheLostLambda, line 141: Stylistically, it might be preferable to use `Self::Options` instead of `()` here? Then you'd only need to change the configuration type in one place (line 139) instead of there and this line!

default-plugins/strider/src/main.rs
```diff
@@ -8,7 +8,9 @@ use zellij_tile::prelude::*;
 register_plugin!(State);
 
 impl ZellijPlugin for State {
-    fn load(&mut self) {
+    type Options = ();
+
+    fn load(&mut self, _: ()) {
```
TheLostLambda, line 13: Same as above if you want to change the `()` to `Self::Options` :)

default-plugins/tab-bar/src/main.rs
```diff
@@ -29,7 +29,9 @@ static ARROW_SEPARATOR: &str = "";
 register_plugin!(State);
 
 impl ZellijPlugin for State {
-    fn load(&mut self) {
+    type Options = ();
+
+    fn load(&mut self, _: ()) {
```
TheLostLambda, line 34: Same as above :)

zellij-tile/src/lib.rs
```diff
@@ -3,10 +3,13 @@ pub mod prelude;
...
 
 use data::*;
+use serde::de::DeserializeOwned;
 
 #[allow(unused_variables)]
 pub trait ZellijPlugin {
-    fn load(&mut self) {}
+    type Options: Default + DeserializeOwned;
+
+    fn load(&mut self, _: Self::Options) {}
```
TheLostLambda, line 12: We don't need to use `_` here, we should probably use a more descriptive name for documentation purposes (something like `config` or `options`?) Even if we aren't using it, I've already added `#[allow(unused_variables)]` to suppress that particular warning.

zellij-tile/src/lib.rs
```diff
@@ -19,8 +22,16 @@ macro_rules! register_plugin {
         }
 
         fn main() {
+            let options = $crate::shim::object_from_stdin().map_err(|err| {
+                eprintln!(
+                    "Your plugins configuration had a derialization error: {:?}.",
```
TheLostLambda, line 27: Just a small English nitpick, I think this should be `plugin's` here?

Also, does this print this error if a plugin is left unconfigured and the default is used? If so, we might want to change the error to make it a little more descriptive.

zellij-utils/src/input/plugins.rs
```diff
@@ -189,6 +196,82 @@ impl Default for PluginTypeFromYaml {
     }
 }
 
+/// PluginOptions are arbitrary options passed into plugins as they are instantiated. The variants
```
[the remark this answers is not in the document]
TheLostLambda: Yeah, I think this is a pretty complex region of the code so I'm fine leaving this as-is for now.

With that being said, this is a somewhat scary amount of boilerplate – how would you feel about adding a little `// FIXME:` note mentioning that this part of the code could use a second look / refactor. Adding a comment with `FIXME` in it helps people find it in the future!

TheLostLambda (approved):
Hi @spacemaison! Just some tiny tweaks overall! If you'd like to make some of those we can merge it right after!!

Issue #164, opened 2021-02-10
Feature: move panes directionally


imsnif:
This will allow the user to move the current pane in a specific direction, replacing its location with another pane that's already there.

### How will this work?
This will add a new input mode: `move`.
In this mode, the arrow keys (as well as hjkl and emacs movements - this behaviour can be copied from other modes) will change the location of the focused pane with the largest overlapping pane in the specified direction. If there are no panes in that direction, this will be a noop.

PR #762, opened 2021-10-03
Feature: Move panes directionally

kunalmohan:
Fix: #164


imsnif:
I'm not familiar with those either I'm afraid :/ If it were me I'd try to debug and compare to main, see what they expect and what's going wrong.

Also btw - be sure to test that with the ctrl-h shortcut we can still use `backspace` - might be awkward otherwise :)

PR #789, opened 2021-10-15
feat: add initial session name to layout template

### LINKED ISSUE #611: Add `session_name` to `layouts`
Add the ability to specify an optional session name to the layout,
make it overridable by the config flags.

jaeheonji:
This PR resolve #611 

**proposal**

```
---
session:
  - name: "SESSION_NAME"
template:
  ...
tabs:
  ...
```

Instead of using `session_name`, a `session` structure is used. (for future scalability)

**expected results**

1. When Zellij is launched (with layouts), the session name is initialized to the `name` in the `layouts`.
2. When the `-s` flag and `layouts` are used together (like `zeliij -s SOMTHING --layout-path SOME.yaml`), the flag of session name takes a priority.


jaeheonji:
@a-kenji  I updated attach options to work only when there is a layout template argument.

And as the main function has more functions, the code seems to be massive. It seems that refactoring is necessary in the future.

a-kenji:
@jaeheonji 
This is still a little off from the desired behavior, but very close.
I think rather than check for the `layout template` argument,
you rather want to check where the `layout name` is coming from,
since you should be able to create a session, with a name,
and a layout path argument.

Thanks for the great work! 
And I agree the main function is in need for some maintenance soon.

a-kenji:
@jaeheonji 
Yes, that is exactly the behavior I think we would expect!
Sorry for not being clearer before!

jaeheonji:
Good! We all don't have to be sorry :)

If the above is implemented, the length of the code will be longer and there will be many overlapping parts. (probably there will be many flows)

So, I'll try the implementation once with refactoring.

a-kenji:
@jaeheonji 
Just a heads up, in  #829 there is a refactor of the main function already.
Do you wish to base upon that refactor?
It seems to make things much clearer.

jaeheonji:
@a-kenji
That would be good. (I have now confirmed that #829 is going on.)

The work on #829 is exactly what I was looking for. After the refactoring is done, it would be nice to implement it in that code base.

ken-matsui:
Thank you all so much for considering using my refactor. It still has a bit confusing parts, so I really would welcome your suggestion!

jaeheonji:
@a-kenji

I added features (https://github.com/zellij-org/zellij/pull/789#issuecomment-956215564) from the latest sources (refactored). please could you do a review?

a-kenji (approved):
LGMT!

a-kenji:
@jaeheonji,
thank you for keeping at it! 
This looks great now I think.

PR #792, opened 2021-10-18
Use the parking_lot crate for Mutex, RwLock and CondVar

fabricedesre:
The parking_lot versions of these types have a more efficient implementation and better ergonomics (less unwrap()s !).


a-kenji:
The standard library implementation is now more efficient in almost all cases, in some by a big margin. And in the case it's not, by a very small margin.

PR #798, opened 2021-10-22
pty: immediately render after a period of inactivity

Gnurou:
The current code waits for 30ms after reading new data to refresh the
screen in order to batch input together. This reduces the number of
refreshes, but also adds perceptible latency as keystrokes will only get
visual feedback after a 30 ms delay.

Improve this by immediately refreshing if more than 30ms have already
passed since the last refresh. That way punctual events like keystrokes
get immediate feedback, while heavier input still gets batched together.


a-kenji:
@Gnurou
Thank you for the contribution, 
I am sure I am seeing the fruits of it already and it is a lot of fun.

PR #801, opened 2021-10-24
feat(ui): add right-click support

tw4452852:


TheLostLambda:
An excellent PR as always! Merged!

Issue #794, opened 2021-10-20
Half-page scroll actions


tailhook:
I would like to propose adding half-page scroll actions. In `less` they are bound to `Ctrl+u`/`Ctrl+d` shortcuts. I think these shortcuts can also be used in scroll mode, but at least actions will be useful (if shortcuts are occupied by something).

It's important because it allows to keep track of what is going on when scrolling (i.e. eyes can match half-screen text shifted, and it's easier to find out if something you're looking for would be on the edge of the screen if scrolling by page). Line-wise scrolling is too slow. Alternative might be something like "scroll up 10 lines" (with configurable number of lines), but usually half-page serves as a good number.

a-kenji:
This is now in `v20.0`, thanks @oromate

PR #813, opened 2021-10-28
 Half-page scroll actions

oromate:
I managed to make it this far. I think there is still a lack of documentation and tests, what do you think? I basically replicated the page scroll code and did the division by two, so maybe it would be nice to avoid code duplication. #794


zellij-server/src/tab.rs
```diff
@@ -2513,6 +2513,31 @@ impl Tab {
...
         }
     }
+    pub fn scroll_active_terminal_up_half_page(&mut self) {
+        if let Some(active_terminal_id) = self.get_active_terminal_id() {
+            let active_terminal = self
+                .panes
+                .get_mut(&PaneId::Terminal(active_terminal_id))
+                .unwrap();
+            // prevent overflow when row == 0
+            let scroll_columns = (active_terminal.rows().max(1) - 1)/2;
```
tailhook, line 2523: Should be `scroll_rows`, right?

imsnif: Yes! But to be fair this is also a mistake in the other methods this was copied from. If you'd like to fix this @oromate that would be cool.

imsnif (changes requested):
Hey @oromate - thank you for your patience and apologies it took me a few days to get to this.

All in all this looks great. There's the issue with the variable name mistake in the code comments if you feel like fixing it there and in the other methods, but no biggie all in all.

Aside from that, the CI tells us this needs to be formatted. You can do that with rustfmt (I run `cargo fmt` you can see more details in our CONTRIBUTING doc), and the e2e tests.

The e2e tests are failing because the controls line changed for Scroll mode and they're comparing it against the previous snapshot and finding a difference. We use `cargo insta` to compare snapshots (https://insta.rs/docs/cli/). You can update the snapshots with their CLI tool once you've run the tests and then commit them to this branch.

Once that is done, I'd be happy to merge this. Please let me know if anything is unclear! I'll try to be quicker on the uptake with the changes.

zellij-server/src/tab.rs
oromate: I can fix this. Do you think it makes sense for me to also correct the variable name in the other method?

imsnif: That would be great, thanks! There are a few of those.

oromate:
I believe I managed to resolve the reviews. I hope everything works out.

Taking the opportunity to share a little of the journey:
- I needed to change the version of docker-compose.yml to 3.3 because of an error with the volume type
- I needed to install the "musl-tools" dependency
- The command to update the snapshots is "cargo insta review --workspace-root src/tests/". As I didn't know that, it could be nice to leave notes to let the next ones know :)

imsnif:
Hey @oromate - thanks for the fixes! I went ahead and ran `cargo fmt` on the code because the CI was still red and I didn't want to do another round just for that :)

Thanks for your observations with docker-compose! Would you like to maybe format them a little and add them to CONTRIBUTING.md? They have the best chance of being seen there.

Thanks for your contribution!

PR #822, opened 2021-10-31
add support for osc 8 escape code

tlinford:
This adds support for osc8, for a quick idea of how it works see: https://iterm2.com/documentation-escape-codes.html, while this [gist](https://gist.github.com/egmontkob/eb114294efbcd5adb1944c9f3cb5feda) has a lot more details covered. 

The case of a newline immediately after a link ends, for example
```shell
printf '\e]8;;\n\nhttp/example.com\e\\LINK1\e]8;;\e\\\n
```
 was  a bit problematic, since the line gets padded with empty characters up to the width of the row:

screenshot-2021-10-31T17:23:28

I handled this by adding a zero width null character to end the link in this situation, not quite sure if that's a reasonable solution though.

### DIFF
--- a/zellij-server/src/panes/grid.rs
+++ b/zellij-server/src/panes/grid.rs
@@ -1490,8 +1490,7 @@ impl Perform for Grid {
                 if params.len() < 3 {
                     return;
                 }
-                self.link_handler
-                    .dispatch_osc8(params, bell_terminated, &mut self.cursor);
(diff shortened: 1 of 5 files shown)


imsnif:
Hey @tlinford - what's the status of this?

tlinford:
Hey, I think it's pretty much good to go, want me to go ahead and merge?

imsnif:
Sure, if you think it's good then go for it.

Issue #803, opened 2021-10-25
Zellij is killed by OOM


tailhook:
**Basic information**

`zellij --version`: 0.18.1
`tput lines`: 42
`tput cols`: 159
`uname -av` or `ver`(Windows): Linux nixos 5.14.10 #1-NixOS SMP Thu Oct 7 05:53:20 UTC 2021 x86_64 GNU/Linux
`alacritty --version`: alacritty 0.9.0

**Further information**
Zellij was killed by OOM while dropping all running processes. I'm not sure:
1. Which process of zellij was dropped first (server?)
2. Why it was killed? Total memory usage on the system wasn't high. is it because of high VM usage? (but looks like that 78Gb is quite usual for zellij having few tabs open).

Prior to OOM killing there was high CPU usage spike. I'm not sure if that was by zellij or kernel trying to compress memory (is usually the case before OOM), or other running process inside zellij.

Here is the dmesg log:
```
[643020.520328] oom-kill:constraint=CONSTRAINT_NONE,nodemask=(null),cpuset=user.slice,mems_allowed=0,global_oom,task_memcg=/user.slice/user-1000.slice/session-2.scope,task=zellij,pid=2377,uid=1000
[643020.520390] Out of memory: Killed process 2377 (zellij) total-vm:78086060kB, anon-rss:555168kB, file-rss:0kB, shmem-rss:0kB, UID:1000 pgtables:1556kB oom_score_adj:0
[643020.551282] oom_reaper: reaped process 2377 (zellij), now anon-rss:0kB, file-rss:0kB, shmem-rss:0kB
```
Here is the panic I in the terminal running alacritty:
```
Originating Thread(s):
1. screen_thread: Render
2. ipc_server: Render

Error: thread 'router' panicked at 'called `Result::unwrap()` on an `Err` value: Io(Error { kind: UnexpectedEof, message: "failed to fill whole buffer" })': /build/source/zellij-utils/src/ipc.rs:167
   0: zellij_utils::errors::handle_panic
   1: std::panicking::rust_panic_with_hook
   2: std::panicking::begin_panic_handler::{{closure}}
   3: std::sys_common::backtrace::__rust_end_short_backtrace
   4: rust_begin_unwind
   5: core::panicking::panic_fmt
   6: core::result::unwrap_failed
   7: <zellij_client::os_input_output::ClientOsInputOutput as zellij_client::os_input_output::ClientOsApi>::recv_from_server
[...]

Issue #796, opened 2021-10-21
Zellij leaks file descriptors to the running processes/shell


tailhook:
**Basic information**

`zellij --version`: 0.8.1
`tput lines`: irrelevant
`tput cols`: irrelevant
`uname -av` or `ver`(Windows): Linux nixos 5.14.10 #1-NixOS SMP Thu Oct 7 05:53:20 UTC 2021 x86_64 GNU/Linux
`alacritty --version`: alacritty 0.9.0

**Further information**
Here is what I get in `lsof -p $$` of the shell started from zellij:
```
zsh     189291   pc    0u   CHR              136,7       0t0      10 /dev/pts/7
zsh     189291   pc    1u   CHR              136,7       0t0      10 /dev/pts/7
zsh     189291   pc    2u   CHR              136,7       0t0      10 /dev/pts/7
zsh     189291   pc    5u  unix 0x0000000000000000       0t0   22962 /run/user/1000/zellij/0.18.1/session1 type=STREAM (LISTEN)
zsh     189291   pc    6u  unix 0x0000000000000000       0t0   22963 /run/user/1000/zellij/0.18.1/session1 type=STREAM (CONNECTED)
zsh     189291   pc    7u  unix 0x0000000000000000       0t0   22963 /run/user/1000/zellij/0.18.1/session1 type=STREAM (CONNECTED)
zsh     189291   pc    8u  unix 0x0000000000000000       0t0   22084 /run/user/1000/zellij/0.18.1/session1 type=STREAM (CONNECTED)
zsh     189291   pc    9u  unix 0x0000000000000000       0t0   22084 /run/user/1000/zellij/0.18.1/session1 type=STREAM (CONNECTED)
zsh     189291   pc   10u   CHR              136,7       0t0      10 /dev/pts/7
zsh     189291   pc   11u   CHR                5,2       0t0      85 /dev/ptmx
zsh     189291   pc   14u  unix 0x0000000000000000       0t0   22098 /run/user/1000/zellij/0.18.1/session1 type=STREAM (CONNECTED)
zsh     189291   pc   15u  unix 0x0000000000000000       0t0   22098 /run/user/1000/zellij/0.18.1/session1 type=STREAM (CONNECTED)
zsh     189291   pc   17r  FIFO               0,13       0t0  637219 pipe
zsh     189291   pc   18w  FIFO               0,13       0t0  637219 pipe
zsh     189291   pc   19u   CHR                5,2       0t0      85 /dev/ptmx
[...]

PR #830, opened 2021-11-01
fix(unix): forkpty => openpty

imsnif:
This is a fix for both https://github.com/zellij-org/zellij/issues/803 and https://github.com/zellij-org/zellij/issues/796

The major user facing effects are that this should significantly improve our resource utilization in regards to process count and memory usage. It also means child processes are not exposed to the open FDs of the main Zellij process.

The major changes:
1. Instead of creating new terminals with `forkpty` we now use `openpty` and do the forking ourselves with Rust's `Command` - which we were doing anyway. This saves considerable amounts of RAM and actually makes a lot of stuff simpler (eg. we don't need to set the process group of the forked process because `Command` does it for us. We also need to do the acrobatics we were doing to find out the shell's pid because it's now the child process itself).
2. We now explicitly reap both parts of the pty on child process end (after `wait`ing for it). I'm not sure if this was implicitly done before, but now its explicit.
3. As pointed out by @tailhook, we now close all fds before forking the child process so as not to leak. For now I elected not to also set the `CLOEXEC` flag every time we open an fd (eg. with our unix pipes) because quite honestly I've been down this rabbit hole for longer than I intended and this is a redundancy in this case. Would be nice to follow up and do this later though.

### DIFF
--- a/zellij-server/src/os_input_output.rs
+++ b/zellij-server/src/os_input_output.rs
@@ -1,5 +1,7 @@
 use std::collections::HashMap;
 
+use crate::panes::PaneId;
+
 #[cfg(target_os = "macos")]
 use darwin_libproc;
 
@@ -8,7 +10,7 @@ use std::fs;
 use std::os::unix::io::RawFd;
 use std::os::unix::process::CommandExt;
 use std::path::PathBuf;
-use std::process::{Child, Command};
+use std::process::{Child, Command, Stdio};
 use std::sync::{Arc, Mutex};
 
 use zellij_utils::{async_std, interprocess, libc, nix, signal_hook, zellij_tile};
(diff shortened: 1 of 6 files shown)


zellij-server/src/os_input_output.rs
```diff
@@ -97,96 +100,63 @@ fn handle_command_exit(mut child: Child) {
...
-        ForkResult::Child => {
-            let child = unsafe {
-                let command = &mut Command::new(cmd.command);
-                if let Some(current_dir) = cmd.cwd {
-                    command.current_dir(current_dir);
+        command
+            .args(&cmd.args)
+            .pre_exec(move || -> std::io::Result<()> {
+                if libc::login_tty(pid_secondary) != 0 {
+                    panic!("failed to set controlling terminal");
```
tailhook, line 121: I'm not sure if panics are safe in pre_exec. You can return some io error, and catch the error code (but not the text) in parent process as far as I remember.

PR #838, opened 2021-11-04
fix(resize): do not ignore minimums when doing directional resizes

imsnif:
The (by now not so new) resizing algorithm unfortunately ignores the minimum pane sizes. Which can sometimes cause rendering issues (eg. cursor being displayed on the pane border). This doesn't fix it 100%, but it at least fixes it for cases when we resize panes directly - which is the vast majority of cases. Otherwise it's just when resizing the whole terminal window.

@TheLostLambda - any chance you can see if I'm missing something critical? This seems good to me.

### DIFF
--- a/zellij-server/src/tab.rs
+++ b/zellij-server/src/tab.rs
@@ -1625,7 +1625,9 @@ impl Tab {
             panes_to_the_right.iter().all(|id| {
                 let p = self.panes.get(id).unwrap();
                 if let Some(cols) = p.position_and_size().cols.as_percent() {
-                    cols - increase_by >= RESIZE_PERCENT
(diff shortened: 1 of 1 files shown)


TheLostLambda:
Hey @imsnif ! This looks good to me! It doesn't touch any of the super scary code inside of the actual resize file, so it's a pretty safe looking change :)

I think this brings us to the most reasonable behavior – I really think that "fixing" this when the whole terminal window is resized isn't a problem we can solve in any useful way (until we can dynamically hide panes that are "less important" or something similar)

TheLostLambda (approved):

imsnif:
Cool, thanks for the second pair of eyes! I have some ideas about the resizer. Will run them by you at some point to see if they make sense.

PR #840, opened 2021-11-06
fix(docker-compose): Use the key-value style for environments to prevent human errors

ken-matsui:
I suggest using the key-value style for environments to prevent human errors in `docker-compose.yml` instead of the array. YAML detects duplicate keys.

I have two more suggestions: one is removing `---` on the top, another is bumping the docker-compose file version to 3. Are there any reasons to use them? If no reasons, I would like to do them.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -6,12 +6,12 @@ services:
     container_name: zellij-e2e
     hostname: zellij-e2e
     environment:
-      - PUID=1000
-      - PGID=1000
-      - TZ=Europe/Vienna
-      - PASSWORD_ACCESS=true
-      - USER_PASSWORD=test
-      - USER_NAME=test
+      PUID: 1000
+      PGID: 1000
+      TZ: Europe/Vienna
+      PASSWORD_ACCESS: true
+      USER_PASSWORD: test
+      USER_NAME: test
     volumes:
       - type: bind
         source: ./target


a-kenji:
Thanks for noticing this!
I think this is a good idea.
I think the `---` are there because otherwise it won't be a valid yaml document?
Maybe `docker-compose` is lenient in it's parsing? I am not sure.
If I recall correctly the version is on purpose not `3` yet,
because some distributions don't have an easy installation path for it.
I could be wrong on that and would like to get more input on that from someone.

ken-matsui:
@a-kenji 
Thank you for your response!

> I think the `---` are there because otherwise it won't be a valid yaml document?
> Maybe `docker-compose` is lenient in it's parsing? I am not sure.

If it is the reason, I think the YAML file format should adapt `docker-compose` unless docker-compose.yml is read programmatically. I can find `---` is not used on the [docker official document](https://docs.docker.com/compose/compose-file/compose-file-v2/#blkio_config).

> If I recall correctly the version is on purpose not `3` yet,
> because some distributions don't have an easy installation path for it.

Assuming that's the case, I totally agree with keeping using `2.1`.

> I could be wrong on that and would like to get more input on that from someone.

Sure! Thank you for considering my questions :)

a-kenji:
@ken-matsui 
Ah, thanks.
It seems that `docker-compose` defines it's own specification:
https://github.com/compose-spec/compose-spec/blob/master/spec.md
that is not really behaving like `yaml`:
https://yaml.org/spec/1.2.2/ , section 2.2 at least as far as I can tell.
If you want you can remove the `---` here, if that is more idiomatic for `docker-compose files`.

a-kenji (approved):

ken-matsui:
@a-kenji 
I see. Thank you. I would like to leave `---` for future scalability.

Also, thanks for your review! This PR is ready to merge :)

a-kenji:
Awesome, thanks for the pr and for the feedback!

PR #842, opened 2021-11-07
fix(envs): Unify operation of Zellij environment variables

ken-matsui:
This PR aims to prevent Zellij environment variable manipulation from being scattered in various places of code and to make adding new variables such as `ZELLIJ_PANE_ID` easy.

### DIFF
--- a/zellij-utils/src/envs.rs
+++ b/zellij-utils/src/envs.rs
@@ -1,0 +1,24 @@
+/// Uniformly operates ZELLIJ* environment variables
+use anyhow::Result;
+use std::env::{set_var, var};
+
+pub const ZELLIJ_ENV_KEY: &str = "ZELLIJ";
+pub fn get_zellij() -> Result<String> {
+    Ok(var(ZELLIJ_ENV_KEY)?)
+}
+pub fn set_zellij(v: String) {
+    set_var(ZELLIJ_ENV_KEY, v);
+}
+
+pub const SESSION_NAME_ENV_KEY: &str = "ZELLIJ_SESSION_NAME";
+pub fn get_session_name() -> Result<String> {
+    Ok(var(SESSION_NAME_ENV_KEY)?)
+}
+pub fn set_session_name(v: String) {
+    set_var(SESSION_NAME_ENV_KEY, v);
+}
+
+pub const SOCKET_DIR_ENV_KEY: &str = "ZELLIJ_SOCKET_DIR";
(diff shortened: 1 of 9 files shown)


zellij-utils/src/envs.rs
```diff
@@ -0,0 +1,24 @@
+/// Uniformly operates ZELLIJ* environment variables
+use anyhow::Result;
```
a-kenji, line 2: Why do we need the `anyhow::Result` here?

a-kenji (approved):
LGTM in general, I have one small question though!

zellij-utils/src/envs.rs
ken-matsui: @a-kenji 
Ah, I just thought it would be preferable to unify and absorb all result types to `anyhow::Result` that would be clear as I did on https://github.com/zellij-org/zellij/pull/824#discussion_r741408603.

If you enact how to handle result types as shown below, I think `anyhow::Result` should be removed from this PR:

1. Use `anyhow::Result` if a function returns various error types like https://github.com/zellij-org/zellij/pull/824#discussion_r741408603
2. If a function returns only one error type, just propagate it

a-kenji: Thank you for the insight, 
I think the list you suggested would be preferable,
unless there is a clear benefit to using `anyhow::Result` everywhere.

What do you think?

ken-matsui: @a-kenji 

> Additionally, the main function also can return `anyhow::Result` as shown [here](https://docs.rs/anyhow/1.0.44/anyhow/#details).

As I mentioned this [here](https://github.com/zellij-org/zellij/pull/824#discussion_r741408603), we can have the main function return `anyhow::Result`. This means all functions can perfectly back forward errors to main and a parent process with almost panicless using `?` operator. I personally think this is the benefit to use `anyhow::Result` everywhere. However, this takes a lot of work, and the benefit may not be worth the effort; in addition, some situations might need intentionally panicking.

To conclude, I think putting `anyhow::Result` everywhere will be preferable when the following conditions are satisfied:
1. If you want to collect all errors into the main function
2. If you want to avoid almost all panics and `process::exit`

a-kenji: Oh, my bad I thought it would still work that way with a normal `Result`.
But it explicitly needs to return the `anyhow::Result`.
Thank you for that explanation again!

PR #845, opened 2021-11-08
fix(docs): Add additional e2e test instruction for macOS users (including m1)

ken-matsui:
I think the e2e test instruction needs more information for macOS.


a-kenji:
Thank you!
I agree this section was needed.

PR #847, opened 2021-11-09
fix(ci): Potential vulnerability on GitHub Actions

ken-matsui:
Current cache policy on GitHub Actions would potentially be vulnerable; for example, we can replace cached code with malware by opening a new PR. Even if you do not merge it, caches are still risky. Therefore, I recommend they be stored in every branch separately. Additionally, versioning caches lets us immediately drop old caches for security reasons and being stale.

What I did:
* Added [`${{ github.ref }}`](https://docs.github.com/en/actions/learn-github-actions/contexts#github-context) to cache keys
* Added cache version to cache keys

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -7,6 +7,7 @@ on:
     branches: [ main ]
 
 env:
+  CACHE_VER: v1
   CARGO_TERM_COLOR: always
 
 jobs:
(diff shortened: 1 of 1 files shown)


a-kenji:
Hey @ken-matsui 
Is this not an issue anymore?

ken-matsui:
@a-kenji 
Ah, I just thought my understanding was incorrect. Let me know if it was unintentional. Also, let me will work on another issue of Zellij when I have enough time :)

a-kenji:
@ken-matsui 
I myself don't know much about Github Actions sadly and the people
that do, currently don't have much time.
So maybe it was correct - maybe it was incorrect?

> Also, let me will work on another issue of Zellij when I have enough time :)
Sure, I would be happy to!

ken-matsui:
@a-kenji 

> I myself don't know much about Github Actions sadly and the people
that do, currently don't have much time.
So maybe it was correct - maybe it was incorrect?

I've lost confidence in my proposal, so if a similar issue comes up again, let's consider it then. Thanks!

a-kenji:
@ken-matsui 
Sure, thank you for all the efforts!

PR #848, opened 2021-11-09
feat(status-bar): add multiple tips

### LINKED ISSUE #808: Add Multiple Tips To Zellij
Currently we just have a single "tip" that shows up in the Zellij status-bar – it would be nice to add a list of tips that are semi-randomly selected (a Did you know? style bit of information)

We should add the fact that `SHIFT` bypasses mouse-capture at run-time!

jaeheonji:
This PR resolve #808

#### Expected Behavior

When the program starts, it randomly shows the pre-defined tips from `tips.rs` to second_line of status-bar.
Also, if possible, changes in tips use cache (maybe `/tmp/zellij-xxxx`) to adjust exposure frequency.

#### TO-DO
- [x] Simplified declaring `tip` data.
- [x] `tip` data is randomly selected.
- [x] Use the local cache for `tip` selection.
- [x] Refactoring


jaeheonji:
I tried to implement the basic concepts discussed in #808.
Of course, this isn't perfect. but we can now start the detailed discussion.

I had focused the following 3 targets:

1. The **tip** should be hard-coded in the code (`tip/data.rs`), but it should be simple to write (using **macros**).
2. There are three types of tips, `short`, `medium`, and `full`, and they are managed as a group.
3. Using the cache, once exposed tips are continuously shown up to the “maximum hit”.

The cache is currently implemented as follows:

- The tip is exposed up to 10 times. it is declared as a constant value in `tip/cache.rs`
- When all tips reach their maximum hit, the cache is cleared and starts over.
- The cache file is in `json` format.

In addition, it would be nice to be able to set whether to use the cache and the maximum hit later through the manifest file.

The reason I changed the status of PR to open is to obtain consensus on the overall implementation.
After that's done, I will do the refactoring and so on :)

imsnif:
This seems sound to me @jaeheonji !

A few questions:
1. What happens after all tips have been shown max_hit times?
2. Would it be possible to somehow reset the cache (fully or for specific tips) when we issue a new version?

jaeheonji:
@imsnif 

1. What happens after all tips have been shown max_hit times? 

The cache data is cleared (delete the all contents in cache file) and new tips are randomly selected.
After that, the newly selected tip is cached and the operation is repeated.

2. Would it be possible to somehow reset the cache (fully or for specific tips) when we issue a new version?

It seems to be possible if some conditions and functions are implemented.
First, add the zellij version information to cache and when the plugin is loaded, we can compare the version of the cache and zellij-client

But in this case, the most important thing that how to get the zellij version in plugin.

The thought in my head, when wasmer invoke the function of `load`, pass the version information to plugin. however, it is occured all plugins to receive unnecessary data.

If my understanding is correct, issue #874 mentioned by @a-kenji seems to be related to this feature.

imsnif:
> The cache data is cleared (delete the all contents in cache file) and new tips are randomly selected. After that, the newly selected tip is cached and the operation is repeated.

So just to make sure I understand - once all tips have been shown then the whole process starts up again? Would it be possible to just select tips randomly once all of them have been shown their minimum times?

> It seems to be possible if some conditions and functions are implemented. First, add the zellij version information to cache and when the plugin is loaded, we can compare the version of the cache and zellij-client
> 
> But in this case, the most important thing that how to get the zellij version in plugin.

Can we maybe solve this from outside the plugins? As in, have the plugins be completely ignorant of which tips to load and be provided the tips from Zellij itself? I have a feeling it will make things easier. What do you think?

a-kenji:
> It seems to be possible if some conditions and functions are implemented. First, add the zellij version information to cache and when the plugin is loaded, we can compare the version of the cache and zellij-client

> But in this case, the most important thing that how to get the zellij version in plugin.

I don't really think this is a current concern, is it?
The plugin that shows the tip is a default plugin that ships with zellij itself,
and updates together with zellij.
And they are updated to have the same version together,
eg `19.0` will have the version of `19.0` plugin.


> Can we maybe solve this from outside the plugins? As in, have the plugins be completely ignorant of which tips to load and be provided the tips from Zellij itself? I have a feeling it will make things easier. What do you think?

It would make it easier in this current case,
but I really do think it might not be needed for this plugin here?

But I do think it is very important to have such a check that plugins can rely on.
For example a vendored zellij version might be a year old on some distributions,
but plugin authors might still want to provide some kind of updates to their plugins - 
without crashing on other zellij versions - or crashing the whole zellij session - 
which is currently the case.

a-kenji:
> If my understanding is correct, issue #874 mentioned by @a-kenji seems to be related to this feature.

It kind of is, I was meaning to write a little more about it, thanks!
But I think providing the version on load, like you suggested is
better in this instance.

imsnif:
I totally agree we should have this functionality @a-kenji, but maybe we can branch this off to a different issue? If @jaeheonji thinks this can be solved from the Zellij side, ofc.

a-kenji:
@imsnif,
Yes I also think this is better suited for a different issue.

a-kenji:
@jaeheonji,
Since you already interact with the file system here,
maybe for this specific reset functionality it is enough to read:
```
cat .local/share/zellij/VERSION
```

jaeheonji:
@imsnif 

> So just to make sure I understand - once all tips have been shown then the whole process starts up again? Would it be possible to just select tips randomly once all of them have been shown their minimum times?

Yes, unless the user intentionally modifies the cache file.

For example, if 3 tips A, B, C exist and the cache file looks like this:

```json
[
  { "name": "A", "hitted": 10 },
  { "name": "B", "hitted": 10 },
  { "name": "C", "hitted": 10 }
]
```

In this state, all tips are maximum hit. Therefore, the tip is randomly selected again among A, B, and C.

```json
[
  { "name": "B", "hitted": 1 }
]
```

If B is selected, the cache file is used until the number of hits of B is 10, and then randomly selects tips from C and A again.
But at this point, if the user edits the file like below:

```json
[
  { "name": "B", "hitted": 10 },
  { "name": "A", "hitted": 1 },
  { "name": "C", "hitted": 1 }
]
```

In current status of impl, the C will be selected. because of `iter::max()` about `[A, C]` (
I did not consider the above situation.)

Except for these situation, it is randomly selected.

default-plugins/status-bar/src/tip/cache.rs
```diff
@@ -0,0 +1,105 @@
...
+#[derive(Debug)]
+pub struct Cache {
+    pub path: PathBuf,
+    pub data: Vec<CachedTip>,
+}
+
+#[derive(Debug, Serialize, Deserialize, Eq, PartialEq, Hash, Clone)]
+pub struct CachedTip {
+    name: String,
+    hitted: usize,
```
a-kenji, line 20: I think `hit`, or `hits` would be cleaner here instead of `hitted`.

jaeheonji: Thank you :) beside this.. there must be many strange name. so i will check all the code again!

zellij-server/src/wasm_vm.rs
```diff
@@ -210,6 +210,8 @@ fn start_plugin(
         .unwrap()
         .map_dir("/data", plugin_own_data_dir.as_path())
         .unwrap()
+        .map_dir("/tmp", ZELLIJ_TMP_DIR.as_path())
+        .unwrap()
```
jaeheonji, line 214: @a-kenji Hi! While implementing the cool method you pointed out, I ran into a couple of situations.

In order for the plugin to access the local file system of the host machine, it must map the host directory to wasi_env. Therefore, in order to read the `VERSION` file, it seems to be necessary to map the user's `$HOME` directory.

This method is not bad. but if #894 is applied, it seems that the feature can be implemented very simply.

So, personally, I want to continue this work after #894 is merged. Coud I hear your opinion?

a-kenji: Oh yeah, that is true. I think #894 looks good to me,
but I would like to have someone with a little more experience in the plugin system look over it as well.

I think it is a good idea, since that function would simplify the work here.

jaeheonji:
Hi! @imsnif @a-kenji 

I've improved some features, and finished refactoring.

> In my idea, the C will be selected. because of iter::max() about [A, C] (
I did not consider the above situation.)
Except for these situation, it is randomly selected.

Also in this part, I modified the Tip to be randomly selected.
Could you please check this again? 🙇

zellij-server/src/wasm_vm.rs
```diff
@@ -212,6 +212,8 @@ fn start_plugin(
         .unwrap()
         .map_dir("/data", plugin_own_data_dir.as_path())
         .unwrap()
+        .map_dir("/tmp", ZELLIJ_TMP_DIR.as_path())
```
a-kenji, line 215: Maybe it would be good to support STATE/CACHE dirs in the future?
For now I would be OK with this.
But probably something we want to restrict again at some point?
cc @TheLostLambda

a-kenji (approved):
Thank you for this awesome work!
This does look good from my side.

a-kenji:
@jaeheonji
Thank you so much for this awesome work!

Issue #196, opened 2021-02-19
Feature: move current pane to new tab


imsnif:
Similar to the tmux `break-pane` command.

Monkeyanator:
Anybody working on this one? If not I'd like to give it a try :)

a-kenji:
@Monkeyanator Awesome! 
Afaik, noone is working on it, so you should be free to to tackle it.
Please feel free to ask any Questions you have!

imsnif:
I just want to mention that I really miss this feature in my day-to-day usage of Zellij. I unfortunately don't have time to devote to it myself right now, but @Monkeyanator - if you still have time for this and I can do anything to guide or help you, I'd be happy to.

prscoelho:
Taking a look at this, seems like we can reuse the Tab::close_pane function by having it return the Pane, and it's good because that takes care of the shifting around of space that should happen when a pane is removed from the tab.

However I'm unsure how to go about creating a new tab with a specific Pane. We want to create a new tab with the provided layout but without any terminal panes, right?

imsnif:
I haven't looked deeply at the code, but to give you a direction, I'd try to create an additional constructor for `Tab` that would allow this (or use the builder method as something like Tab::new(..).with_pane(pane)` if you want to be more rusty).

Getting the pane from `close_tab` sounds promising, but I'd be a little worried about doing this every time even when we don't need it. Perhaps we can duplicate the logic somehow?

wviana:
Started using Zellij in my flow. Just missed this feature. Move a panel into its own tab.
Any updates on that?

imsnif:
Implemented finally! Will be released in the next version.

imsnif:
Incidentally, thanks @tlinford for laying down the groundwork for this.

utkarshgupta137:
This is not visible in the documentation, had to search github to find it!

JimitSoni18:
can someone please tell what keybind does this

PR #853, opened 2021-11-10
refactor(tab): close_pane returns closed pane

tlinford:
Prepare for break pane command (#196) by making close_pane return the closed
pane. Also split the function up to hopefully be more readable.


TheLostLambda (approved):
Looks good to me! Nice to have a bit of refactoring done as well! The tab file definitely has a lot of duplication and every bit of cleanup helps!

PR #855, opened 2021-11-10
add(feature): toggle boolean options with cli flag

a-kenji:
add the ability to toggle boolean options with a cli flag:

example:
    if the pane frames are turned off in the config file,
    then passing in the `--no-pane-frames` flag will toggle the
    pane frames on

### DIFF
--- a/src/commands.rs
+++ b/src/commands.rs
@@ -176,7 +176,7 @@ pub(crate) fn start_client(opts: CliArgs) {
     })) = opts.command.clone()
     {
         let config_options = match options {
-            Some(SessionCommand::Options(o)) => config_options.merge(o),
+            Some(SessionCommand::Options(o)) => config_options.merge_from_cli(o.into()),
             None => config_options,
         };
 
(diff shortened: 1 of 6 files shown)


jovandeginste:
Am I reading this right that the flag's behavior depends on the current configuration value? If so, would it not make more sense to have two flags, that each override whatever is in the configuration? Ie. `--pane-frames` and `--no-pane-frames`, or perhaps `--pane-frames=[true|false|1|0]` (depends on what is easiest to implement)

a-kenji:
@jovandeginste 
Thank you for the input!
Yes, you are reading it right.

The `--pane-frames true|false` would have been the easiest to implement,
I wasn't sure if that was a good UX. Since I like having it as a flag, but also as a specific option. But maybe it would be better to make it more explicit.
What do you think?

a-kenji:
@jovandeginste 
On second thought, maybe it makes sense to rename the config option itself to:
```
pane_frames: true|false
```
that way it makes more sense having a cli option that has:
```
--pane-frames true|false
```
or mutual exclusive
```
--no-pane-frames
```

What do you think?

jovandeginste:
Most Linux tools use the `--flag`, `--no-flag` combo, which is a good one. Alternatives are fine too. However, i would never expect `--no-flag` to invert whatever is in the config file... It would disabled the feature "flag", irrespective of the configuration entry for "flag". I hope my explanation is not confusing... 😁

a-kenji:
@jovandeginste 
I think that makes sense.
Do you mind going over this behavior for a quick sanity check #859 ?

PR #857, opened 2021-11-10
doc: add INSTALL.md

jaeheonji:
A lot of people love Zellij ❤️  

Currently, people who like zellij are managing it through various `package managers`.
So I've put together the documentation to make it easily accessible to a variety of people.

There is still little content, but I believe this will grow over time.

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -19,7 +19,7 @@ To get the best experience either install nerdfonts, or use the simplified ui by
 
 ## Packages
 
-** :warning: These packages are not affiliated with the Zellij maintainers and are provided here for convenience.**
+ :warning: **These packages are not affiliated with the Zellij maintainers and are provided here for convenience.**
 
 [![Packaging status](https://repology.org/badge/vertical-allrepos/zellij.svg)](https://repology.org/project/zellij/versions)
 
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -1,0 +1,48 @@
+# Installation
+
+* [Cargo](#cargo)
+* [Packages](#package)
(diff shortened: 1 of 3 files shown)


imsnif:
Hey, this is great! Just a small request, can you add something like: "These packages are not affiliated with the Zellij maintainers and are provided here for convenience" somewhere near the top?

jaeheonji:
@imsnif Yes! you're right :) I added that phrase to the `packages` section.

a-kenji:
@jaeheonji 
We have an installation guide at:
https://zellij.dev/documentation/installation.html
do you think it would make sense move them together?

Maybe the one from the documentation here,
or the one from here to the documentation? 
And we could link to them?

jaeheonji:
@a-kenji 
Oh... sorry! I didn't consider its existence. then, I have an idea.

* First, rollback `README.md` and add link to `THIRD_PARTY_INSTALL.md` (need to create)
* `THIRD_PARTY_INSTALL.md` includes `pacakging status` and `installation method` using each package manager.
* And add link to `THIRD_PARTY_INSTALL.md` in [this section](https://github.com/zellij-org/zellij-org.github.io/blob/207002438c87e4be4c53e0a9789fbe338b64e5e9/docs/src/installation.md?plain=1#L66)

This will make it easier for people to contribute (added more package installation methods).
And contributors also seem to have easy document management.

How about this option?

a-kenji:
@jaeheonji 
No problem. I just want it to be in one place.
I like your suggestion!

jaeheonji:
I tried rewriting the README content :)

a-kenji (approved):
Thanks, looks good!

PR #859, opened 2021-11-10
fix(options): handling and overwriting cli opts

a-kenji:
* previously it was only possible to turn off certain features with a
  command line option, now it is possible to also overwrite this
  behavior in a sane way, for that some breaking changes happened:

  following options got renamed and inverted:
  ```
  disable_mouse_mode -> mouse_mode
  no_pane_frames -> pane_frames
  ```

  following cli options got added:
  ```
  mouse-mode [bool]
  pane-frames [bool]
  simplified-ui [bool]
  ```

  the following cli flag got removed:
  ```
  simplified-ui
  ```

  They can be specified in the following way:
  ```
  zellij options --mouse-mode true
  ```
  in order to enable the mouse mode, even if it is turned off in the
  config file:
  ```
  mouse_mode: false
  ```

  The order is now as follows:
  1. corresponding flag (`disable-mouse-mode`)
  2. corresponding option (`mouse-mode`)
  3. corresponding config option (`mouse_mode`)

### DIFF
--- a/zellij-client/src/input_handler.rs
+++ b/zellij-client/src/input_handler.rs
@@ -63,7 +63,7 @@ impl InputHandler {
         let mut err_ctx = OPENCALLS.with(|ctx| *ctx.borrow());
         err_ctx.add_call(ContextType::StdinHandler);
         let alt_left_bracket = vec![27, 91];
-        if !self.options.disable_mouse_mode.unwrap_or_default() {
+        if self.options.mouse_mode.unwrap_or(true) {
             self.os_input.enable_mouse();
         }
         loop {
--- a/zellij-server/src/screen.rs
+++ b/zellij-server/src/screen.rs
@@ -560,7 +560,7 @@ pub(crate) fn screen_thread_main(
     config_options: Box<Options>,
 ) {
     let capabilities = config_options.simplified_ui;
-    let draw_pane_frames = !config_options.no_pane_frames.unwrap_or_default();
+    let draw_pane_frames = config_options.pane_frames.unwrap_or(true);
 
     let mut screen = Screen::new(
         bus,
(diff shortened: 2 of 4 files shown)


jovandeginste:
The last part is a bit confusing me. I think what you are saying, is:

`--mouse-mode true --no-mouse-mode`: mouse mode is disabled

Correct? Even when the flags are reversed in order?

Flags should always have precedence over configuration files, so that's good! In case of multiple flags controlling the same behavior (or repeated flags), the expected behavior is less predictable, but usually "last flag wins". Is such a thing doable without a large refactor? Alternatively, an error is also acceptable... (Conflicting flags)

As a side note, is it possible to make `true` implicit (but allowed as explicit value)?

By this, i mean that  `--mouse-mode=true`, `--mouse-mode true` and `--mouse-mode` are all valid and do the exact same thing, enable mouse mode.

Respectively, you can also use: `--mouse-mode=false`, `--mouse-mode false` and `--no-mouse-mode` to disable mouse mode.

jovandeginste:
Oh, and to be clear: I have zero experience in rust; every time I look at rust code, my head starts spinning 😁

Just saying, I'm basing my reply on your text, not your code. If you want a once-over of your code, someone else should really do it...

a-kenji:
@jovandeginste 
Thanks for the input,
and no worries I just wanted to hear what you had to say about the behavior!

```
--mouse-mode true --no-mouse-mode: mouse mode is disabled
```
> Correct? Even when the flags are reversed in order?

Yes, that is correct.

>  In case of multiple flags controlling the same behavior (or repeated flags), the expected behavior is less predictable, but usually "last flag wins". Is such a thing doable without a large refactor?

Yes, that is possible, 
currently not without a huge change to how we parse the cli options.
We are waiting on this front for a new version of a library that is still in a beta
version currently - once we move over it should be a single option we specify.

Similar goes for the implicit action.


> Alternatively, an error is also acceptable... (Conflicting flags)

I think that is reasonable, thanks!

jovandeginste:
Good luck 🤞

Issue #877, opened 2021-11-17
[BUG] text selection very slow with large output


markus-wa:
**In Case of Graphical or Performance Issues**

> Please attach the files that were created in `/tmp/zellij-1000/zellij-log/` to the extent you are comfortable with.

[zellij-debug.tar.gz](https://github.com/zellij-org/zellij/files/7554004/zellij-debug.tar.gz)

**Basic information**

`zellij --version`: 20.1
`tput lines`: 51
`tput cols`: 104
`uname -av` or `ver`(Windows): 
```
Linux mark-ThinkPad-X1-Extreme-2nd 5.11.0-40-generic #44~20.04.2-Ubuntu SMP Tue Oct 26 18:07:44 UTC 2021 x86_64 x86_64 x86_64 GNU/Linux
```

List of programs you interact with as, `PROGRAM --version`:
happens in both alacritty and gnome-terminal - neither of them has this issue when not using `zellij`
`alacritty --version`: `alacritty 0.9.0`
`gnome-terminal --version`: `# GNOME Terminal 3.36.2 using VTE 0.60.3 +BIDI +GNUTLS +ICU +SYSTEMD`

**Further information**
```
$ zellij
$ while :; do; tr -dc A-Za-z0-9 </dev/urandom | head -c 200 ; echo ''; done
wait a while
CTRL+C

select some text - it will be very slow
```

another quicker way to repro is usually just a simple `rg a` as it's not as slow as generating random data - but that would have included sensitive stuff in the debug log
```
$ zellij
$ rg a
CTRL+C

select some text - it will be very slow
```

it gets really bad at around 50k lines

imsnif:
Thank you!! I have been trying to reproduce this one for a short while. @tlinford - any ideas?

markus-wa:
no problem :smile: 

some more notes:
- it affects all panes
- exiting the pane with tons of output solves the issue for other panes :slightly_smiling_face: (but I would expect zellij to handle these volumes without a slowdown, as alacritty and gnome-terminal on their own can handle it fine)

tlinford:
@imsnif I'm thinking it may not be directly related to mouse mode, running `rg a` shows the same problem also with mouse mode off.
Seems like at a certain point the scrollback stops being capped at 10000 lines for some reason, and after that scrolling is really slow.

imsnif:
[...]

PR #881, opened 2021-11-18
keep track of scrollback buffer size

tlinford:
initial attempt at solving #877 

missing handling of when the viewport gets switched to alternative_viewport

### DIFF
--- a/zellij-server/src/panes/grid.rs
+++ b/zellij-server/src/panes/grid.rs
@@ -606,9 +606,12 @@ impl Grid {
             };
 
             let dropped_line_width = bounded_push(&mut self.lines_above, line_to_push_up);
+            let dropped_line_height = calculate_row_display_height(dropped_line_width, self.width);
 
-            self.scrollback_buffer_lines -=
-                calculate_row_display_height(dropped_line_width, self.width);
+            self.scrollback_buffer_lines = self
+                .scrollback_buffer_lines
+                .saturating_sub(dropped_line_height);
+
             transfer_rows_from_lines_below_to_viewport(
                 &mut self.lines_below,
                 &mut self.viewport,
--- a/zellij-server/src/panes/grid.rs
+++ b/zellij-server/src/panes/grid.rs
@@ -599,12 +599,6 @@ impl Grid {
     pub fn scroll_down_one_line(&mut self) {
         if !self.lines_below.is_empty() && self.viewport.len() == self.height {
             let mut line_to_push_up = self.viewport.remove(0);
-            log::info!("line_to_push_up: {:?}", line_to_push_up);
-            log::info!(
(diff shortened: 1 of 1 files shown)


zellij-server/src/panes/grid.rs
```diff
@@ -170,6 +177,7 @@ fn transfer_rows_from_viewport_to_lines_above(
             viewport.insert(0, row);
         }
     }
+    transferred_rows_height
```
imsnif, line 180: We need to account for the `bounded_push` above somehow. There's a chance that it drops a line if it overflows the buffer... maybe also return a value from it?

zellij-server/src/panes/grid.rs
```diff
@@ -557,6 +575,8 @@ impl Grid {
     pub fn scroll_down_one_line(&mut self) {
         if !self.lines_below.is_empty() && self.viewport.len() == self.height {
             let mut line_to_push_up = self.viewport.remove(0);
+            self.scrollback_buffer_lines +=
+                (line_to_push_up.width() as f64 / self.width as f64).ceil() as usize;
             if line_to_push_up.is_canonical {
                 bounded_push(&mut self.lines_above, line_to_push_up);
```
imsnif, line 581: We also need to account for bounded_push here, it might drop lines

imsnif (commented):
Looks like this turned out to be like a fair bit of work! Happy to see this though.
I have not tried this and only glanced over the code. Left some comments. I trust that you went through the basic smoke tests to make sure this works... only thing I can add (if you didn't try) is to make sure this doesn't slow pane resizing too much when the scrollbuffer is very full.

About alternative grid: I think we can totally ignore that use case (as long as it doesn't break the main one). Ideally scrolling should be completely disabled when the alternative grid is set (both in the UI and functionality).

tlinford:
i should have addressed the bounded push cases.
perfomance on resize seems to be unchanged.
the alternative grid case does need handling i think, seems to be used at least by nvim :)

imsnif:
What happens right now with the alternative grid, in this current state of affairs?

tlinford:
well i noticed something like this can happen. where you keep scrolling:
nvim-scrolled

this seems to happen regardless of these changes though.

imsnif:
Yeah, this is what I meant. We should ideally not allow scrolling when in alternative grid mode. So if this change leaves things as they are, I think it's fine.

tlinford:
@imsnif  I've handled a couple more edge cases, and am seeing numbers that match what we have in main.
Performance while resizing seems comparable.
I hope I haven't missed anything, let me know if you think this is good to go.

imsnif (approved):
Great work! Thanks for doing this. This LGTM! (only thing is to remove the log messages :) )

PR #894, opened 2021-11-23
feat: provide the zellij version to plugins

### LINKED ISSUE #888: Feature: give plugins version information on load
But in this case, the most important thing that how to get the zellij version in plugin.

The thought in my head, when wasmer invoke the function of `load`, pass the version information to plugin. however, it is occured all plugins to receive unnecessary data.

_Originally posted by @jaeheonji in https://github.com/zellij-org/zellij/issues/848#issuecomment-976309075_

jaeheonji:
resolves #888

### DIFF
--- a/zellij-server/src/wasm_vm.rs
+++ b/zellij-server/src/wasm_vm.rs
@@ -238,6 +238,8 @@ fn start_plugin(
     let start = instance.exports.get_function("_start").unwrap();
 
     // This eventually calls the `.load()` method
+    // TODO: pass the zellij version to the plugins.
+    // zellij version can be obtained from [`zellij-utils::consts`] by calling [`zellij-utils::consts::VERSION`]
     start.call(&[]).unwrap();
 
     (instance, plugin_env)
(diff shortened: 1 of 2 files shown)


jaeheonji:
I did some research on this feature.
The initial (**my stupid**) idea was to provide the zellij version when wasmer invokes the plugin's load function.

For some reason I changed the implementation direction.

1. The data type that can be provided as a plugin through wasmer's call is only integer. Of course other types are possible with "slightly" complicated methods.
2. Already in the `wasm_vm.rs` there is a nice way called `plugin API` !!

So I made a pluginAPI that simply outputs the zellij version. The plugin only needs to call the `get_zellij_version` function.

a-kenji (approved):
LGTM, but would like to have a 2nd pair of eyes here.

TheLostLambda:
I think we could improve the code here a little, let me experiment for a moment!

TheLostLambda:
Sorry for the delay in getting back to you!

I think we can reuse the `object_from_stdin()` function we already have, since `String` is deserializable!

Here is the diff that worked for me!

image

TheLostLambda:
Otherwise things look stellar! Thanks for doing this @jaeheonji !

jaeheonji:
@TheLostLambda Thanks for reviewing! I applied the part you suggested :)

a-kenji (approved):
Awesome, thank you!

Issue #797, opened 2021-10-21
Flakey Scrolling Issue


markus-wa:
Video of the bug:
https://user-images.githubusercontent.com/5138316/138323913-ee4e48ae-9d7c-408a-a0b6-92aa87b524f3.mp4

I keep randomly getting the scrolling issue in the attached video at the bottom. Basically after some large outputs the scrolling stops working properly and it skips a huge chunk of output.

In the video I use my cursor to mark the gap in output - you can see the git status from my earlier command is just cut off in the middle.

Note that it seems to cut off at `tput lines` right after the issue happens but after resizing or running additional command scrolling still just cuts off at the same lines (i.e. not at the resized `tput lines`)

**In Case of Graphical or Performance Issues**

I don't know how to reproduce this bug exactly, but it keeps happening to me every day
If I manage to get the debug info I will attach it afterwards

1. Delete the contents of `/tmp/zellij-1000/zellij-log`, ie with `cd /tmp/zellij-1000/` and `rm -fr zellij-log/`
2. Run `zellij --debug`
3. Recreate your issue.
3. Quit Zellij immediately with ctrl-q (your bug should ideally still be visible on screen)

Please attach the files that were created in `/tmp/zellij-1000/zellij-log/` to the extent you are comfortable with.

**Basic information**

`zellij --version`: `zellij 0.19.0`
`tput lines`: 28 / 71
`tput cols`: 86 / 282
`uname -av`: `Linux mark-ThinkPad-X1-Extreme-2nd 5.11.0-38-generic #42~20.04.1-Ubuntu SMP Tue Sep 28 20:41:07 UTC 2021 x86_64 x86_64 x86_64 GNU/Linux`

List of programs you interact with as, `PROGRAM --version`: output cropped meaningful, for example:
`alacritty --version`: `alacritty 0.9.0`
`zsh --version`: `zsh 5.8 (x86_64-ubuntu-linux-gnu)`

**Further information**
I'm not sure how to reproduce it

Thanks for making Zellij :heart:

raphCode:
I also hit this bug occasionally without any clue on what it might cause.

Can we record the terminal session with `script` until we hit the bug and then try to replay it?

How can we start a `script` session in each pane opened?
I naively tried `SHELL=script zellij` and `zellij options --default-shell script` and it didn't work...

jovandeginste:
[...]

PR #915, opened 2021-12-01
fix(compatibility): full screen scroll region

imsnif:
Fixes https://github.com/zellij-org/zellij/issues/797

This was happening because when the scroll region is set to the full viewport, it should send lines to the scrollbuffer rather than swallowing them.

In our codebase, we treat the scroll region as an `Option` - and this is not entirely correct according to https://vt100.net/docs/vt510-rm/DECSTBM.html

At some point we should change the scroll region to always be there and default to the first and last line of the viewport. This is however quite a big change, so for now I solved this case with a conditional.


Lillecarl:
Good work, not that anyone's asking but I'd say this deserves a release on it's own, scrollback is quite essential 😄

imsnif:
Definitely planning on it for next week, sorry - I'm just quite swamped at the moment.
Now that I know what the problem is, as a temporary workaround you can do `echo -e "\033[r"` whenever this happens. If for some reason that doesn't work, you could also do "reset" - the latter will kill the current scrollback but at least you won't have to restart the pane.

mhristache:
I am still getting this issue with zellij 0.40.1, not sure what the trigger is but I mainly get into it when using ssh sessions (zellij running on my machine from which I ssh to other machines). `reset` is fixing it, `echo -e "\033[r"` doesn't

PR #936, opened 2021-12-13
Feature: Configurable scroll buffer

### LINKED ISSUE #240: Feature: Limit the scrollback buffer by default, make it configurable.
Currently the scrollback buffer is unlimited.
Change it to a sensible default. Maybe 10k lines?
Make it a configuration option to set it either to unlimited, or to a different number.

kunalmohan:
resolve #240

### DIFF
--- a/zellij-utils/assets/config/default.yaml
+++ b/zellij-utils/assets/config/default.yaml
@@ -392,3 +392,10 @@ plugins:
 #   - true (default)
 #   - false
 #mouse_mode: false
+
+# Configure the scroll back buffer size
+# This is the number of lines zellij stores for each pane in the scroll back
+# buffer. Excess number of lines are discarded in a FIFO fashion.
+# Valid values: positive integers
+# Default value: 10_000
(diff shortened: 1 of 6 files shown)


zellij-utils/assets/config/default.yaml
```diff
@@ -392,3 +392,10 @@ plugins:
 #   - true (default)
 #   - false
 #mouse_mode: false
+
+# Configure the scroll back buffer size
+# This is the number of lines zellij stores for each pane in the scroll back
+# buffer. Excess number of lines are discarded in a FIFO fashion.
+# Valid values: positive integers
+# Default value: 10_000
```
a-kenji, line 400: Our `yaml` parser seems to have problems with this way of specifying integers, it seems to work well with `10000` for me.

a-kenji (approved):
LGTM, it works well, I tried a few different buffer sizes.

The only thing is that you can't specify large numbers like that.

Issue #804, opened 2021-10-25
ZelliJ crashes when run at home


duongdominhchau:
## Description

Don't laugh at my silly title, it actually happens. You can see it here:

https://user-images.githubusercontent.com/56427890/138713571-0c4b4600-30a6-4d9f-bff1-cbe974ed71fd.mp4

## Information

- `zellij --version`: `zellij 0.19.0`
- `tput lines`: `40`
- `tput cols`: `159`
- `uname -av`: `Linux chau-laptop 5.14.14-arch1-1 #1 SMP PREEMPT Wed, 20 Oct 2021 21:35:18 +0000 x86_64 GNU/Linux`
- `/tmp/zellij-1000/zellij-log/zellij.log`:

```
INFO   |zellij_client            | 2021-10-25 21:19:29.051 [main      ] [zellij-client/src/lib.rs:107]: Starting Zellij client!
INFO   |zellij_server            | 2021-10-25 21:19:29.066 [main      ] [zellij-server/src/lib.rs:178]: Starting Zellij server!
INFO   |zellij_server::wasm_vm   | 2021-10-25 21:19:29.120 [wasm      ] [zellij-server/src/wasm_vm.rs:74]: Wasm main thread starts
```

har7an:
Hello,

does this still occur with current releases of zellij?

duongdominhchau:
Sorry, I forgot to update on this one. Right now I'm using zellij 0.31.4 and this problem does not happen anymore.

PR #950, opened 2021-12-17
chore: upgrade wasmer, wasmer-wasi

ahkrr:
I resolved a dependency conflict with `darwin-libproc` (because of its requirement of memchr = "~2.3") by depending on `darwin-libproc-sys` directly and adding the 2 necessary functions from darwin-libproc to zellij.
A pull request was opened 6 months ago for this issue with `darwin-libproc` but no response was made since then. 
I ran `cargo test --all --all-features` and everything passed. 
I haven't figured out yet how to run the tests in the docker container.

### DIFF
--- a/zellij-server/Cargo.toml
+++ b/zellij-server/Cargo.toml
@@ -18,8 +18,8 @@ daemonize = "0.4.1"
 serde_json = "1.0"
 unicode-width = "0.1.8"
 url = "2.2.2"
-wasmer = "1.0.0"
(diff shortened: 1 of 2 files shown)


ahkrr:
After upgrading wasmer, 'executing zellij -l strider' in the home directory doesn't crash anymore (at least for me). 
This may be a way to resolve https://github.com/zellij-org/zellij/issues/804.

PR #953, opened 2021-12-18
chore: upgrade wasmer, wasmer-wasi

ahkrr:
I resolved a dependency conflict with darwin-libproc (because of its requirement of memchr = "~2.3") by depending on darwin-libproc-sys directly and adding the 2 necessary functions from darwin-libproc to zellij.
A pull request was opened 6 months ago for this issue with darwin-libproc but no response was made since then.

I use linux and I couldn't get `darwin-libproc-sys` to typecheck because it only gets compiled on macos. 
I hope I didn't miss anyhting, and it compiles without problem on macos.

### DIFF
--- a/zellij-server/Cargo.toml
+++ b/zellij-server/Cargo.toml
@@ -18,8 +18,8 @@ daemonize = "0.4.1"
 serde_json = "1.0"
 unicode-width = "0.1.8"
 url = "2.2.2"
-wasmer = "1.0.0"
-wasmer-wasi = "1.0.0"
+wasmer = "2.1.0"
+wasmer-wasi = { version = "2.1.0", features = ["enable-serde"] }
 cassowary = "0.3.0"
 zellij-utils = { path = "../zellij-utils/", version = "0.22.1" }
 zellij-tile = { path = "../zellij-tile/", version = "0.22.1" }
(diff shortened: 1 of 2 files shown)


imsnif:
Hey - while I very much appreciate the work you've put into this, I must admit I'm very wary of this change.
This is quite an untested part of the code base that has the potential to cause a lot of havoc.

So before proceeding with this, I'd like to understand: what problem are we solving here? What's the benefit Zellij users (or developers) will get from this upgrade?

ahkrr:
The last update to `wasmer 1.x.x` was `1.0.2 Feb 5, 2021` since then `wasmer 2.1.0` was released. My thought process was, that zellij would probably not stick with 1.0.2 perpetually. So I looked into what was needed to upgrade this dependency. 
After going to 2.1.0 I saw cargo dependency resolution conflicts because of darwin-libproc and its strict dependency on `memchr = ~2.3`. 
I looked into the use of darwin-libproc in zellij and noticed that only one function was used and wondered what it would take to no longer depend on darwin-libproc directly. 
The result is this pull-request with 3 functions taken from darwin-libproc and a dependency on darwin-libproc-sys.
If this change is unwanted, this PR can be closed.
If zellij wants to upgrade wamser in the future, this problem will probalby come up again, as long as darwin-libproc doesn't bump its dependency on memchr or this dependency resolution conflict only happen for me for some kind of reason.

ahkrr:
The changelog for wasmer is [changelog](url)

imsnif:
So - I went through the changelog and I don't think there's anything there we particularly need. Also, it has one or two breaking changes that will make me lose sleep :)

Again, thanks for your work on this, but unless there's a severe compatibility or security issue that we need to address - I think this ain't broke. I'll close this for now and would be happy to revisit it when the need arises.

PR #958, opened 2021-12-20
feat: add focus attribute in layout

### LINKED ISSUE #852: Feature: add focus attribute to the layout
Add a `focus` or `focused` attribute to the layout, to specify which tab, or pane should have the focus,
once the layout is fully loaded.

### LINKED ISSUE #942: Feature: keep state of focusd pane in each tabs
`zellij` currently does not keep the state of the previously focused pane like below:

[asciicast](https://asciinema.org/a/DZi8hSWJYNtk4xruxQ1SUfmTV)

Even if we move tabs, if the previous state is maintained, it will be more convenient when a complex environment is configured.

One thing to consider is that `zellij` can move tabs with `alt+hjkl`, so this is actually effective when we use the `mouse` or `ctrl-t` to move tabs.

jaeheonji:
This PR resolve #852 , resolve #942

### Description

The `focus` attribute is added to the layout like this:
```yaml
---
template:
  ~
tabs:
- direction: Vertical
- direction: Vertical
  focus: true # this
  parts:
    - direction: Vertical
      split_size:
        Percent: 50
    - direction: Vertical
      focus: true # this
      split_size:
        Percent: 50
```

First, each `TabLayout` has a `focus` attribute. The focus value of the top-level TabLayout determines "which `tab` to focus on."
The sub-level TabLayout refers to the `pane` of each tab, it determines "which `pane` to focus on".

If focus attribute is duplicated, tab or pane with the first declared value is focused.

Also, to keep the state of the last focused pane within each tab, the `Tab` struct has a `last_active_pane_id` variable.
This saves the state of the focused pane from the layout, and at the same time preserves the state of the previous pane when the tab is moved. (Previously, it was initialized to the first (Index 0) pane)

I declared the `last_active_pane_id` variable as an `Option` in case there was any problem with the system. If it's no problem, I will remove the `Option` type.

#### Other thing..

To focus the tab, internally call `GoToTab` Instruction after all layouts are loaded.
When zellij starts with multiple tabs layout, it currently has the following process.

```
newTab -> render -> newTab -> render -> newTab -> render ... -> GoToTab -> render
```

Of course, this is not a big problem but,

```
newTab -> newTab -> newTab -> ... -> GoToTab -> render
```

It might be a good idea to render only once, **only when the program is started**. Of course this is a different issue.
 

**EDIT: I will update the test-related errors when this feature is determined to be ok.**


djpate:
That's exactly what I need. +1

djpate:
If this was passing specs, would you consider merging this? @imsnif @tlinford

imsnif:
If it passes CI and @jaeheonji thinks it's good, I'm happy.

jaeheonji:
@djpate Thanks for your interest!

@imsnif The original concept for this feature was presented by @a-kenji, so I respect his opinion and await a review.
However, since he was already agreed on the overall spec (although review has not been done yet), I'll merge this features after CI and strict testing.

As an aside, I'm currently busy until this week, so please be patient if the work is a little late 🙏

imsnif:
@jaeheonji - there is no rush! Please take your time.

zellij-server/src/pty.rs
```diff
@@ -43,6 +44,7 @@ pub(crate) enum PtyInstruction {
     SpawnTerminalVertically(Option<TerminalAction>, ClientId),
     SpawnTerminalHorizontally(Option<TerminalAction>, ClientId),
     UpdateActivePane(Option<PaneId>, ClientId),
+    FocusOnTab(TabIndex, ClientId),
```
a-kenji, line 47: Maybe `FocusTab` would be sufficient here? I am not sure.

zellij-server/src/pty.rs
```diff
@@ -114,6 +117,12 @@ pub(crate) fn pty_thread_main(mut pty: Pty, layout: Box<LayoutFromYaml>) {
             PtyInstruction::UpdateActivePane(pane_id, client_id) => {
                 pty.set_active_pane(pane_id, client_id);
             }
+            PtyInstruction::FocusOnTab(tab_index, client_id) => {
+                pty.bus
+                    .senders
+                    .send_to_screen(ScreenInstruction::GoToTab(tab_index, Some(client_id)))
```
a-kenji, line 123: Have both Instructions named similarly? They seem to be doing the same thing? So maybe
`PtyInstruction::FocusOnTab` and `ScreenInstruction::FocusOnTab` ?

a-kenji (approved):
Sorry that it took this long.
This is awesome!
LGTM!

zellij-server/src/pty.rs
jaeheonji: Thanks for the comment!
Yes, it does the same thing. It's just simple wrapper instruction.

> Maybe FocusTab would be sufficient here? I am not sure.

Similar to the feedback above, I thought it would be good to rename the Instruction.
Instead of `FocusOnTab`, how about unifying with `GoToTab`? (like `ScreenInstruction::GoToTab`)

a-kenji: I like it. That is a good idea!

jaeheonji:
**History Note:**

I did not fully implement the function I initially expected.

In particular, the ability to save the state of the focused `Pane` in each `Tab` is not implemented at this time due to several problems.

(Because I found a conflict with the multiple user function and an unexpected bug internally. This should be managed as a separate issue.)

So, I only implement focusing on specific `Tab` and `Pane` through Layout.

PR #959, opened 2021-12-20
fix: update switch to tab bug in multiple users

jaeheonji:
Attach users cannot change tabs by clicking the mouse.
Because `tab-bar` plugin call the `host_switch_tab_to` function which it does not return `client_id`.

Since each client passes the `client_id` in the `PluginEnv` structure, I added it.


imsnif (approved):
Great! I totally missed this one :D Thanks!

PR #962, opened 2021-12-21
add simplified example with usage instructions

matu3ba:
justification:
- 1. ca. 200 LOC vs ca. 350 LOC
- 2. 1 common prefix instead of 2
- 3. 2 ways instead of 4 to get back to normal mode (Esc+Enter)
- 4. mode switching only from and to normal mode
- 5. remove 4 modes without loss of functionality
- 6. second key for mappings to switch to mode memorable

### DIFF
--- a/example/simpleconfig.yaml
+++ b/example/simpleconfig.yaml
@@ -1,0 +1,235 @@
+###########################################################
+####  ####  Simplified configuration for Zellij  ####  ####
+###########################################################
+# - 1 prefix key (Alt) for most common actions in normal mode
+# - less common actions require switching mode or are removed
+# - switching back to normal mode with escape and enter
+# - Alt-q for quitting and Alt-d for detaching
+
+# To access other modes from normal mode press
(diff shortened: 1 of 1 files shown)


matu3ba:
TODO
1. link to the nonexisting wiki (we had 2 issues with easy to workaround keymap problems the last 2 days)
2. clarify, if the locked mode should stay or not (I believe that it is nice, but not part of simple)

imsnif:
Hey @matu3ba - sorry it took me some time to get to this. I'm good with adding this as an additional example. But maybe change the name to something less, erm... judgemental? :) Like not "simple" config, but maybe "ALT-centered" config or some such? Like, what differentiates this config in your mind? What do you think?

imsnif:
Also, I'm not sure what you mean by point 1 in the TODO... We have a documentation website, it even has a troubleshooting section. Would you like to add stuff there, or? What are you referring to?

matu3ba:
superseded by #989

PR #972, opened 2021-12-24
ensure clippy runs on all targets

danieleades:
clippy is not currently configured to run on tests or examples when using cargo make.

This PR addresses that


tlinford:
Looks good to me, thanks!

PR #977, opened 2021-12-31
fix: left click on tabs when pane is on fullscreen mode

### LINKED ISSUE #944: Changing tabs via mouse click does not work when pane is in fullscreen
The title basically says it all :D
- open two tabs
- open two panes in one tab
- put one pane in fullscreen
- clicking on the tab bar does not change tabs anymore

Using Ctrl-T and arrows does still work although.

dswij:
closes #944

### DIFF
--- a/zellij-server/src/tab.rs
+++ b/zellij-server/src/tab.rs
@@ -1,8 +1,8 @@
 //! `Tab`s holds multiple panes. It tracks their coordinates (x/y) and size,
 //! as well as how they should be resized
 
+use zellij_utils::position::{Column, Line};
 use zellij_utils::{position::Position, serde, zellij_tile};
-use zellij_utils::position::{Line, Column};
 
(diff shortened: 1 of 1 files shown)


dswij:
I was playing around with the idea that handling mouse events should start at `Screen` instead of `Tab`, but I realized that there'll be a lot of effort for this, but I'm not sure if this is a better way for the moment.

imsnif (approved):
This looks great! Sorry it took me some time to get to this.
I very much agree with you about the tab/screen relationship. I'm doing some work on breaking stuff out now as well. I'm going to merge a refactor PR later today or tomorrow which I think makes sense of some of these things, but there's definitely more work to be done in tidying up our internal logic if you're interested...

dswij:
Sure, I will gladly see what I can help. I will take a further look as soon as I have the time :slightly_smiling_face:

PR #981, opened 2022-01-03
Implement passing the bell signal to the terminal

### LINKED ISSUE #776: Feature: Terminal Notification/Highlighting
Imagine you have an app in a terminal and your attention is on another virtual desktop (say looking at your browser). An app can make the terminal get highlighted in your desktop environment or window manager so you know it needs attention. Is there a way that if you are running the app in zellij that zellij will do the same?

auronandace:
Closes #776.

Detects the bell signal from an application within zellij and passes it to the terminal.


imsnif (approved):
This is great, thanks for your work on this!

PR #989, opened 2022-01-03
add ALT-centered example with usage instructions

matu3ba:
justification:
- 1. ca. 200 LOC vs ca. 350 LOC
- 2. 1 common prefix instead of 2
- 3. 2 ways instead of 4 to get back to normal mode (Esc+Enter)
- 4. mode switching only from and to normal mode
- 5. remove 4 modes without loss of functionality(except adjusting titles)
- 6. second key for mappings to switch to mode memorable


imsnif (approved):
Cool, thanks for this example!

PR #999, opened 2022-01-10
feat: allow renaming session

### LINKED ISSUE #998: Feature: rename an existing session
First of all, thanks a lot for this great project!
Currently it's only possible to set the name when creating the session. The possibility to rename an existing session would be great though.

dswij:
closes #998 

This PR adds `rename-session` command to rename existing session to a new one

Usage:
```sh
zellij rename-session <existing-session-name> <new-session-name>
```


imsnif:
Looking forward to this feature!

dswij:
This appears to be a bigger change than I thought :smiling_face_with_tear:

DerFetzer:
This is a feature that I would like to see in zellij as well.
@dswij Is there something that I can do to support you?
What is holding this PR back? If it is your lack of time then I would like to see if I can push this further.

imsnif:
@DerFetzer - I think it's safe for you to work on this feature if you'd like, but note that the way things are set up right now, this is no small change!

Off the top of my head (there might be more I'm not thinking about) it would require:
1. Somehow renaming the IPC pipe (which uses the same session name), or breaking the connection between the IPC file name and the session name (and dealing with the fallout of all other commands who assume both are identical)
2. Creating a full path, including some basic UI (even if it's just "typing" where the session name is now like we do with renaming panes/tabs) for doing the actual name change.
3. Moving all connected clients to the new pipe from point 1

If you're up for it, that'd be awesome. I'd be happy to answer any questions. Just be sure you're coming into this with your eyes open :)

dswij:
@DerFetzer Yes, unfortunately, I haven't gotten too much time since I wrote this PR.

For anyone that's interested to work in this feature - feel free to take this up :).

Adding some info to @imsnif's pointers -- one of the issues here (at least at the time I wrote this PR) is that it's not trivial to change the session info in already-running shells inside Zellij. Of course, I think the current state of Zellij is significantly different from this PR, so this issue might not be valid anymore.

DerFetzer:
@dswij @imsnif Thank you very much for your hints and remarks! They are a good starting point.

@dswij The first step would be to rebase your branch to current main or do you suggest to start from scratch?

dswij:
@DerFetzer I would personally create a new PR since zellij has changed a lot. 
Points 1 and 2 were addressed in this PR though, maybe that can be an inspiration.

I will be closing this PR now 🙂

PR #1001, opened 2022-01-10
feat: change dependency for `process_cwd`

jaeheonji:
This issue related the `darwin-libproc` dependency.

The darwin-libproc library has not been updated for a long time, including dependency issues.
(reference: https://github.com/heim-rs/darwin-libproc/pull/3 and current problem is similar to this: https://github.com/zellij-org/zellij/pull/953#issuecomment-997219613)

So this PR simplifies the `get_cwd` function while using the new library [sysinfo](https://docs.rs/sysinfo/latest/sysinfo/).

This library solves dependency problems and supports various OSes.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -416,6 +416,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "core-foundation-sys"
+version = "0.8.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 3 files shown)


imsnif (approved):
@jaeheonji - this looks good! I'm not sure what caused the behaviour change with the tests... but if it passes consistently now, it's fine. I think it's only about which pane is focused on attach, which doesn't make a big difference. You can merge this whenever you feel ready.

PR #1015, opened 2022-01-19
Create tokyonightDark.yaml

Vanderscycle:
added tokyonight dark-storm-ish that I hacked from my config and wanted to share.

### DIFF
--- a/example/tokyonightDark.yaml
+++ b/example/tokyonightDark.yaml
@@ -1,0 +1,17 @@
+# not super color accurate but its my hack of tokyonight from https://github.com/enkia/tokyo-night-vscode-theme
+# changes are welcomed
+
+themes:
+  tokyonights:
+    gray: [86,95,137] #565f89
+    fg: [169,177,214]
+    bg: [26,27,38]
+    black: [65,72,104]
+    red: [247,118,142]
+    green: [158,206,106]
+    yellow: [242,191,0]
+    blue: [122,162,247]
+    magenta:  [187,154,247]
+    cyan: [42,195,222]
+    white: [213,214,219]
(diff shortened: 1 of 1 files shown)


jaeheonji:
Hi, @Vanderscycle!
Thank you for your contribution!

TokyoNight! This is one of my favorite themes too.
But in my personal opinion, I think adding only a specific theme to the `example` folder is different from the meaning of the folder.

However, it is also true that there are no examples related to themes. So, it seems like a good idea to rename the file to something like `custom-themes.yaml`.

And it would be even better if an example for 256color was added. (https://zellij.dev/documentation/themes.html)

jaeheonji:
Thanks for updating!. I've been busy for a while, so I'll just check it out now.

I have seen that some colors are custom. personally, I think it's good to make examples with colors that are often used in the community.

If you're okay with it, I'd like to merge it after adding some updates and a storm theme as well. How about this?

Vanderscycle:
Sounds good, I have been working on a tokyonight theme pr on highlight.js. I'll run another pass on the colors today as I had to make tweaks.

Vanderscycle:
Ready, can you have a test on your side for as a sanity check (I think that I have it right on my machine)
```  zellij options --theme tokyonightDark```

What it looks like on my machine
image

Vanderscycle:
I still use tmux because of tmuxinator as I haven't transferred my scripts over yet. So the irony isn't lost on me that I used tmux to run zellig haha

Vanderscycle:
Sorry I forgot to push my changes. Ready!

jaeheonji:
Really good! I'll merge it as soon as CI is finished. thank you for your contributing 😄

PR #1017, opened 2022-01-19
Update arg parsing to clap v3

tranzystorek-io:
Updates arg parsing to the recently released `clap` 3 with the `derive` feature (the new `structopt`).

Let me know if these should be changed:
- `clap` now colors its help output by default
- `structopt` used to propagate version to subcommands by default, `clap` doesn't


zellij-utils/src/setup.rs
```diff
@@ -376,16 +377,16 @@ impl Setup {
 
         Ok(())
     }
-    fn generate_completion(shell: String) {
-        let shell = match shell.parse() {
+    fn generate_completion(shell: &str) {
+        let shell: Shell = match shell.parse() {
```
jaeheonji, line 381: Since the `Shell` can only be in lower-case, it would be nice to have a part to convert it to lower-case.

```rust
let shell: Shell = match.shell.to_lowercase().parse() {
...
}
```

jaeheonji (approved):
This upgrade is very good!
I've tested and everything is fine. 

There is only one feedback for UX.
Thank you for your effort!

zellij-utils/src/setup.rs
tranzystorek-io: I wanted to automate this `parse` away by using `ArgEnum`, but since the `generate_completion` is a flag in a `Serializable` struct, it won't work...

It seems this used to be a separate subcommand before #503, is there any reason it was moved to a flag?

jaeheonji: Hmm... I don't know the history of this, maybe @a-kenji knows..?

tranzystorek-io: Ah, but AFAICS all the cli-related structs are serialized and sent to client, so `clap_complete::Shell` couldn't be used as an argument anyway without some redesign, so I'll just stick with this change for now 😅

PR #1019, opened 2022-01-20
Add option to default config for copy_command (PR #996)

JCallicoat:
Document the new --copy-command / copy_command option (#996) in the default config with examples.


tlinford:
Thanks, this will be useful, could you also add an example with pbcopy for mac?

JCallicoat:
Done!

tlinford (approved):

Issue #952, opened 2021-12-18
Extra Character Row for full Screen TUIs ("Scroll 0/1")


KyleKing:
**Description**

When launching a full screen TUI in Zellij, the screen itsefl scrolls, when it should have a 100% fill. This doesn't appear to be terminal emulator or library-specific because I tried TUI applications implemented in several languages ([bpytop](https://github.com/aristocratos/bpytop), [lazygit](https://github.com/jesseduffield/lazygit), htop, etc.) in iTerm2 and Alacritty

**Further information**

To reproduce, launch zellij, launch any fullscreen TUI (i.e. `lazygit` or `htop`), then attempt to scroll and you should see the whole interface move up or down one row when it instead should fill the screen and not move

Demo with asciinema:

[![asciicast](https://asciinema.org/a/457069.svg)](https://asciinema.org/a/457069)

Logs from testing with lazygit:

[zellij-11.log](https://github.com/zellij-org/zellij/files/7739241/zellij-11.log)
[zellij.log](https://github.com/zellij-org/zellij/files/7739242/zellij.log)

**Device information**

`zellij --version`: zellij 0.22.1
`tput lines`: 58
`tput cols`: 192
`uname -av` or `ver`(Windows): Darwin MacBook-Pro.local 19.6.0 Darwin Kernel Version 19.6.0: Sun Nov 14 19:58:51 PST 2021; root:xnu-6153.141.50~1/RELEASE_X86_64 x86_64
`alacritty --version`: `alacritty 0.9.0 (fed349a)`
`lazygit --version`: commit=, build date=, build source=homebrew, version=0.31.4, os=darwin, arch=amd64
(Reproducible in latest versions of iTerm2 or Kitty and with htop, bpytop, etc.)

**Experimentation**

I tried: `zellij options --no-pane-frames` and/or the most generic layout possible and the error persisted:

```yaml
---
template:
  direction: Horizontal
  parts:
    - direction: Vertical
      no_pane_frames: true
```

KyleKing:
This is a blocker for me, so tried to take a look in more depth

This appears to be introduced in Zellij 0.18 and isn't present in 0.17 or 0.16. I think it might be related to this change (#746), but I don't know Go or the codebase. This was the full diff between releases: https://github.com/zellij-org/zellij/compare/1868816791caca222f8a17ab96e3d514dc2ae79f...50452b53cbd9f50001a9662fba3faef393b73c8b

KyleKing:
[...]

PR #1025, opened 2022-01-26
fix(scroll): scroll pane by actual amount of rows

### LINKED ISSUE #1002: Zellij skips one extra line with Page-{Up,Down} when frames are enabled
**Basic information**

`zellij --version`: 0.24.0
`tput lines`: 57
`tput cols`: 237
`uname -av` or `ver`(Windows): 5.13.0-23-generic

**Further information**
1. start zellij, make sure frames are enabled (add panes as you wish, it remains reproducible)
2. print numbered lines (I did: `for i in {1..200}; do echo $i; done`); make sure you have >2x as many lines printed as you have lines on your screen
3. switch to scroll mode
4. use page-up / page-down, and see how one line (number) is skipped
    - op line on my screen says `147`
    - I press page-up
    - bottom line on my screen says `145`

Bonus:
5. disable frames
6. scroll works correctly (no skipped lines)

tlinford:
use pane content_rows to take account of frame enabled/disabled

fix #1002

### DIFF
--- a/zellij-server/src/tab/mod.rs
+++ b/zellij-server/src/tab/mod.rs
@@ -1722,7 +1722,7 @@ impl Tab {
                 .get_mut(&PaneId::Terminal(active_terminal_id))
                 .unwrap();
             // prevent overflow when row == 0
-            let scroll_rows = active_terminal.rows().max(1) - 1;
+            let scroll_rows = active_terminal.get_content_rows();
             active_terminal.scroll_up(scroll_rows, client_id);
         }
     }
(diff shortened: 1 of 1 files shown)


jaeheonji (approved):
I've seen issues related to this before. So I tested it through several programs (`bat-cat`, `btop`). The result is perfect 😄 

I think this can solve this issue as well (#952)

tlinford:
> I think this can solve this issue as well (#952)
Unfortunately that's another problem, but I'm making progress :)

jaeheonji:
> Unfortunately that's another problem

Oh.. your right, it's my mistake It looked similar, but it must be a different issue. After running some tests again, I found that it was definitely different. thank you 😅

PR #1028, opened 2022-01-26
Initial solution to issue #195

RobWalt:
This is my first try to solve issue #195 

I kind of built it on top of @a-kenji s idea.

The current path is displayed as a relative path starting from the origin of where zellij was originally started.

PR1

I've made sure that the path is always displayed at the top of the plugin pane, even when scrolling.
PR2

Here is an example of how a deeper path looks.
PR3

I don't really know about the colors here. Some feedback on that would be nice.
Also, in my first attempt I wanted to display the absolute path. But I had no idea how to get the "pwd" info inside of the plugin.
Is there any way to do this.

Also what is the "/host" prefix in the path supposed to achieve?

The proposed idea from @imsnif to display the path in the pane title is also kind of hard, since the title is part of the pane and I don't know how to communicate information out from the plugin to make it available in the pane.

While writing I noticed that I should think about the case again, where the path doesn't fit in the pane.

### DIFF
--- a/default-plugins/strider/src/main.rs
+++ b/default-plugins/strider/src/main.rs
@@ -87,7 +87,13 @@ impl ZellijPlugin for State {
     }
 
     fn render(&mut self, rows: usize, cols: usize) {
-        for i in 0..rows {
+        let display_current_dir = FsEntry::DisplayDir(self.current_dir.clone())
+            .as_line(cols)
+            .normal()
+            .on_yellow()
+            .black();
+        println!("{}", display_current_dir);
+        for i in 0..rows - 1 {
             if self.selected() < self.scroll() {
                 *self.scroll_mut() = self.selected();
             }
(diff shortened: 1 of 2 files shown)


RobWalt:
Now the end of the current path gets displayed instead of its start if there isn't sufficient space to show it as a whole.
PR4
PR5

tlinford:
Nice, I don't know a ton about it - but if you want to send info to a plugin you can use the `Event` enum: https://github.com/zellij-org/zellij/blob/a0791e35048dc5f7262328fc9a6ed61d78bd8c1d/zellij-tile/src/data.rs#L73.

From the server thread something like https://github.com/zellij-org/zellij/blob/a0791e35048dc5f7262328fc9a6ed61d78bd8c1d/zellij-server/src/tab/mod.rs#L1939 can be used to send it, and then they can be subscribed in the plugin setup.

I'm not sure where it's done but the pane title has the pwd so it should be doable to send it over.

RobWalt:
Thanks I'm going to look into that :+1:

RobWalt:
I was trying to implement it, but I failed because I just don't get how to state of the plugins transforms over time.

I propose, that the PR gets merged in its current state. Getting the `pwd` in front of the displayed path doesn't provide that much of an extra benefit imo.

tlinford:
Hey @RobWalt, sorry for taking so long to get to this!
Did a quick run with it (after rebasing on main), and it looks like mouse clicks are off by a couple of rows somehow.
Are you still interested in finishing this up?

RobWalt:
@tlinford not really, unfortunately I'm quite busy lately with other stuff

PR #1032, opened 2022-01-27
fix(compatibility): do not set scroll buffer in alternate screen

tlinford:
initial attempt at fixing #952.

The 0/1 problem was due to creating the viewport for the alternate screen app with an empty line, and then in fill_viewport pushing current lines to lines above.

I also found that if resizing a window containing a full-screen tui app, the various calls to `transfer_rows_from_viewport_to_lines_above` were causing the pane to become scrollable.

I'm still a bit unsure of how to handle things in `change_size`, but have not been able to see any misbehaving apps so far.


tlinford:
in alternate screen mode change size now handles width changes by truncating extra characters, and height changes by dropping excess lines. @imsnif does that make sense to you?

ahkrr:
@tlinford. After this commit, calling something like `fzf` crashes zellij. 
After investigating, I found that emptying the `viewport` in this line
``` rust
// zellij-server/src/panes/grid.rs +1871
let current_viewport = std::mem::take(&mut self.viewport);
```
leads to
```Error: thread 'screen' panicked at 'called `Option::unwrap()` on a `None` value': zellij-server/src/panes/grid.rs:```
because 
``` rust
// zellij-server/src/panes/grid.rs +1145
fn pad_current_line_until(&mut self, position: usize) {
    let current_row = self.viewport.get_mut(self.cursor.y).unwrap();
```
tries to index into an empty `viewport`.

When I reverted `mem::take` back to 
``` rust 
let current_viewport = std::mem::replace(
     &mut self.viewport,
     vec![Row::new(self.width).canonical()],
);
```
the crash was gone.

tlinford:
@ahkrr  Thanks for figuring this out, glad you caught it before release. I merged the fix and also added a test case for `fzf`.

PR #1038, opened 2022-02-06
feat: improve error reporting system

jaeheonji:
**This is the current concept stage.**

There have been concerns from some users about `application errors` from the past.
This was due to errors from some `panic` codes.

Whether or not the `panic` code is used has nothing to do with the stability of the program. When `panic` is used, it means that the program **must be stopped**.

The problem isn't the use of panic. the problem is that errors from panic are inconvenient for humans to read.
So, we have to include the "**more information**" for solving the error and print the error more "**pretty**"

**Before:**
before

**(UPDATED) After:** 
스크린샷 2022-03-08 오후 10 35 36

with **Backtrace**
스크린샷 2022-03-08 오후 10 34 36


imsnif:
@jaeheonji - this looks awesome! I left 2 minor comments, and otherwise you can merge this whenever you want. Great work!

jaeheonji:
@imsnif Thank you so much for your review :)

Issue #1010, opened 2022-01-15
Confusing tips on `RenamePane` or `RenameTab` mode


dswij:
When in `RenamePane` or `RenameTab` mode, the tips is `<Enter> when done / <ENTER> Select pane`, which is confusing since there are two `<ENTER>` action

![image](https://user-images.githubusercontent.com/44697459/149617878-a14c1454-519d-4767-aa2e-643fbfbc30e3.png)

jaeheonji:
Obviously, the `<ENTER> Select pane` part can be confusing.
It seems that that phrase can be modified in the `status-bar` plugin.

In the details, the following functions in `seconds_line.rs`:
* `full_shortcut_list`
* `shortened_shortcut_list`
* `best_effort_shortcut_list`

I'll make this a **good first issue** for now!

RobWalt:
@jaeheonji This can be closed

PR #1045, opened 2022-02-07
Solve Issue #1010

RobWalt:
This solves #1010 

- The instructions are now displayed correctly
- It slightly unifies the style (the angle brackets weren't colored green before)
- I refactored the code a bit, since I noticed, that we do the same thing over and over again in this function

Here are some pictures:
Before:
Issue1010-1
After:
Issue1010-2

### DIFF
--- a/default-plugins/status-bar/src/second_line.rs
+++ b/default-plugins/status-bar/src/second_line.rs
@@ -100,7 +100,9 @@ fn locked_interface_indication(palette: Palette) -> LinePart {
     }
 }
 
+/// Creates hints for usage of Rename Mode in Pane Mode
 fn select_pane_shortcut(is_first_shortcut: bool, palette: Palette) -> LinePart {
+    // get the colors
     let white_color = match palette.white {
         PaletteColor::Rgb((r, g, b)) => RGB(r, g, b),
         PaletteColor::EightBit(color) => Fixed(color),
@@ -113,7 +115,10 @@ fn select_pane_shortcut(is_first_shortcut: bool, palette: Palette) -> LinePart {
         PaletteColor::Rgb((r, g, b)) => RGB(r, g, b),
         PaletteColor::EightBit(color) => Fixed(color),
     };
+    // dynamic separator
     let separator = if is_first_shortcut { " " } else { " / " };
+    // define text with according style
+    // tuple is (displayed text, color, bold [true=yes,false=no])
     let text_with_style = [
         (separator, white_color, false),
         ("Alt", orange_color, true),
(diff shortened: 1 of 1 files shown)


jaeheonji:
@RobWalt Thank you for your contribution!
it make sense! and also code readability is good 👍 

The asset update is missing, so I'll update it and merge it.
Thank you!

RobWalt:
@jaeheonji Thanks for the feedback. How would I update the assets myself? You know... just for the next time 😄

jaeheonji:
@RobWalt Sorry! There was one thing I couldn't check.

The `select_pane_shortcut` function changes the contents of all tips. In most cases, it make sense. But in some cases, the meaning becomes strange like below.

화면 캡처 2022-02-08 023501

So, it seems good to keep the original text `<ENTER> select pane`, but apply the new text only in the `RenamePane` mode.
The `mode` value can be passed as follows.

```rust
fn full_shortcut_list(help: &ModeInfo, tip: TipFn) -> LinePart {
    match help.mode {
        InputMode::Normal => tip(help.palette),
        InputMode::Locked => locked_interface_indication(help.palette),
        _ => {
            ...
            let select_pane_shortcut = select_pane_shortcut(help.keybinds.is_empty(), help.palette, help.mode); // pass the mode
            ...
        }
    }
}
```

This could show a clearer meaning. Can you please proceed with this?

RobWalt:
Yeah, I already noticed that. I fixed it, but I still don't get why the e2e tests fail. These changes are not very substantial.

jaeheonji:
> I still don't get why the e2e tests fail. These changes are not very substantial.

The e2e test compares to previous snapshots. And the snapshot has a buffer of the last successful Zellij UI.
The reason the test failed is that the output buffer of status-bar has changed, but the snapshot is not updated.

To verify this, I recommend that you first run the e2e test locally. This will give you a guide on what has changed and how to change it.

jaeheonji:
Thanks for the really quick update!
Unfortunately, currently my time zone is at dawn, so I'll check again as soon as possible in the morning!

RobWalt:
It finally works 🎉 

So @tlinford and I were searching for the issue here and this came up:

- Apparently the `focus_tab_with_layout` test is very sensitive w.r.t. timing since it seems to be too slow to fulfill the test condition in time. To fix this, we compiled in release mode, replaced the wasm module in the asset folder and restarted the local docker.

jaeheonji (approved):

jaeheonji:
Thanks! LGMT 👍

Issue #771, opened 2021-10-07
Replace yaml with KDL


imsnif:
[KDL](https://kdl.dev/) is a new document language that supports nesting, custom type annotations and more. It is Human-readable, easy to pick up and is actively maintained through a [living specification](https://github.com/kdl-org/kdl/blob/main/SPEC.md).

A Zellij layout KDL configuration might look something like this:

```KDL
layout "template" {
  direction "Horizontal"
  parts {
    layout {
      direction "Vertical"
      borderless true
      split_size Fixed=1
      run plugin=true location="zellij:tab-bar"
    }
    layout "body" {
      direction "Vertical"
    }
    layout {
      direction "Vertical"
      borderless true
      split_size Fixed=2
      run plugin=true location="zellij:status-bar"
    }
  }
}

layout "tabs" {
  direction "Vertical"
}
```

A Zellij configuration might look something like this:

```KDL
keybinds unbind=true {
  normal {
    action SwitchToMode="Locked" key {
       Ctrl "g"
    }
    action SwitchToMode="Pane" key {
      Ctrl "p"
    }
    action SwitchToMode="Resize" key {
      Ctrl "n"
    }
    action SwitchToMode="Tab" key {
      Ctrl "t"
    }
    action SwitchToMode="Scroll" key {
      Ctrl "s"
    }
    action SwitchToMode="Session" key {
      Ctrl "o"
    }
    action Quit key {
      Ctrl "q"
    }
    action NewPane key {
       Alt "n"
    }
    action MoveFocusOrTab="Left" key {
       Alt "h"
    }
    action MoveFocusOrTab="Right" key {
       Alt "l"
    }
    action MoveFocus="Down" key {
       Alt "j"
    }
    action MoveFocus="Up" key {
       Alt "k"
    }
    action FocusPreviousPane key {
       Alt "["
    }
    action FocusNextPane key {
       Alt "]"
    }
  }
  locked {
    action SwitchToMode="Normal" key {
      Ctrl "g"
    }
  }
  // ... (yes, it supports comments!!)
}
```

The only blocker I see right now is that there is no Serde support, but it looks like it's in the works: https://github.com/kdl-org/kdl-rs/issues/17

Would love to hear thoughts on this. Pinging some people of interest (though I'd love to hear anyone's take on this!) @a-kenji , @TheLostLambda, @spacemaison

TheLostLambda:
[...]

PR #1046, opened 2022-02-08
Initial KDL support

tailhook:
Here is my approach to #771.

I'm not sure it parses exactly the same format that is shown in #1036, but should be generally close to that. One of my goals was to make the PR small, so most Rust structures are similar to Yamlish ones, but KDL is more powerful, so config should be quite nice nevertheless.

I've not tested all the settings, but simple configs look like work. This is obviously just a draft, but it shows how easy the integration might be.

Here is a non-comprehensive list of questions and things that could be changed:
1. Keybindings can be grouped by mode, in the current implementation thay have `key ... mode="normal" {...}`. This layout makes multi-modal bindings easier.
2. Themes and plugins are scattered as `theme "xx"`, `plugin "xx"` items in the single section. They might be in the `themes` and `plugins` sections. But with KDL I prefer more flat.
3. All names are in `kebab-case` for simplicity (including node names, properties and enum values). But we may tweak a style for some of them if it's needed. But I feel that single style should be less confusing (although not like in #1036)
4. Duplicate theme and duplicate plugins check is not handled in the current implementation
5. Serializing config back is not implemented (I think it worked with Yaml, right?)
6. We should figure out plugin config. My approach would be to serialize KDL nodes with `minicbor` (is already implemented in the `knuffel`) and then plugin can deserialise nodes and use their own derive macro to validate and process the config.

<details>
<summary>Example config</summary>

```kdl
keybinds {
    unbind-all mode="normal"
    key "F1" {
        go-to-tab 1
    }
    key "F2" {
        go-to-tab 2
    }
    key "F3" {
        go-to-tab 3
    }
    key "F4" {
        go-to-tab 4
    }
    key "F5" mode="normal" {
        new-tab
    }
    key "Alt+h" mode="normal" {
        move-focus "left"
    }
    key "Alt+l" mode="normal" {
        move-focus "right"
    }
    key "Alt+j" mode="normal" {
        move-focus "down"
    }
    key "Alt+k" mode="normal" {
        move-focus "up"
    }
    key "n" mode="pane" {
        switch-to-mode "normal"
    }
    key "N" mode="pane" {
        new-pane
    }
}
```

</details>

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4,11 +4,11 @@ version = 3
 
 [[package]]
 name = "addr2line"
-version = "0.15.2"
+version = "0.17.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "e7a2e47a1fbe209ee101dd6d61285226744c6c8d3c21c8dc878ba6cb9f467f3a"
+checksum = "b9ecd88a8c8378ca913a680cd98f0f13ac67383d35993f86c90a70e3f137816b"
 dependencies = [
- "gimli 0.24.0",
+ "gimli 0.26.1",
 ]
 
 [[package]]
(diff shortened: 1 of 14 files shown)


TheLostLambda:
Whoops @tailhook ! I think we've been duplicating some work! I've been working through the same issue (starting with defining the format) over at #1036 !

With that being said, I'd not yet properly started on the code half and wasn't aware of knuffle, something definitely worth looking into!

tailhook:
> Whoops @tailhook ! I think we've been duplicating some work! I've been working through the same issue (starting with defining the format) over at #1036 !
> 
> With that being said, I'd not yet properly started on the code half and wasn't aware of knuffle, something definitely worth looking into!

Looks like we were not duplicating too much work, because I've not researched on the actual format too much. And you haven't invested into the code too much. Which is fortunate, as I've not seen your PR until yesterday too.

Also, you could not have seen knuffel before, as I've just published it yesterday :sunglasses:  That said, I've done this PR mostly to figure out whether knuffel is powerful enough to represent quite complex config like this one, and if it's simple enough to integrate. And overall I'm fine with results.

@TheLostLambda, do you want to carry on from this point? I'll be available to help you discover the features of knuffel and fix bugs. But I'm not sure I'll have enough time to work on this singleheartedly.

TheLostLambda:
@tailhook I'd be super happy to carry on the work you've started here! Your library looks stellar and I'm eager to try it out!

PR #1047, opened 2022-02-08
fix(tests): ensure e2e tests use updated plugins

tlinford:
Spent some time debugging some strange test failures in #1045 with @RobWalt, and we realized that currently
the e2e tests are using the pre-built plugins from the assets folder. Also they are installed into a directory inside the container
that persists between runs when testing locally, so even updating the plugins in `assets` was not giving reliable results.
 
Solve this issue by:
- launching zellij with `--data-dir` flag in e2e tests, to avoid using prebuilt
  plugins from assets dir, and instead use the ones from the docker
  mount.
- making the tests use plugins built in release mode to avoid some test
  failures due to timing issues.

Other changes are auto-formatting of Makefile.toml.

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -18,7 +18,7 @@ jobs:
       ssh:
         image: ghcr.io/linuxserver/openssh-server
         env:
-          PUID: 1000
+          PUID: 1001
           PGID: 1000
           TZ: Europe/Vienna
           PASSWORD_ACCESS: true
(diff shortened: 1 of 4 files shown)


tlinford:
note: had to change the config for the ssh service to use UID 1001 to match the runner's UID

TheLostLambda:
Thrilled to hear this fix is finally possible! I tried this ages ago with @imsnif in #676, but we were encountering some bizarre behaviors and never sorted it out. Thanks for wrapping this up!

RobWalt:
🎉  nice, congrats :)

PR #1048, opened 2022-02-09
feat(update): Add dependabot targets

ken-matsui:
Some of Zellij's dependencies that should be updated are not up-to-date.
This PR introduces adding all directories to dependabot targets.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -4,3 +4,41 @@ updates:
     directory: "/"
     schedule:
       interval: "weekly"
+
+  - package-ecosystem: "cargo"
+    directory: "/"
+    schedule:
+      interval: "weekly"
+  - package-ecosystem: "cargo"
+    directory: "/zellij-client/"
+    schedule:
+      interval: "weekly"
+  - package-ecosystem: "cargo"
+    directory: "/zellij-server/"
+    schedule:
+      interval: "weekly"
+  - package-ecosystem: "cargo"
+    directory: "/zellij-tile-utils/"
+    schedule:
+      interval: "weekly"
+  - package-ecosystem: "cargo"
+    directory: "/zellij-tile/"
+    schedule:
+      interval: "weekly"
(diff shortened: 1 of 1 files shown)


a-kenji:
Thank for this pr!
If I remember correctly we decided against doing this for now @imsnif, is that correct?

imsnif:
Hey! While I appreciate the work on this - unless there are some specific upgrades that are urgent for reasons of security or something not working, I think keeping things constantly up to date will be a challenge for us as it is.

Our tests are unfortunately not comprehensive enough to give enough confidence that a specific upgrade didn't break anything. It's a situation we're working on changing - but we're not there yet. So we would have to do a lot of manual checking for every such upgrade.

We're still a pre-1.0 - so I think we can be cut a little bit of slack on this front :)

ken-matsui:
@a-kenji @imsnif 
Thank you all for responding to my PR!

I understood your situation well, and I totally agree with your decision.
So then I need two additional instructions:
1. Which would you like me to either close or keep this PR?
2. Could I bump up the version of [this](https://github.com/zellij-org/zellij/blob/10a22c479ffb4d76627eadbee2dc4b53ae4309c3/Cargo.toml#L24) dependency I maintain which does not bring breaking changes?

imsnif:
Hey @ken-matsui - thanks for understanding! Would be happy to upgrade `suggestion` and I just opened a separate issue about it so it won't get lost: #1068 

Gods willing, we will release a new version next week which is going to be quite big as is (and there's still some more stuff I'd like to add to it). So I hope you won't mind if we postpone upgrading suggestion to the next version.

I assigned the issue to myself though, so plan to personally take care of it and not let it be forgotten.

As for this PR: you can close it if you want, yeah.

ken-matsui:
Hi @imsnif,

Thanks for your support!
I close this PR because I can expect the structure to have changed significantly by the time we can safely update dependencies.

I am looking for the new version :)

PR #1071, opened 2022-02-20
chore: bump internal autocfg dependency to 1.1.0

tranzystorek-io:
`autocfg` 1.0.1 can cause problems when running cross compilation with the `--locked` configuration, for example in the Void Linux builds: https://github.com/void-linux/void-packages/issues/34889


jaeheonji:
@tranzystorek-io LGTM! Thank you 😄

PR #1073, opened 2022-02-21
feat(ux): tmux mode

imsnif:
This adds a hidden "tmux mode" which can be triggered (just like in tmux) with `ctrl-b`. This mode offers basic built-in tmux shortcuts: 
`<←↓↑→> Move focus / <"> Split Down / <%> Split Right / <z> Fullscreen / <c> New Tab / <,> Rename Tab / <p> Previous Tab / <n> Next Tab`

So that (for example): pressing `ctrl-b` + `"` would split the focused pane down.

This is intended to help people migrating from tmux to feel more at home.


jarod:
use `ctrl-b` is really a bad idea, `ctrl-b` is hotkey for readline `move backward`. we can opt-out with config of clause.

PR #1075, opened 2022-02-21
Simplify the conversion from `bool` to `i32`

ken-matsui:
`true` and `false` can respectively be `1` and `0` using the type cast expression simply.


imsnif:
Nice :)

PR #1077, opened 2022-02-21
Remove unnecessary `continue`

ken-matsui:
This PR simplifies the `if` statement in `zellij-server/src/wasm_vm.rs`.


ken-matsui:
@tlinford 
I appreciate your review :)
I agree that would be better and more robust against changes.
I committed new change you've suggested.

PR #1078, opened 2022-02-21
default.yaml: Reorder key bindings

har7an:
to reflect the order in which they appear in the UI and add some spacing
to visually group related items.

I hope I didn't lose entries on the way...
I also reordered some of the keybindings due to consistency.


har7an:
Oh darn, I should've rebased before I started working on this... :joy:

har7an:
Interestingly I can't seem to activate the tmux mode from #1073... Is this expected?

imsnif:
Hey @har7an - I think not being able to switch to tmux mode is caused because Zellij is using an old configuration (for example in your ~/.config/zellij folder). This might also be the case with your installed Zellij version, the release notes of `0.25.0` have an "upgrading" section with more info on how to fix this.

About this PR - while I appreciate the work and think this is definitely better - I'm a little concerned about introducing large diffs to people's already existing configs. We're about to make a very large change to our configuration - moving everything from YAML to KDL. Maybe you can incorporate this change as part of those efforts? https://github.com/zellij-org/zellij/pull/1036

a-kenji:
> I think not being able to switch to tmux mode is caused because Zellij is using an old configuration (for example in your ~/.config/zellij folder).

Just a small addition @har7an , this is not just because zellij is using an old configuration, but this is most likely because you configured zellij to [unbind](https://zellij.dev/documentation/keybindings.html#to-unbind-the-default-keybindings) every single default keybinding with:
```
keybinds:
    unbind: true
```

har7an:
> We're about to make a very large change to our configuration - moving everything from YAML to KDL

Yes, I've already heard from kenji. I don't mean to steal anyones show, but as I understand it KDL is another markup language, right? How exactly is it superior to YAML? Because I believe that merely adding new sections for general keybindings for all modes would be possible in YAML as well, or not?

> Maybe you can incorporate this change as part of those efforts?

Well if the author of #1036 is fine with other people dropping commits into his branch, I can surely look into it. :)

har7an:
> Just a small addition @har7an , this is not just because zellij is using an old configuration, but this is most likely because you configured zellij to [unbind](https://zellij.dev/documentation/keybindings.html#to-unbind-the-default-keybindings) every single default keybinding with:

Good point, I admit... Thanks!

imsnif:
Hey, @har7an - I'm pretty close to moving us to KDL and so I think I'm going to close this one for now.

PR #1082, opened 2022-02-22
Add `#[allow(dead_code)]` attribute to unused variables

ken-matsui:
I would suggest deleting these variables unless you think they will be used.
If you would so, I would be pleased to remove them in this PR.

### DIFF
--- a/zellij-server/src/ui/overlay/mod.rs
+++ b/zellij-server/src/ui/overlay/mod.rs
@@ -22,6 +22,7 @@ pub trait Overlayable {
     fn generate_overlay(&self, size: Size) -> String;
 }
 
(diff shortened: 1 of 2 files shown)


ken-matsui:
This PR conflicts with https://github.com/zellij-org/zellij/pull/1087.
I would suggest merging either @orhun's or mine.

orhun:
Ah I totally missed that this PR already exists :/

ken-matsui:
Never mind :)

ken-matsui:
Thanks for your opinion. I close this PR, so any changes should be continued on https://github.com/zellij-org/zellij/pull/1087.

PR #1083, opened 2022-02-22
Replace redundant `map` with `if let`

ken-matsui:
I mainly replaced the redundant `map` with `if let` to make them much simpler and easy to understand, which was also pointed out by clippy.
Additionally, I unified duplicate codes in the `operate_active_terminal_scroll` function.
Through the unification, I found the following possible problems:
1. `scroll_active_terminal_to_bottom` & `clear_active_terminal_scroll` are perfectly the same
2. `scroll_active_terminal_up_page` are the only function that uses different action between floating_pane and active_pane

These might be problems, so I hope you can check them out :)

### DIFF
--- a/zellij-server/src/tab/mod.rs
+++ b/zellij-server/src/tab/mod.rs
@@ -1,6 +1,7 @@
 //! `Tab`s holds multiple panes. It tracks their coordinates (x/y) and size,
 //! as well as how they should be resized
 
+mod active_terminal_scroll;
 mod clipboard;
 mod copy_command;
 pub mod floating_pane_grid;
@@ -28,6 +29,7 @@ use crate::{
     wasm_vm::PluginInstruction,
     ClientId, ServerInstruction,
 };
+use active_terminal_scroll::ActiveTerminalScroll;
 use serde::{Deserialize, Serialize};
 use std::cell::RefCell;
 use std::os::unix::io::RawFd;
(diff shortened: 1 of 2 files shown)


zellij-server/src/tab/mod.rs
```diff
@@ -2196,212 +2196,114 @@ impl Tab {
                 .unwrap();
         }
     }
-    pub fn scroll_active_terminal_up(&mut self, client_id: ClientId) {
+
+    fn operate_active_terminal_scroll<ClosureT>(
```
imsnif, line 2200: Hey, this is great! As long as we're here, how about we do something like:

```rust
pub fn scroll_active_terminal_up(&mut self, client_id: ClientId) {
    // ...
    ScrollActiveTerminal::new(/* .. */).up(1)
}
```

imsnif: Or do you think that'll be too involved? I just feel it'll save us passing closures around and mucking about with generics.

ken-matsui: Hi @imsnif,

Thank you for your suggestion!
I agree we want to avoid closures and generics as possible. However, your suggestion initializes a struct as a closure initializes a struct as described [here](https://rustc-dev-guide.rust-lang.org/closure.html). So I'm thinking there may not be that much difference in terms of memory usage (but the code looks beautiful!).

I would like to additionally investigate if this code can be improved without closures and generics in three aspects: memory usage, speed, and code cleanliness :)

imsnif: Sure - as you wish. Just bear in mind that I don't think performance or resource utilization are super important in this flow. When scrolling the terminal we're bound by STDIN which is considerably slower.

My suggestion was mostly because I thought it would read better.

ken-matsui: Oh, that's certainly right... I'll follow your suggestion! Thank you!

ken-matsui:
Hi @imsnif,
Sorry for taking the time! I've updated the code as you suggested, so please take a look when you have time :)

ken-matsui:
It seems this PR is stale, so closing.

PR #1087, opened 2022-02-22
Fix unused code warnings

orhun:
This PR is doing what title suggests :bear:

I came across this while updating the Arch Linux package and thought they should be fixed :)

Log:

```
warning: field is never read: `rows`
  --> zellij-server/src/ui/overlay/mod.rs:27:5
   |
27 |     rows: usize,
   |     ^^^^^^^^^^^
   |
   = note: `#[warn(dead_code)]` on by default

warning: field is never read: `cols`
  --> zellij-server/src/ui/overlay/mod.rs:28:5
   |
28 |     cols: usize,
   |     ^^^^^^^^^^^

warning: field is never read: `plugin_own_data_dir`
  --> zellij-server/src/wasm_vm.rs:76:5
   |
76 |     plugin_own_data_dir: PathBuf,
   |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^

warning: `zellij-server` (lib) generated 3 warnings
```

### DIFF
--- a/zellij-server/src/ui/overlay/mod.rs
+++ b/zellij-server/src/ui/overlay/mod.rs
@@ -23,12 +23,6 @@ pub trait Overlayable {
 }
 
 #[derive(Clone, Debug)]
-struct Padding {
-    rows: usize,
-    cols: usize,
-}
-
-#[derive(Clone, Debug)]
 pub enum OverlayType {
     Prompt(prompt::Prompt),
 }
--- a/zellij-server/src/wasm_vm.rs
+++ b/zellij-server/src/wasm_vm.rs
@@ -73,6 +73,7 @@ pub(crate) struct PluginEnv {
     pub subscriptions: Arc<Mutex<HashSet<EventType>>>,
     pub tab_index: usize,
     pub client_id: ClientId,
+    #[allow(dead_code)]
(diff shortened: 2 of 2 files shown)


imsnif:
Thanks @orhun - you are always super fast updating the Arch version!

@a-kenji - if I'm not mistaken this is part of some ongoing work you're doing. Are you still counting on this? Maybe we can add a dead_code exception as @ken-matsui suggested instead?

jaeheonji:
@imsnif Oh.. I didn't know this @a-kenji was working on. So I suggested to @ken-matsui that getting rid of this is better than using dead_code.

If this is part of @a-kenji work, it would be good to include dead_code.

a-kenji:
@orhun,
Yes it is part of some work I am doing, but it is fine to delete it for now, since I think I wont have merge conflicts inside the file.

I do recommend however to apply the dead code to the plugin data dir, since that is something that Brooks is still working on.

orhun:
> I do recommend however to apply the dead code to the plugin data dir, since that is something that Brooks is still working on.

Done! 631fa06b4680c3c1ab64c3541c29b6d69722a44a

a-kenji:
@orhun,
Awesome, thanks!

PR #1088, opened 2022-02-22
fix(docs): replace the list with a table that is more readable rather than strikethrough

ken-matsui:
Since strike-through texts are difficult to read, I made them easier to read by putting them in a table.


imsnif:
Looks great, thanks!

PR #1090, opened 2022-02-22
fix(ci): run clippy with `-D warnings`

ken-matsui:
I hope I resolved all `clippy` lint warnings in my previous unmerged PRs.
So then it should be the next step to enforce failing CI by default when any warning is found to keep the quality of Zellij.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -75,4 +75,4 @@ jobs:
     - name: Install cargo-make
       run: test -x "${HOME}/.cargo/bin/cargo-make" || cargo install --debug cargo-make
     - name: Check Lints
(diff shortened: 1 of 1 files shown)


imsnif:
Tbh, I turned off the warnings last week because I found myself spending over an hour fixing them when everything else was green. IMO they're a little excessive and slow down development more than they help, but since you were so kind to do the work on this (thank you!) - I'm fine with bringing this back (will check out your PRs before). However... I can't promise I won't get into the same situation in the future...

imsnif:
I agree with everything you said! This sounds perfect.

ken-matsui:
Glad to hear that! I appreciate your understanding and will work on them :)

Issue #880, opened 2021-11-18
Support for hex-colors


fedoranvar:
Good day to you!

Is there support for hex-colors in `theme`-section?

 like: 
```yaml
green: '#102030'
```
^tried that for 'green' component - it shows error

Thank you!

imsnif:
Not yet! I would love to merge something like this.

ramonGonzEdu:
The PR was merged. just use colors like `"#FF0000"` in config

a-kenji:
Thank you @ramonGonzEdu.

PR #1092, opened 2022-02-22
Add basic support for hex codes in theme

ramonGonzEdu:
This pull request adds the ability to use hex codes in the config as strings.
The theme should follow the pattern:
`green: '#00FF00'`

This supports the `#FFFFFF` format and the `#FFF` format.

Please let me know if there are any issues as this is my first PR.

Possible Issues:
I had to remobe the Copy trait on `PaletteFromYaml` and `PaletteColorFromYaml` in order to ask for a string, but I am looking into serde's `deserialize_with` in order to move the verification step out of the `PaletteColor` conversion.

### DIFF
--- a/zellij-utils/src/input/theme.rs
+++ b/zellij-utils/src/input/theme.rs
@@ -15,7 +15,7 @@ struct Theme {
 }
 
 /// Intermediate deserialization struct
-#[derive(Clone, Copy, Debug, Serialize, Deserialize, PartialEq, Eq, Hash, Default)]
+#[derive(Clone, Debug, Serialize, Deserialize, PartialEq, Eq, Hash, Default)]
 pub struct PaletteFromYaml {
     pub fg: PaletteColorFromYaml,
     pub bg: PaletteColorFromYaml,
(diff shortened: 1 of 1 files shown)


ramonGonzEdu:
This implements #880

ramonGonzEdu:
This assumes 255 as the color value if there is a parsing error. It also simply checks for string length (4 or 7) and ignores the first character

zellij-utils/src/input/theme.rs
```diff
@@ -99,6 +100,29 @@ impl From<PaletteColorFromYaml> for PaletteColor {
...
+                    let red = u8::from_str_radix(hex_string.get(1..2).unwrap(), 16);
+                    let green = u8::from_str_radix(hex_string.get(2..3).unwrap(), 16);
+                    let blue = u8::from_str_radix(hex_string.get(3..4).unwrap(), 16);
+                    (
+                        red.unwrap_or(15) * 16,
+                        green.unwrap_or(15) * 16,
+                        blue.unwrap_or(15) * 16,
+                    )
+                } else {
+                    panic!("Invalid hex string");
```
a-kenji, line 123: Please convert this to an error, rather than a panic.

zellij-utils/src/input/theme.rs
```diff
@@ -99,6 +100,29 @@ impl From<PaletteColorFromYaml> for PaletteColor {
         match yaml {
             PaletteColorFromYaml::Rgb(color) => PaletteColor::Rgb(color),
             PaletteColorFromYaml::EightBit(color) => PaletteColor::EightBit(color),
+            PaletteColorFromYaml::RgbHex(hex_string) => {
+                PaletteColor::Rgb(if hex_string.len() == 7 {
+                    let red = u8::from_str_radix(hex_string.get(1..3).unwrap(), 16);
```
a-kenji, line 105: Same here, maybe check for the length prior to this step and return an error?

a-kenji (changes requested):
Thanks for this work! This already looks very promising and is something we have been looking forward to for a while.

ramonGonzEdu:
Sorry for the late response, i was busy with classes, but I will work on those changes Tomorrow. :)

ramonGonzEdu:
Move to #1536

PR #1094, opened 2022-02-23
Ensure Title doesn't overrun

tfgast:
The set title escape code needs to be terminated with a bell character. See https://tldp.org/HOWTO/Xterm-Title-3.html
Before this change I was seeing the following
Screen Shot 2022-02-22 at 4 50 56 PM


imsnif:
Good catch! And thanks for this. Also - apologies for the CI email spam. We've recently introduced a test that ended up being a little flaky. Will take care out of out of scope.

PR #1103, opened 2022-02-23
WIP: Refactor/streamline keybindings

har7an:
Streamlines the way the keybindings are read. While working on #1078 I realized that a lot of the bindings are redundant/repetitions of what is already bound in `normal`. This also means that merely changing keybindings to switch modes in `normal` isn't sufficient, because the changes won't carry to other modes.

This MR attempts to changes this by always applying the keybindings from the `normal` mode/section to all modes. This still gives the individual modes the ability to overwrite single keybindings with custom ones where needed.

It's marked WIP because:

1. I uncommented the tests because I didn't want to adapt those before I heard your feedback on the general idea
2. The way its currently implemented it doesn't respect the `unbind`s, hence I had to exclude `InputMode::Locked` from taking the `normal` bindings. I assume that this is due to the fact that I am overwriting the bindings after the `unbind`s have been handled already? But I'm not quite sure about that because I find it a little hard to understand what exactly is going on in parts of the code with all the `From` traits and stuff. :D


a-kenji:
Thank you for this work.

I don't think we will benefit from this keybinding change, as #1036 proposes a much better solution, IMO.

If you want to refactor the config part, I will gladly look at a refactor pr.

har7an:
> I don't think we will benefit from this keybinding change, as #1036 proposes a much better solution, IMO.

I don't understand how #1036 does anything to deduplicate keybindings in the configuration. Could you elaborate on that?

a-kenji:
> I don't understand how https://github.com/zellij-org/zellij/pull/1036 does anything to deduplicate keybindings in the configuration. Could you elaborate on that?

Sure!

To quote directly from #1036:
```
// Configuration for zellij.

modes clear-defaults=true {
  // Can also use `shared` with a list of modes to include
  shared-except "locked" "renametab" "renamepane" {
    bind "Ctrl-g" {
      SwitchToMode "Locked"
    }
    bind "Ctrl-p" {
      SwitchToMode "Pane"
    }
```

The `shared-except` configuration option would allow all modes
except for the specified modes to share configuration values.
The `shared` configuration options would allow all modes that are listed
to share configuration values. This would allow deduplication of keybindings in the configuration.

har7an:
Hmm, I guess then #1078 is obsolete as well, right?

har7an:
Well, on to something new then...

Issue #883, opened 2021-11-18
Feature: pass title through to terminal.


a-kenji:
Currently zellji swallows the tile, that is set for a pane.
And doesn't pass it through to the terminal.

- [ ] pass the title through
- [ ] make the title adjustable (zession name, title, host name)

KerfuffleV2:
Has anyone looked at this recently? As far as I can tell, passing the title through just doesn't work at all in spite of a pull being merged that was supposed to fix it. My terminals just have `zellij` as the title, although my prompt normally sets a title with the user/path.

Also nothing happens when running a command such as: 
```sh
echo -ne '\e]0;Yo\a';sleep 5
```

Tested in Konsole, xterm and WezTerm with identical behavior.

imsnif:
Can this be closed? Zellij already sets the title of the terminal according to the current session and pane

KerfuffleV2:
The problem I posted about here was resolved when my PR got merged, so that wouldn't be a reason to keep it open anyway.

PR #1113, opened 2022-02-24
fix: Allow terminal title passthrough even when not drawing pane frames.

### LINKED ISSUE #1111: Terminal title passthrough only works when pane frames are enabled.
Reproduced in `zellij 0.26.0`.

This appears to be caused because the title is only set in the pane frame render function, which doesn't get called when pane frames are disabled. Expected behavior would be to always pass through the title.

KerfuffleV2:
Currently the terminal title is set in the pane frame render function, which does not seem to run when displaying pane borders is disabled. 

This pull adds a `render_terminal_title` method to the `Pane` trait and a `render_terminal_title_if_needed` method to the `PaneContentsAndUi` struct. The latter can be called from `Tab::render` to ensure the title is always set when needed.

***Note***: First time looking at the Zellij codebase so I don't know if this is the correct approach. However, it does fix the issue for me with pane frames disabled (and terminal title setting still works with them enabled.)

Closes #1111
Other related issue: #883

### DIFF
--- a/zellij-server/src/panes/plugin_pane.rs
+++ b/zellij-server/src/panes/plugin_pane.rs
@@ -16,6 +16,7 @@ use zellij_utils::zellij_tile::prelude::{Event, InputMode, Mouse, PaletteColor};
 use zellij_utils::{
     channels::SenderWithContext,
     pane_size::{Dimension, PaneGeom},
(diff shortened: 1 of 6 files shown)


a-kenji (approved):
LGTM, thank you very much.

PR #1116, opened 2022-02-25
add tmux ctrl+b d shortcut for detach

casonadams:
The default keybinding for tmux detach is `ctrl+b` `d` adding this to the default might be nice for `tmux` users.


a-kenji:
Thank you!

PR #1122, opened 2022-02-25
Use the `clippy-check` action

ken-matsui:
This PR is created because of [this discussion](https://github.com/zellij-org/zellij/pull/1090#issuecomment-1051007086).
I could confirm that the clippy command I use is the same as the `cargo make clippy` command:
https://github.com/zellij-org/zellij/runs/5334524327?check_suite_focus=true#step:5:44


a-kenji:
This is so cool, thanks!

PR #1126, opened 2022-02-26
fix(docs): add information about clippy lints

ken-matsui:
This PR adds information about clippy lints as being discussed [here](https://github.com/zellij-org/zellij/pull/1090).


a-kenji:
Thank you!

PR #1131, opened 2022-02-26
init: nix support

a-kenji:


a-kenji:
Hey @cololemickens as promised here already a draft of initial nix support. I would be happy if you would look over it.

colemickens:
Looks good to me, but two thoughts:

* (nit, not blocker): I'm personally of the opinion that the ecosystem can do better than needing `flake-utils`, as I just opened here: https://github.com/NixOS/nix/issues/6172.
* (maybe blocker? or another quick PR?) `helix` and `jj` both added GitHub workflows to keep their flake inputs up to date as well (I can't remember off the top of my head if they also push the built nix artifacts to a cache. That's also very easy to add down the line with cachix.org etc)

a-kenji:
Thank you!

> (nit, not blocker): I'm personally of the opinion that the ecosystem can do better than needing flake-utils

It would be awesome, if something like that would be more straight forward in nix!

> (maybe blocker? or another quick PR?) helix and jj both added GitHub workflows to keep their flake inputs up to date as well (I can't remember off the top of my head if they also push the built nix artifacts to a cache. That's also very easy to add down the line with cachix.org etc)

I agree it is kind of a blocker, I want to address it in a different pr.
I am kind of torn on the implementation still. I want to add simple checks to the ci for the nix build, but I also don't want to make it too intrusive for people that don't care about the builds.

I am thinking about updating the `inputs` automatically on a weekly basis, what do you think about that?

Kind of like I do on https://github.com/a-kenji/zellij-nix, there I update all the inputs on a weekly basis and I advance the zellij input nightly.

I just looked at `jj` and `helix` and I can't see any automated updates there. Maybe I am looking at the wrong place there?

a-kenji:
Also what do you think of splitting the files up, like I did in 
https://github.com/a-kenji/zellij-nix/blob/main/nix/default.nix,
is that generally a bad idea?

colemickens:
:confetti_ball: ! And yes, I'm sorry, my comment was inaccurate - I totally missed that they added a build check without updating inputs. I'll have to make a note to loop back around on that.

Personally, the faster the better -- nixpkgs-wayland updates every 30-60 minutes or something like that -- all package sources and nix inputs are pulled up.

For now, it's not a huge issue, users can override nixpkgs on importing these repos to build latest, but I do want to circle back around and adding updates, and then maybe in nix-community have a centralized job that:

1. builds zellij/helix/jj flakes as-they-are
2. builds them with their nixpkgs updated to latest nixos-unstable

That way we can provide cache hits for as many folks as possible, and also try to help catch build-breaks early, etc. Just a idea.

PR #1140, opened 2022-02-27
Fix nix

a-kenji:


happysalada:
quick note, I think libiconv is included by default in buildRustPackage for darwin.

a-kenji:
Thank you, that is good to know!
Just getting everything set up, since I finally have gotten the green light for adding nix to this project.
If you have any more suggestions, please don't hesitate in reaching out.

PR #1161, opened 2022-03-03
Improve: more explicit detach instruction 

### LINKED ISSUE #1132: Session restore after detach always default to first tab
Steps to reproduce
- `zellij -s 0`
- create 2nd tab
- detach
- reattach

Expected results
Reattach in same state as just before detachment (Tab #2)

Actual results
Reattached in (Tab #1)

jaeheonji:
This PR resolve #1132 

### Objective

Currently, session detach and quit call the same `Remove Client` Instruction.
I wanted to make the difference between the two functions a little clearer.

Detach does not clear the state of an existing session. on the other hand, Quit **certainly** clears the existing state.

This change ensures that the existing state (location of the tab or pane) is maintained when the user re-attach to the session after detaching the session.


src/tests/e2e/snapshots/zellij__tests__e2e__cases__detach_and_attach_session.snap
```diff
@@ -5,7 +5,7 @@ expression: last_snapshot
 ---
  Zellij (e2e-test)  Tab #1                                                                                            
 ┌ Pane #1 ─────────────────────────────────────────────────┐┌ Pane #2 ─────────────────────────────────────────────────┐
-│$ █                                                       ││$ I am some text                                          │
+│$                                                         ││$ I am some text█                                         │
```
a-kenji, line 8: :+1: , makes much more sense.

jaeheonji: Thanks for the review!

jaeheonji:
@a-kenji I have one small issue with this change, could you please let me know what you think?

The zellij's `client_id`  is the same as the session connection sequence. (gradually increase from 1)
This may cause the user to feel a little strange in the following situations:

1. start server and `client_id 1` connected.
2. `client_id 2` connected in terminal A, `client_id 3` connected in terminal B
3. `client_id 2` and `client_id 3` detached
4. In terminal B, try to attach and assign `client_id 2`.

This can eventually appear as if the session has changed. This is even more pronounced because of the color of multiple sessions.

I have a little difficulty dealing with this alone. Is it right to apply it first and improve it later? Or is it not something to consider?

a-kenji:
> This may cause the user to feel a little strange in the following situations:

Oh yeah, good catch! That is a very strange situation.

> I have a little difficulty dealing with this alone. Is it right to apply it first and improve it later? Or is it not something to consider?

This is definitely something we need to fix, I personally think it should be OK to add now for 2 reasons:
- the consistency of attaching to a session is important and would affect more users
- in my opinion we need to have a different story for connecting two different users for the same client, we shouldn't rely solely on the `client_id`, if I want to have 2 clients myself connected to the same server, I don't want all that visual overhead this represents. So I would be happy if there could be a solution that tackles both of these problems. Maybe add a flag to attach that says it is a new user?

jaeheonji:
@a-kenji  I'm really happy to hear from you. thank you!

Yeah! you're right. We need something other than `client_id` to identify the user.
I've been thinking about a few things. For example, combination of username and tty number, etc.

But, it would be better to discuss this as a different issue!
Anyway, I will merge this PR when I have time!

Really Thanks :)

PR #1177, opened 2022-03-05
build: strip debug symbols

a-kenji:
Stripping the symbols for the release build drops binary size
currently by 20% and the plugin binaries by up to 50%.

Alternative: Keep the debug symbols also on releases


matu3ba:
What does this mean in total size? The binary is currently 16MB. 
1. How much space do plugins need?
2. Is it possible to remove plugins to save space? I guess debugging plugin is a very frequent use case, since plugins are "to customize things".

a-kenji:
The plugins take:
```
545k status-bar.wasm -> 306k
581k strider.wasm -> 302k
444k tab-bar.wasm -> 216k
```
The `zellij` binary has `12M` after stripping.

> Is it possible to remove plugins to save space? I guess debugging plugin is a very frequent use case, since plugins are "to customize things".

Not yet, that is certainly planned.`

For now you could load a layout without specifying plugins:

This is a layout that omits the loading of plugins:
```
---
template:
  direction: Horizontal
  parts:
    - direction: Vertical
      body: true
tabs:
  - direction: Vertical
```

TheLostLambda:
I'm 100% on board with merging this! I've been warned off of some cutting-edge features by @imsnif in the past (as they can break for people who haven't updated their Rust in a while), but I'm personally up for merging whenever others are comfortable!

imsnif:
> I'm 100% on board with merging this! I've been warned off of some cutting-edge features by @imsnif in the past (as they can break for people who haven't updated their Rust in a while), but I'm personally up for merging whenever others are comfortable!

How do you mean? Will people need a higher version of rust than is in eg. apt?

TheLostLambda:
> > I'm 100% on board with merging this! I've been warned off of some cutting-edge features by @imsnif in the past (as they can break for people who haven't updated their Rust in a while), but I'm personally up for merging whenever others are comfortable!
> 
> How do you mean? Will people need a higher version of rust than is in eg. apt?

I've not actually checked if it breaks older versions, so it might just be ignored, but it's a feature of the latest stable – 1.59.0 ([Blog](https://blog.rust-lang.org/2022/02/24/Rust-1.59.0.html#creating-stripped-binaries))

a-kenji:
> > I'm 100% on board with merging this! I've been warned off of some cutting-edge features by @imsnif in the past (as they can break for people who haven't updated their Rust in a while), but I'm personally up for merging whenever others are comfortable!
> 
> How do you mean? Will people need a higher version of rust than is in eg. apt?

They will "need" a higher version of `cargo` than what is in `apt`.

TheLostLambda:
@imsnif I've confirmed it breaks for at least 1.57 which was the version on my Ubuntu docker image
image

Maybe we just offer to migrate all of  our Ubuntu / Debian users to Arch? :wink:

imsnif:
> Maybe we just offer to migrate all of our Ubuntu / Debian users to Arch? wink

Best solution

a-kenji:
This is the update policy of `cargo` and `rustc` of ubuntu:

> We maintain rustc and cargo in ubuntu almost entirely to support building firefox. If you're a rust developer looking to work on your own rust programs, you are probably going to have a better experience (for now, at least) using rustup.  

https://wiki.ubuntu.com/FoundationsTeam/RustUpdates

raphCode:
Out of curiosity, why do Panic reports still include line numbers?
I expected them to get stripped away with the debug symbols, but they seem to have a different mechanism?

a-kenji:
Yes, the panic unwind is not yet really configurable on rust, especially stable rust.
It only strips the debug symbols / doesn't include them, which decreases binary size and can improve compilation time.

PR #1183, opened 2022-03-06
fix(comp): typo in long option '--generate-completion' for fish

FollieHiyuki:


FollieHiyuki:
@a-kenji I think you misread my comment https://github.com/zellij-org/zellij/commit/e2ce26121072adb8e4632f1f7de29a06f7e0662b#r68041394, so here is the PR.

a-kenji:
Thank you!

PR #1190, opened 2022-03-08
fix(compatibility): home and end in cursor keys mode

### LINKED ISSUE #1182: Home and End keys do not work in `helix` while using `zellij`
`zellij --version`: `zellij 0.25.0`
`tput lines`: `45`
`tput cols`: `unknown terminfo capability 'col'`
OS: `arch linux` `5.16.11-arch1-2`

List of programs interacted with: 
HELIX `helix --version`
`helix v0.6.0`
ALACRITTY `alacritty --version`
`alacritty 0.10.1 (2844606d)`

log displays all `INFO` events, nothing stands out as wrong.

Running `helix` in `alacritty` without `zellij` I am able to use `Home` and `End` keys while in `helix`'s insert mode to navigate the cursor to the beginning and end of the line I am editing. Running `zellij` and then `helix`, using those keys in the same way appear to do nothing. I also tried locking `zellij`, but this does not change the behavior.

tlinford:
When `DECCKM` is set, `Home` and `End` should also be adjusted. This builds upon the fix in https://github.com/zellij-org/zellij/pull/815 for vim compatibility, but should be also compatible with applications that don't expect the keys in `DECCKM` format.
Fixes #1182.

cc @imsnif

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -667,27 +667,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "dirs"
-version = "2.0.2"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "13aea89a5c93364a98e9b37b2fa237effbb694d5cfe01c5b70941f7eb087d5e3"
-dependencies = [
- "cfg-if 0.1.10",
- "dirs-sys",
-]
-
-[[package]]
-name = "dirs-sys"
-version = "0.3.6"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "03d86534ed367a67548dc68113a0f5db55432fdfbb6e6f9d77704397d95d5780"
-dependencies = [
- "libc",
- "redox_users",
- "winapi",
-]
-
-[[package]]
 name = "dirs-sys-next"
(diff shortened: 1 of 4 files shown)


imsnif:
Ah, awesome! I was going to take a look at it this week. I'll review it in the next day or two

tlinford:
Adding some details to clarify how this pr hopefully solves the problem without creating regressions.
With the changes in #815,  these sequences are always transformed:
- Home: `ESC [ H` -> terminfo entry for `khome`
- End: `ESC [ F` -> terminfo entry for `kend`

Most terminfos seem to define:
- `khome`: `ESC O H`
- `kend`: `ESC O F`

This fixed vim not working, and apparently many other programs read both these encodings fine. Helix on the other hand (via crossterm), does not recognize these `ESC O` variants.

With this change the transformation is only done when `DECCKM` has been set (this was already the case for the other cursor keys).

I removed the terminfo lookups since it would have been slightly annoying to attach it in that part of the code, and also because it's not being done for the existing `DECCKM` transformations we have, but it's worth keeping in mind that it's probably the right thing to do.

PR #1250, opened 2022-03-21
fix selection rendering of unstyled characters

### LINKED ISSUE #1208: Visual selecting selects to end of row, actual select is correct.
**In Case of Graphical or Performance Issues**

[zellij.log](https://github.com/zellij-org/zellij/files/8233945/zellij.log)
[zellij-9.log](https://github.com/zellij-org/zellij/files/8233946/zellij-9.log)

**Basic information**
`zellij --version`: zellij 0.26.0
`tput lines`: 33
`tput cols`: 88
`uname -av` or `ver`(Windows): Linux marvin 5.16.12-200.fc35.x86_64 #1 SMP PREEMPT Wed Mar 2 19:06:17 UTC 2022 x86_64 x86_64 x86_64 GNU/Linux
`alacritty --version`: alacritty 0.10.1

**Further information**
Selecting part of line visually selects to end of line but actually selects the characters as shown in the example gif
example

tlinford:
With a basic bash session the character style diffing was not emitting a reset when rendering the first character after the selection. By starting the cursor off with `RESET_STYLES`, the problem goes away, and have noticed no problems otherwise.

fixes #1208

cc @imsnif

### DIFF
--- a/zellij-server/src/panes/terminal_character.rs
+++ b/zellij-server/src/panes/terminal_character.rs
@@ -685,7 +685,7 @@ impl Cursor {
             x,
             y,
             is_hidden: false,
-            pending_styles: CharacterStyles::new(),
+            pending_styles: RESET_STYLES,
             charsets: Default::default(),
             shape: CursorShape::Initial,
         }


imsnif:
I haven't tested this, but I'm not sure this behaviour is correct. Looking at this briefly, it seems that if we have:

Two characters after each other, the first being red and the second being red and bold. This would:
1. Render the first character as red
2. Diff the two characters, find that they are both red and so do nothing with the foreground color
3. Find that the second one is bold, and add bold to the diff

This would then mean that we get a diff with the bold bit turned on and everything else reset. Meaning the second character would only be rendered as bold but not red.

I haven't tested this though, so maybe I'm missing something.

tlinford:
Does not seem to be the case - Otherwise I don't think we would properly handle an 'ESC [ 0 m` sequence? 

This is what I get with the example above:
screenshot-2022-03-21T11:26:55

imsnif:
As discussed offline - this change should be safe. I didn't notice it's happening in the `Cursor` constructor rather than the `CharacterStyles` constructor. That means it only happens at the first character when clearing a pane (eg. when starting a new pane or moving to an alternate screen).

PR #1264, opened 2022-03-24
Add navigation with alt + arrow keys

imsnif:
Recent changes in our STDIN parsing have broken the `Alt + []` navigation method. After some investigation, we could not find a method to bring these back without causing crashes for various systems. While it is theoretically possible to get these to work again, it would require a non-trivial amount of work and likely force us to write our own STDIN parser.

Because of this, in a previous PR we moved from parsing STDIN with termion to parsing it with termwiz. Termwiz supports alt + arrow key codes, which allow us to step out of this problem by providing a user friendly alternative.

This PR adds this alternative - allowing pane navigation with `Alt` + `<arrow keys>`.

The implementation is a little bit hacky in the configuration parsing front. I'm hoping @a-kenji can have a look and maybe find a better way.


a-kenji (approved):
I think this should be ready now!

imsnif:
Thanks @a-kenji ! I think it wasn't working for me because I tried to make it:

```rust
pub enum CharOrArrow {
    Char(char),
    Right,
    Left,
    Up,
    Down,
}
```

And gave up trying to figure what Serde wanted from me :) Much appreciated!

Issue #1261, opened 2022-03-23
Feature: `Actions` should give feedback.


a-kenji:
Actions that are launched should have the capability to communicate back in a oneshot fashion.

For example:
- Running a command -> exit codes
- Moving focus -> was it successful
- Querying state -> do we have hidden panes / how many panes are on tab 3?

This should be possible in:
- The configuration
- Plugins
- The client/server itself.

jaeheonji:
Personal opinion:

* Run specific command like `ls` / Move, Resize, Swap, Scroll -> return exit codes (0 or 1)
* Rename Pane or Tab -> return the changed name if successful or exit code 1
* New Pane or Tab -> return the created Pane/Tab ID if successful or exit code 1
* Query -> return result if successful or exit code 1 (the result would be good to use if it was output to the terminal in a specific format such as JSON?)

Issue #1195, opened 2022-03-09
Clearer path for multiple users in a session


a-kenji:
Currently every connected client is treated as a distinct user. This is not always desired. We can improve on this assumption.

Make it possible to specify a user on attach?
Take the `USER` in to account?
Optionally show a username in the pane frames?

PR #1265, opened 2022-03-24
feat: add capability to dispatch actions from cli

a-kenji:
Add capability to dispatch actions from the cli.

Can be invoked through `zellij action [actions]`

Automatically sends the action either to the current session,
or if there is only one session to the single session.

If there are multiple sessions, and no session is specified it will
error out.

Special Consideration for the format of the actions:

The format should not be changed since they are handled by our action parser.
They will automatically get upgraded once we move over to kdl in #771 ,
which is inherently suited for that task.
So if someone wants to improve the action parsing here, they would need to implement #771.

Example:
1.
```
zellij action "NewTab: , NewTab: "
```
2.
```
zellij -s fluffy-cat action 'NewPane: , WriteChars: "echo Purrr\n"'
```
3.
```
zellij -s fluffy-cat action CloseTab
```
4.
```
zellij action NewTab:
```
This is still an early draft and a lot of errors are not yet handled.

Most actions here should work, some need special care from this pr.
Others should probably be changed regardless, so that they will work in combination to this pr. 

For example:
The `action` `ScrollUp` does work, but panics, if we send the action and there is no room to scroll anymore. The `action` itself should not be able to panic here and the server should just recognize that - we could log here for now, until something like #1261 is implemented.

These actions are tested and work as expected:
- [x] Quit
- [x] Detach
  - [x] works with one client
  - [x] still needs handling for more than one client
- [x] MovePane: Direction
- [x] DumpScreen: File
- [x] SwitchToMode: InputMode
- [x] Resize: Direction
- [x] FocusNextPane
- [x] FocusPreviousPane
- [x] SwitchFocus
- [x] MoveFocus: Direction
- [x] ScrollUp
- [x] ScrollDown
- [x] PageScrollUp
- [x] PageScrollDown
- [x] ToggleFocusFullscreen
- [x] TogglePaneFrames
- [x] ToggleActiveSyncTab
- [x] ToggleTab
- [x] GoToTab
- [x] ToggleFloatingPanes
- [x] TogglePaneEmbedOrFloating
- [x] NewPane: Direction
- [x] CloseFocus
- [x] NewTab: TabLayout
  - [x] 'NewTab: ' works mostly
  - [x] Plugins in `NewTab`, don't get their state propagated correctly yet
  - [x] Selected Pane frame, doesn't have correct colors
  - [x] NewTab: TabLayout is wonky, probably the options need to be mapped in the `fake_client`, or maybe has to do with: #1267 ?
- [x] GoToNextTab
- [x] GoToPreviousTab
- [x] CloseTab
- [x] Run: {cmd: path}
- [x] Write: bytes
- [x] WriteChars: characters

Improve cli:
- [x] Pretty print error message
- [x] Make `[]` opaque to the user
- [x] Automatically pick session, if there is only one
- [ ] Maybe make it possible to send actions to specific clients?
- [x] Automatically send action to current session
- [x] Automatically send action to current client

Other:
- [ ] Hide `fake_client` from ui

@jaeheonji 
I would be happy about any input here!


jaeheonji:
I did some tests on each item and found out that both "working well" and "not working" are related to `client_id`.

As you may already know, for all actions to operate normally, the correct `client_id` must be sent to the server. Therefore, it seems that two things are needed.

1. A request from the server to return a list of `client_id` connected to the current session.
2. `Action` requires `client_id`

```rust
// zellij-utils/src/ipc.rs

// Types of messages sent from the client to the server
#[allow(clippy::large_enum_variant)]
#[derive(Serialize, Deserialize, Debug, Clone)]
pub enum ClientToServerMsg {
    ...
    Action(Action, Option<ClientId>),
    ...
}

...

#[derive(Serialize, Deserialize, Debug, Clone)]
pub enum ServerToClientMsg {
    ...
    SessionState(Vec<ClientId>)
}
```
If it seems to be ok, I will try to update the code. Or please let me know if you are already in the process or have any other comments.

a-kenji:
I will look at it later, everything you said so far makes sense to me. I would be happy if you implemented your ideas.

a-kenji:
@jaeheonji,
We probably shouldn't surface the different client_id's to the user, or?

I feel that is maybe a too complicated setup for people.

jaeheonji:
I understand your concern.

However, if we do not expose the client_id, we will need a different approach.

First of all, the problem is that client_id is simply a continuous number. We talked about this once before.

When client_id 1 is already connected to the currently running session, if the action command is executed, the fake_client performs the operation with client_id 2. And, when the operation is performed and terminated, information about client_id 2 disappears from the session. For this reason, commands such as GoToNextTab cannot be executed at this time. Because the command is executed for client_id that do not exist. And this becomes more complicated in shared sessions.

If we don't consider shared sessions, it might be easy to solve, but otherwise we'll have to think a little more.

a-kenji:
> I understand your concern.
> 
> However, if we do not expose the client_id, we will need a different approach.

Offline, imsnif and me talked about maybe adding a headless client, that would by default not be considered a different user. 

> First of all, the problem is that client_id is simply a continuous number. We already talked about this once before.

> When client_id 1 is already connected to the currently running session, if the action command is executed, the fake_client performs the operation with client_id 2. And, when the operation is performed and terminated, information about client_id 2 disappears from the session. For this reason, commands such as GoToNextTab cannot be executed at this time. Because the command is executed for client_id that do not exist. And this becomes more complicated in shared sessions.
> 
> If we don't consider shared sessions, it might be easy to solve, but otherwise we'll have to think a little more.

Yes, I agree. This is a problem. Maybe it makes sense to fix it here? Currently it might make more sense to send the actions to each `client_id`, that is not a fake client. 
And once we have more intricate differentiation between:

- different clients by the same user
- different clients per user

We can improve upon that further. Maybe we do find a good solution here, that would of course be preferable!

I'm a different branch I tried setting the `ZELLIJ` environment variable inside the addition of the `client_id`'s. Maybe that could be a good starting point?

jaeheonji:
> Yes, I agree. This is a problem. Maybe it makes sense to fix it here? Currently it might make more sense to send the actions to each client_id, that is not a fake client.
> And once we have more intricate differentiation between:
> 
> different clients by the same user
> different clients per user
> We can improve upon that further. Maybe we do find a good solution here, that would of course be preferable!

Yes! I agree 100% and for each case, I will think deeply about it.

> I'm a different branch I tried setting the ZELLIJ environment variable inside the addition of the client_id's. Maybe that could be a good starting point?

Good news! Actually I was already thinking about this.
Yes your right, If the client_id is assigned to an environment variable inside the session, the user can perform the action without worrying about client_id. So, in this case, I thought that it could be implemented without any other problems 👍

a-kenji:
> Yes your right, If the client_id is assigned to an environment variable inside the session, the user can perform the action without worrying about client_id. So, in this case, I thought that it could be implemented without any other problems 👍 

This would solve the case for running it inside the already running client and also if we just have one client. Which should cover many cases already. 
I guess if there are more client id's we can use flags for them.

a-kenji:
> To make this a little bit clearer, I think this is equivalent to #1195. right?

Yes, it can be! 

> As you already mentioned in #1195, I think it doesn't seem like a big problem to determine the client based on `$USER`.

Agreed, but in that case we will need (IMO), a flag that let's a person
connect as a distinct client, even if they are the same `$USER`.
If only to make it possible for us to debug the interface.

> When the same user connects to the same session, it is treated as a `mirror session`. In case of different users, it will be treated as a `shared session`.
> 
> What do you think of this plan?

It would make it easier, tough I like the idea of 2 distinct clients from the same user attaching to a session that is not mirrored. (2 screens) for example. So if we at some point could toggle between `shared` and `mirrored` that would be optimal.

a-kenji:
Oh I just realized in #1287, we can't distinguish by `$USER`, because it only works with one user.

Apart from that I think your suggestion is excellent! And I think we should go that route. Then we should let the people specify the user as a flag for now, and in the future when we actually support multiple `$USER`, it is already compatible.

jaeheonji:
Oh.. Using $USER was also related to security issue. I got it.
then, we'll need a different approach. cool, it is a pretty fun challenge 😸

a-kenji:
@jaeheonji,
I would be happy if you could look over this again!
This is still not fully cleaned up, but the only actions, that still need figuring out are:
`Detach` and `NewTab`.

jaeheonji:
Ok! I'll check it out too :)

jaeheonji:
@a-kenji 

I looked into the reasons why `Detach` and `NewTab` do not work normally.

First, About color for Pane frame in `NewTab` action, This is a problem caused by setting option to default when calling fake_client.
https://github.com/zellij-org/zellij/blob/5dc43e91e4d859998b41dc630fafad733df6cd19/src/commands.rs#L153-L163

So, I don't think this is a big deal. We can easily fix this later.

But, the biggest problem is `Detach`. This problem is also related to the `client_id`. The reason why `Detach` is not working properly is that when the `input_handler` processes the `Detach` action, it **detach** the **last client** in the client list.

And since the last client_id is always the `fake_client`, the client connecting to the actual server is not detached, and the fake_client is detached.

Of course, there is no problem if we modify the specific `client_id` to be input. However, this leads back to the original problem that `client_id` cannot be specified.

So I'm thinking about creating a random name like a session name instead of a `client_id`. For example:
```bash
zellij # start client using random name such as `geust-01`
zellij --user awesome-zellij # start client using specific user name
```

Of course this is a different subject. And I'm not sure if these ideas are really a good way yet.

a-kenji:
@jaeheonji, 
Thank you for looking over the pr!

> First, About color for Pane frame in NewTab action, This is a problem caused by setting option to default when calling fake_client.

Yes, but Ideally this option wouldn't matter at all. And I can't figure out why it does.
Even when sending the `NewTab` action to the correct client. It depends on the options that are specified in the `fake_client`.


> But, the biggest problem is Detach. This problem is also related to the client_id. The reason why Detach is not working properly is that when the input_handler processes the Detach action, it detach the last client in the client list.

Yes, but that is also not 100% the issue here. I switched the client id's around here for testing, if I detach from the correct one then we get some weird state problems. 

Thanks for checking it out!

a-kenji:
@jaeheonji,

Do you mind looking over it again?
Most of the issues are addressed now.
But there are still edge cases that need to be addressed.
I am thinking about merging this soon, so people can test the feature
and more people can improve it, what do you think?

jaeheonji (approved):

a-kenji:
@jaeheonji,
Awesome, thank you for the review and the guidance through the whole pr!
I hope we can improve this functionality further.

Issue #1134, opened 2022-02-27
stdin input before rendered crashes zellij


DieracDelta:
# In Case of Graphical or Performance Issues

1. Delete the contents of `/tmp/zellij-1000/zellij-log`, ie with `cd /tmp/zellij-1000/` and `rm -fr zellij-log/`
2. Run `zellij --debug`
3. Recreate your issue.
3. Quit Zellij immediately with ctrl-q (your bug should ideally still be visible on screen)

log output:

```
INFO   |zellij_client            | 2022-02-26 23:07:11.464 [main      ] [zellij-client/src/lib.rs:121]: Starting Zellij client!
INFO   |zellij_server            | 2022-02-26 23:07:11.466 [main      ] [zellij-server/src/lib.rs:188]: Starting Zellij server!
INFO   |zellij_server::wasm_vm   | 2022-02-26 23:07:11.517 [wasm      ] [zellij-server/src/wasm_vm.rs:86]: Wasm main thread starts
INFO   |zellij_server::wasm_vm   | 2022-02-26 23:07:11.520 [wasm      ] [zellij-server/src/wasm_vm.rs:225]: wasm main thread exits
```

# Basic information

`zellij --version`: on both 0.24.0 and 01f7f4f3d2a83b8f3af90d9e6a85371d988c9f2a (so, HEAD as of when this issue was created)
`tput lines`:67
`tput cols`: 126
`uname -av` or `ver`(Windows): Linux desktop 5.16.10 #1-NixOS SMP PREEMPT Wed Feb 16 11:58:51 UTC 2022 x86_64 GNU/Linux

List of programs you interact with as: N/A/none

# To reproduce

`zellij`

Then mash the keyboard immediately after. This results in unwrapping a none value with the following stack trace:

```
as
Error occurred in server:
Originating Thread(s):
1. stdin_handler_thread: AcceptInput
2. screen_thread: ClearScroll

Error: thread 'screen' panicked at 'called `Option::unwrap()` on a `None` value': zellij-server/src/screen.rs:1062
   0: zellij_utils::errors::handle_panic
   1: std::panicking::rust_panic_with_hook
             at /rustc/9d1b2106e23b1abd32fce1f17267604a5102f57a/library/std/src/panicking.rs:610:17
   2: std::panicking::begin_panic_handler::{{closure}}
             at /rustc/9d1b2106e23b1abd32fce1f17267604a5102f57a/library/std/src/panicking.rs:500:13
[...]

Issue #805, opened 2021-10-26
Error: thread 'screen' panicked at 'called `Option::unwrap()` on a `None` value': zellij-server/src/screen.rs:226


happysalada:
Thank you for taking the time to file this issue! Please follow the instructions and fill the missing parts below the instructions, if it is meaningful. Try to be brief and concise.

Please attach the files that were created in `/tmp/zellij-1000/zellij-log/` to the extent you are comfortable with.

**Basic information**

`zellij --version`: 0.18.1
`tput lines`:
`tput cols`:
`uname -av` or `ver`(Windows): Darwin yt 20.6.0 Darwin Kernel Version 20.6.0: Mon Aug 30 06:12:21 PDT 2021; root:xnu-7195.141.6~3/RELEASE_X86_64 x86_64

List of programs you interact with as, `PROGRAM --version`: output cropped meaningful, for example:
`alacritty --version`: alacritty 0.9.0

**Further information**
This happens upon trying to close a tab. When hitting the `x` of closing a particular tab, sometimes I just get the crash

I can't reproduce reliably, this bug happens just from time to time. The logs without the debug flag contain the following
```
INFO   |zellij_client            | 2021-10-24 21:00:29.051 [main      ] [zellij-client/src/lib.rs:107]: Starting Zellij client! 
INFO   |zellij_server            | 2021-10-24 21:00:29.065 [main      ] [zellij-server/src/lib.rs:174]: Starting Zellij server! 
INFO   |zellij_server::wasm_vm   | 2021-10-24 21:00:29.122 [wasm      ] [zellij-server/src/wasm_vm.rs:73]: Wasm main thread starts 
INFO   |zellij_server::wasm_vm   | 2021-10-25 22:17:59.637 [wasm      ] [zellij-server/src/wasm_vm.rs:148]: Bye from plugin 4 
INFO   |zellij_server::wasm_vm   | 2021-10-25 22:17:59.672 [wasm      ] [zellij-server/src/wasm_vm.rs:148]: Bye from plugin 5 
INFO   |zellij_server::wasm_vm   | 2021-10-25 22:18:00.630 [wasm      ] [zellij-server/src/wasm_vm.rs:148]: Bye from plugin 2 
INFO   |zellij_server::wasm_vm   | 2021-10-25 22:18:00.640 [wasm      ] [zellij-server/src/wasm_vm.rs:148]: Bye from plugin 3 
[...]

PR #1269, opened 2022-03-25
fix(screen): handle various edge cases rather than crashing

imsnif:
Fixes https://github.com/zellij-org/zellij/issues/1134, https://github.com/zellij-org/zellij/issues/805 and probably a few others

When this code was written, the Zellij architecture was a little bit different and it didn't make a whole lot of sense to perform checks to make sure the Tab a certain Action is operating on actually exists. As we moved stuff around (namely to a client/server architecture but not just), these edge-cases became more and more possible.

This fixes a lot (all?) of these potential pitfalls by performing a no-op in case the tab an action would like to use doesn't exist.

### DIFF
--- a/zellij-server/src/screen.rs
+++ b/zellij-server/src/screen.rs
@@ -312,60 +312,63 @@ impl Screen {
     /// A helper function to switch to a new tab at specified position.
     fn switch_active_tab(&mut self, new_tab_pos: usize, client_id: ClientId) {
         if let Some(new_tab) = self.tabs.values().find(|t| t.position == new_tab_pos) {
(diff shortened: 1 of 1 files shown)


a-kenji:
> This fixes a lot (all?) of these potential pitfalls by performing a no-op in case the tab an action would like to use doesn't exist.

Can we log these instances, maybe just with `debug`?

imsnif:
> > This fixes a lot (all?) of these potential pitfalls by performing a no-op in case the tab an action would like to use doesn't exist.
> 
> Can we log these instances, maybe just with `debug`?

Good call, thanks. Added.

PR #1274, opened 2022-03-26
Improve theme usage + add default ones

kunalmohan:
- We no longer use gray color in the plugins and is no longer required in the theme.

Samples-

Solarized Dark
solarized-dark

One Half Dark
onehalf-dark

Molokai Dark
molokai-dark

Gruvbox Light
gruvbox-light

Gruvbox Dark
gruvbox-dark

Dracula
dracula


imsnif:
This is fantastic! One minor issue: this changed the default colors for me.

Before:
img-2022-03-31-091133

After:
img-2022-03-31-091110

Was this intentional?

kunalmohan:
> Was this intentional?

Yeah I switched the background color of the status bar from gray to black (since gray is removed no longer specified in the theme). Maybe we can set a lighter shade of black in the [default_palette](https://github.com/zellij-org/zellij/blob/843919863610d9a5426bf6c734aeda69d127723c/zellij-utils/src/shared.rs#L55-L72), if that's better?

imsnif:
Hum - would that affect other stuff? Do we use this black color somewhere else in our UI?

kunalmohan:
Ah..right. Black is used for text on the status bar. Using a lighter color may reduce the visibility of the text as well. But we can still change it to 236, it is lighter and still properly visible.

image

imsnif:
Hmm... still not great. I just feel it would be a pity if we have to change our UI because we changed our API. Do you feel there's no other way?

imsnif:
Okay, that's fair. I understand the tradeoffs now and when you put it this way, I totally agree. Let's go with 236. Otherwise this LGTM. Great work on this!

imsnif (approved):

bluss:
Thanks for adding themes. My Zellij usage can now start - with a theme that works for a light terminal. :slightly_smiling_face:

PR #1298, opened 2022-04-05
feat(copy): add copy_on_select option

### LINKED ISSUE #1192: How to disable copy on selection and change the color of selection?
Can I disable copy-on-selection without disabling the mouse mode in zellij? Besides, how can I change the color of selection?

Thanks

tlinford:
Add `copy_on_select` option:
-  With default value of `true`, keep current behavior of automatically
copying selection to clipboard when done selecting.
-  when `false`, on mouse release the selection remains, and
can be copied with the `Copy` action.

Resolve #1192


tlinford:
@a-kenji is there anything that should be added in default.yaml for the `Copy` action?
I don't think we should enable it by default but wondering if it should maybe be there commented?

a-kenji:
@tlinford,
Maybe we can add it as a comment.
But it should definitely be added to the documentation.

PR #1304, opened 2022-04-08
add: nord theme to example

smonv:


tlinford:
awesome, thanks for this!

PR #1320, opened 2022-04-13
Fix #1286 by sending SIGHUP when closing a pane

### LINKED ISSUE #1286: Zombie Process on Tab/Pane Close
I'm seeing zombie processes on tab/pane close.  Found out about it through vim open buffers on reopening a previously closed tab that had the same file open. (duplicate/closed issue https://github.com/zellij-org/zellij/issues/518)

## Basics
zellij: 0.26.1
os/arch: Darwin/arm64 (M1)
alacritty: 0.11.0-dev
shell: zsh 5.8

## Repro Steps:
- start zellij
- create new pane or tab
- open vim
- close pane or tab
- ps aux | grep vim

raphCode:
SIGHUP correctly states the intention behind sending a signal when a
pane is closed: The controlling terminal is "hung up".
Also, SIGHUP is better suited than SIGTERM since bash ignores the
latter. This led to the zombie processes observed by some users.
Furthermore, SIGHUP has a special meaning in bash's job control, namely
re-sending the signal to all owned jobs before exiting.

### DIFF
--- a/zellij-server/src/os_input_output.rs
+++ b/zellij-server/src/os_input_output.rs
@@ -300,7 +300,7 @@ impl ServerOsApi for ServerOsInputOutput {
         Box::new((*self).clone())
     }
     fn kill(&self, pid: Pid) -> Result<(), nix::Error> {
-        let _ = kill(pid, Some(Signal::SIGTERM));
+        let _ = kill(pid, Some(Signal::SIGHUP));
         Ok(())
     }
     fn force_kill(&self, pid: Pid) -> Result<(), nix::Error> {


imsnif:
Thank you very much for this! I'm not sure I'll get a chance to fully test this today - worst case will tomorrow.

imsnif:
Tested this a little bit and it LGTM. Thanks once again for looking into this issue!

raphCode:
Thanks for merging!

> LGTM

I am sure you mean this in a positive way, but I found it surprising to find out that actually multiple and contrary definitions exist:
- lets good to me
- lets get this merged
- literally garbage to me

https://www.urbandictionary.com/define.php?term=LGTM

imsnif:
:D Looks Good To Me

PR #1351, opened 2022-04-27
Fix empty ipc msg crash

imsnif:
This fixes some edge cases in which the ipc message buffer (between client and server) was corrupted and Zellij would crash as a result. Now, instead of crashing we will log the error and attempt to re-read the channel.

This might be masking other problems (maybe the channel buffer is too small? and there are times when it is filled up?) but in any case these errors should be almost invisible to the user (one render missed or one keypress missed when the load is heavy, etc.)

If this is indeed masking other problems, it'll be much easier to find them and fix them individually now that we log these.

### DIFF
--- a/src/sessions.rs
+++ b/src/sessions.rs
@@ -58,10 +58,13 @@ fn assert_socket(name: &str) -> bool {
         Ok(stream) => {
             let mut sender = IpcSenderWithContext::new(stream);
             sender.send(ClientToServerMsg::ConnStatus);
-
             let mut receiver: IpcReceiverWithContext<ServerToClientMsg> = sender.get_receiver();
-            let (instruction, _) = receiver.recv();
-            matches!(instruction, ServerToClientMsg::Connected)
+            match receiver.recv() {
+                Some((instruction, _)) => {
+                    matches!(instruction, ServerToClientMsg::Connected)
+                },
+                None => false
+            }
         }
         Err(e) if e.kind() == io::ErrorKind::ConnectionRefused => {
             drop(fs::remove_file(path));
--- a/src/sessions.rs
+++ b/src/sessions.rs
@@ -62,8 +62,8 @@ fn assert_socket(name: &str) -> bool {
             match receiver.recv() {
                 Some((instruction, _)) => {
                     matches!(instruction, ServerToClientMsg::Connected)
-                },
-                None => false
+                }
+                None => false,
             }
         }
         Err(e) if e.kind() == io::ErrorKind::ConnectionRefused => {
--- a/zellij-client/src/lib.rs
+++ b/zellij-client/src/lib.rs
@@ -280,9 +280,11 @@ pub fn start_client(
                         if should_break {
                             break;
                         }
-                    },
+                    }
                     None => {
-                        send_client_instructions.send(ClientInstruction::UnblockInputThread).unwrap();
+                        send_client_instructions
+                            .send(ClientInstruction::UnblockInputThread)
+                            .unwrap();
                         log::error!("Received empty message from server");
                     }
                 }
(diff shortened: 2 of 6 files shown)


ahirner:
I `cargo install --git https://github.com/zellij-org/zellij --branch fix-empty-ipc-msg-crash zellij` and ran the command in the issue ~20 times without error. If it wasn't fixed, I expected it to crash ca. every 2nd time.

`zellij.log` ended without error logs starting with `zellij --debug` (as far as I understand how errors would be logged):
```
INFO   |zellij_client            | 2022-04-27 10:47:27.209 [main      ] [zellij-client/src/lib.rs:123]: Starting Zellij client!
INFO   |zellij_server            | 2022-04-27 10:47:27.215 [main      ] [zellij-server/src/lib.rs:189]: Starting Zellij server!
INFO   |zellij_server::wasm_vm   | 2022-04-27 10:47:27.269 [wasm      ] [zellij-server/src/wasm_vm.rs:87]: Wasm main thread starts
```

I no try, `CTRL-C` was ignored/dropped. It fixes the issues for me.

imsnif:
Thanks for giving it a try (and so quickly!) @ahirner 
I'm not sure why this didn't log - but at least happy to see the issue is fixed for you.

a-kenji:
> I'm not sure why this didn't log - but at least happy to see the issue is fixed for you.

https://github.com/zellij-org/zellij/pull/1351/commits/8fbd167b0f75997a4c4d715fa988daafb8a715d7#diff-946b5111f582dca9fe6de768ae1636061984c7037c612e83c35980d9326218deR190

The error message is ignored here, or? So it can't be logged.

imsnif:
> > I'm not sure why this didn't log - but at least happy to see the issue is fixed for you.
> 
> [8fbd167#diff-946b5111f582dca9fe6de768ae1636061984c7037c612e83c35980d9326218deR190](https://github.com/zellij-org/zellij/commit/8fbd167b0f75997a4c4d715fa988daafb8a715d7#diff-946b5111f582dca9fe6de768ae1636061984c7037c612e83c35980d9326218deR190)
> 
> The error message is ignored here, or? So it can't be logged.

I was hoping it would be logged here: https://github.com/zellij-org/zellij/commit/8fbd167b0f75997a4c4d715fa988daafb8a715d7#diff-6f77a56f2f7ebfaa89005876b659af0aa348b809cf0eebda1814e1c7454a637cR480

Becuase we return None there - but maybe I'm missing something

a-kenji:
> > > I'm not sure why this didn't log - but at least happy to see the issue is fixed for you.
> > 
> > 
> > [8fbd167#diff-946b5111f582dca9fe6de768ae1636061984c7037c612e83c35980d9326218deR190](https://github.com/zellij-org/zellij/commit/8fbd167b0f75997a4c4d715fa988daafb8a715d7#diff-946b5111f582dca9fe6de768ae1636061984c7037c612e83c35980d9326218deR190)
> > The error message is ignored here, or? So it can't be logged.
> 
> I was hoping it would be logged here: [8fbd167#diff-6f77a56f2f7ebfaa89005876b659af0aa348b809cf0eebda1814e1c7454a637cR480](https://github.com/zellij-org/zellij/commit/8fbd167b0f75997a4c4d715fa988daafb8a715d7#diff-6f77a56f2f7ebfaa89005876b659af0aa348b809cf0eebda1814e1c7454a637cR480)
> 
> Becuase we return None there - but maybe I'm missing something

Ah, that makes sense. Thanks!

raphCode:
Great to see this fixed, good job!

Did you understand why a hang did occur in the original bug report?
In my understanding, some `read()` was blocking somewhere, the diff doesn't explain why that is avoided now.

imsnif:
My guess is that more than one message was written to the channel buffer before it could be read, but I honestly don't know.

Issue #1323, opened 2022-04-13
Thread 'screen' panicked when attempting to rename a floating panel


DannyBen:
**Basic information**

`zellij --version`: `0.28.1`
`tput lines`: `34`
`tput cols`: `137`
`uname -av` or `ver`(Windows): `Linux ubuntu-focal 5.4.0-107-generic #121-Ubuntu SMP Thu Mar 24 16:04:27 UTC 2022 x86_64 x86_64 x86_64 GNU/Linux`

**Further information**

I have seen many panic issues, so I hope this is not a duplicate. If it is, feel free to close it.

**Error**

```
  × Thread 'screen' panicked.
  ├─▶ Originating Thread(s)
  │     1. stdin_handler_thread: AcceptInput
  │     2. screen_thread: UpdatePaneName
  │
  ├─▶ At zellij-server/src/tab/mod.rs:1740:18
  ╰─▶ called `Option::unwrap()` on a `None` value
  help: If you are seeing this message, it means that something went wrong.
        Please report this error to the github issue.
        (https://github.com/zellij-org/zellij/issues)

        Also, if you want to see the backtrace, you can set the `RUST_BACKTRACE` environment variable to `1`.
```

then I rerun with backtrace on, hoping it would help the debugging effort, but it seems to not add any new infomration:

```
Error occurred in server:

  × Thread 'screen' panicked.
  ├─▶ Originating Thread(s)
  │     1. stdin_handler_thread: AcceptInput
  │     2. screen_thread: UpdatePaneName
  │
  ├─▶ At zellij-server/src/tab/mod.rs:1740:18
  ╰─▶ called `Option::unwrap()` on a `None` value
         0: <unknown>
        ..: <...trimmed - all unknown>
        13: <unknown>
```

**Reproduction steps**

1. Start zellij after a fresh install (no config)
2. Create a new floating pane <kbd>Ctrl</kbd>+<kbd>P</kbd> <kbd>W</kbd>
3. Rename the floating pane <kbd>Ctrl</kbd>+<kbd>P</kbd> <kbd>C</kbd>
4. Observe the panic

raphCode:
Thanks for reporting this, server stability aka not-crashing-sessions a big concern for us! :)

I can reproduce this an Arch.

Link to crash location:
https://github.com/zellij-org/zellij/blob/v0.28.1/zellij-server/src/tab/mod.rs#L1735-L1749

imsnif:
[...]

PR #1357, opened 2022-04-28
Fix crash on renaming a floating pane (#1323)

raphCode:


raphCode:
Ideally we add some tests to avoid regressions, but to be honest, I hate writing tests and also the huge e2e test functions scared me away.
It looks like some sort of [Tcl expect](https://en.wikipedia.org/wiki/Expect) just for terminal UI that updates via ANSI sequences?

imsnif:
> Ideally we add some tests to avoid regressions, but to be honest, I hate writing tests and also the huge e2e test functions scared me away. It looks like some sort of [Tcl expect](https://en.wikipedia.org/wiki/Expect) just for terminal UI that updates via ANSI sequences?

That would be great! The e2e tests are indeed a bit of a handful (as such tests tend to be), but for this maybe you can add an integration test? https://github.com/zellij-org/zellij/blob/main/zellij-server/src/tab/unit/tab_integration_tests.rs

Wdyt?

raphCode:
Added some test
---

In case you are interested in my experiences:

It looked difficult at first, because I did not find a definiton of `assert_snapshot!`. The insta testing library make things actually much easier and also allows to view the rendered zellij UI. Nice!

imsnif (approved):
Breaking `Tab` into like... 17 different modules is totally on my list for the not too far future, I very much agree.

Otherwise about the tests - this is good feedback. Do you have suggestions about what we should change that would have made your experience easier?

This LGTM. Please feel free to merge, and don't forget to update the changelog.

raphCode:
> Do you have suggestions about what we should change that would have made your experience easier?

Some high level dev docs or a guide on e.g. how to add a test would be super nice. But I totally understand if this maybe is too much to ask for, many projects do not have this and it can also be a maintenance burden since dev docs must be kept up-to-date on code changes.

In general, when doing something for the first time and having no idea on how to approach it, it costs me quite a bit willpower and motivaton to work my way into it. A guide lowers this barrier because it shows a potential path through the problem space and familiarizes me with any new tools involved.
In fact, https://insta.rs/docs/quickstart/ was enough to get to know insta and become comfortable with its workflow. From there, I worked my way backwards through the tests by reverse engineering them: seeing what each line did change in the snapshot.

Issue #1169, opened 2022-03-04
Faster way or conveniences for attaching to session by name


colemickens:
Hi,

It looks like clap does not *yet* support runtime-provided completion support (https://github.com/clap-rs/clap/issues/1232), hence I'm not sure there's an easy path to providing completion for session names.

However, attaching to a session implies that someone is loading the entire zellij UI interactive inactively anyway.

What if `zellij attach` or `zellij attach -m` offered an interactive menu for the user to select a session?

Alternatively, would it be possible to allow `zellij attach f` to match a session named `foobar-awesome`, assuming there are no other sessions starting with `f` ?

I guess one last no-zellij-change-needed option would be to setup some sort of shell keybind so that I could pipe `zellij list-sessions` into `skim` or something? Does someone else have something they like for this?

a-kenji:
Hey! Thank you for taking the time for opening this issue.

>hence I'm not sure there's an easy path to providing completion for session names. 

There is sadly not, but we could append functions to the completions.
The negative here would be that this is very shell dependent, but it could be reasonable to support the shells as good as we can.
There is an issue that has a great example: #1030


> What if `zellij attach` or `zellij attach -m` offered an interactive menu for the user to select a session?

That would be awesome, I think connecting to your last point, maybe we could even support fzf/skim and family choosers similarly.
That way the maintenance burden would be not as high with still full functionality.

> Alternatively, would it be possible to allow `zellij attach f` to match a session named `foobar-awesome`, assuming there are no other sessions starting with `f` ?

So, that is another possibly good idea. Just for the first letter, and maybe restrict the possibility of automatically generating the same first letter, in order to make this functionality more useful.

> I guess one last no-zellij-change-needed option would be to setup some sort of shell keybind so that I could pipe `zellij list-sessions` into `skim` or something? Does someone else have something they like for this?

In my window manager I have bound something similar to this
(I have removed some very specific integration I use):
```
#!/usr/bin/env bash

ZJ_SESSIONS=$(zellij list-sessions)
NO_SESSIONS=$(echo "${ZJ_SESSIONS}" | wc -l)
TERMINAL=alacritty

if [ "${NO_SESSIONS}" -ge 2 ]; then
    "${TERMINAL}" -e zellij attach \
    "$(echo "${ZJ_SESSIONS}" | rofi -dmenu -l "${NO_SESSIONS}")"
else
   "${TERMINAL}" -e zellij attach -c
fi
```
[...]

PR #1360, opened 2022-04-28
feat: Attach sessions by unique prefix name (#1169)

raphCode:
This missing feature bothered me for a while, see also #1169.
Maybe I also just wanted to have some fun with Rust's fantastic `match` blocks :D

Ideas what could be improved:
- generate sessions names that avoid having the same starting characters (probably won't happen often in practice anyway)
- eliminate `session_exists()` by using my new `match_session_name()` at callsites (no idea if this gives a benefit)
- add tests
- you tell me

---

Commit message body:

This make attaching to sessions more convenient since only the first character(s) of the
session name must be typed.
If this prefix matches multiple sessions and is therefore ambiguous, zellij will complain
and show all sessions names starting with these characters.
If any session name matches the given string exact, it is attached immediately, therefore
it is always possible to attach to every session, even if the set of session names is not
prefix-free.


raphCode:
wtf, a 2 character string message change made the e2e test pass again

PR #1373, opened 2022-05-03
more verbose `unwrap()` expressions in mod.rs

amir-arad:
added `unwrap!` macro that writes a message to log before (or even, instead of) panicking.

### DIFF
--- a/zellij-server/src/tab/mod.rs
+++ b/zellij-server/src/tab/mod.rs
@@ -56,6 +56,44 @@ macro_rules! resize_pty {
         }
     };
 }
+macro_rules! unwrap {
+    (option $value:expr; log $message:expr) => {
+        match $value {
+            None => {
+                log::error!($message);
+            }
+            Some(p) => p,
+        }
+    };
+    (option $value:expr; panic $message:expr) => {
+        match $value {
+            None => {
+                log::error!($message);
+                log::error!("panicking!");
+                panic!($message);
+            }
+            Some(p) => p,
+        }
+    };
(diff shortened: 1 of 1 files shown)


jaeheonji:
Leaving a log when an unexpected error occurs will definitely help with troubleshooting.

But personally, I think it will be easier to access by using `panic hook`. And we can leave a detailed explanation using `expect `instead of `unwrap`.

For example, the client and server entry have hooks for all panic codes as follows:
https://github.com/zellij-org/zellij/blob/7b0a46f812f0c0f60068742ea2984b88b6ad8ed5/zellij-server/src/lib.rs#L207-L213

imsnif:
Hey @jaeheonji - this PR is part of some work me and @amir-arad have started doing to lay the groundwork for an error handling and recovery system.

This macro is a temporary measure and will (probably) eventually lead to a `handle_error` method of some kind in which, depending on the error, we will decide whether to panic or handle the error in some way (eg. just log it and move on).

I understand your wish to place this in a panic hook - for sure, if this were the only thing we were planning to do it would make sense - but this is not necessarily the way we will handle errors in the future. So I think for now we can do it this way and slowly move forward to a better system. Wdyt?

jaeheonji:
@imsnif Oh ok, there was a history I didn't know about 😅. If this is an intermediate step towards a better error handling and recovery system, No problem, I like it.

imsnif:
Hey, what's the status of this?

amir-arad:
Sorry, I didn't follow through. closing.

PR #1384, opened 2022-05-06
fix: macos ansi instruction parsing

### LINKED ISSUE #1381: Characters written to console when Terminal.app is resized
I can't tell if this is specific to my setup or not given I just noticed it today. When resizing the Terminal.app on macOS the following string is printed out `10;rgb:c5c5/c8c8/c6c6`

image

tlinford:
fix #1381


imsnif:
Hey @tlinford - I think I fixed the ctrl-g issue. Would you mind checking if it's fixed and also giving this a thorough whirl to make sure the issue is still fixed?

tlinford:
@imsnif tested on macos Terminal + Wezterm, LGTM.

PR #1387, opened 2022-05-09
fix(deps): Rename `suggestion` crate to `suggest`

ken-matsui:
I changed the name of the `suggestion` crate to `suggest` to be shortened.


a-kenji:
Thanks for updating!

PR #1426, opened 2022-05-21
Merge --layout and --layout-path

### LINKED ISSUE #1412: Consider merging --layout and --layout-path
When trying to use layouts for the first time and reading through the docs, I missed the distinction between these two arguments. I had tried to use `--layout` with a specific path which I knew pointed to a valid layout file. Thinking this must be a bug, I looked through Github and Discord. On Discord I found someone else had also run into the same issue.

Worse, the error message for `--layout` made me think that it simply couldn't find the file, as if there was a permission problem or a spelling error. Additionally, if you happen to use an absolute path with `--layout` then it works, only adding to the confusion.

My understanding is that there used to only be one argument, and that `--layout-path` was made after #322 brought up a very specific situation which resulted in unexpected behavior when no argument was used. It seems to me the only reason to have two arguments is to make a distinction between a path relative to the current directory, vs a path relative to the layout-dir. But I'm not sure why this distinction is necessary.

If others feel it best to keep them both, then I'd like to at least see the error message improved to assist users. There are a few possible solutions:
* Change `--layout`'s IoError to specify the absolute path it attempted to search for the file. (hints to the user that it's expecting a path relative to layout-dir)
* Add a suggestion to both `--layout` and `layout-path`'s IoErrors to try using the opposite argument. (e.g. instead of `IoError: The layout was not found, File: {}`, use `IoError: The layout was not found. Did you mean '--layout-path {}'?`, and vis-versa).
* The above, but only if zellij detects that it would have found a file if the user had used the opposite argument.

If others feel merging them again is best, then we'd need to decide which location gets priority when a relative path is used. Or else throw an error message if the same filename is found in both/multiple locations.

SpyrosRoum:
closes #1412 

A layout is considered local if:
`layout.extension().is_some() || layout.components().count() > 1`

This check produces the following results:
"foo": Abs
"foo.yml": Local
"./foo": Local
"foo/bar": Local
"foo/bar.yml": Local
 
TODO:
- [x] Fix tests

### DIFF
--- a/src/tests/e2e/remote_runner.rs
+++ b/src/tests/e2e/remote_runner.rs
@@ -127,7 +127,7 @@ fn start_zellij_with_layout(channel: &mut ssh2::Channel, layout_path: &str) {
     channel
         .write_all(
             format!(
-                "{} --layout-path {} --session {} --data-dir {}\n",
+                "{} --layout {} --session {} --data-dir {}\n",
                 ZELLIJ_EXECUTABLE_LOCATION, layout_path, SESSION_NAME, ZELLIJ_DATA_DIR
             )
             .as_bytes(),
--- a/zellij-utils/src/cli.rs
+++ b/zellij-utils/src/cli.rs
@@ -26,13 +26,10 @@ pub struct CliArgs {
     #[clap(long, short, overrides_with = "session")]
     pub session: Option<String>,
 
-    /// Name of a layout file in the layout directory
+    /// If it's a bare name it's interpreted as the name of a layout file in the layout dir,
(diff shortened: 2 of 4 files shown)


imsnif:
Hey @SpyrosRoum - cool that you're picking this up! Please reach out if you have any questions or concerns.

SpyrosRoum:
Thank you @imsnif, basic implementation is done without issues, I just need to update the e2e test as soon as I find time (hopefully today)!

imsnif (approved):
Cool work on this! I hope you didn't mind I reworded the help text a little. Hope to see more contributions from you.

PR #1430, opened 2022-05-22
Updated names and dialoguer crates dependencies

Laiot:
I tried to build and run zellij with the updated dependencies and they seem to show no issues.
I need this in order to package zellij for Fedora.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -688,12 +688,11 @@ dependencies = [
 
 [[package]]
 name = "dialoguer"
-version = "0.9.0"
+version = "0.10.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "61579ada4ec0c6031cfac3f86fdba0d195a7ebeb5e36693bd53cb5999a25beeb"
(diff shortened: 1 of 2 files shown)


Laiot:
I don't understand why the build fails on the Nix OS. Any idea?

a-kenji:
You bumped the dependencies solely in the `Cargo.toml` file, which is sufficient for libraries, but shouldn't be sufficient for applications.

https://doc.rust-lang.org/cargo/guide/cargo-toml-vs-cargo-lock.html

I expect it will work, if you run:
```
cargo update -p dialoguer -p names
```
And commit the `Cargo.lock` file.

Laiot:
Sorry about that, I just forgot to commit the Cargo.lock
It should be updated since I ran a local build.

a-kenji:
No worries!
It happens, luckily we have a CI that can catch these things.
I hope I can get time to look at the crates in more depth soon.

a-kenji:
@Laiot,
Thank you!
Please do reach out in the future.

PR #1432, opened 2022-05-23
fix: add checking for missing extensions

### LINKED ISSUE #1431: The default layout is not loaded.
The default layout is not loaded on the current main branch.

`zellij --version`: zellij 0.30.0

To reproduce:
- create a `default.yaml` inside the `ZELLIJ_LAYOUT` directory.
- run `zellij`

jaeheonji:
Fixes #1431 

Under the circumstance of checking the existence of a `layout` file, the default assets are always loaded due to missing extensions.

### DIFF
--- a/zellij-utils/src/input/layout.rs
+++ b/zellij-utils/src/input/layout.rs
@@ -262,7 +262,7 @@ impl LayoutFromYamlIntermediate {
         match layout_dir {
             Some(dir) => {
                 let layout_path = &dir.join(layout);
-                if layout_path.exists() {
+                if layout_path.with_extension("yaml").exists() {
                     Self::from_path(layout_path)
                 } else {
                     LayoutFromYamlIntermediate::from_default_assets(layout)


a-kenji (approved):
LGTM,
thank you for the quick fix!

jaeheonji:
@a-kenji Thanks :)

raphCode:
I think this commit slightly changed the behavior:
Previously, I had a layout file `standard` without file extension in the layout directory that I could load with `zellij --layout standard`.
This command now failed with `IoError: The layout was not found, File: standard` until I renamed the file to `standard.yaml`.

Not sure if I did something unusual but I feel we should mention the behavior change and fix (rename to `.yaml`) in the changelog?

jaeheonji:
I think the commit that changed the behavior is related to #1413 .

This commit is close to solving the problem that the existing function did not work properly because there was a missing part due to behavioral changes.

Issue #1414, opened 2022-05-12
Zellij does not exit on reboot/shutdown


langbamit:
**Current Behavior**
Zellij doesn't quit. 
So I get the nice message of Systemd waiting for a stop job, for 90 seconds. Until zellij gets killed. In the log it looks like that:
```
May 12 14:20:11 systemd[1]: session-3.scope: Failed with result 'timeout'.
May 12 14:20:11 systemd[1]: session-3.scope: Killing process 3103 (input_handler) with signal SIGKILL.
May 12 14:20:11 systemd[1]: session-3.scope: Killing process 3002 (zellij) with signal SIGKILL.
May 12 14:20:11 systemd[1]: session-3.scope: Stopping timed out. Killing.
May 12 14:18:42 systemd[1]: Finished Save Hardware Clock.
May 12 14:18:42 systemd[1]: save-hwclock.service: Deactivated successfully.
May 12 14:18:42 systemd[1]: Stopped Network Manager Script Dispatcher Service.
```
If I quit zellij before reboot/shutdown, everything is fine (very fast on reboot/shutdown).

**Expected Behavior**
When rebooting or shuting down the computer, zellij should quit properly.

**Basic information**

`zellij --version`: zellij 0.29.1
`uname -av` or `ver`(Windows): Linux thinkpad 5.17.5 #1-NixOS SMP PREEMPT Wed Apr 27 12:41:17 UTC 2022 x86_64 GNU/Linux
`alacritty --version`: alacritty 0.10.1 ()

raphCode:
I can see similar messages appearing in my journal of my server in the past.
I tested a bit around and was able to pinpoint the issue to a stuck zellij client process.
This is already tracked here: #1326, but that it also impacts local sessions is new to me. I'll update the referenced issue.

AjiBuster499:
Hi, I encountered this issue recently, and tracked it down to applications open inside Zellij as well (i.e. leaving a Neovim instance open). I'm on v0.30, so the PR that closed this should be in there. I don't know if something changed or what.

raphCode:
Hey thanks for the report, @AjiBuster499!
Can you check if you got the same messages in `journalctl` as in the first post?

AjiBuster499:
> Did you got the same messages in `journalctl` as in the first post?

Yes, I just rebooted my machine to check, and I got the same messages.

```
Jun 21 11:05:19 arch audit: BPF prog-id=0 op=UNLOAD
[...]

Issue #1326, opened 2022-04-14
Client process lingers around after ssh disconnect


raphCode:
This problem was already noticed in #1029 and now moved here.

Reproduction steps:
- ssh to remote or localhost
- start zellij or attach to existing session (I advise using `--session unique_string` to find it more easily in a process listing)
- disconnect ssh, e.g. by using `<Enter>~.` escape
- Observe that the zellij client process did not exit, e.g. with `ps` or `htop`

The client process linger around, no matter what. They even do not react to SIGHUP, SIGINT or SIGTERM.

[I did a bisect](https://github.com/zellij-org/zellij/issues/1029#issuecomment-1087920070) and found 0b74604a9f069086aa491f46d0f4a3f34520d153 to be the first bad commit

raphCode:
#1414 brought to attention that this is also a problem on local machines without involving ssh at all:
- open new terminal emulator window
- start zellij
- close terminal window
- zellij client stuck as outlined in first post

Also, on shutdown/reboot systemd waits for the process to exit upon SIGTERM, which does not happen. This introduces a 90 second wait time.

raphCode:
I think there might be race-y things going on, since sometimes the client exits when I try the local reproducer.
strace when the client process exits:
```
futex(0x55ef8935aa58, FUTEX_WAIT_PRIVATE, 4294967295, NULL) = ? ERESTARTSYS (To be restarted if SA_RESTART is set)
--- SIGHUP {si_signo=SIGHUP, si_code=SI_USER, si_pid=83768, si_uid=1000} ---
sendto(10, "X", 1, MSG_DONTWAIT, NULL, 0) = 1
rt_sigreturn({mask=[]})                 = 202
futex(0x55ef8935aa58, FUTEX_WAIT_PRIVATE, 4294967295, NULL) = 0
write(7, "\7\0\0\0\4\0\0\0\7\0\0\0\3\0\0\0\0\0\0\0\7\0\0\0\7\0\0\0\7\0\0\0"..., 36) = 36
futex(0x7f7aca8a8910, FUTEX_WAIT_BITSET|FUTEX_CLOCK_REALTIME, 83831, NULL, FUTEX_BITSET_MATCH_ANY) = 0
write(3, "INFO   |zellij_client           "..., 122) = 122
ioctl(0, SNDCTL_TMR_START or TCSETS, {B38400 opost isig icanon echo ...}) = 0
[...]

PR #1433, opened 2022-05-23
Fix double panic lockup in clients panic handler

raphCode:
https://github.com/zellij-org/zellij/issues/1326
https://github.com/zellij-org/zellij/issues/1414

### DIFF
--- a/zellij-client/src/lib.rs
+++ b/zellij-client/src/lib.rs
@@ -6,6 +6,7 @@ mod stdin_ansi_parser;
 mod stdin_handler;
 
 use log::info;
+use log::error;
 use std::env::current_exe;
 use std::io::{self, Write};
 use std::path::Path;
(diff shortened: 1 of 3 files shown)


raphCode:
@jaeheonji in case you want to have a look, I solved the problem by returning a `Result` from `unset_raw_mode()`, which must be handled at callsites. I unwrapped it everywhere except in the panic handler.
I also log every panic to the logfile now, since we now know we can't rely on printing to the terminal.

jaeheonji (approved):
LGMT. This can prevent `panic` from falling into a loop.
Of course, perhaps, the output for errors in `unset_raw_mode` can be unkind, but this would be a very rare case. And I don't think it's a big deal.

raphCode:
> perhaps, the output for errors in `unset_raw_mode` can be unkind

What do you mean? Ugly to handle for the programmer?

jaeheonji:
> > perhaps, the output for errors in `unset_raw_mode` can be unkind
> 
> What do you mean? Ugly to handle for the programmer?

Almost the same, maybe compared to the `handle_panic`, I thought there would be some lack of information, and no pretty output. However, as I have already mentioned, I do not think it is such a big problem. What do you think?

jaeheonji:
Cool! I agree :)

PR #1449, opened 2022-06-02
add(assets): `bottom-tab-bar` layout

a-kenji:
Add layout, that loads the `tab-bar` plugin as the bottom pane,
by default.

It can be run by specifying:
```
zellij -l bottom-tab-bar
```

### DIFF
--- a/zellij-utils/assets/layouts/bottom-tab-bar.yaml
+++ b/zellij-utils/assets/layouts/bottom-tab-bar.yaml
@@ -1,0 +1,13 @@
+---
+template:
+  direction: Horizontal
+  parts:
+    - direction: Vertical
+      body: true
+    - direction: Vertical
+      borderless: true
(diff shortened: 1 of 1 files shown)


imsnif:
@a-kenji - iirc you have a plugin that shows tabs as well as the current mode, right? Maybe we can use that instead? Give it a nice name like "compact" so you can load it with `zellij --layout compact`?

a-kenji:
I have a very very slightly adjusted version of the `tab-bar`, I could add it as an extra plugin. Should I add it with the zellij part, or without it? It is also a little rough, so I think there might be one or two good first issues in there.

imsnif:
I think by default all our layouts should be branded to help us with exposure (no advertizing budget and all). And for sure, I think this is a great start and we can improve the UI with some Good First Issues.

PR #1454, opened 2022-06-04
Exit client-loop on empty message receive

### LINKED ISSUE #1410: Logging may fill up /tmp mount until all zellij instances are closed
I did something with zellij and it started spamming the logfile with `empty ipc` messages.
This filled up the maximum capacity of the /tmp mount, which is backed by RAM or swap, consuming those ressources.

image
```
$ df -h /tmp
Filesystem      Size  Used Avail Use% Mounted on
tmpfs            12G   12G     0 100% /tmp
$ free -h
               total        used        free      shared  buff/cache   available
Mem:            23Gi       9,5Gi       498Mi        11Gi        13Gi       1,6Gi
Swap:          4,0Gi       2,4Gi       1,6Gi
```
I noticed because the build process fails when it can't write to /tmp.

The actual problem is that deleting the offending log file does not remedy the situation as long as _any_ zellij instance is running: All zellij processes have an open file descriptor to that file and the kernel holds on to the file content as long as there are any references to it.
```
$ rm /tmp/zellij-1000/zellij-log/zellij.log
$ du -hs /tmp
6,7M	/tmp
$ echo "aaa" > /tmp/test
bash: echo: write error: No space left on device
$ lsof | grep zellij.log
zellij     7363                 raph    4w      REG               0,36 12567707648         60 /tmp/zellij-1000/zellij-log/zellij.log (deleted)
zellij     7363  7368 stdin_han raph    3w      REG               0,36 12567707648         60 /tmp/zellij-1000/zellij-log/zellij.log (deleted)
...
```
This effectively forces me to close my working session which is bad.

Possible solutions:
- every zellij instance / version / session uses a new logging file, which makes it possible to delete offending files and only closing that / some zellij process(es) to free up the space
- special handling of repeated logging messages ("last message repeated x times") so they don't fill up so fast (possibly combine with previous point)
- limit log filesize (deleting old entries when size grows)
- logging somewhere else, like `/var/log` (potentially filling up disk space, ouch my SSD lifetime!)
- log to systemd or equivalent?
- since data is only appended to the log (assuming `open(2)` with `O_APPEND`), in theory it should be possible to free up space once the filename is deleted since no fd refers to the "early parts" of the file. But I don't think the kernel supports something like this.

### LINKED ISSUE #1419: Received empty message from client repeated in logfile
I sometimes get zellij into a state where it spams the logfile [until the /tmp mount is full](https://github.com/zellij-org/zellij/issues/1410), and even then it doesn't stop: I currently have a zellij server thread consuming 100% CPU, desperately trying to write this to the logfile:
`ERROR  |zellij_server::route     | 2022-05-16 20:24:20.269 [server_router] [zellij-server/src/route.rs:510]: Received empty message from client`

This happens to me when starting two different zellij builds with the same official version number, but from different commits.
- Specifically, I had the zellij-git AUR package (at 82822374b41db83989af7b9a14cc49e21aa53f52) installed as my daily driver to test things before the upcoming release. I happened to have a session running from this package.
- I then started a debug build with `cargo make run` at d3ed8a295bfab5ec1226f0afcf2927a25579c57a. This build did not start correctly and just sat around consuming no ressources.
- At the moment I terminate the debug build with `Ctrl-C`, the original session's server goes crazy spamming the logfile
- The original session is still useable / re-attachable. The logging is happening in a separate thread
- The server is somehow latched into the spamming mode, since it also happens if it is the only remaining zellij process. Also, closing the last Tab / Pane does not exit that server process, it silently continues spamming.

Probably not an issue for end users, but I think it hints at some broken logic somewhere.

SpyrosRoum:
This closes #1419.

By extension it should fix #1410 as well bu I think it would be a good idea to implement a limit for the log file so this doesn't happen again

### DIFF
--- a/zellij-server/src/route.rs
+++ b/zellij-server/src/route.rs
@@ -16,7 +16,7 @@ use zellij_utils::{
         command::TerminalAction,
         get_mode_info,
     },
-    ipc::{ClientToServerMsg, IpcReceiverWithContext, ServerToClientMsg},
+    ipc::{ClientToServerMsg, ExitReason, IpcReceiverWithContext, ServerToClientMsg},
 };
 
 use crate::ClientId;
@@ -514,6 +514,13 @@ pub(crate) fn route_thread_main(
             }
             None => {
                 log::error!("Received empty message from client");
+                os_input.send_to_client(
+                    client_id,
+                    ServerToClientMsg::Exit(ExitReason::Error(
+                        "Received empty message".to_string(),
+                    )),
+                );
+                break;
             }
         }
     }


imsnif (approved):
Good catch!

Issue #1403, opened 2022-05-10
Open pane scrollback in default editor


imsnif:
The experience should be:
1. Press a Zellij shortcut (something like ctrl+s + e)
2. Have the current pane replaced with your default editor, opened to a temporary file containing all the scrollback of the terminal pane and scrolled to the same line
3. Search, save, copy/paste, do whatever you want
4. When you quit the editor (or press ctrl+s + e again), the original pane comes back

Depends on: https://github.com/zellij-org/zellij/pull/1375

imsnif:
@cosminadrianpopescu - you mentioned you might like to work on this. I wanted to open this as a separate issue because it's part of our roadmap.

cosminadrianpopescu:
Yes, I would like to work on this.

cosminadrianpopescu:
How would I go about it? What is the idea? How do I replace the content of the pane with a command without putting that command in the history of the current pane's shell?

Also, when running the command, is there something to know about it regarding different operating systems?

My idea would be to create a new temporary pane which would replace the current one. Is there a better approach? If this is the right approach, how do I know when this temporary panel is closed, to react and put back the old pane?

Any hints?

imsnif:
I think the basic approach should be:

1. When we receive the `EditScrollback` action (which we should create), we first dump the scrollback of the current terminal pane into a temporary file. This is something we do in the `screen_thread`, in `Tab` in a similar function to the one you already know here: https://github.com/zellij-org/zellij/blob/main/zellij-server/src/tab/mod.rs#L1385
2. Then, we need to open a pane with the default editor and point it to that file. In our `pty_thread` we already have a facility for that. So if we send it the right instruction (`PtyInstruction::SpawnTerminal` with the `TerminalAction::OpenFile(path)` parameter), it'll do all this work for us. Here's an example where our plugins send it this instruction: https://github.com/zellij-org/zellij/blob/c53b3235c848bbf1c6f56e4c00f46e99b47f9cd6/zellij-server/src/wasm_vm.rs#L395-L399
3. The above function will send back a message to `Screen` (and eventually to the relevant `Tab`) once the pane is open, giving us the pane ID. The problem is that it'll send the message to us as a new pane and we'll open it as a new pane instead of replacing it with the current pane. 
[...]

PR #1456, opened 2022-06-05
Edit scrollback

cosminadrianpopescu:
PR for #1403


imsnif:
@a-kenji - any clues as to why the nix build is failing?

a-kenji:
@imsnif,
I assume this is because the `Cargo.lock` file is in a weird state.
`cargo update` and committing the lock file might alleviate this issue,
If not I think I can look more into this tomorrow.

imsnif:
@a-kenji - FYI, it works now. The `Cargo.lock` was updated from a local build (I update and push it all the time as I'm building, testing manually and running tests). What fixed this was merging from main and then building (which indeed included a change to the Cargo.lock file over the one from this branch). Could it be that the nix test uses some sort of external state combined with main?

imsnif:
Thank you for your excellent work on this @cosminadrianpopescu - I'm stoked about this feature and think many others will be as well! Hope to see more contributions from you - even if Rust might not be your cup of tea :)

cosminadrianpopescu:
> Thank you for your excellent work on this

You welcome.

> I'm stoked about this feature and think many others will be as well! 

I also think it's a killer feature. This would be enough to make me switch from tmux. And I never had any big issues with tmux's find and copy paste. Just small annoyances. But to be able to open all that buffer in `vim`, it cannot be beaten by any paradigm impementation. `vim` is the best at this kind of things. 

> Hope to see more contributions from you

I was thinking now of finishing this by adding a nice name to the editor panel, like we discussed. Then I was thinking at looking into the buffer's rename (what we discussed before this: adding an option to keep the tab's shortcuts and by default re-naming the tabs when closing one of the tabs which is not the last one). What do you think?

> even if Rust might not be your cup of tea

Yeah... I can see the advantages, and I can understand if someone wants to program something for routers for example (with limited resources). But for doing something in the user space (no matter how low level), I think a garbage collector is really nothing compared with a steep learning curve. I think that this steep learning curve is going to make a lot of people just not want to contribute.

imsnif:
> I was thinking now of finishing this by adding a nice name to the editor panel, like we discussed. Then I was thinking at looking into the buffer's rename (what we discussed before this: adding an option to keep the tab's shortcuts and by default re-naming the tabs when closing one of the tabs which is not the last one). What do you think?

Sounds great.

a-kenji:
@cosminadrianpopescu 
I agree, this is awesome, thanks for all the work.

@imsnif,
The `nix build` shouldn't use external state from main, but it might be possible, I will look into it a bit more.
Basically it uses the nix files, the `Cargo.toml`, `Cargo.lock` and the `rust-toolchain.toml`, which should be contained inside the branch.

Issue #1445, opened 2022-06-01
Feat: Log Reason, why Empty Message is send from a client.


a-kenji:
Log Reason, why Empty Message is send from a client.

raphCode:
As far as I understood the problem, noone really knows why this happens.
In the past we only ever saw messages like [`Io(Error { kind: UnexpectedEof, message: "failed to fill whole buffer" })`](https://github.com/zellij-org/zellij/issues/1326)

But yes, we could log that exact message.

PR #1459, opened 2022-06-06
Add logging to ipc receiver

raphCode:
#1445 

@a-kenji does this look like you imagined?

Example log message:
```
WARN   |zellij_utils::ipc        | 2022-06-07 01:51:22.924 [server_router] [zellij-utils/src/ipc.rs:194]: Error in IpcReceiver.recv(): Io(Error { kind: UnexpectedEof, message: "failed to fill whole buffer" })
ERROR  |zellij_server::route     | 2022-06-07 01:51:22.925 [server_router] [zellij-server/src/route.rs:522]: Received empty message from client
```


a-kenji (approved):
LGTM!

PR #1464, opened 2022-06-07
move: `test-template` layout to `layouts` directory

a-kenji:


imsnif:
I think I once committed this by accident and as far as I'm aware it can totally be deleted

a-kenji:
Sounds good!

PR #1485, opened 2022-06-10
fix: tab selection by left click in compact-bar

### LINKED ISSUE #1477: Fix: Clicking Tabs inside the `compact-bar`
The `compact-bar` interprets mouse clicks on tabs with an incorrect offset.

`zellij --version`: 0.31.0

amaihoefner:
Fixes #1477

One thing I wasn't able to figure out:
It works when running "cargo make run -l compact"
but not when running "cargo make install ./test && ./test -l compact".

The latter seems to re-build the old version of the plugin and replace the wasm file in the "plugins" directory?
I don't know enough about WebAssembly / cargo make / the build process to be able to figure this one out.

Any pointers on how to fix this?
Or maybe someone more qualified can take over from here.

### DIFF
--- a/default-plugins/compact-bar/src/main.rs
+++ b/default-plugins/compact-bar/src/main.rs
@@ -115,11 +115,11 @@ impl ZellijPlugin for State {
             if self.should_render
                 && self.mouse_click_pos > len_cnt
                 && self.mouse_click_pos <= len_cnt + bar_part.len
-                && idx > 2
+                && idx > 3
             {
-                // First three elements of tab_line are "Zellij", session name and empty thing, hence the idx > 2 condition.
-                // Tabs are indexed starting from 1, therefore we need subtract 2 below.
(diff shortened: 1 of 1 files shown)


a-kenji:
@amaihoefner,
Thanks for the contribution!

The install just installs `zellij` itself, not the plugins. 
They only get installed on actual releases.

You can run `cargo make build-plugins-release`
And then run `zellij` with `zellij --data-dir ./target/dev-dir`

amaihoefner:
@a-kenji
Ah, that makes sense. Thanks for the info!
Following your instructions, it now works as expected.

a-kenji (changes requested):
LGTM, please remove the asset itself, it will get build, optimized and included by our release script.

a-kenji:
Thank you!

a-kenji (commented):
LGTM

PR #1515, opened 2022-06-15
feat: fullscreen focus swapping

### LINKED ISSUE #1466: Support for switching between panes when in full screen
It would be nice to be able to switch between panes in the current tab when in pane full screen mode.  This would make the pane you switched to full screen and send the previous pane to the background.

nacairns1:
Adds support for swapping between panes in fullscreen mode. This can be accomplished by either using the next pane command in pane mode or navigating focus left and right in normal mode.
Closes #1466


a-kenji (approved):
LGTM, thank you very much!

PR #1529, opened 2022-06-19
Fix installation with MacPorts

ken-matsui:
MacPorts installed in a general way requires `sudo`.
Ref: https://superuser.com/a/1413502


a-kenji:
Thank you @ken-matsui .

PR #1535, opened 2022-06-21
Exit client empty rcv

### LINKED ISSUE #1414: Zellij does not exit on reboot/shutdown

raphCode:
fix #1414 

did you solve it the same way @tlinford?

### DIFF
--- a/zellij-client/src/lib.rs
+++ b/zellij-client/src/lib.rs
@@ -315,7 +315,7 @@ pub fn start_client(
         let restore_snapshot = "\u{1b}[?1049l";
         os_input.disable_mouse();
         let error = format!(
-            "{}\n{}{}",
+            "{}\n{}{}\n",
             restore_snapshot, goto_start_of_last_line, backtrace
         );
         let _ = os_input
--- a/zellij-client/src/lib.rs
+++ b/zellij-client/src/lib.rs
@@ -297,6 +297,12 @@ pub fn start_client(
                             .send(ClientInstruction::UnblockInputThread)
                             .unwrap();
                         log::error!("Received empty message from server");
+                        send_client_instructions
+                            .send(ClientInstruction::Error(
+                                "Received empty message from server".to_string(),
+                            ))
+                            .unwrap();
+                        break;
                     },
                 }
             }


tlinford:
yep! same exact thing.

tlinford (approved):

tlinford:
As a sidenote, initially I wanted to do it a bit differently, by having the server react to a sigterm by instructing all clients to exit, to implement a "controlled" shutdown let's say, do you think it's worth looking into that at some point?

raphCode:
Since we need to handle ungraceful disruptions of the communication channel anyways (and now we do on both sides), I don't see a lot of sense in handling SIGTERM on the server side different.
There is only one "official" way to cleanly exit a session and that is initiated through the client. For this case it is okay to notify all clients so they don't report any Errors.

PR #1536, opened 2022-06-21
feat(theme): allow hex colors for themes

RGFTheCoder:
This is a redo of a previous PR which I had made on a different account. It does the hex conversion in a much nicer way, and uses a custom Deserialize impl.

This adds support for hex colors in the form
```yaml
fg: ff4500
bg: 3f7
```
because hashtags are recognized as comments.

Should solve #880

### DIFF
--- a/zellij-utils/src/input/theme.rs
+++ b/zellij-utils/src/input/theme.rs
@@ -1,5 +1,8 @@
-use serde::{Deserialize, Serialize};
-use std::collections::HashMap;
+use serde::{
+    de::{Error, Visitor},
+    Deserialize, Deserializer, Serialize, Serializer,
+};
+use std::{collections::HashMap, fmt};
 
 use super::options::Options;
 use crate::shared::detect_theme_hue;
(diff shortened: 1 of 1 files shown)


zellij-utils/src/input/theme.rs
```diff
@@ -48,6 +51,65 @@ pub struct PaletteFromYaml {
...
+            ))
+        } else if s.len() == 6 {
+            Ok(HexColor(
+                u8::from_str_radix(&s[0..2], 16).map_err(E::custom)?,
+                u8::from_str_radix(&s[2..4], 16).map_err(E::custom)?,
+                u8::from_str_radix(&s[4..6], 16).map_err(E::custom)?,
+            ))
+        } else {
+            Err(Error::custom("Hex color must be of form #RGB or #RRGGBB"))
+        }
```
jaeheonji, line 94: It is really cool feature. I like it. But, I think it would be better to express the colors more specifically like '#FF4500'' or ''#3F7''. It seems to make more sense even if the user is presented with an error.

RGFTheCoder: I added a commit which now allows the forms:
`ff0000`
`f00`
`"ff0000"`
`"f00"`
`"#ff0000"`
`"#f00"`

The deserializer is passed the string directly, so I can't check for quotes around the hex string.

jaeheonji: LGTM!

jaeheonji:
There is a small issue with formatting. Could you check it out?

RGFTheCoder:
Should be fixed.

zellij-utils/src/input/theme.rs
```diff
@@ -48,6 +51,71 @@ pub struct PaletteFromYaml {
...
+        write!(formatter, "a hex color in the format #RGB or #RRGGBB")
+    }
+
+    fn visit_str<E>(self, s: &str) -> Result<Self::Value, E>
+    where
+        E: Error,
+    {
+        if s.starts_with("#") {
+            return self.visit_str(&s[1..]);
+        }
```
jaeheonji, line 82: Hey, almost done!

Could you please apply clippy's recommendations as follows?

```rust
if let Some(stripped) = s.strip_prefix('#') {
    return self.visit_str(stripped);
}
```

jaeheonji (changes requested):

zellij-utils/src/input/theme.rs
RGFTheCoder: done

jaeheonji (approved):

PR #1541, opened 2022-06-24
Refactor all shared contents from zellij tile to zellij utils

har7an:
This PR incorporates a refactoring that will hopefully enable an easier implementation of https://github.com/zellij-org/zellij/pull/1242. Currently, `zellij-utils` is supposed to contain contents that are shared between the `client` and `server` crate, respectively. However, it also re-exports `zellij-tile`, in particular the `data` module of `zellij-tile`, which holds a lot of lower-level data structures used throughout the whole code.

This has the implication that `zellij-tile` itself cannot import `zellij-utils` to access the shared data structures there, since that would cause a cyclic dependency. This is currently worked around by e.g. converting all keys and keybindings to strings before sending them to the plugins (which rely only in `zellij-tile`) since `String` as a base type is understood by all plugins. However, handling strings isn't ergonomic and completely foregoes all advantages of Rusts type system.

Since the only mode inside `zellij-tile` that is used by **all** other parts of the code is the `data` module, this PR moves the data submodule to `zellij-utils`. This breaks `utils` dependency on `tile` and allows us in turn to import `utils` and use it in `zellij-tile`. In order not to break the plugins, we then re-export the `data` modules contents from `tile`.

This will significantly ease extensions to the code base (such as in https://github.com/zellij-org/zellij/pull/1242) that relate to the plugins since the plugins can now selectively be granted access to shared data structures from `zellij-utils` by re-exporting them through `zellij-tile`.

Note that since `zellij-utils` includes some crates that cannot compile against the `wasm` target, a new feature "full" was added to aleviate this. As such `zellij-tile` currently has only access to `zelij_utils::data`, but other modules that do not depend on crates incompatible with `wasm` could be added, too.

The new feature is currently still undocumented. I wanted to get feedback from you first before polishing it. Also I wasn't sure where/how to document this feature (Because this is the first time I work with features in cargo).

Looking forward to your feedback!


har7an:
I played around a little with basing #1242 on this. While in theory it allows us to use the `ModeKeybinds` directly and send them to the plugin, in practice the interdependencies between all the structures involved in the code are massive.

I'm currently trying to sort out the bits we can use, but it doesn't look very promising so far. I also must confess that the sheer amount of code to go through is a bit over my head...

imsnif (changes requested):
Hey @har7an - great work on this!

This looks good in principle, but I'm wondering about a few things:

1. Why the upgrade to rust 1.60.0? Did you need some specific feature that was missing in 1.59.0? Generally I prefer to only upgrade if we absolutely have to, because it has lots and lots of other consequences (eg. running the dev build on this PR is noticeable slower - not sure why, but before I start debugging I'd like to minimize the amount of changes)

2. Would it be possible to use the `target_family` (https://doc.rust-lang.org/reference/conditional-compilation.html#target_family) cargo feature in the zellij-utils crate instead of adding a custom feature? I think that'll make it a little clearer and include less clutter. What do you think?

har7an:
> 1. Why the upgrade to rust 1.60.0?

That's explained in 40a93426401350f05b9f3590543ab03d9c0f9774:

*As a side effect, due to the dependency notation for the
optional dependencies (See
https://doc.rust-lang.org/cargo/reference/features.html#optional-dependencies),
we bump the rust toolchain version to 1.60.0.*

I'm not a huge fan of it either, but I saw no other way to achieve this truly conditional compilation. :/

I need this to exclude all the non-wasm compatible crates from the compilation. Without it, cargo will compile *all* crates listed in `Cargo.toml`, even if they aren't needed with the current config.



> 2. Would it be possible to use the `target_family`

I don't know about that, see the explanation for why I had to raise the MSRV. If there is a way to tell in cargo.toml that some dependencies are only for certain compilation targets we can sure do that, but I haven't really found something like that yet. If you know a way to do this I'd be happy to change it!

a-kenji:
@har7an,
you could also use the pre `1.60` way of conditional compilation, use features.

imsnif:
Here's another example of how mio does it with dependencies: https://github.com/tokio-rs/mio/blob/master/Cargo.toml#L48

I'll say again - ideally if possible this should be a `target_family` thing, rather than defining a custom feature. I think with the `cfg` keyword we have a lot of room to play with.

har7an:
> you can look at main and how `unstable` is implemented.

Iirc that propagates the feature flag from the base crate to the child crates. What I need in order for zellij-utils to compile is to disable some dependencies from ever being touched by cargo when targeting `wasm`. You can try for yourself and see what I mean: Check out the changes here, go to `zellij-utils/Cargo.toml` and change `full = [...]` to `full = []`.

I have made all dependencies that don't cooperate with `wasm` `optional = true`, so unless my feature pulls them in, they will not be compiled for e.g. zellij-server or client. If you remove the `optional` part of the dependencies on the other hand, they are *always compiled*, even if they're not used by the `wasm` target. However, since they are compiled unconditionally, the plugins can then no longer be built: `utils` will not compile without errors (since some crates don't support `Wasm`), and the plugins can't be built.

har7an:
> Here's another example of how mio does it with dependencies: https://github.com/tokio-rs/mio/blob/master/Cargo.toml#L48

See, that's what I meant: I had no idea this existed. This is great, and that should work without bumping the MSRV. Thank you. ^^

a-kenji:
> > you can look at main and how `unstable` is implemented.
> 
> Iirc that propagates the feature flag from the base crate to the child crates. What I need in order for zellij-utils to compile is to disable some dependencies from ever being touched by cargo when targeting `wasm`. You can try for yourself and see what I mean: Check out the changes here, go to `zellij-utils/Cargo.toml` and change `full = [...]` to `full = []`.
> 
> I have made all dependencies that don't cooperate with `wasm` `optional = true`, so unless my feature pulls them in, they will not be compiled for e.g. zellij-server or client. If you remove the `optional` part of the dependencies on the other hand, they are _always compiled_, even if they're not used by the `wasm` target. However, since they are compiled unconditionally, the plugins can then no longer be built: `utils` will not compile without errors (since some crates don't support `Wasm`), and the plugins can't be built.

No, you can specify that in the crate itself.

imsnif:
For the record, I think this could have been done with @a-kenji 's suggestion as well and that is what he meant - but I'm happy we resolved this :) Once again though: let's try to use the `target_family` if we can?

har7an:
It looks muuuuch better now, thank you for the suggestion and examples!

Since supporting Windows isn't actively pursued but not strictly out of the question I went with the more verbose `#[cfg(not(target_family = "wasm"))]` instead of `#[cfg(unix)]` to make it clear to anyone inspecting the code that these specific bits of code *aren't meant for wasm*. I have not looked at what happens under Windows and I never will. But if one day windows support does land, some poor soul will have to shift through all of this anew in any case, so I may just as well leave it as is.

Or do you strongly prefer `[cfg(unix)]` here?

imsnif:
This looks great! I totally agree it's much better.
The only concern I have now before merging this is that looking through `Cargo.lock` we seem to have (inadvertently?) upgraded some of our dependencies. I wonder if there's a way to roll them back to the way they are in main? I'm concerned this might have unknown (and frankly, unknowable) side-effects. 90% all is well, but if we can keep the older versions this will make me calmer while releasing.

har7an:
> looking through `Cargo.lock` we seem to have (inadvertently?) upgraded some of our dependencies.

That's just stupidity on my part, it looks like at some point throughout developing this I did a `cargo update`. Thanks for spotting it, I reverted the version changes!

PR #1548, opened 2022-06-27
roll over logs / fix filling up /tmp with logs

### LINKED ISSUE #1410: Logging may fill up /tmp mount until all zellij instances are closed

raphCode:
- rolls over at 20 MB into `zellij.log.old.0`
- fix #1410
- unsure what happens with multiple zellij instances, probably they all copy over the same old log?
- do we need to update the issue template to tell users to include the old log?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -52,6 +52,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 * fix: properly handle in-place editor in full-screen (https://github.com/zellij-org/zellij/pull/1544)
 * Terminal compatibility: properly trim whitespace in lines with wide-characters when resizing panes (https://github.com/zellij-org/zellij/pull/1545)
 * fix: reset scroll properly when typing in certain edge cases (https://github.com/zellij-org/zellij/pull/1547)
+* fix: logging may fill up /tmp, now logs are capped at 100 kB (https://github.com/zellij-org/zellij/pull/1548)
 
 ## [0.30.0] - 2022-06-07
 * fix: right and middle clicks creating selection (https://github.com/zellij-org/zellij/pull/1372)
--- a/zellij-utils/src/logging.rs
+++ b/zellij-utils/src/logging.rs
@@ -63,10 +63,7 @@ pub fn configure_logger() {
         )))
         .build(
             &*ZELLIJ_TMP_LOG_FILE,
-            Box::new(CompoundPolicy::new(
-                Box::new(trigger),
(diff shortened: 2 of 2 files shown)


imsnif:
Cool! Honestly, I cannot imagine a scenario in which we need (or even will need) more than 100KB of logs. I think we can cap it there and delete the rest.

raphCode:
changed the cap, now we have max. 100 kB in a file.
In case the log just rolled over, we still have the old 100 kB log so the user can always post enough logs in an issue.

PR #1552, opened 2022-07-01
fix: fallback to default values when terminal rows/cols are 0

### LINKED ISSUE #1551: always crashes on resize
Is zellij supposed to do something reasonable if you resize the terminal? I find that it always, 100% of the time, crashes if you resize the terminal.

```
Error occurred in server:

  × Thread 'screen' panicked.
  ├─▶ Originating Thread(s)
  │   	1. stdin_handler_thread: AcceptInput
  │   	2. screen_thread: TerminalPixelDimensions
  │
  ├─▶ At zellij-server/src/screen.rs:490:46
  ╰─▶ attempt to divide by zero

```

**Basic information**

`zellij --version`:0.29.1
`uname -av` or `ver`(Windows): Linux micc 3.10.0-693.el7.x86_64 #1 SMP Thu Jul 6 19:56:57 EDT 2017 x86_64 x86_64 x86_64 GNU/Linux

tlinford:
fix #1551

fallback to default values (80 cols, 24 rows) if either terminal dimension is reported as 0 by TIOCGWINSZ.


tlinford:
well another case of the e2e tests being weird - this is not failing locally on my setup :confused:

tlinford:
@imsnif one e2e test kept failing here (scrolling_inside_a_pane_with_mouse), until I increased its `retry_pause_ms`. Is this cause for concern?

imsnif:
Yes. But I doubt it has to do with this PR. I think that particular case causes some trouble because we fill up the terminal screen through STDIN one character at a time.

The best thing to do here (I think) would be to instead get the tests to `cat` a fixture that would fill in the viewport. I think that'll cause it to be far more consistent in this and other cases.

tlinford:
@imsnif added a `load_fixture` method to `RemoteTerminal`, and was able to remove the extra long delays from the two scrolling tests.

imsnif:
You're the best :)

PR #1573, opened 2022-07-08
Let the exit message be different when detaching

### LINKED ISSUE #1572: Print "Detached" instead of goodbye-message on detach
Hi.

Would be nice to get a "Detached" message instead of the normal goodbye-message when detaching.
This way a user recognizes that the thing is still running and they did not fat-finger quit their session.

What do you think?

matthiasbeyer:
This patch changes the exit message printed to the user, so the user
does not get the impression that they fat-fingered an "exit" instead of
what was intended (a detach).

For this, the InputHandler::exit() function was refactored, to get the
reason as a parameter. As this function is not pub, this is considered
okay.

Closes #1572


imsnif:
Looks good in principle, but I see this is still a draft... are you planning more work on this, or?

matthiasbeyer:
I posted this as draft because I am not sure whether such a thing is desirable and because I didn't yet do a test-run. But code-wise, I don't know whether there is more to do :laughing:

imsnif:
I think it's definitely desirable :) Let me know when you'd like me to review

imsnif:
Hey @matthiasbeyer - is this working for you? I just took this out for a spin and I still see "Bye from Zellij!" when detaching...
What I did:
1. `cargo make run`
2. `Ctrl+o` + `d`
3. I see "Bye from Zellij!"

matthiasbeyer:
Indeed I missed one spot. I fixed it and tested it - this works for me now!

raphCode:
Very nice feature!

Just a small change: I would suggest rephrasing the message to "Session detached" to hint that the session is still running. Also the wording would more consistent with `zellij --help`:
```
...
SUBCOMMANDS:
    attach               Attach to a session [aliases: a]
...
```

matthiasbeyer:
@imsnif ready for review IMO.

imsnif:
LGTM!

Issue #1574, opened 2022-07-08
`zellij setup --generate-auto-start fish` is misleading without the documentation about shell integration


tsdh:
While exploring the command line interface of zellij I found out about `zellij setup --generate-auto-start fish` and just did `zellij setup --generate-auto-start fish > ~/.config/fish/conf.d/99_zellij.fish`.  The result of this exercise was that when I wanted to ssh into the machine where I did that, my local terminal hang with no way to break out of that hang except by killing it.

Thereafter I've read the [integration docs](https://zellij.dev/documentation/integration.html) which explains well how it's supposed to be done.  Basically, with `fish` the crucial thing seems to be the `if status is-interactive ... end` which makes the hang go away.

I'd suggest to somehow improve the command line docs for `zellij setup --generate-auto-start fish` by linking the integration docs.  Or maybe the command could output a comment before the fish code, e.g.,
```fish
# The following snippet is meant to be used like this in your fish configuration:
# if status is-interactive
#     # Maybe set ZELLIJ_AUTO_ATTACH / ZELLIJ_AUTO_EXIT
#     eval (zellij setup --generate-auto-start fish | string collect)
# end
if not set -q ZELLIJ                                                                                                                                                           
    if test "$ZELLIJ_AUTO_ATTACH" = "true"                                                                                                                                     
        zellij attach -c                                                                                                                                                       
    else                                                                                                                                                                       
        zellij                                                                                                                                                                 
    end                                                                                                                                                                        
                                                                                                                                                                               
    if test "$ZELLIJ_AUTO_EXIT" = "true"                                                                                                                                       
        kill $fish_pid                                                                                                                                                         
    end                                                                                                                                                                        
end
```

By the way: I've had the very same problem with `tmux` and `fish` before and simply gave up on auto-attaching.  Your good documentation made me aware of `if status is-interactive ... end` which makes both zellij and tmux auto-attach work.  Thanks!

jaeheonji:
Good point. Thank you for your interest.

I think it's good about the way to add comments.
If you are interested in contributing to this, would you like to try making PR?
(Of course, if you are busy, I will update this later 😄)

tsdh:
@jaeheonji Yes, I can provide a PR.  Don't hold your breath but it should come within a week. :-)

tsdh:
@jaeheonji Thank you!

PR #1583, opened 2022-07-11
fix: add usage comment to fish shell auto-start snippet (#1574)

tsdh:
This PR adds a usage comment before the auto-start snipped for the fish shell which is emitted by `zellij setup --generate-auto-start fish` as discussed in #1574.


jaeheonji:
@tsdh Thank you! LGTM!

PR #1631, opened 2022-08-02
fix: theme not loading without config

jaeheonji:
Fixed the theme not being loaded from `themes` folder if there is no `config.yaml` file or `themes:` does not exist in the file.


jaeheonji:
@imsnif 

I found out one problem with the themes feature included in a recent release.
I happened to find this problem and was able to solve it. (Fortunately, there seems to be no issue related to this issue yet.)

So, After the CI, if possible, could you make a minor release that includes the patch?

PR #1644, opened 2022-08-09
forward OSC52 events

### LINKED ISSUE #1199: forward osc 52 copy events
Handle osc 52 copy that is sent from applications running inside zellij, to allow for example a nested `zellij` instance to set clipboard contents.

apexo:
Fixes #1199

### DIFF
--- a/zellij-server/src/panes/grid.rs
+++ b/zellij-server/src/panes/grid.rs
@@ -342,6 +342,7 @@ pub struct Grid {
     scrollback_buffer_lines: usize,
     pub mouse_mode: bool,
     pub search_results: SearchResult,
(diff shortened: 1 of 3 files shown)


imsnif (approved):
This looks great, apologies for taking a few days to get to it.
Would have been nice to add a test, but given the nature of this change, we'll need to add some infrastructure to do so. I'm doing this elsewhere for some other mouse interaction stuff, so will add a test for this as well then.

Thanks for this!

tombh:
I'm really excited about this! Thank you so much 😊

apexo:
Well, thank you for this excellent piece of software. Apparently it's good enough that people use and care to improve it :-)

PR #1648, opened 2022-08-10
screen: Don't crash when first tab doesn't exist

### LINKED ISSUE #1645: Crash when attaching when only "Tab 2" exists
Thank you for taking the time to file this issue! Please follow the instructions and fill in the missing parts below the instructions, if it is meaningful. Try to be brief and concise.

**Basic information**

`zellij --version`: 0.31.1
`stty size`: 64 252
`uname -av` or `ver`(Windows): Linux ********** 5.10.0-14-amd64 #1 SMP Debian 5.10.113-1 (2022-04-29) x86_64 GNU/Linux

**Further information**

1. Run zellij.
2. Create a new tab, then close the first tab. There should be only 1 tab, called `Tab #2`.
3. Detach ungracefully by force-closing the terminal window
4. Launch a new terminal window, attempt to attach
5. Zellij crashes:

```
Error occurred in server:

  × Thread 'screen' panicked.
  ├─▶ Originating Thread(s)
  │     1. ipc_server: AttachClient
  │     2. screen_thread: AddClient
  │
  ├─▶ At /home/lucretiel/.cargo/registry/src/github.com-1ecc6299db9ec823/zellij-server-0.31.1/src/screen.rs:719:14
  ╰─▶ called `Option::unwrap()` on a `None` value
         0: <unknown>
         1: <unknown>
         2: <unknown>
         3: <unknown>
         4: <unknown>
         5: <unknown>
         6: <unknown>
         7: <unknown>
         8: <unknown>
         9: <unknown>
        10: <unknown>
        11: <unknown>
        12: <unknown>
        13: <unknown>
        14: <unknown>
        15: start_thread
        16: clone

```

In case it matters, I'm running zellij on a server that I access via ssh from Windows Terminal on a Windows machine. The server itself is real hardware, not a VM.

har7an:
while trying to attach a new client. Instead, check whether the first
tab does exist and if not, take the first tab index from the tabs
present in the session. If no tabs exist, panic with a better error
message.

Fixes #1645

### DIFF
--- a/zellij-server/src/screen.rs
+++ b/zellij-server/src/screen.rs
@@ -641,11 +641,6 @@ impl Screen {
     pub fn new_tab(&mut self, layout: Layout, new_pids: Vec<RawFd>, client_id: ClientId) {
         let tab_index = self.get_new_tab_index();
         let position = self.tabs.len();
-        let client_mode_info = self
-            .mode_info
-            .get(&client_id)
-            .unwrap_or(&self.default_mode_info)
-            .clone();
         let mut tab = Tab::new(
             tab_index,
             position,
@@ -657,7 +652,7 @@ impl Screen {
             self.bus.senders.clone(),
             self.max_panes,
             self.style,
-            client_mode_info,
+            self.default_mode_info.clone(),
             self.draw_pane_frames,
             self.connected_clients.clone(),
             self.session_is_mirrored,
--- a/zellij-server/src/screen.rs
+++ b/zellij-server/src/screen.rs
@@ -702,21 +702,29 @@ impl Screen {
     }
 
     pub fn add_client(&mut self, client_id: ClientId) {
-        let mut tab_index = 0;
         let mut tab_history = vec![];
-        if let Some((_first_client, first_active_tab_index)) = self.active_tab_indices.iter().next()
-        {
-            tab_index = *first_active_tab_index;
-        }
         if let Some((_first_client, first_tab_history)) = self.tab_history.iter().next() {
             tab_history = first_tab_history.clone();
         }
+
+        let tab_index = if let Some((_first_client, first_active_tab_index)) =
+            self.active_tab_indices.iter().next()
+        {
+            *first_active_tab_index
+        } else if self.tabs.contains_key(&0) {
+            0
+        } else if let Some(tab_index) = self.tabs.keys().next() {
+            tab_index.to_owned()
+        } else {
+            panic!("Can't find a valid tab to attach client to!");
+        };
+
         self.active_tab_indices.insert(client_id, tab_index);
         self.connected_clients.borrow_mut().insert(client_id);
         self.tab_history.insert(client_id, tab_history);
         self.tabs
             .get_mut(&tab_index)
-            .unwrap()
+            .unwrap_or_else(|| panic!("Failed to attach client to tab with index {tab_index}"))
             .add_client(client_id, None);
     }
     pub fn remove_client(&mut self, client_id: ClientId) {
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 
 ## [Unreleased]
+* fix: crash when attaching to a session without the first tab (https://github.com/zellij-org/zellij/pull/1648)
 
 ## [0.31.1] - 2022-08-02
 * add: `solarized-light` theme to the example theme directory (https://github.com/zellij-org/zellij/pull/1608)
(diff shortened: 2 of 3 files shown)


har7an:
Curiously, when attaching now the status bar believes we are in tab mode when we are in fact not. Investigating...

Also I think this deserves a test to prevent regressions, but where/how would I test that? Is this a candidate for an e2e test?

imsnif:
> Curiously, when attaching now the status bar believes we are in tab mode when we are in fact not...

Is this something introduced by this fix?

har7an:
> Is this something introduced by this fix?

Hard to say, previously it would just crash and die when there wasn't a first tab. I guess you could say so.

imsnif:
Cool, happy you're looking into this.

About tests, maybe try here: https://github.com/zellij-org/zellij/blob/main/zellij-server/src/unit/screen_tests.rs

har7an:
@imsnif It seems the `mode_info` sent to the plugins has `mode` set to `Tab`. It happens in `server/src/screen.rs` around line 644, where the Screen queries itself for the `ModeInfo` of the client that is attached. I would expect that to turn up empty and fallback to the default, but it finds one with `ModeInfo.mode` set to `Tab`.

Do you have an idea how that can be? Are the client IDs shared in a session?

har7an:
Been fiddling with it but I cannot make out the reason for this behavior. Maybe it's because something in the code sees that the tab index is different from what it expects, and then just thinks "Oh we switched to a new tab (as in: We made a new tab), so we are in tab mode now"? Other than that I'm out of ideas for now.

Or maybe zellij starts out with the client being in "Tab" mode, sends that to the plugins, immediately updates that to be "Normal" mode and for some reason that doesn't propagate to the plugins anymore? Just guessing...

FYI: I'll be out of town until Tuesday, so if someone else wants to pick this up in the meantime that's fine by me.

zellij-server/src/screen.rs
```diff
@@ -702,21 +702,29 @@ impl Screen {
     }
 
     pub fn add_client(&mut self, client_id: ClientId) {
-        let mut tab_index = 0;
```
tlinford, line 705:
```suggestion
let mut tab_index = *self.tabs.keys().next().expect("no tabs found");
```

I think this is maybe easier than changing the logic of the ifs below?

zellij-server/src/screen.rs
```diff
@@ -702,21 +702,29 @@ impl Screen {
...
+            panic!("Can't find a valid tab to attach client to!");
+        };
+
         self.active_tab_indices.insert(client_id, tab_index);
         self.connected_clients.borrow_mut().insert(client_id);
         self.tab_history.insert(client_id, tab_history);
         self.tabs
             .get_mut(&tab_index)
-            .unwrap()
+            .unwrap_or_else(|| panic!("Failed to attach client to tab with index {tab_index}"))
```
tlinford, line 727: couldn't this just be an expect?

har7an:
On Wed, 2022-08-10 at 09:41 -0700, Thomas Linford wrote:
> @tlinford commented on this pull request.
> 
> 
> 
> >          self.active_tab_indices.insert(client_id, tab_index);
>          self.connected_clients.borrow_mut().insert(client_id);
>          self.tab_history.insert(client_id, tab_history);
>          self.tabs
>              .get_mut(&tab_index)
> -            .unwrap()
> +            .unwrap_or_else(|| panic!("Failed to attach client to tab with
> index {tab_index}"))
> 
> couldn't this just be an expect?

Originally it was, but that makes clippy unhappy, because `expect()` takes a
`&str`, so you end up doing `expect(format!(...))`. Apparently that's bad. :)

>

har7an:
I wouldn't call it easier, because IMO it makes the code less readable to the
extent that there are two places where the `tab_index` is handled separately.
Also this gets rid of the `mut`, since I'm assigning the whole thing in one go.
The idea of making things `mut` and then assigning them inside an `if` branch
just looks/feels too much like C to me, and I don't like that.

But you've been a project member for longer, so if you prefer this I'll accept
it, of course.

tlinford:
> Originally it was, but that makes clippy unhappy, because `expect()` takes a `&str`, so you end up doing `expect(format!(...))`. Apparently that's bad. :)

ha, I see.

> I wouldn't call it easier, because IMO it makes the code less readable to the
extent that there are two places where the `tab_index` is handled separately.
Also this gets rid of the `mut`, since I'm assigning the whole thing in one go.
The idea of making things `mut` and then assigning them inside an `if` branch
just looks/feels too much like C to me, and I don't like that.

I can get on board with that reasoning!

The one thing I still don't get though is the special casing for tab id 0: ` else if self.tabs.contains_key(&0) `? wouldn't that be covered by `else if let Some(tab_index) = self.tabs.keys().next()`?

PR #1658, opened 2022-08-14
fix(tab-bar): fix tab bar click when tabs are collapsed, fix tab bar length calculation

### LINKED ISSUE #1603: Probably incorrect calculation of tab length in tab-bar plugin
https://github.com/zellij-org/zellij/blob/9dc392e75b960f53b2b39bf40fca3eb912a91862/default-plugins/tab-bar/src/tab.rs#L46-L47

prscoelho:
Fixes #1603 , as well as another bigger issue. I noticed that clicking on a tab used index in the linepart vector, which will be incorrect for every tab if there's collapsed tabs. We have to save tab index in LinePart, and then use that to switch tabs.

Clicking on the collapsed tabs brings the user to the first hidden tab on that side.

Fixing the issue in #1603 was as discussed in the issue, as well switching the click comparison to be `click >= length && click < length` instead of `click > length && click <= length`

### DIFF
--- a/default-plugins/tab-bar/src/line.rs
+++ b/default-plugins/tab-bar/src/line.rs
@@ -26,8 +26,24 @@ fn populate_tabs_in_tab_line(
     loop {
         let left_count = tabs_before_active.len();
         let right_count = tabs_after_active.len();
-        let collapsed_left = left_more_message(left_count, palette, tab_separator(capabilities));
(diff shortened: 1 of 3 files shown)


prscoelho:
I split it into two commits, but let me know if you'd prefer this squashed.

imsnif (approved):
Sorry for taking so long to get to this! This looks great. I gave it a go and it fixes the problem. Thanks for your work on this.

PR #1663, opened 2022-08-14
A little refactoring

TornaxO7:
I went a little bit through the code and refactored some little stuff.

### DIFF
--- a/zellij-server/src/panes/csi.rs
+++ b/zellij-server/src/panes/csi.rs
@@ -1,6 +1,5 @@
 /// A trait for values which have a csi-representation
 pub trait Csi {
-
     /// Returns the csi representation of the value
     fn get_csi_str(&self) -> &str;
 }
--- a/zellij-server/src/panes/csi.rs
+++ b/zellij-server/src/panes/csi.rs
@@ -1,0 +1,6 @@
+/// A trait for values which have a csi-representation
+pub trait Csi {
+
+    /// Returns the csi representation of the value
+    fn get_csi_str(&self) -> &str;
+}
--- a/zellij-server/src/panes/csi.rs
+++ b/zellij-server/src/panes/csi.rs
@@ -1,5 +1,0 @@
-/// A trait for values which have a csi-representation
-pub trait Csi {
-    /// Returns the csi representation of the value
-    fn get_csi_str(&self) -> &str;
-}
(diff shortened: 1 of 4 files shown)


zellij-server/src/panes/terminal_pane.rs
```diff
@@ -118,63 +162,39 @@ impl Pane for TerminalPane {
...
-            },
-            _ => {},
-        };
+                HOME_KEY => {
+                    return AnsiEncoding::Home.as_vec_bytes();
+                },
+                END_KEY => {
+                    return AnsiEncoding::End.as_vec_bytes();
+                },
+                [27, 91, 50, 48, 48, 126] | [27, 91, 50, 48, 49, 126] => {
```
jaeheonji, line 186: Do you plan on refactoring this? If not, what about `BRACKETED_PASTE_BEGIN`
and `BRACKETED_PASTE_END` ?

zellij-server/src/panes/csi.rs
```diff
@@ -0,0 +1,5 @@
+/// A trait for values which have a csi-representation
+pub trait Csi {
```
jaeheonji, line 2: Is there a reason the trait is needed at this point? Do you need it for future development?

jaeheonji (changes requested):

zellij-server/src/panes/terminal_pane.rs
TornaxO7: Good point! I didn't knew what those stand for and asked you this when I looked at it.
Should be fixed now.

zellij-server/src/panes/csi.rs
TornaxO7: No, I don't need it for the future development, it's just in my opinion suitable to create a `trait` for this since it emphasizes the context of `get_csi_str`. I can remove it if you want.

jaeheonji: OK, I got it. 

However, this may be a slightly different topic, but as you know, in zellij, we use a lot of CSI code in various crates as well as `zellij-server`. 

So I think the end goal of this refactoring should be to provide all the CSI code from `zellij-utils`. And if we provide a variety of CSI codes, this trait will definitely make sense. (Of course, this is not to say that this PR covers all of this.)

Anyway, personally, for now, it seems like a good idea to refactor with less code.

TornaxO7: Alright, I'll move it back then.

TornaxO7: Done

jaeheonji (approved):
LGTM. When CI is finished, I'll merge it. Thank you!

Issue #1633, opened 2022-08-03
Mouse reporting/tracking and mouse encoding/coordinates behave incorrectly


AutumnMeowMeow:
zellij mouse modes have the following issues.  (See reference: https://www.invisible-island.net/xterm/ctlseqs/ctlseqs.html#h3-X10-compatibility-mode)  (Sorry a bit scatterbrained, hopefully the gist makes it...)

* Mouse tracking and coordinate encoding are independent of each other.  Selecting SGR coordinates (e.g. CSI ? 1006h) should not cause anything to be emitted unless one of the tracking modes is also enabled.  For example, "CSI ? 9 ; 1006h" would cause SGR mouse button down events, but nothing else.
* "CSI ? 1002 ; 1006h" would cause SGR mouse button down, drag, and release events, which is what zellij appears to be doing with just "CSI ? 1006h".
* zellij is not recognizing DECSET with multiple parameters, e.g. "CSI ? 1002 ; 1006h" is not doing anything at all, while "CSI ? 1006h" is doing the equivalent of "CSI ? 1002 ; 1006h".  I think [at this spot](https://github.com/zellij-org/zellij/blob/main/zellij-server/src/panes/grid.rs#L2020) it is reading only the first param; it needs to loop over all of the params.

grid.rs needs to add a mouse encoding (which can be just SGR (1003 h) and X10 or NONE (1003 l) in today's world) which is distinct from the mouse events to report (button down only (9), button down and up (1002), button down/up and motions (1003)).

(It would be _really_ sweet if any-event (1003) was also supported, then a moving text mouse cursor could work happily.  But that's a feature request, not a bug.  :) )

imsnif:
Thanks for the details!
While I get what you're saying about not assuming any `1006h` is `CSI ? 1002 ; 1006h` - I think I don't fully understand the rest of your suggestions.

What other type of behaviour, aside from any-event*, should we implement? I'm just trying to avoid implementing 4-5 behaviours that no-one uses anymore :) So which ones exactly, aside from the `1002` which we're doing right now, do you suggest we implement?

*Which I think would be cool to implement for focused panes who requested it!

AutumnMeowMeow:
(Sorry about my rambling, it was a long few weeks. :) )

Widest coverage of modern applications would be modes 1000, 1002, 1003, 1005, and 1006.  BTW these can all be tested quickly in vttest under menu 11.8.5 "XTERM mouse features".

[...]

PR #1664, opened 2022-08-14
Fix SGR/UTF8 mouse reporting in terminal panes

imsnif:
This is a fix for: https://github.com/zellij-org/zellij/issues/1633

Main changes:
1. We now track which mouse button was released when collecting user input (rather than sending a generic "release" event on to the zellij server).
2. DECSET/DECRST now loop through all arguments rather than just taking the first one (oops!)
3. Mouse mode is now tracked as "NoEncoding", "SGR" or "UTF8" rather than true/false
4. We now also track the mouse tracking method as "Off", "Normal" and "ButtonEventTracking", reporting nothing, button press + release and button press + release + hold respectively.
5. We send the relevant events from 4 with the relevant encoding from 3 to the terminal panes as needed. Defaulting to utf8 encoding.

There's room to implement AnyEvent tracking in supported terminals, but that's out of scope for this PR.


imsnif:
@tlinford - if you could give this a spin to make sure I didn't break any mouse stuff that I'm not aware of that would be great :)

@AutumnMeowMeow - if you have the time and feel like it, would you like to check this change out and let me know what you think?

AutumnMeowMeow:
@imsnif Testing inside vttest, I see the following issues:

* Off-by-one coordinates: SGR mode (1006), button-event (1002), during a drag-with-button-down: the coordinates are incorrectly +1 rows +1 cols.  The button-down and button-up coordinates are correct.
* Off-by-one coordinates: UTF8 mode (1005), normal-event (1000), during a button-release: the coordinates are incorrectly +1 rows +1 cols.  The button-down coordinates are correct.
* Off-by-one coordinates: UTF8 mode (1005), button-event (1002), during a button-release and during a drag-with-button-down: the coordinates are incorrectly +1 rows +1 cols.  The button-down coordinates are correct.
* When a test is exited via 'q', mouse tracking is still on and not off.  You are missing the corresponding cases [from here](https://github.com/zellij-org/zellij/blob/fix-mouse-reporting/zellij-server/src/panes/grid.rs#L2345) for 1000, 1002, 1003 (TBD) that belong [here](https://github.com/zellij-org/zellij/blob/1b2e52cfc323be0f8414af7010a4fd53e87da8ab/zellij-server/src/panes/grid.rs#L2270).

imsnif:
Thanks @AutumnMeowMeow ! I guess I was pretty tired when I submitted this. I even glanced over these in the tests :) All fixed.

AutumnMeowMeow:
@imsnif It's better now, but I still see the following:

* Button 2 (usually middle button) is not reporting on button-down or button-down-and-drag, but is reporting on button-up/release.
* In SGR coordinates, when button 2 reports button-up, it incorrectly reports it as button 1 (code 0) button-up.
* In both UTF-8 and SGR coordinates, button-down-and-drag is not being reported as a drag (button code + 32), but rather as button-down (button code).

imsnif:
@AutumnMeowMeow - all fixed. These were part misunderstanding on my part, and part Zellij not previously fully supporting middle-clicks. Thanks again!

tlinford:
@imsnif gave it a test, all of these seem fine to me:

- pane/tab selection with mouse
- moving floating pane
- selecting text
- scrolling with mouse wheel
- interacting with nvim with `set mouse=a`

AutumnMeowMeow:
I confirm normal-event and button-event are working correctly for me:

https://user-images.githubusercontent.com/99296476/184955817-b1c97eff-e816-4746-a251-2bcd46533714.mp4

Any-event is not yet working, so no tracking mouse cursor tracking, but that is expected.  Click/drag/release and wheel-up/wheel-down are working well.

I would say #1633 is fixed with this. :)

imsnif:
Fantastic! Thank you both for your help.

Issue #1607, opened 2022-07-27
[Feature request] send-prefix behaviour?


budimanjojo:
In tmux, there's a `send-prefix` command to set a prefix key when running as "nested client".
This gives user the ability to run commands in the nested instance instead of the "server". Workflow in tmux:
`Ctrl+b` is the normal prefix key, while `Ctrl+z` is the client prefix key. To create new window in the outer layer, run `Ctrl+b`, then use the key to create new window, And when I want to create a new window in the nested layer, I press `Ctrl+z`, then run the key to open new window.
It looks like there's no way to do it like `tmux` does, because there's no "mode" in tmux, everything is done with prefix key.
Is it reasonable to have a client mode that will do everything in the nested instance when it's available? So I can do something like an action `SwitchToMode: Client` and just add the keybindings to that mode.
Related issue: https://github.com/zellij-org/zellij/issues/387

har7an:
@budimanjojo It's not exactly what you asked for, but does `Locked` mode work for you? By default that's bound to `Ctrl+g`, and will then forward all keystrokes (except `Ctrl+g`) to the active pane, which is the nested session in your case. Unlike the `send-prefix`, `Locked` mode is more of a toggle. So to send a single key through and work in the outer session after that you must press `Ctrl+g, <Key>, Ctrl+g`.

In theory (I haven't tried this myself), you can give your nested zellij session an entirely custom set of keybindings. The outer session should forward all keybindings it doesn't know to the application, which would be the nested zellij session then. As long as they don't share keybindings, that should work. Although I agree that it's probably awkward to learn 2 sets of keybindings... However, the latest release shows your current keybindings in the status bar, so you can always peek. :)

But that is just a bunch of workarounds for the time being, really. Having a `send-prefix` sounds like a good idea (for example to forward `Ctrl+g` to vim, or something like this).

budimanjojo:
Thanks! I'm currently using `Locked` mode to work like tmux "prefix key". Maybe I can tweak a little bit to make it work like `send-prefix`, I have some ideas, thanks for the hint!

budimanjojo:
Ah I just read the `As long as they don't share keybindings` part after so many tries with no success

budimanjojo:
I found a workaround for this. I now have another key bound to `SwitchToMode: Locked`. and inside `locked mode` I set the same key as `SwitchToMode: Tmux`. This is how my config looks like:
```
  normal:
    - action: [SwitchToMode: Tmux]
[...]

Issue #775, opened 2021-10-11
missing feature: passthrough?


colemickens:
What happens when I run `zellij` on my host, and `zellij` on my remote machine, thus rendering two sets of frames? How do I pass through keystrokes to another instance.

This already sort of works with only-two sessions, because you can lock the top layer and trigger the inner second layer, but obviously this doesn't work for everything, doesn't generalize past two layers, etc.

It seems that, as `tmux` offers, there needs to be a way to say "the next keystroke specifically goes to the focused pane".

imsnif:
So, this is actually a big problem in my personal workflow. When I develop Zellij, I am forced to open my "development" copy in a different terminal window because of this. So I feel strongly about solving it :)

I like the idea of having a shortcut that would say "the next keystroke specifically goes to the focused pane" as you suggest. I kind of feel though that we might be able to make this an additional "escape hatch" to an otherwise more intuitive out-of-the-box workflow.

Here's my suggestion:
When we have nested Zellij sessions, the commands will *always* go to the deepest session, unless:
1. If the deepest session is locked, it will go one layer up
2. If all sessions are locked and the user presses `ctrl-g` it will unlock all sessions

What do you think?

Lillecarl:
@imsnif Is zellij able to do two-way communication between nested sessions in a generic way? If so a ctrl-$ N would be an option as well, where $ is some key we didn't decide on yet and N is the depth you want to execute upcoming commands at.

This way you can easily jump back and forth between levels, which is useful when sshing through jumpboxes/bastions.

imsnif:
@Lillecarl - if I'm being honest this sounds a bit complex both to implement and to use... do you find yourself very often with more than 2-3 nested Zellij sessions?

Lillecarl:
@imsnif 3 isn't entirely uncommon, one on my host, one on a bastion, one on the server i'd like to remote.

An action to unlock all the way down rather than having to cycle through locks could probably work nice too. But relying on pressing something multiple times in a sequence will 100% lead to people "overshooting" the lock status quite a bit. So that might be a good middle ground.

Though if the framework of bidi communication is in place that could find a lot of other cool usecases in the future (But beyind the scope of this issue).

I think I've nested 4x at most. usually the panes are fullscreen then and I mostly use Zellij/tmux to not lose state.

a-kenji:
I think that is a solution that would fit well I think.
It could be bound to the session-mode. And for example 
Up/Down - J/K - Could be session up / session down.

[...]

Issue #418, opened 2021-04-30
Unable to send ctrl-g with default config


muffin4:
## Description

With the default config I can't seem to find a way to send `Ctrl-g` to any programs I run in zellij because it's the key to both enter and exist the locked mode.

You can send any other key that's already used by the default config by entering the locked mode with `Ctrl-g`, then hitting the key you want to send, and leaving locked mode with `Ctrl-g` again, but that doesn't work for `Ctrl-g` itself.

## Information

- zellij 0.6.0
- Arch Linux 5.11.16-zen1-1-zen

imsnif:
True. This is indeed an issue. I think it's one of the reasons `Ctrl-g` was chosen, since there aren't that many uses for it. Out of curiosity: is there a specific use this is blocking for you?

Also: do you have a suggestion on a UX for this?

muffin4:
I noticed that I'm unable to send `Ctrl-g` when I tried to use it in vim to print the current file name.

Some solution that come to mind are these:

- Change the keybind to leave locked mode to `esc`. You can already send `esc` in normal mode, though you might want to send `esc` in locked mode without wanting to leave locked mode.

- If `Ctrl-g` is the first key entered in locked mode: send `Ctrl-g` and leave locked mode. This behavior is similar to `Ctrl-b Ctrl-b` with the default tmux config, but this prevents users from leaving locked mode without sending anything if they entered it by accident.

- Add a keybind that sends only the next character, like `Ctrl-v` in vim's insert mode.

imsnif:
>     * Add a keybind that sends only the next character, like `Ctrl-v` in vim's insert mode.

This is my favorite solution. I would also:
1. Make this keybind `Ctrl-g` and change the current behaviour to happen with `Ctrl-G` (uppercase `G`). This way we can make just a minor change to the interface by changing it to `<g/G> LOCK`
2. Change the `-- INTERFACE LOCKED --` indication for the lowercase variant (the temporary one) to `-- INTERFACE TEMPORARILY LOCKED --` to make it clear to the user.

What do you think?

a-kenji:
I also think that this is the best solution.

imsnif:
(I'm unsure if terminals are able to detect ctrl-shift-g, btw - if that doesn't work we might have to change it to a different char)

SaintFenix:
This would improve on a feature that tmux has that Zellij doesn't, the ability to pass through the prefix key. 

[...]

Issue #387, opened 2021-04-26
Feature request: nested zellij instances (e.g. over ssh)


alxhub:
Currently, running `zellij` inside `zellij` does work, but is not specially handled. A really good experience could be provided though, especially in combination with `ssh`:

1. I'm inside a tab in `zellij` on my local machine.
2. I `ssh` to a remote machine.
3. I launch `zellij` on the remote machine.

Rather than running an independent instance of `zellij` that's tricky to control as it requires escape sequences to operate, `zellij` could instead recognize that it's running inside another instance and coordinate. It would be possible to:

* distinguish local vs remote tabs/panes, e.g. through color or name prefixes
* open new tabs/panes for the remote machine, coordinated by the inner `zellij` instance
* navigate seamlessly between local and remote

Lillecarl:
There's a really cool thing that Zellij could do that'd "remove" the need for running Zellij on the remote machine.

The terminal emulator Wezterm supports running SSH through it, and then you can spawn new wezterm tabs they'll be in the same SSH session because of something called SSH channels, Wezterm is also written in rust and uses libssh2 for this so i guess quite a bit of code could be reused from that project (both WezTerm and Zellij are MIT licensed too so copypasta is cool)

This would kinda imply adding a command mode to zellij to be able to run the ssh command, or if there's a way for the zellij binary to know which pane it runs in, which then can be used to instruct the zellij server to initiate an ssh connection in that pane. Think zellij ssh user@host.

When i saw this feature in Wezterm my jaw dropped, it's such an obviously awesome feature.

```
ssh2 = {version="0.9.3", features=["openssl-on-win32"]}
libssh-rs = {version="0.1.1", features=["vendored"]}
```

https://wezfurlong.org/wezterm/ssh.html

milanglacier:
Both tmux and zellij treats nested session as is...

But most of the time, nested tmux works just fine because its keybinding scheme is centered around the prefix key. Say if you want to control the nesteed tmux session, you just need to press `Control b` twice to send the real `Control b` to send the real `control b` to the terminal which then forwards to the nesteed tmux session.

In zellij, you need to firstly lock the outer level session by `C-g` and then you are able to send keys like `C-p`, `C-n` into the inner level zellij session. But here comes the problem, you can never send a real `C-p`, `C-n` stuffs into the nested tmux session because you can never really lock the nested zellij session because you can never send a real `C-g` into the terminal.

[...]

PR #1668, opened 2022-08-16
Feature: key passthrough

har7an:
The idea for this feature came to my mind after reading through #1607, #775, #418 and #387. I'm sure there are more open issues that may benefit from having this "fix", but I don't feel like combing through all of them at this very moment. :)

*The following discussion about keybindings etc. assumes you use the default keybindings as of the moment of writing this.*

## Description

This PR adds a new input mode called `Passthrough`, which does exactly what it says: It passes the next key-combination you enter through to the application, verbatim.

It is currently bound to `Ctrl+z`, but that's just a proposal and open for discussion.

## The underlying problem

Unlike other terminal multiplexers like tmux, zellij doesn't have the notion of a **prefix key**.

**In tmux**, this prefix key *must* be pressed before performing any tmux-specific action (such as creating a split, a new tab, detaching, etc). This means that, conversely, any key-combination entered by the user that isn't the prefix key is passed directly to the currently active pane. Hence, there is only one real collision in the keybindings: When the underlying application has a keybindings that is identical to the tmux prefix key.

This is usually worked around in one of two ways:

1. Declare a keybinding that performs the `send-prefix` "command", and sends the prefix key to the nested application, or
2. Declare a custom keybinding that acts as prefix key for the inner session directly

**Zellij** takes a different approach in this regard: It has a plethora of keybindings that are directly accessible without having to press a prefix key first. While this makes a lot of tasks quicker to execute by default (e.g. changing focus = `Alt + <hjkl>`, for example), it also creates much more collisions for application keybindings.

This issue becomes even more apparent when nesting zellij sessions: Assuming both sessions use the same keybinding, controlling the inner session is only sensibly possible when entering Locked mode in the outer session. Note that it is currently **impossible** to send the Locked mode binding to an application. Hence, the inner zellij session cannot be locked. Also, changing focus in the inner session while leaving control to the outer session requires the user to press `Ctrl+g, Alt+<hjkl>, Ctrl+g`, which is very awkward.

## Limitations of this solution

- In its current implementation, the outermost zellij session is dropped back into normal mode after using the passthrough mode, regardless of the input mode the user was in previously.
- It doesn't scale: One level of nesting is perfectly manageable this way, but anything below that becomes unwieldy. Think of having three zellij sessions nested, to send `Alt + h` to the innermost one requires the following input: `Ctrl+z, Ctrl+z, Ctrl+z, Alt+h`.

Still I think that it's a good step in the right direction and will likely benefit most users already.

Feedback is very welcome!

### DIFF
--- a/default-plugins/status-bar/src/first_line.rs
+++ b/default-plugins/status-bar/src/first_line.rs
@@ -864,4 +864,36 @@ mod tests {
 
         assert_eq!(ret, " Ctrl +  a  b  c ".to_string());
     }
+
+    #[test]
+    fn first_line_passthrough_wide() {
+        #[rustfmt::skip]
+        let mode_info = ModeInfo{
+            mode: InputMode::Passthrough,
+            keybinds : vec![
+            ],
+            ..ModeInfo::default()
+        };
+
+        let ret = first_line(&mode_info, 30, "");
+        let ret = unstyle(ret);
+
+        assert_eq!(ret, " PASSTHROUGH ".to_string());
+    }
+
+    #[test]
+    fn first_line_passthrough_narrow() {
+        #[rustfmt::skip]
+        let mode_info = ModeInfo{
+            mode: InputMode::Passthrough,
+            keybinds : vec![
+            ],
+            ..ModeInfo::default()
+        };
+
+        let ret = first_line(&mode_info, 10, "");
+        let ret = unstyle(ret);
+
+        assert_eq!(ret, "".to_string());
+    }
 }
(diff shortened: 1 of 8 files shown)


har7an:
## Current state:

- Passes the next pressed key-combination through to the nested application
- Has status indication in the status bar
    - In the first line (prints a single, highlighted ` PASSTHROUGH >` tile)
    - In the second line (prints an explanation of what happens here)


## TODO

- Needs unit tests
- Maybe add a status bar tip?

imsnif:
@har7an - I'm sorry, I know you put a lot of work and thought into this - but I'm very much against adding any sort of keybinding to the "upper" level. This would be doubling down on a mistake we made as a tool by having so many keybindings on the upper level. A mistake we're scheduled to correct (https://github.com/zellij-org/zellij/issues/1399).

That being said, I think this is a great feature to add as part of the above correction (maybe a "transient" mode to locked mode that can be toggled?) - but I think it should be approached as part of that solution and not now.

PR #1669, opened 2022-08-16
add(config): send ^g in session mode

a-kenji:
Bind sending `^g` to the session mode of the default configuration.
That way analogous to how the tmux mode can be nested, now the locked
mode can also be nested in the default configuration.

Example: `^o-^g` Will lock/unlock the inner session, if running the
default layout.

[send^g.webm](https://user-images.githubusercontent.com/65275785/184886869-63774628-3f62-4912-a940-1d21b5dd293e.webm)

### DIFF
--- a/zellij-utils/assets/config/default.yaml
+++ b/zellij-utils/assets/config/default.yaml
@@ -467,6 +467,8 @@ keybinds:
           key: [Ctrl: 'o', Char: "\n", Char: ' ', Esc]
         - action: [SwitchToMode: Scroll,]
           key: [Ctrl: 's']
+        - action: [Write: [7,], SwitchToMode: Normal]
+          key: [Ctrl: 'g']
         - action: [Quit,]
           key: [Ctrl: 'q',]
         - action: [Detach,]


a-kenji:
@har7an,

Is this an acceptible alternative to #1668 for now for you?

imsnif:
That's a cool idea, but it kind of breaks both the consistency of the UI and the actual UI, as one would expect `ctrl-g` to send the outer session to locked mode. I feel this behaviour would be very unexpected unless it's explained through the UI.

har7an:
@a-kenji And it still requires `Ctrl+g, Alt+h, Ctrl+g` to change focus in the inner zellij, or create a new pane, etc.
But it definitely solves the issue of sending the locked mode keyshortcut to a panes application.

a-kenji:
@imsnif,
Oh yes I didn't see that the `^g` was already used in the session mode, good catch! How about `g` in session mode to send the key?

@har7an,
This doesn't have all the nice functionality, yes. The tmux mode is a little bit easier to use for that, because it can also pass through the keys to it's inner session and it reverts to the normal session then.

imsnif:
@har7an and I have just been discussing this. He's going to look into implementing https://github.com/zellij-org/zellij/issues/775 which I think is the more holistic solution here.

a-kenji:
Sure, sounds good.

PR #1681, opened 2022-08-19
feat(#195): display current path in the strider

knidarkness:
This pull request addresses the #195 issue and resolves the problem of @RobWalt's solution with clicks 1-row offset.  

As in Rob's PR, the path displayed is a relative path from the `zellij`'s start location.

Display the current path on top of the directory files in the strider default plugin. As of now, no configuration
is available for this feature.

Examples of how it looks:

image

image

### DIFF
--- a/default-plugins/strider/src/main.rs
+++ b/default-plugins/strider/src/main.rs
@@ -58,10 +58,14 @@ impl ZellijPlugin for State {
                 Mouse::ScrollUp(_) => {
                     *self.selected_mut() = self.selected().saturating_sub(1);
                 },
-                Mouse::Release(line, _) => {
-                    if line < 0 {
+                Mouse::Release(mut line, _) => {
+                    // we need to shift the line with line -=1 because one line
+                    // is occupied by the current path, and all clicks on it are
+                    // to be ignored
+                    if line < 1 {
                         return;
                     }
+                    line -= 1;
                     let mut should_select = true;
                     if let Some((Event::Mouse(Mouse::Release(prev_line, _)), t)) = prev_event {
                         if prev_line == line
@@ -85,7 +89,14 @@ impl ZellijPlugin for State {
     }
 
     fn render(&mut self, rows: usize, cols: usize) {
-        for i in 0..rows {
+        let display_current_dir = FsEntry::DisplayDir(self.current_dir.clone())
+            .as_line(cols)
+            .normal()
+            .on_yellow()
+            .black();
+        println!("{}", display_current_dir);
+
+        for i in 0..rows - 1 {
             if self.selected() < self.scroll() {
                 *self.scroll_mut() = self.selected();
             }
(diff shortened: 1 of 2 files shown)


default-plugins/strider/src/state.rs
```diff
@@ -50,38 +52,89 @@ impl State {
...
-        path.file_name().unwrap().to_string_lossy().into_owned()
+        match self {
+            FsEntry::File(..) | FsEntry::Dir(..) => {
+                // only use the filename
+                path.file_name().unwrap().to_string_lossy().into_owned()
+            },
+            FsEntry::DisplayDir(..) => {
+                // use full path, but we need to remove the host part
+                let path = path.to_string_lossy().into_owned().replace("/host", "");
+                ".".to_string() + &path
```
tlinford, line 73: Think this can be a bit simpler:
```suggestion
                path.to_string_lossy().replace("/host", "")
```

default-plugins/strider/src/main.rs
```diff
@@ -58,10 +58,14 @@ impl ZellijPlugin for State {
...
+                    // we need to shift the line with line -=1 because one line
+                    // is occupied by the current path, and all clicks on it are
+                    // to be ignored
+                    if line < 1 {
                         return;
                     }
+                    line -= 1;
                     let mut should_select = true;
                     if let Some((Event::Mouse(Mouse::Release(prev_line, _)), t)) = prev_event {
                         if prev_line == line
```
tlinford, line 71: since line has been decremented, I think prev_line needs to be adjusted too?
I noticed double clicks to enter a directory aren't working reliably.

tlinford (changes requested):
Thanks for picking this up!
I gave it a quick run and found two small things to adjust :)

default-plugins/strider/src/main.rs
knidarkness: oops, my bad, had the `prev_line` decrement, but accidentally deleted it.

knidarkness: @tlinford , done, double clicks seem to be working fine now.

tlinford (approved):

default-plugins/strider/src/state.rs
```diff
@@ -50,38 +52,88 @@ impl State {
...
     }
 
     pub fn as_line(&self, width: usize) -> String {
+        let name = self.name();
         let info = match self {
             FsEntry::Dir(_, s) => s.to_string(),
             FsEntry::File(_, s) => pb::convert(*s as f64),
+            FsEntry::DisplayDir(..) => {
+                let current_path = name.clone();
+                FsEntry::display_with_shortened_name_start(current_path, width)
```
tlinford, line 84: should just return empty string? just did another run and i'm seeing extra output on the right

tlinford (changes requested):
sorry I missed one other thing.
screenshot-2022-08-23T20:17:02

default-plugins/strider/src/state.rs
knidarkness: oh, thanks for pointing this out. I will check this today in the evening.

imsnif:
Hey, what's the status here?

davidfufu:
Want to echo @imsnif , what's the status on this? This is an important feature that makes the file tree more useful, unless there's another way to switch dir within the strider window.

chanson02:
Looks like this PR can be closed, as it's functionality has been implemented in #3200.

Issue #1460, opened 2022-06-07
Add action to dump the whole screen


a-kenji:

evolvomind:
Hi, @a-kenji
I am a new contributor and would like to start from here.
Could you let me know if this issue has been resolved? May I work on this issue?

majiayu000:
Hi, I've been looking into this and traced the root cause.

**Proposed approach:** Add a new action (e.g. DumpScrollback or extend DumpScreen) that captures the full scrollback buffer rather than just the visible area. Follow the existing DumpScreen action-dispatch pattern. Avoid renaming existing actions (lesson from PR #1698 being closed). Wire the new action through the keybinding config, action enum, and server handler. Add integration tests.

I'll open a draft PR shortly. Happy to adjust based on your preference.

PR #1698, opened 2022-08-30
PR For Issue "Add: action to dump the whole screen #1460"

jafriyie1:
This PR contains a new feature for dumping the entire screen's buffer into a file. I have renamed `DumpScreen` to `DumpPane` (since the original `DumpScreen` only dumped the buffer from an active pane), and created a new Action `DumpScreen`. This action will take the current Tab and dump all of the panes scroll-back buffer, besides the status-bar and tab-bar, to a file. This PR addresses the issue presented here: https://github.com/zellij-org/zellij/issues/1460 . As this is my first PR on this project please let me know if I missed anything! Thanks!


imsnif:
ping @a-kenji ?

jafriyie1:
@a-kenji When you get the chance, could you please take a look at this PR?

jafriyie1:
@imsnif @a-kenji sorry to be a bother but what is the status on getting this reviewed? Just wanted to double check, no rush :) .

a-kenji (changes requested):
LGTM, but please remove the binaries from the commit.

imsnif:
My two cents here, because I don't know if @a-kenji has time these days?

I would like to understand the use-case for this action before we merge it. I realize @jafriyie1 did a lot of work here and was very patient with us reviewing. I don't want to take a way this clearly valuable work, but I'm also wary of clogging up our internal and external interface with actions that we'll have to maintain if there isn't at least one person who says "I need this for my daily workflow!". So... is there such a person? :)

jafriyie1:
@imsnif no worries at all i understand :) !!! i'm always happy to pivot and contribute to other parts of Zellij if we find better use cases :) ! so i looked online and it seems that some people wanted the option of dumping their Zellij session to a file (either for a given pane or the whole screen). if we find that we want to incorporate this `Dump` action, then i can delete this PR, incorporate the newer changes of Zellij, and put in my changes in a new PR. otherwise, i would be happy to delete this PR and pickup a different issue that we think would be helpful for people

jafriyie1:
@imsnif yes that sounds good :) ! I'll look at some other issues and/or reach out to you on Discord. Thanks :) !

PR #1724, opened 2022-09-11
fix(nix) missing frameworks on darwin

mdavezac:
Zellij would not link without the DiskArbitraction and Foundation frameworks on macos (Big Sur). It seems sysinfo requires them. They must be passed as `buildInputs` in the nix flake.


a-kenji:
Thank you @mdavezac!

a-kenji (approved):

PR #1728, opened 2022-09-13
test: simplify tab integration tests

tlinford:
Reduce boilerplate needed to write tab integration tests that need to check what is being written to pty.

### DIFF
--- a/zellij-server/src/tab/unit/tab_integration_tests.rs
+++ b/zellij-server/src/tab/unit/tab_integration_tests.rs
@@ -160,11 +160,11 @@ impl MockPtyInstructionBus {
         handle.join().unwrap();
     }
 
-    fn sender(&self) -> SenderWithContext<PtyWriteInstruction> {
+    fn pty_write_sender(&self) -> SenderWithContext<PtyWriteInstruction> {
         self.pty_writer_sender.clone()
     }
 
-    fn get_output(&self) -> Vec<String> {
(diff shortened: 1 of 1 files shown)


zellij-server/src/tab/unit/tab_integration_tests.rs
```diff
@@ -104,6 +106,69 @@ impl ServerOsApi for FakeInputOutput {
...
+
+    fn exit(&mut self) {
+        self.pty_writer_sender
+            .send(PtyWriteInstruction::Exit)
+            .unwrap();
+        let handle = self.handle.take().unwrap();
+        handle.join().unwrap();
+    }
+
+    fn sender(&self) -> SenderWithContext<PtyWriteInstruction> {
```
imsnif, line 163: Can we call this pty_write_sender?

zellij-server/src/tab/unit/tab_integration_tests.rs
```diff
@@ -104,6 +106,69 @@ impl ServerOsApi for FakeInputOutput {
...
+        self.pty_writer_sender.clone()
+    }
+
+    fn get_output(&self) -> Vec<String> {
```
imsnif, line 167: Maybe we can call this `clone_output` to be explicit?

imsnif (changes requested):
Looks great! Just some minor suggestions with naming :)

Issue #1734, opened 2022-09-16
Crash in tab module


sorenrade:
**Basic information**

`zellij --version`: 0.31.3
`stty size`: 50 196
`uname -av`: Linux fresh 5.15.0-46-generic # 49-Ubuntu SMP Thu Aug 4 18:03:25 UTC 2022 x86_64 x86_64 x86_64 GNU/Linux

alacritty 0.11.0-dev (ebc6922e)
helix 22.08.1 (66276ce6)
nushell: 0.68.1

###  Error
```
Error occurred in server:

  × Thread 'screen' panicked.
  ├─▶ Originating Thread(s)
  │   	1. screen_thread: HandlePtyBytes
  │
  ├─▶ At /home/ahe/.cargo/registry/src/github.com-1ecc6299db9ec823/zellij-server-0.31.3/src/tab/mod.rs:1031:75
  ╰─▶ called `Option::unwrap()` on a `None` value
```

### Tail of /tmp/zellij-1001/zellij-log/zellij.log
``` 
... lots of similar warnings ...
WARN   |zellij_server::panes::gri| 2022-09-15 20:53:24.920 [screen    ] [/home/ahe/.cargo/registry/src/github.com-1ecc6299db9ec823/zellij-server-0.31.3/src/panes/grid.rs:2744]: Unhandled esc_dispatch: 92->[] 
WARN   |zellij_server::panes::gri| 2022-09-15 20:53:24.920 [screen    ] [/home/ahe/.cargo/registry/src/github.com-1ecc6299db9ec823/zellij-server-0.31.3/src/panes/grid.rs:2207]: Unhandled osc: [[49, 51, 51], [68], [48]] 
WARN   |zellij_server::panes::gri| 2022-09-15 20:53:24.920 [screen    ] [/home/ahe/.cargo/registry/src/github.com-1ecc6299db9ec823/zellij-server-0.31.3/src/panes/grid.rs:2744]: Unhandled esc_dispatch: 92->[] 
ERROR  |zellij_utils::errors     | 2022-09-15 20:53:28.052 [screen    ] [/home/ahe/.cargo/registry/src/github.com-1ecc6299db9ec823/zellij-utils-0.31.3/src/errors.rs:94]: Panic occured:
             thread: screen
             location: At /home/ahe/.cargo/registry/src/github.com-1ecc6299db9ec823/zellij-server-0.31.3/src/tab/mod.rs:1031:75
             message: called `Option::unwrap()` on a `None` value 
[...]

PR #1748, opened 2022-09-22
Server: Remove `panic`s in `tab` module

har7an:
Removes all calls to `unwrap()` in the tab module of `zellij-server`. Makes all relevant functions return `Result`s instead, attaching error context where possible to trace error origins better. Also updates the code in `server::screen` to accept the new `Result`s and attach context there, too.

Related to #1734.

I want to test what it looks like first by provoking some errors in the code to see whether it's really any good.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -40,6 +40,9 @@ name = "anyhow"
 version = "1.0.57"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "08f9b8508dccb7687a1d6c4ce66b2b0ecef467c94667de27d8d7fe1f8d2a9cdc"
+dependencies = [
+ "backtrace",
+]
 
 [[package]]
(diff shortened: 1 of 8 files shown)


har7an:
Note to self: Must replace calls to `context()` with `with_context(|| format!(...))` to get the implicit formatting.

har7an:
ping, @imsnif

har7an:
I'll merge this now (if CI passes) since with the recent activity around error handling I expect merge conflicts to become much more frequent. We can still revert it before the next release when necessary, or I'll patch it as required.

Note to self: Discuss whether we should drop `strip = yes` in Cargo.toml to keep stack traces useful.

imsnif:
@har7an - please please please wait for the KDL PR before merging!
I have a very long TODO list before we can release the next version and merging this will very likely block a lot of it.

Issue #1735, opened 2022-09-16
Zellij layout with session name called inside the session causes recoursive calamity


alekspickle:
**Basic information**

`zellij --version`: 0.31.4
`stty size`: 36 137
`uname -av` or `ver`(Windows): Linux pickle 5.15.0-47-generic #51-Ubuntu SMP Thu Aug 11 07:51:15 UTC 2022 x86_64 x86_64 x86_64 GNU/Linux

`nvim --version`: NVIM v0.7.0 (appimage installation)
`alacritty --version`: alacritty 0.10.1

**Further information**
Have this layout in `~/.config/zellij/layouts/default.yaml`:
(I believe single `session` field is enough to reproduce, but just for the context I am providing my env)
```

---
session:
  name: "nevermind"
template:
  direction: Horizontal
  parts:
    - direction: Vertical
      borderless: true
      split_size:
        Fixed: 1
      run:
        plugin:
          location: "zellij:tab-bar"
    - direction: Vertical
      body: true
    - direction: Vertical
      borderless: true
      split_size:
        Fixed: 1
      run:
        plugin:
          location: "zellij:status-bar"
tabs:
  - direction: Vertical
    name: dev
  - direction: Vertical
    name: misc


```

Now if you do `zj` and run it again inside the session it obviously will try to attach session with such name and you see it doing just that but hundreds of times. After you close this calamity with Ctrl-C, zellij CLI is completely unresponsive until you do `killall zellij`.

I am not sure this should be handled at all, I just thought I let you know about this behavor.

Maybe there should be some command to fix such garbaged environments? Like `zellij clean`(inspired by `cargo clean`)

har7an:
Thanks for reporting this!

That's surely an interesting bug you found. I don't have the slightest suspicion what's causing it, but we'll investigate.

har7an:
This shouldn't have happened in the first place, and in fact we explicitly check in the code whether you try to attach to a zellij session you're currently already in. That's why e.g. `zellij attach nevermind` from inside your layout wouldn't have worked, it gives an error instead. But the session names defined in layouts are treated differently, hence it wasn't catched.

The next release will not contain this bug any longer and give you an error instead.

Well spotted, @alekspickle, and thanks again for reporting!

PR #1766, opened 2022-10-01
zellij/commands: Prevent recursive sessions

har7an:
with session names specified in layout files. A "recursive session" is created when, while running inside some zellij session, a user attempts to spawn zellij and make it attach to that same session.

When attaching via CLI (`zellij attach`) we explicitly check for this condition and error out when needed.

However, besides `zellij attach` it is also possible to declare the session to attach to in layout files like so:

```yaml
session:
  name: "foo"
```

This takes a different codepath when starting zellij, and hence bypases the checks we already have in place to avoid recursive sessions. Hence, we implement the check in the other codepath, too, and prevent recursive sessions from happening for good.

Fixes: #1735


har7an:
@imsnif or @a-kenji could either of you have a quick look over this? I'm not sure whether this is the right place to fix the issue. I did some tests, but I'd like to avoid users being unable to start their sessions in the next release.

PR #1768, opened 2022-10-01
Add SoftClear action to replicate macOS Cmd+K behavior

fcoury:
Replicates the behavior of the `Cmd+K` command used on macOS's Terminal.app and iTerm.

https://user-images.githubusercontent.com/1371/193431625-276f38a7-cf3c-4ce3-a7b7-0f32472f7d15.mp4

This PR also adds a new configuration setting `soft_clear_offset` that allows the user to specify an offset, counted from the top of the screen, of how many additional lines he/she wants to see  when performing this soft clear. This overcomes a problem I face with iTerm with the same command, where it trims the first line of my prompt (see below), while retaining a sane default for most uses.

https://user-images.githubusercontent.com/1371/193431677-aa8942f5-6db9-4d1d-b685-a6336584f7fb.mp4


imsnif (changes requested):
Hey @fcoury - thank you for your patience!

I'm happy to accept this, but would ask for a few changes:
1. How about if instead of `roll_to_top` we clear the viewport and then add canonical lines to it similar to how we initialize it in the constructor (`viewport: vec![Row::new(columns).canonical()]`)? I think that will make things a little clearer
2. Now that we've added actions through the CLI, let's add this one as well so the user has an immediate interface to it without having to configure it? Something like `zellij action soft-clear`?
3. Given `2`, I think we can remove the configuration option and add it as a flag to either the CLI or the bound key.
4. Maybe we can rename the action to `clear`?

fcoury:
@imsnif not a problem! I'll got right to it as soon as time permits and get back to you!

Thank you!

fcoury:
@imsnif just rebased my branch from current `main` branch and gonna start working on addressing your feedback.

fcoury:
Hey there @imsnif, finally decided to take another look at solving this. Let me tell you what I found so far:

>     1. How about if instead of `roll_to_top` we clear the viewport and then add canonical lines to it similar to how we initialize it in the constructor (`viewport: vec![Row::new(columns).canonical()]`)? I think that will make things a little clearer

The problem with this approach is that the current prompt is lost if we do that. Here's what the command looks like with the viewport solution you propose:

Resetting viewport:

https://github.com/zellij-org/zellij/assets/1371/f358c5a1-2a40-46d7-b3a4-73eca85e269b

Adding canonical lines:

https://github.com/zellij-org/zellij/assets/1371/2a3c87b4-5a10-478f-a4c7-bbcfad624b68

Do you think there's a better way to achieve this?

Thanks!

fcoury:
@imsnif afer further investigation I came up with this alternative version that I think looks a bit better:

```
    /// Clears the scrollback and positions the current active line on top of the pane.
    pub fn clear_scrollback(&mut self) {
        // HERE
        if self.alternate_screen_state.is_some() {
            log::warn!("Tried to clear scrollback of pane with alternate_screen_state");
            return;
        }

        let offset = 2; // SOFT_CLEAR_OFFSET.get().unwrap()

        log::info!("viewport: {:#?}", self.viewport);
        let new_viewport = self
            .viewport
            .clone()
            .into_iter()
            .skip(self.viewport.len().saturating_sub(offset + 1))
            .collect::<Vec<_>>();
        log::info!("new viewport: {:#?}", new_viewport);

        self.clear_viewport_before_rendering = true;
        self.lines_above = VecDeque::new();
        self.lines_below = vec![];
        self.viewport = vec![Row::new().canonical()];
        self.scroll_region = None;
        self.scrollback_buffer_lines = 0;
        self.viewport = new_viewport;
        self.cursor.y = offset;
        self.output_buffer.update_all_lines();

        self.mark_for_rerender();
    }
```

I will force push this branch with my new code if you think this is a good idea.

Thanks again!

fcoury:
@imsnif ready for review 😊

imsnif:
Hey @fcoury - thanks for this, but it has already been implemented: https://github.com/zellij-org/zellij/pull/2239#discussion_r1137594454

In the docs: https://zellij.dev/documentation/keybindings-possible-actions#clear

fcoury:
> Hey @fcoury - thanks for this, but it has already been implemented: https://github.com/zellij-org/zellij/pull/2239
> 
> 
> 
> In the docs: https://zellij.dev/documentation/keybindings-possible-actions#clear

Hey there, they don't seem to do the same think. This perfectly emulates what happens when you press Cmd+K on macOS. 

If this doesn't seem like something we want to support, can it be introduced as a plugin? If so, do you have any examples?

Thank you!

imsnif:
What's the difference?

fcoury:
Sorry replying from my phone but will try to explain in detail with some
screencasts tomorrow.

Issue #1722, opened 2022-09-10
Could not find the SHELL variable: NotPresent


altair-albert:
**Basic information**

`zellij --version`: 0.31.4
`stty size`: 52 192

`uname -av` or `ver`(Windows):  run in docker (host is Linux Master 5.15.60-1-MANJARO)

**Further information**

![image](https://user-images.githubusercontent.com/5851078/189477377-88e2a774-a207-49c6-9e38-9e37b0e35f3b.png)

```
Error occurred in server:

  × Thread 'pty' panicked.
  ├─▶ Originating Thread(s)
  │   	1. ipc_server: NewClient
  │   	2. pty_thread: NewTab
  │
  ├─▶ At zellij-server/src/pty.rs:206:54
  ╰─▶ Could not find the SHELL variable: NotPresent
  help: If you are seeing this message, it means that something went wrong.
        Please report this error to the github issue.
        (https://github.com/zellij-org/zellij/issues)

        Also, if you want to see the backtrace, you can set the `RUST_BACKTRACE` environment variable to `1`.
```

imsnif:
Inside the docker container, if you do `echo $SHELL`, what do you get?

altair-albert:
> Inside the docker container, if you do `echo $SHELL`, what do you get?

it's /bin/bash. 

when i  run zellij in tmux session inside docker container, it works.

imsnif:
Hmm, very odd. Could you help me understand how you are running this? What computer you are on, how are you connecting to the docker host, etc. I want to figure out where this env variable gets dropped.

altair-albert:
1. I used docker img base ubuntu 20.04
2. I started a container with docker start -i
3. execute zellij

karczex:
I encountered the same issue on Fedora 36 docker image run on podman. I was logged in the container as root and `SHELL` environ variable was set to /bin/bash.

imsnif:
@altair-albert , @karczex - could this be an issue of timing or environment? Instead of running Zellij with `zellij`, could you maybe do something like `echo $SHELL > /tmp/logfile && zellij`, and then see what you get in `/tmp/logfile`?

imsnif:
Btw, a workaround for this would be to specify a `default_shell` in the zellij configuration: https://zellij.dev/documentation/options.html

But I still want to get to the bottom of this

karczex:
[...]

PR #1769, opened 2022-10-02
fix(pty): use /bin/sh when SHELL env variable not found (#1722)

sux2mfgj:
In some environment, the SHELL environemnt variable is not exists. It causes a panic that is reported as #1722. Use generic shell (/bin/sh) to prevent the panic.

### DIFF
--- a/zellij-server/src/pty.rs
+++ b/zellij-server/src/pty.rs
@@ -433,9 +433,16 @@ impl Pty {
         }
     }
     pub fn get_default_terminal(&self, cwd: Option<PathBuf>) -> TerminalAction {
+        let shell = PathBuf::from(env::var("SHELL").unwrap_or_else(|_| {
+            log::warn!("Cannot read SHELL env, falling back to use /bin/sh");
+            "/bin/sh".to_string()
+        }));
+        if !shell.exists() {
+            panic!("Cannot find shell {}", shell.display());
+        }
         TerminalAction::RunCommand(RunCommand {
             args: vec![],
-            command: PathBuf::from(env::var("SHELL").expect("Could not find the SHELL variable")),
+            command: shell,
             cwd, // note: this might also be filled by the calling function, eg. spawn_terminal
             hold_on_close: false,
             hold_on_start: false,


zellij-server/src/pty.rs
```diff
@@ -201,9 +201,13 @@ impl Pty {
         }
     }
     pub fn get_default_terminal(&self) -> TerminalAction {
+        let shell = PathBuf::from(env::var("SHELL").unwrap_or_else(|_| "/bin/sh".to_string()));
```
har7an, line 204: I think we should add a log entry for this, something like "Cannot read 'SHELL' env, falling back to ...". This way we (and the user) can diagnose why zellij didn't pick up some shell configured by the user.

har7an (changes requested):
Thanks for picking this up!

Apart from a minor remark I think it's good to go. I'd like to wait and see where the discussion in #1722 leads to, though. Maybe we can give the user a hint as to why this happened in the `panic!` message.

zellij-server/src/pty.rs
sux2mfgj: Ok, I'll do that. Thanks.

imsnif:
Hey - I think this change is desirable. We do the same thing with the `EDITOR` now, defaulting to /usr/bin/vi if none is found. IMO this is a good user experience. Wdyt @har7an ?

sux2mfgj:
Hi @imsnif,

I agree with you. The default behavior is required. At lease panic is not better.

But, I think we have to create a new issue and discuss it. What do you think?

imsnif:
Why not just merge this? :)

imsnif:
Great! @sux2mfgj - would you like to add the log message so we can get this merged? You can even remake this PR if the merge conflict is too much. I think a `log::warn` would be good to add here inside the closure.

sux2mfgj:
I fixed the conflict and added the warn message. Please review it. 

Thanks a lot.

har7an:
Sorry for taking so long to respond, and thanks for your contribution!

raphCode:
This PR seems to break `SHELL=sh zellij` which used to work previously.

sux2mfgj:
Oops, I didn't care about the case...

Discuss it in the issue you opened.

Issue #1656, opened 2022-08-13
Feature: move tab to other position


jaeheonji:
Reference #1654 

1. Tab can be moved forward or backward
```
<tab 1> | <tab 2 *> | <tab 3>

# backward
<tab 1> | <tab3> | <tab 2 *>

# forward
<tab 2 *> | <tab 1> | <tab 3> 
```

2.  Move tabs to the front or back
```
<tab 1 *> | <tab 2> | <tab 3>

# back
<tab 2> | <tab 3> | <tab 1 *>
```

TornaxO7:
I'll try to implement this

jaeheonji:
@TornaxO7 Awesome! If you want, you can do it anytime. No need to rush!

TornaxO7:
@jaeheonji [`ScreenContext`](https://github.com/zellij-org/zellij/blob/main/zellij-utils/src/errors.rs#L228) is only used for error messages, right? Because I'm thinking of creating just a `MoveTab` entry for this instead of `MoveTabRight`, `MoveTabRight` and so on

TornaxO7:
@jaeheonji Is it fine if I'm also trying to refactor a little bit the Screen struct? I'd like to move the tabs attribute to an extra-struct and move some methods of screen as well.

TornaxO7:
@jaeheonji I also have another question: Why are you picking up the history of the first entry according to [this line](https://github.com/zellij-org/zellij/blob/main/zellij-server/src/screen.rs#L701)? Why not creating a new history for the new client?

jaeheonji:
> [ScreenContext](https://github.com/zellij-org/zellij/blob/main/zellij-utils/src/errors.rs#L228) is only used for error messages, right? Because I'm thinking of creating just a MoveTab entry for this instead of MoveTabRight, MoveTabRight and so on

Yes! The context is used for errors. In order to implement this feature, you will need to implement a new action. So you will have to add something new to Context and Action such as `MoveTabRight`, `MoveTabLeft`

> Is it fine if I'm also trying to refactor a little bit the Screen struct? I'd like to move the tabs attribute to an extra-struct and move some methods of screen as well.

Of course, If you need it you can. However, if there are a lot of refactoring contents, it would be good for a review to make a separate PR.

> I also have another question: Why are you picking up the history of the first entry according to [this line](https://github.com/zellij-org/zellij/blob/main/zellij-server/src/screen.rs#L701)? Why not creating a new history for the new client?

[...]

PR #1774, opened 2022-10-03
Feature move tab to other position 1656

TornaxO7:
This should fix https://github.com/zellij-org/zellij/issues/1656


jaeheonji (commented):
Sorry for the late reply!

I looked up why the error occurred in the current code.

First, the conclusion, as I mentiond in [the issue](https://github.com/zellij-org/zellij/issues/1656#issuecomment-1236362695). You need to load the layout using the rebuilt plugin.

Of course, you didn't modify the plugin's code, but all plugins refer to `zellij_utils::input::actions` internally. Therefore, since a new action is added to `actions.rs`, when the local zellij plugin is loaded, the plugin conflicts with the newly added action `MoveTab`.

imsnif:
Also, please see my comment from https://github.com/zellij-org/zellij/pull/1784#issuecomment-1300457983

Issue #1753, opened 2022-09-27
Tracking issue to improve error handling


har7an:
# The goal

We are looking to improve the general state of error handling in zellij. As the zellij code-base changed, a lot of the previous places where `unwrap` made sense can now potentially cause errors which we'd like to handle. We want to eliminate calls to `unwrap` and use `Result` types instead, so we can handle errors where appropriate.

You can read more about the background and the tools we have to achieve this [in the docs][1].

# Help wanted

The zellij codebase has become very large, and few functions use `Result` types to propagate errors up to calling functions. We are looking for contributors to help us out with successively converting the code to change this.

If you want to make a contribution, this may be the ideal start for you. While working on error handling you:

- get familiar with different parts of the code, without having to integrate new features while you're at it,
- get a feeling for how individual portions of the code interact and
- help users and developers alike by
    - making more sense of application crashes, or
    - preventing crashes from occurring in the first place

# Contributing

**You want to help us out with error handling?** That's great!

You can get going right away. Keep in mind while working on this that it's very important to understand the piece of code you are changing. A good way to help you understand what is happening would be to try to run the application, reach that place in the code and see what happens. You can also try to manually generate errors in different places (for example with the `anyhow::bail!` macro) to see if you're on the right track.

In order to help us coordinate efforts, we would ask you to indicate that you're working on some aspect of error handling by either:

- leaving us a comment here in this issue, or
- opening a Pull Request ahead of time and mentioning this issue from there.

This allows us to keep the list below up-to-date, and prevents you from working on something that another person has picked up before you.

If you want to help us, but aren't sure where to start or what to do, you can either:

- leave a comment in this issue, or
- reach out to us via [Discord][discord] or [Matrix][matrix].

We'll help you find a suitable topic to work on and guide you as questions arise.

Thanks in advance!

# List of efforts regarding error handling

- #1748

[discord]: https://discord.gg/MHV3n76PDq
[matrix]: https://matrix.to/#/#zellij_general:matrix.org

naosense:
I want to work on this, any hints or suggestions where I can start.

har7an:
Hi @naosense,

glad to hear it! Here's a bunch of suggestions (along with the number of calls to `unwrap()`):

- `zellij-server/src/route.rs` (77): This one should be mostly self-contained I think (Doesn't require lots of changes in other files across the codebase)
- `zellij-server/src/pty.rs` (33): This may require changes in other files
[...]

PR #1775, opened 2022-10-04
zellij-server: improve thread_bus error handling

arriven:
WIP: #1753 

As suggested in my previous [PR](https://github.com/zellij-org/zellij/pull/1770#issuecomment-1266518037), moving thread_bus error changes to a separate one

### DIFF
--- a/zellij-server/src/thread_bus.rs
+++ b/zellij-server/src/thread_bus.rs
@@ -4,6 +4,7 @@ use crate::{
     os_input_output::ServerOsApi, pty::PtyInstruction, pty_writer::PtyWriteInstruction,
     screen::ScreenInstruction, wasm_vm::PluginInstruction, ServerInstruction,
 };
+use zellij_utils::errors::prelude::*;
 use zellij_utils::{channels, channels::SenderWithContext, errors::ErrorContext};
 
 /// A container for senders to the different threads in zellij on the server side
@@ -20,10 +21,7 @@ pub(crate) struct ThreadSenders {
 }
 
 impl ThreadSenders {
-    pub fn send_to_screen(
-        &self,
-        instruction: ScreenInstruction,
-    ) -> Result<(), channels::SendError<(ScreenInstruction, ErrorContext)>> {
+    pub fn send_to_screen(&self, instruction: ScreenInstruction) -> Result<()> {
         if self.should_silently_fail {
             let _ = self
                 .to_screen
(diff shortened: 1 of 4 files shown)


zellij-server/src/thread_bus.rs
```diff
@@ -32,14 +30,15 @@ impl ThreadSenders {
                 .unwrap_or_else(|| Ok(()));
             Ok(())
         } else {
-            self.to_screen.as_ref().unwrap().send(instruction)
+            self.to_screen
+                .as_ref()
+                .unwrap()
+                .send(instruction)
+                .context("failed to send message to screen")
```
har7an, line 37: Do you think we can get rid of the `unwrap()` attached to `as_ref`, too? Now that we already return a `Result` we may as well report that error case.

Also what do you think about adding the instruction we sent to the error message, similar to how you already do in `send_to_plugin`? It may not be helpful for the user, but maybe a developer can make sense of all the content.

har7an (changes requested):
Thanks for continuing your work on this!

I left a remark below, do you mind applying it to your other changes, too? That'd be awesome (and very helpful for debugging).

zellij-server/src/thread_bus.rs
arriven: upd: crossbeam::channel::SendError doesn't format the content so yes, that pattern would be helpful for devs. Also I've noticed you've added it as a function on the type so I've moved the usage of `.to_anyhow()` into those `thread_bus` functions, hope you don't mind

zellij-server/src/thread_bus.rs
```diff
@@ -32,14 +30,16 @@ impl ThreadSenders {
                 .unwrap_or_else(|| Ok(()));
             Ok(())
         } else {
-            self.to_screen.as_ref().unwrap().send(instruction)
+            self.to_screen
+                .as_ref()
+                .context("failed to get screen sender")?
+                .send(instruction)
+                .to_anyhow()
```
har7an, line 37: Good catch, initially I implemented it as a crutch for the changes in `tab`, as you noticed correctly. Thank you for removing it in `tab`!

arriven: I mean, it refused to compile otherwise since `anyhow::Result<T> != Result<T, crossbeam::channel::SendError>` but thanks! :sweat_smile:

Issue #1311, opened 2022-04-11
[Feature] Automatic tab renaming


Lillecarl:
**Basic information**

`zellij --version`: zellij 0.25.0 (Realized I'm 2 versions behind, time to update my flakes 😄 )
`tput lines`: 43
`tput cols`: 176
`uname -av` or `ver`(Windows): Darwin hostname.local 21.4.0 Darwin Kernel Version 21.4.0: Fri Mar 18 00:45:05 PDT 2022; root:xnu-8020.101.4~15/RELEASE_X86_64 x86_64

List of programs you interact with as, `PROGRAM --version`: output cropped meaningful, for example:
`iterm2 version`: Build 3.4.15

**Further information**
If I open a bunch of tabs and leave them with their default naming then close the first one, the new tab 1 will be named Tab #2 which doesn't matter so much for the first one, but if you have 5-6 tabs open it adds a bit of overhead when you're switching around.

I'd like to suggest something that auto resolves this by renaming tabs upon closing. At least if their name is following the default pattern. (I never name my tabs).

Another solution could be to decouple the indexing and the name, if the name should be before or after the configured name would probably be a heated discussion though (configureable maybe?)

I've been reading through the source-code but I can't seem to find where this would be. I'm not rust-native either so for now I'll leave it as a feature request.

As always, great job guys! Been daily-driving Zellij since 0.19 or 0.20 :)

cole-h:
I've taken a stab at this in https://github.com/zellij-org/zellij/pull/1784. If you're so inclined, please give it a shot and let me know what you think.

PR #1784, opened 2022-10-08
feat(screen): rename and reorder tabs when closed

cole-h:
My goal is to close https://github.com/zellij-org/zellij/issues/1311, but since this is my first time experimenting with the zellij code base, I imagine there are some things I'm doing improperly / misunderstanding.

One of my immediate questions is... why is there a `position` and `index` field on the `Tab` struct? They seem to fulfill the same purpose to me, but I also don't use terminal multiplexers like this in a different way than most (I use tabs more than panes), so maybe in my usecase they are the same, but when using other features they are distinct fields.

My solution to re-order the `BTreeMap` (so that the index into `self.tabs` is the same as the tab's `index` field) may not be the best. In fact, everything here is likely sub-par, and potentially has unintended side effects.

Please tell me everything that's wrong with this code, and I'll do my best to fix it (unless someone sees a better solution and wants to supersede this with another PR -- I promise I won't be mad).

---

It might be possible to not do as many hacky things (and potentially drop my "fix tests" commit) if I steal some of the implementation of `self.move_clients_from_closed_tab` so that it doesn't affect the active indices / tab history.

It might also be a "better" solution to have the tabs have an unchangeable index in the title like tmux does, so we don't have to worry about handling manually-renamed tabs, but that might be more invasive.

Happy to discuss the reasons behind any of my changes, just ask.

(I did my best to adhere to conventional commits, feel free to tell me how I can improve there as well :)


imsnif:
Hey @cole-h - thanks for starting this up and apologies it took me so long to get to it!

I think the main issue here stems from something you said: the `position` and `index` essentially do the same thing. In my mind, and that of those who created these features in the first place, `Tab #<index>` is more of a name than a position. This made sense before we had actions such as "go to tab <x>". Once those were added, we had to add the `position` attribute and then everything started getting messy.

I think the correct solution to this problem would be to merge these two fields. Initially I thought people would be surprised by this behaviour, since their tabs would essentially be renamed when they close a tab in the middle, but I think I'm the only one who really thinks that. I'm happy with introducing this behaviour.

I know this might be a bigger change than you signed up for, but I'm hoping you'd be willing to do it?

cole-h:
Honestly, I was hoping you'd say that!

That said, my time is a bit sparse right now as I deal with work and finishing up school. So yes, I'm definitely interested in making that (only slightly more invasive...) change, but it, uh, might take a little bit for me to get around to it. :bow: 

If anybody comes along and is tired of waiting for this, feel free to reach out to me for push access to my branch (or just open another PR and ping me so I can take a peek myself!). I'm in the Zellij matrix room, or you can glean my email from my commit.

cole-h:
Just wanted to drop a small update: I expect I'll have some time to explore this in about 2 weeks (maybe a little less, maybe a little more). Offer still stands if anybody wants to take over (via push access to my fork, or a fresh PR), though.

cole-h:
Whew, finally got around to this! I guess "a little more" is what it ended up being :laughing: 

I've pushed up 2 additional commits that address the `tab`-`position` situation (one that actually does the removal / logic fixing, and another that does the same for the tests). All the tests pass (including the `e2e` ones), so I think this is ready for another review, @imsnif! Thanks for your patience!

imsnif:
Hey @cole-h - great work on this! I'm happy you stuck with this even through lack of time and holidays :)

Things looks good to me on the surface - would you like to get the tests to pass so that I can do a final go-over and merge if all is well?

cole-h:
Absolutely (a bit embarrassing there were so many failures, I could have sworn I was more thorough than that...)! I'll be able to take a look later today / early tomorrow.

cole-h:
OK, should be ready for your final go-over -- just missed a rename :upside_down_face:

imsnif:
Hey @cole-h - looks great, but I found a small bug: when there are multiple users connected and focused on different tabs, if one closes a tab the other users seem to focus on the new tab even if they were in an entirely different one.

To reproduce:
1. start zellij in terminal window number 1
2. attach to the same session in terminal window number 2
3. in terminal window number 2, open 2 new tabs
4. in terminal window number 2, close the current tab with `Ctrl t` + `x`
5. Observe that terminal window number 2 is focused on tab number 2 (expected) but that terminal window number 1 is also focused on tab number 2 (not expected - it should still be focused on tab number 1).

Would you like to take a look?

cole-h:
Sorry, I forgot to update:

I did take a look at that bug and it was relatively easy to solve. However, I noticed that the opposite direction was also problematic: closing the first tab while another client has the last tab focused breaks due to the reordering. I'm still playing with ideas to fix both, but it's a little bit of a doozy...

cole-h:
It's been about 2 months and I really haven't done anything on this since...

Recently, I've found myself just not in the mood to hack on things (may or may not be related to me re-discovering Factorio...). More importantly, I haven't found the brain power to work through the problems the current implementation is causing (I usually just wander in circles until I give up for the day). I have some uncommitted changes that I'll push as a  "WIP" commit (which also includes some comments documenting the current issues, as well as my very robust debugging of `log::error!()` lines), but I don't think I'll have the mental capacity to get back to this any time soon.

With all that said, I'm going to mark this PR as a draft and "WIP". If anybody feels like they want to pick this up, I'm happy to give push access to my fork and let you take up the mantle in the meantime. (You might even want to restart the work with the learnings from this PR in mind...)

cole-h:
(It might also be worth extracting my "deduplicate `index` and `position`" commit(s): bbee675ef838798f5d627ae32b6feb3654ab2bd2, 37426577e41a0843e42e207734af4197559d4f34)

PR #1804, opened 2022-10-17
feat(cli): zellij run improvements

imsnif:
This PR does several things:
1. The `command` part of the various CLI pane-opening commands has been moved to the last part after double-dashes as per convention. eg `zellij run --floating -- tail -f /tmp/my_nice_output.log` - this is in order to prevent ambiguity and mistakes when providing commands with arguments.
2. We can now name panes through the CLI as well when opening them using `--name`, eg. `zellij run --name "my tests" -- cargo test`
3. I added shell aliases for 3 major shells (bash, zsh and fish) that do the following:
`zp` (short for "zellij pane") ===> shortcut for `zellij run -- <arguments>`, eg. `zp cargo test` (notice the lack of quotations!)
`zpf` (short for "zellij pane floating") - same as above with the `--floating` flag to open a floating pane.
`zo` (short for "zellij open") ===> shortcut for `zellij edit <arguments>`, eg. `zo /tmp/my-file.txt` will open `my-file` in your default editor (eg. vim)
`zof` (short for "zellij open floating") - same as above with the `--floating` flag to open a floating pane.

A "hidden" feature of the shell completions from point 3 is that they wrap the provided command with the relevant shell's `-c` flag (eg. bash -c "cargo test") - with this, shell aliases and built-ins such as `ll` in fish will work out of the box.

@raphCode - if you have a chance (even after I merge this) to check out the bash and zsh functions I created to make sure I didn't do any faux pas, that would be great. :)

EDIT: the reason for the alias name change (eg. `zp` instead of `zr` and `zo` instead of `ze`) is that I felt these aliases will be easier and quicker to type, because the command is spread over both hands on the keyboard. Since these are commands that advanced users will probably be using *a lot*, I thought it would be nice to start with a sane default.


raphCode:
> @raphCode - if you have a chance (even after I merge this) to check out the bash and zsh functions I created to make sure I didn't do any faux pas, that would be great. :)

For me (bash) it needs the `--interactive` / `-i` flag to source my `.bashrc`
Test yourself:
```bash
bash -c ll
bash: line 1: ll: command not found
bash -ic ll
# output when .bashrc is executed, after that the ll alias is executed
```
I have confirmed the same behavior with `zsh`, it also needs the `-i` flag.
I haven't found anything meaningful in fish's manpage what its `-i` flag does. Even without it, fish seems to have its own definition of `ll` so you might want to test that yourself with a custom alias.

raphCode:
@imsnif Just to make sure you didn't miss my comment above, I think you might want to change this before the release?

PR #1805, opened 2022-10-17
docs: Describe how to handle Options as errors

har7an:
Extend the error handling documentation.

@imsnif: You said you wanted to mention the bit about returning an `anyhow::Result` mentioned in `CONTRIBUTING.md`. I'm honestly not sure where it fits in there, and I don't want to bloat it needlessly... So I extended the error handling docs and added a `TL;DR`:

image

Is that alright for you, too?

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -113,6 +113,21 @@ something interesting to work on and guide through.
 [discord-invite-link]: https://discord.gg/feHDHahHCz 
 [good-first-issue]: https://github.com/zellij-org/zellij/labels/good%20first%20issue
 
+
+## Tips for Code Contributions
+
+### Prefer returning `Result`s instead of `unwrap`ing
+
+- Add `use zellij_utils::errors::prelude::*;` to the file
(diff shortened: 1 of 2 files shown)


har7an:
I feel like it's a little out of place in `CONTRIBUTING.md` to be honest. This is sort of a very concrete tip for how to code. Like, where do you draw the line for things we care to mention there?

Since it's likely the first file (after `README.md`) that potential contributors will read, I wouldn't want to strike them with a wall of text as it grows. I agree we should somehow reference it, but I'm not sure how. Maybe we add a section `Tips for code contributions`, or `Zellij coding best practices` and then stick a few very brief examples in there along with links to "further reading" (Like the new section in `ERROR_HANDLING.md` in this case)?

imsnif:
> Since it's likely the first file (after `README.md`) that potential contributors will read, I wouldn't want to strike them with a wall of text as it grows. I agree we should somehow reference it, but I'm not sure how. Maybe we add a section `Tips for code contributions`, or `Zellij coding best practices` and then stick a few very brief examples in there (the `TL;DR` in this case) along with links to "further reading" (Like the new section in `ERROR_HANDLING.md` in this case)?

Sounds good!

PR #1808, opened 2022-10-18
improve error handling in route

naosense:
Replacing `unwrap` in `route.rs` with error propagation.

Related to #1753

### DIFF
--- a/zellij-server/src/lib.rs
+++ b/zellij-server/src/lib.rs
@@ -275,7 +275,8 @@ pub fn start_server(mut os_input: Box<dyn ServerOsApi>, socket_path: PathBuf) {
                                             to_server,
                                             receiver,
                                             client_id,
-                                        ).fatal()
+                                        )
+                                        .fatal()
                                     })
                                     .unwrap(),
                             );
--- a/zellij-server/src/lib.rs
+++ b/zellij-server/src/lib.rs
@@ -35,7 +35,6 @@ use crate::{
     wasm_vm::{wasm_thread_main, PluginInstruction},
 };
 use route::route_thread_main;
-use zellij_utils::errors::prelude::Context;
 use zellij_utils::{
     channels::{self, ChannelWithContext, SenderWithContext},
     cli::CliArgs,
@@ -276,10 +275,9 @@ pub fn start_server(mut os_input: Box<dyn ServerOsApi>, socket_path: PathBuf) {
                                             to_server,
                                             receiver,
                                             client_id,
-                                        )
+                                        ).fatal()
                                     })
-                                    .context("Failed to spawn server_router thread")
-                                    .fatal(),
+                                    .unwrap(),
                             );
                         },
                         Err(err) => {
--- a/zellij-server/src/lib.rs
+++ b/zellij-server/src/lib.rs
@@ -35,6 +35,7 @@ use crate::{
     wasm_vm::{wasm_thread_main, PluginInstruction},
 };
 use route::route_thread_main;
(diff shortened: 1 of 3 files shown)


har7an:
Hi again and thanks for working on this!

Are you interested in expanding this PR a bit and helping us to remove all calls to `unwrap` from `route.rs`? In particular I think it would be great if you change `route_thread_main` to return a `Result<()>` and then call `.fatal()` [here][1]. Then we can remove `route` from the list of files to improve error handling in for good. What do you say?


[1]: https://github.com/zellij-org/zellij/blob/12d35bded55756c951e86212a3c72305a26043d6/zellij-server/src/lib.rs#L280

naosense:
Thanks for your detailed explanation, it's a pleasure to do it

naosense:
hi @har7an , could you please review my code again?

zellij-server/src/route.rs
```diff
@@ -26,7 +27,7 @@ pub(crate) fn route_action(
     _os_input: &dyn ServerOsApi,
     to_server: &SenderWithContext<ServerInstruction>,
     client_id: ClientId,
-) -> bool {
+) -> Result<bool> {
     let mut should_break = false;
```
har7an, line 32: Sorry for not raising this earlier... I think we should add an error context to all the `?` inside this function. I missed that this actually *isn't* the threads main function, but only called from `route_thread_main`. I'm fine with sticking to `?` in `route_thread_main` because we will call `.fatal()` on it in the next step and it's totally clear where the error comes from and what we tried to do.

I think an error context like this:

```suggestion
let err_context = || format!("failed to route action for client {client_id}: {action:?}");
```

should do, and then you replace all `?` in this function with `.with_context(err_context)?`. Is that alright for you?

zellij-server/src/lib.rs
```diff
@@ -277,7 +278,8 @@ pub fn start_server(mut os_input: Box<dyn ServerOsApi>, socket_path: PathBuf) {
                                             client_id,
                                         )
```
har7an, line 279: Sorry, I mixed up the braces here. The call to `.fatal()` goes here, the `.unwrap()` below can stay.

har7an (changes requested):
Thanks for continuing to work on this!

It seems last time I looked over your code I was a bit sloppy, and I want to apologize for that. If you apply the changes below this is good to go.

zellij-server/src/route.rs
naosense: It seems that putting the `action` in the `err_context` will report an error: cannot move out of `action.0` because it is borrowed [E0505] move out of `action.0` occurs

har7an: That's true, my bad! Then maybe this:

```rust
let err_context = || format!("failed to route action for client {client_id}");
```

naosense: Thanks for your patience, I'll do that.

zellij-server/src/route.rs
```diff
@@ -607,16 +639,17 @@ pub(crate) fn route_thread_main(
     to_server: SenderWithContext<ServerInstruction>,
     mut receiver: IpcReceiverWithContext<ClientToServerMsg>,
     client_id: ClientId,
-) {
+) -> Result<()> {
     let mut retry_queue = vec![];
+    let err_context = || format!("failed to handle instruction for client {client_id}");
```
har7an, line 644: Great, thank you!

har7an (approved):
Well done, thanks for your effort!

PR #1827, opened 2022-10-20
errors: Don't panic in `wasm_vm`

har7an:
Removes calls to `unwrap` and `panic` in the server API of `wasm_vm`, replacing explicit calls to `panic!` with `.fatal()` instead. Leaves the plugin API unchanged.

The most notable change is that now, when we load incompatible plugins (older versions, whatever) we get pretty error messages. In the terminal it looks like this:

```
$ ./target/debug/zellij

Error occurred in server:

  × Thread 'screen' panicked.
  ├─▶ Originating Thread(s)
  │   	1. ipc_server: NewClient
  │   	2. pty_thread: NewTab
  │   	3. screen_thread: NewTab
  │   
  ├─▶ At zellij-server/src/panes/plugin_pane.rs:164:42
  ╰─▶ called `Result::unwrap()` on an `Err` value: RecvError
        
        Also, if you want to see the backtrace, you can set the `RUST_BACKTRACE` environment variable to `1`
```

which is just an effect caused by the initial error (and the next thing on my TODO). The real error is now sitting in the logs:

```
INFO   |zellij_client            | 2022-10-20 10:29:21.592 [main      ] [zellij-client/src/lib.rs:133]: Starting Zellij client! 
INFO   |zellij_server            | 2022-10-20 10:29:21.597 [main      ] [zellij-server/src/lib.rs:213]: Starting Zellij server! 
INFO   |zellij_server::wasm_vm   | 2022-10-20 10:29:21.658 [wasm      ] [zellij-server/src/wasm_vm.rs:104]: Wasm main thread starts 
DEBUG  |tab-bar                  | 2022-10-20 10:29:21.825 [id: 0     ] thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: Error("expected value", line: 1, column: 372)', default-plugins/tab-bar/src/main.rs:30:1 
DEBUG  |tab-bar                  | 2022-10-20 10:29:21.825 [id: 0     ] note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace 
ERROR  |zellij_utils::errors::not| 2022-10-20 10:29:21.840 [wasm      ] [zellij-utils/src/errors.rs:435]: Panic occured:
             thread: wasm
             location: At zellij-server/src/lib.rs:722:92

Caused by:
    0: If you're seeing this error the most likely cause is that your plugin versions
       don't match your current zellij version.
       
       If you're a user:
           to them.
       
       If you're a developer:
           Please run zellij with the updated plugins. The easiest way to achieve this
       
    1: failed to update plugin
    2: RuntimeError: unreachable
           at <unnamed> (<module>[190]:0x1d557)
           at <unnamed> (<module>[91]:0xe98a)
           at <unnamed> (<module>[96]:0xf010)
           at <unnamed> (<module>[451]:0x4cb23)
    3: unreachable 
ERROR  |zellij_utils::errors::not| 2022-10-20 10:29:21.840 [screen    ] [zellij-utils/src/errors.rs:435]: Panic occured:
             thread: screen
             location: At zellij-server/src/panes/plugin_pane.rs:164:42
             message: called `Result::unwrap()` on an `Err` value: RecvError 
ERROR  |zellij_utils::errors::not| 2022-10-20 10:29:21.844 [async-std/runti] [zellij-utils/src/errors.rs:435]: Panic occured:
             thread: async-std/runtime
             location: At zellij-server/src/terminal_bytes.rs:124:14
             message: called `Result::unwrap()` on an `Err` value: failed to send message to screen

Caused by:
    0: Originating Thread(s)
       
    1: failed to send message to channel
```

The runtime error about having reached unreachable code is still as useless as before, but that's out of our hands.

@imsnif Are you fine with the wording I chose? In particular this bit:

>       If you're seeing this error the most likely cause is that your plugin versions
>       don't match your current zellij version.
>       
>       If you're a user:
>           Please contact the distributor of your zellij version and report this error
>           to them.
>       
>       If you're a developer:
>           Please run zellij with the updated plugins. The easiest way to achieve this
>           is to build zellij with `cargo make install`. Also refer to the docs:
>           https://github.com/zellij-org/zellij/blob/main/CONTRIBUTING.md#building

### DIFF
--- a/zellij-server/src/lib.rs
+++ b/zellij-server/src/lib.rs
@@ -718,7 +718,9 @@ fn init_session(
             );
             let store = Store::default();
 
-            move || wasm_thread_main(plugin_bus, store, data_dir, plugins.unwrap_or_default())
+            move || {
+                wasm_thread_main(plugin_bus, store, data_dir, plugins.unwrap_or_default()).fatal()
+            }
         })
         .unwrap();
 
--- a/zellij-server/src/panes/plugin_pane.rs
+++ b/zellij-server/src/panes/plugin_pane.rs
@@ -161,7 +161,9 @@ impl Pane for PluginPane {
                 .unwrap();
 
             self.should_render = false;
-            let contents = buf_rx.recv().unwrap();
+            let contents = buf_rx
+                .recv()
+                .expect("Failed to receive reply from plugin. Please check the logs");
             for (index, line) in contents.lines().enumerate() {
                 let actual_len = ansi_len(line);
                 let line_to_print = if actual_len > self.get_content_columns() {
--- a/zellij-server/src/wasm_vm.rs
+++ b/zellij-server/src/wasm_vm.rs
@@ -54,6 +54,9 @@ If you're a developer:
     Please run zellij with the updated plugins. The easiest way to achieve this
     is to build zellij with `cargo make install`. Also refer to the docs:
     https://github.com/zellij-org/zellij/blob/main/CONTRIBUTING.md#building
+
+A possible fix for this error is to remove all contents of the 'PLUGIN DIR'
+folder from the output of the `zellij setup --check` command.
 ";
 
 #[derive(Clone, Debug)]
--- a/zellij-server/src/wasm_vm.rs
+++ b/zellij-server/src/wasm_vm.rs
@@ -30,12 +30,12 @@ use crate::{
 use zellij_utils::{
     consts::{VERSION, ZELLIJ_CACHE_DIR, ZELLIJ_PROJ_DIR, ZELLIJ_TMP_DIR},
     data::{Event, EventType, PluginIds},
-    errors::{ContextType, PluginContext},
-};
-use zellij_utils::{
-    input::command::TerminalAction,
-    input::layout::RunPlugin,
-    input::plugins::{PluginConfig, PluginType, PluginsConfig},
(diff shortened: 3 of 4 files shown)


imsnif:
This is very cool!!!
I didn't look at the code, but I'm very happy to see this.

About the wording: I'd add something about clearing the plugin cache as a possible solution (in the `PLUGIN_DIR` from `zellij setup --check`). Otherwise it looks great.

Also: is there any chance we can show this in STDOUT as well as the logs? A lot of users wouldn't know to look there.

har7an:
> Also: is there any chance we can show this in STDOUT as well as the logs? A lot of users wouldn't know to look there.

Not quite yet. As you see on stdout, zellij dies trying to receive a message from the `wasm_vm`, which of course fails (because by then the VM is dead). My next task after this one will be to rewrite `plugin_pane` and find a way to handle this. I'll do that outside this PR, because it requires me to touch a trait function from `Pane`, which requires changes in all the pane handling code, you can imagine...

Ideally I'll manage to send a message to the `plugin_pane` and then handle it there. I'll play around with it. As a quickfix (in case I don't make it until the release, and I don't want to block that), we can change the existing call to `unwrap` in `plugin_pane` with something like `.expect("Failed to receive message from plugins. Please check the logs.")`.

PR #1838, opened 2022-10-23
plugins: Improve error handling on plugin version mismatch

har7an:
This PR picks up a long-standing issue with respect to the plugin system. Since we do not make guarantees about plugin version compatibility, we always require the plugin versions to match the zellij version.

## Current situation

This is usually not an issue for people that only install zellij from official releases, because in this case zellij takes care of updating the plugins proper itself. Things start to get hairy when developing zellij, because once a user has built the plugins themselves from `main`, they will not be updated on the next release automatically.

Previously we had no mechanism to check for this: The plugins were just loaded and executed. Errors always cropped up in the plugins `update` functions, where an `Event` is deserialized from stdin on the plugin side (i.e. in the WASM sandbox/VM). At this point, the serde error is `unwrap`ped, and the plugin crashes along with the application.

Symptomatic for this sort of failure are multiple things:

1. The actual `unwrap` inside the plugin is masked by `WASM`, which is evident from the logs where messages like this would occur (Note the log level "DEBUG"):
```
DEBUG  |tab-bar                  | 2022-08-27 19:56:48.928 [id: 0     ] thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: Error("invalid type: sequence, expected a string", line: 1, column: 53)', default-plugins/tab-bar/src/main.rs:29:1 
```
2. The error shown on stdout contains lots of garbage:
```
called `Result::unwrap()` on an `Err` value: RuntimeError { source: Trap(UnreachableCodeReached), wasm_trace: [FrameInfo { module_name: "<module>", func_index: 196, function_name: None,
      func_start: SourceLoc(118399), instr: SourceLoc(121927) }, FrameInfo { module_name: "<module>", func_index: 97, function_name: None, func_start: SourceLoc(59952), instr: SourceLoc(60199) },
      FrameInfo { module_name: "<module>", func_index: 102, function_name: None, func_start: SourceLoc(61748), instr: SourceLoc(61869) }, FrameInfo { module_name: "<module>", func_index: 509,
      function_name: None, func_start: SourceLoc(409244), instr: SourceLoc(427678) }], native_trace:    0: <unknown>
         1: <unknown>
         2: <unknown>
         3: <unknown>
         4: <unknown>
         5: <unknown>
         6: <unknown>
       }
```
3. We have no way of handling this, because in the code WASM gives us some obscure type of `RuntimeError` without any further information.

2 above is particularly annoying because it's the only hint the user has to go on and it starts with `Trap(UnreachableCodeReached)`. Not helpful.

## Improving the situation

This PR approaches the problem in a number of ways.

### Don't mask the underlying error cause

#1827 laid the groundwork for this PR, in that it first dealt with the issue of version mismatches and reporting them. This was achieved by appending a descriptive error message before unwrapping on a failed call to `update` a plugin. Unfortunately, in the stdout that was shown to the user, this error was frequently covered by a related crash happening in the `screen` thread when trying to `render` the offending plugins pane. What happens there is that we dispatch a request to the plugin to render its output, and expect to receive a message [here][1]. However, the plugin may already have crashed at this point, closing the `mpsc::Channel` we're trying to receive from.

What we now do instead is:

- Attach a context to the `RecvError`,
- Print the error to the zellij logs, and
- show a default text "No output from plugin received. See logs" in the plugin pane

This keeps the `screen` thread from crashing and allows us to display the real error cause on stdout.

### Check zellij/plugin version compatibility

Furthermore, in order to prevent such situations from happening in the first place, we're now embedding a plugin version into the plugins at build time. As of currently, this is the zellij version (taken from `zellij_utils::consts::VERSION`), taken from the version in the `Cargo.toml`.

Since we're not making guarantees about backwards compatibility yet, the requirement is that Plugin version == Zellij version. We check the versions when starting the plugin (after loading it, but before updating it for the first time). If this check fails, we output an error such as this:

```
$ ./target/debug/zellij

Error occurred in server:

  × Thread 'wasm' panicked.
  ├─▶ Originating Thread(s)
  │   	1. ipc_server: NewClient
  │   	2. pty_thread: NewTab
  │   	3. screen_thread: NewTab
  │   	4. plugin_thread: Load
  │   
  ├─▶ At zellij-server/src/wasm_vm.rs:433:19
  ╰─▶ If you're seeing this error your plugin versions don't match your current
      zellij version. Detected versions:
      
      - Plugin version: Unavailable
      - Zellij version: 0.32.0
      - Offending plugin: tab-bar
      
      If you're a user:
          to them.
      
      If you're a developer:
          Please run zellij with the updated plugins. The easiest way to achieve this
      
      A possible fix for this error is to remove all contents of the 'PLUGIN DIR'
      folder from the output of the `zellij setup --check` command.
      
```

In this case, I tried to run `zellij` compiled from main against plugins compiled on v0.31.4. Since the old plugins don't have any version information, we show the plugin version as "Unavailable".

### Catching mismatches during development

The previous method of catching errors works wonderfully for users that just update their application from release to release. However, developers compiling zellij off of main will not update the zellij application version after every compilation. This is particularly tricky, as one may spend multiple hours developing on the latest main before performing commits, so a comparison such as "zellij git-rev" == "plugin git-rev" won't do the trick either. However, every modification increases the chance of a change in the `Event` serialization structure, making the plugins incompatible.

As mentioned above, the typical symptom tor this error is that deserialization of the `Event` on the plugin fails. At this point it would have been amazing if wasmer could propagate errors from an exported WASM function called inside the plugin onto the host. The WASM docs do not talk about this at all, and it isn't considered an error when one does change the called function inside the WASM module (plugin) to return a `Result`. As far as I could gather, however, this `Result` is silently dropped.

It is stated somewhere on the wasmer GitHub that a `panic` inside a WASM module will be converted to a `RuntimeError`, and there is even a `RuntimeError::downcast`, but that also doesn't give any hints about the underlying error cause.

Instead, we're overriding the [panic hook][2] to send a message (as string) to the host. Overriding the panic hook is okay in this case, because the default wasmer panic hook in our case doesn't contribute any meaningful information (likely due to `wasm-opt` used during building). Also, sending the panic message as string is the only sensible way to go: Since a panic is likely to occur upon deserialization, we obviously cannot expect to match the hosts data types any more. Hence, serializing to some other type and deserializing on the host successfully likely isn't possible at all.

Here's the resulting error for this error case:

```
$ ./target/debug/zellij -l devel

Error occurred in server:

  × Thread 'wasm' panicked.
  ├─▶ Originating Thread(s)
  │   	1. ipc_server: NewClient
  │   	2. pty_thread: NewTab
  │   	3. screen_thread: NewTab
  │   	4. plugin_thread: Update
  │   
  ├─▶ At zellij-server/src/wasm_vm.rs:629:5
  ╰─▶ 
      An error occured while receiving an Event from zellij. This is most likely
      caused by your plugins not matching your current zellij version.
      
      The most likely explanation for this is that you're running either a
      self-compiled zellij or plugin version. Please make sure that, while developing,
      you also rebuild the plugins in order to pick up changes to the plugin code.
      
      Please refer to the documentation for further information:
      
      
      Caused by:
          0: failed to deserialize object from stdin
          1: unknown variant `Foo`, expected one of `PageDown`, `PageUp`, `Left`, `Down`, `Up`, `Right`, `Home`, `End`, `Backspace`, `Delete`, `Insert`, `F`, `Char`, `Alt`, `Ctrl`, `BackTab`,
      `Null`, `Esc` at line 1 column 3476
```

The error in this case was provoked by introducing a new key "Foo", that the plugins didn't know about. As an added bonus compared to the previous situation, we now also get to see why exactly the deserialization failed in the output of stdout (no more hiding as "DEBUG" message in the logs).

[1]: https://github.com/zellij-org/zellij/blob/788bcd6151431222325800e0a46c58384fe0bd22/zellij-server/src/panes/plugin_pane.rs#L165
[2]: https://doc.rust-lang.org/stable/std/panic/fn.set_hook.html

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -35,6 +35,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 * feat: allow defining tab cwd in layouts (https://github.com/zellij-org/zellij/pull/1828)
 * debugging: Remove calls to `unwrap` from plugin WASM VM (https://github.com/zellij-org/zellij/pull/1827)
 * debugging: Improve error handling in `server/route` (https://github.com/zellij-org/zellij/pull/1808)
+* debugging: Detect plugin version mismatches (https://github.com/zellij-org/zellij/pull/1838)
 
 ## [0.31.4] - 2022-09-09
 * Terminal compatibility: improve vttest compliance (https://github.com/zellij-org/zellij/pull/1671)
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3405,6 +3405,7 @@ dependencies = [
  "highway",
  "insta",
  "log",
+ "semver",
  "serde_json",
  "sixel-image",
  "sixel-tokenizer",
--- a/zellij-server/Cargo.toml
+++ b/zellij-server/Cargo.toml
@@ -31,6 +31,7 @@ sixel-tokenizer = "0.1.0"
 sixel-image = "0.1.0"
 arrayvec = "0.7.2"
 uuid = { version = "0.8.2", features = ["serde", "v4"] }
+semver = "0.11.0"
 
 [dev-dependencies]
 insta = "1.6.0"
--- a/zellij-server/src/panes/plugin_pane.rs
+++ b/zellij-server/src/panes/plugin_pane.rs
@@ -15,6 +15,7 @@ use zellij_utils::shared::ansi_len;
 use zellij_utils::{
     channels::SenderWithContext,
     data::{Event, InputMode, Mouse, PaletteColor},
+    errors::prelude::*,
     pane_size::{Dimension, PaneGeom},
     shared::make_terminal_title,
 };
@@ -137,12 +138,16 @@ impl Pane for PluginPane {
     fn render(
         &mut self,
         client_id: Option<ClientId>,
-    ) -> Option<(Vec<CharacterChunk>, Option<String>, Vec<SixelImageChunk>)> {
+    ) -> Result<Option<(Vec<CharacterChunk>, Option<String>, Vec<SixelImageChunk>)>> {
         // this is a bit of a hack but works in a pinch
-        client_id?;
-        let client_id = client_id.unwrap();
+        let client_id = match client_id {
+            Some(id) => id,
+            None => return Ok(None),
+        };
         // if self.should_render {
         if true {
+            let err_context = || format!("failed to render plugin panes for client {client_id}");
+
             // while checking should_render rather than rendering each pane every time
             // is more performant, it causes some problems when the pane to the left should be
             // rendered and has wide characters (eg. Chinese characters or emoji)
@@ -158,12 +163,16 @@ impl Pane for PluginPane {
                     self.get_content_rows(),
                     self.get_content_columns(),
                 ))
-                .unwrap();
+                .to_anyhow()
+                .with_context(err_context)?;
 
             self.should_render = false;
+            // This is where we receive the text to render from the plugins.
             let contents = buf_rx
                 .recv()
-                .expect("Failed to receive reply from plugin. Please check the logs");
+                .with_context(err_context)
+                .to_log()
+                .unwrap_or("No output from plugin received. See logs".to_string());
             for (index, line) in contents.lines().enumerate() {
                 let actual_len = ansi_len(line);
                 let line_to_print = if actual_len > self.get_content_columns() {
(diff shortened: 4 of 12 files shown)


har7an:
## TODO

- [ ] rustfmt
- [ ] check whether this has any performance impact
- [ ] check whether this somehow increases binsize

har7an:
@imsnif Are you fine with the new error formats posted above? You needn't read all the text, but you can skim the text in the code fences.

imsnif:
@har7an - skimming over the code blocks, this looks great. Minor nitpick - I think it looks better if the error voice is passive "the plugins do not match the current Zellij version".

Otherwise - pretty awesome. Will this happen for any sort of deserialization error?

har7an:
> Otherwise - pretty awesome. Will this happen for any sort of deserialization error?

It will happen for all unhandled panics in the plugins. The "version mismatch" message in particular will be shown for all deserialization failures in the call to `update`. I know that some of the other plugin API calls also deserialize things from stdin, they're simply unwrapping (without the pretty text block).

PR #1840, opened 2022-10-24
improve error handling in pty

naosense:
Replacing unwrap in pty.rs with error propagation.

Related to #1753 .

BTW, I don't know how to handle error in `task.spawn`, any advise? @har7an

### DIFF
--- a/zellij-server/src/pty.rs
+++ b/zellij-server/src/pty.rs
@@ -12,6 +12,7 @@ use std::{collections::HashMap, env, os::unix::io::RawFd, path::PathBuf};
 use zellij_utils::nix::unistd::Pid;
 use zellij_utils::{
     async_std,
+    errors::prelude::*,
     errors::{ContextType, PtyContext},
     input::{
         command::{RunCommand, TerminalAction},
@@ -86,7 +87,8 @@ pub(crate) struct Pty {
     default_editor: Option<PathBuf>,
 }
 
-pub(crate) fn pty_thread_main(mut pty: Pty, layout: Box<Layout>) {
+pub(crate) fn pty_thread_main(mut pty: Pty, layout: Box<Layout>) -> Result<()> {
+    let err_context = || "failed in pty thread main".to_string();
     loop {
         let (event, mut err_ctx) = pty.bus.recv().expect("failed to receive event on channel");
         err_ctx.add_call(ContextType::Pty((&event).into()));
@@ -115,7 +117,7 @@ pub(crate) fn pty_thread_main(mut pty: Pty, layout: Box<Layout>) {
                                 should_float,
                                 client_or_tab_index,
                             ))
-                            .unwrap();
+                            .with_context(err_context)?;
                     },
                     Err(SpawnTerminalError::CommandNotFound(pid)) => {
                         if hold_on_close {
@@ -127,7 +129,7 @@ pub(crate) fn pty_thread_main(mut pty: Pty, layout: Box<Layout>) {
                                     should_float,
                                     client_or_tab_index,
                                 ))
-                                .unwrap();
+                                .with_context(err_context)?;
                             if let Some(run_command) = run_command {
                                 send_command_not_found_to_screen(
                                     pty.bus.senders.clone(),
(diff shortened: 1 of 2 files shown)


zellij-server/src/pty.rs
```diff
@@ -291,43 +293,43 @@ pub(crate) fn pty_thread_main(mut pty: Pty, layout: Box<Layout>) {
...
-                    .unwrap();
+                    .with_context(err_context)?;
             },
             PtyInstruction::NewTab(terminal_action, tab_layout, tab_name, client_id) => {
                 pty.spawn_terminals_for_layout(
                     tab_layout.unwrap_or_else(|| layout.new_tab()),
                     terminal_action.clone(),
                     client_id,
-                );
+                )?;
```
har7an, line 303: Is this an oversight?

zellij-server/src/pty.rs
```diff
@@ -468,7 +471,8 @@ impl Pty {
         layout: PaneLayout,
         default_shell: Option<TerminalAction>,
         client_id: ClientId,
-    ) {
+    ) -> Result<()> {
+        let err_context = || format!("fail to spawn terminals layout for client {client_id}");
```
har7an, line 475: Sorry, this is picky, but I think we should be consistent with the error phrasing

```suggestion
        let err_context = || format!("failed to spawn terminals for layout for client {client_id}");
```

har7an (changes requested):
Thanks for picking this up!

About `task::spawn`:

This spawns a different task, which, from out point of view, is an independent unit of execution. If you follow the code around the calls to `task::spawn` a little you'll find that we store the `JoinHandle` somewhere, but we never `join` on it. In other words: This is pretty much the same thing as we have with e.g. `move || pty_thread_main`.

So my suggestion is we treat it that way. Here's a proposal what this may look like (This is somewhere inside `spawn_terminal`):

```rust
        let terminal_bytes = task::spawn({
            let err_context = || format!("failed to run async task for terminal {terminal_id}");

            let senders = self.bus.senders.clone();
            let os_input = self.bus.os_input.as_ref().with_context(err_context).fatal().clone();
            let debug_to_file = self.debug_to_file;
            async move {
                TerminalBytes::new(pid_primary, senders, os_input, debug_to_file, terminal_id)
                    .listen()
                    .await;
            }
        });

```

zellij-server/src/pty.rs
naosense: no, because I change `spawn_terminals_for_layout` to return `Result`

zellij-server/src/pty.rs
```diff
@@ -659,46 +660,52 @@ impl Pty {
...
         }
+        Ok(())
     }
-    pub fn close_tab(&mut self, ids: Vec<PaneId>) {
-        ids.iter().for_each(|&id| {
-            self.close_pane(id);
-        });
+    pub fn close_tab(&mut self, ids: Vec<PaneId>) -> Result<()>{
+        for id in ids {
+            self.close_pane(id)?;
```
har7an, line 716: Could you please add an error context here?

zellij-server/src/pty.rs
```diff
@@ -731,11 +739,13 @@ impl Pty {
                     self.bus
                         .os_input
                         .as_mut()
-                        .unwrap()
+                        .with_context(err_context)
+                        .map_err(|_e|SpawnTerminalError::GenericSpawnError("os input is none"))?
```
har7an: These two lines should be swapped, i.e. first we convert the error type (with `map_err`), and then we attach the context to that instead.

har7an (commented):
Seems you found another few, thank you!

I see there are a few places where you use a plain `?` instead of the usual `with_context(...)`. Is that on purpose?

zellij-server/src/pty.rs
naosense: Oh, I see, I'll fix that.

zellij-server/src/pty.rs
naosense: Yeah, I'll fix that.

zellij-server/src/pty.rs
naosense: It report error after i swapped it, so i changed it to `ok_or_else`.

har7an (approved):
Good job, thanks for the contribution!

PR #1842, opened 2022-10-24
client: disallow nested sessions

leoppro:
Disallow nested sessions

Before this PR:
image

After this PR:
image

### DIFF
--- a/src/commands.rs
+++ b/src/commands.rs
@@ -351,7 +351,7 @@ pub(crate) fn start_client(opts: CliArgs) {
             attach_with_session_name(session_name, config_options.clone(), create)
         };
 
-        if let Ok(val) = std::env::var(envs::SESSION_NAME_ENV_KEY) {
+        if let Ok(val) = envs::get_session_name() {
             if val == *client.get_session_name() {
                 eprintln!("You are trying to attach to the current session (\"{}\"). Zellij does not support nesting a session in itself.", val);
                 process::exit(1);
--- a/zellij-client/src/lib.rs
+++ b/zellij-client/src/lib.rs
@@ -11,7 +11,7 @@ use log::info;
 use std::env::current_exe;
 use std::io::{self, Write};
 use std::path::Path;
-use std::process::Command;
+use std::process::{Command, self};
 use std::sync::{Arc, Mutex};
 use std::thread;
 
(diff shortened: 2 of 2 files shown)


har7an:
Hello,

what exactly is the problem you're trying to fix with this PR? I know that we have a few users that nest zellij sessions from time to time, me included.

sploders101:
FWIW, Tmux has similar behavior. My personal opinion (as someone who also nests zellij sessions from time to time) is that as long as there is an escape hatch, I think it's a good idea.

That being said, given that the default config shows keybinds at the bottom and clearly labels the "LOCK" keybind, I'm not sure this is necessary either. It seems more like a feature parity tweak than a solution to an actual problem.

imsnif:
The absolute best solution (IMO) is for Zellij to detect this and then (by default):
1. Open the new session in a new tab with a visual indication that this is a remote session tab (color? text?)
2. The new tab is entirely the other session, even the plugins themselves
3. Only visual difference is instead of `Zellij` it should say `Zellij - Remote Session` for clarity

karczex:
@har7an 
From perspective of the user who do not need nested zellij sessions,  logic similar to one in this PR have to be implemented somewhere. It may be implemented in zellij itself or may be implemented i.e. in oh-my-zsh plugin   

The basic functionality it provides is to allow user to start zellij automatically in each terminal and do not nest it in every pane or tab. Currently I'm using [zsh plugin from my fork of oh-my-zsh]( https://github.com/karczex/ohmyzsh/blob/zellij_plugin/plugins/zellij/zellij.plugin.zsh) and it works well, but I would rather to use mainline oh-my-zsh than maintaining my own fork or patches. 

Another solution to my problem would be to make possible to set zellij as default shell (via chsh) and add setting to zellij to specify which shell should be run by default.  
 
So please decide what should be the fate of this PR. If you decide to close it, I may open PR to oh-my-zsh.

@sploders101 
What do you use for nested zellij sessions?

har7an:
> The basic functionality it provides is to allow user to start zellij automatically in each terminal and do not nest it in every pane or tab.

I do get that, but if you're only concerned about autostarting zellij as part of your shell profile/`.${SHELL}rc`, we have a solution for that: `zellij setup --generate-auto-start`. Does that help with your particular problem?

t18n:
Any update on this? I personally never nest Zellij and I bind Ctrl+E to the `zellij` command. All I want is some sort of notification/nothing when I accidently click Ctrl+E from within a session.

Issue #1817, opened 2022-10-19
`search term` in scroll mode can only be used once


naosense:
**Basic information**

zellij --version:zellij 0.31.4
stty size:22 100
uname -av or ver(Windows):Darwin ZBMAC-C02C21EH2.local 19.6.0 Darwin Kernel Version 19.6.0: Mon Apr 18 21:50:40 PDT 2022; root:xnu-6153.141.62~1/RELEASE_X86_64 x86_64

**Further information**
`search term` in scroll mode can only be used once, see below.  First i type `file`, it worked, then i type `will`, it didn't work any more.

https://user-images.githubusercontent.com/2225426/196627313-09edce28-29ac-462b-9351-4bd6e08f5423.mp4

imsnif:
Hey - I think this behavior makes sense if you see the UI (eg. with pane frames enabled), no?

imsnif:
I think there's a misunderstanding here that comes from part of the UI being disabled (the pane frames).
Could you try doing the same thing with pane frames enabled (eg. `ctrl+p` + `z`) and seeing if it makes more sense to you? If you do that you see the search term still on the pane frame and see that you're actually just adding to it and not replacing it.

naosense:
Ah, I see it after the panel frames enable. But the keyword is not clear when the panel frames are disabled, does it have a better way to show the keyword?

imsnif:
I agree we should probably show this in the status bar's tooltip as well

naosense:
To display the keyword in the status bar or to press `s` to clear the previous keyword like `/` in vim? I think the latter is more intuitive. How do you think?

naosense:
Fixed by #1853 , close this issue.

PR #1853, opened 2022-10-26
allow terminating char to clear search term

naosense:
Hi @imsnif, I found something interesting after I walked around the code base.
Because `\0` is not allowed here,
https://github.com/zellij-org/zellij/blob/5793af7655b327f6cbd6f68a26afe59e470dc366/zellij-server/src/tab/mod.rs#L2610-L2611
it wouldn't step in this branch below.
https://github.com/zellij-org/zellij/blob/5793af7655b327f6cbd6f68a26afe59e470dc366/zellij-server/src/panes/terminal_pane.rs#L653-L656
I don't know if this is on purpose or just an oversight, if this is on purpose, I will close this pr, thanks.

Related to #1817


imsnif:
Good find @naosense !! This seems to make sense to me (and is working for me). @msirringhaus - what do you think? Do you see any downsides to implementing this?

msirringhaus:
LGTM!

PR #1870, opened 2022-10-28
improve error handling in ui module

naosense:
Replacing unwrap in `ui` module with error propagation.

Related to #1753

### DIFF
--- a/zellij-server/src/panes/floating_panes/mod.rs
+++ b/zellij-server/src/panes/floating_panes/mod.rs
@@ -17,6 +17,7 @@ use std::rc::Rc;
 use std::time::Instant;
 use zellij_utils::{
     data::{ModeInfo, Style},
+    errors::prelude::*,
     input::command::RunCommand,
     pane_size::{Offset, PaneGeom, Size, Viewport},
 };
@@ -234,7 +235,8 @@ impl FloatingPanes {
             resize_pty!(pane, os_api);
         }
     }
-    pub fn render(&mut self, output: &mut Output) {
+    pub fn render(&mut self, output: &mut Output) -> Result<()> {
+        let err_context = || "failed to render output";
         let connected_clients: Vec<ClientId> =
             { self.connected_clients.borrow().iter().copied().collect() };
         let mut floating_panes: Vec<_> = self.panes.iter_mut().collect();
@@ -242,8 +244,16 @@ impl FloatingPanes {
             self.z_indices
                 .iter()
                 .position(|id| id == *a_id)
-                .unwrap()
-                .cmp(&self.z_indices.iter().position(|id| id == *b_id).unwrap())
+                .with_context(err_context)
+                .fatal()
+                .cmp(
+                    &self
+                        .z_indices
+                        .iter()
+                        .position(|id| id == *b_id)
+                        .with_context(err_context)
+                        .fatal(),
+                )
         });
 
         for (z_index, (kind, pane)) in floating_panes.iter_mut().enumerate() {
@@ -266,23 +276,27 @@ impl FloatingPanes {
                     .get(client_id)
                     .unwrap_or(&self.default_mode_info)
                     .mode;
-                pane_contents_and_ui.render_pane_frame(
-                    *client_id,
-                    client_mode,
-                    self.session_is_mirrored,
-                );
+                pane_contents_and_ui
+                    .render_pane_frame(*client_id, client_mode, self.session_is_mirrored)
+                    .with_context(err_context)?;
                 if let PaneId::Plugin(..) = kind {
-                    pane_contents_and_ui.render_pane_contents_for_client(*client_id);
+                    pane_contents_and_ui
+                        .render_pane_contents_for_client(*client_id)
+                        .with_context(err_context)?;
                 }
                 // this is done for panes that don't have their own cursor (eg. panes of
                 // another user)
-                pane_contents_and_ui.render_fake_cursor_if_needed(*client_id);
+                pane_contents_and_ui
+                    .render_fake_cursor_if_needed(*client_id)
+                    .with_context(err_context)?;
             }
             if let PaneId::Terminal(..) = kind {
                 pane_contents_and_ui
-                    .render_pane_contents_to_multiple_clients(connected_clients.iter().copied());
+                    .render_pane_contents_to_multiple_clients(connected_clients.iter().copied())
+                    .with_context(err_context)?;
             }
         }
+        Ok(())
     }
     pub fn resize(&mut self, new_screen_size: Size) {
         let display_area = *self.display_area.borrow();
--- a/zellij-server/src/panes/plugin_pane.rs
+++ b/zellij-server/src/panes/plugin_pane.rs
@@ -231,10 +231,10 @@ impl Pane for PluginPane {
         _client_id: ClientId,
         frame_params: FrameParams,
         input_mode: InputMode,
-    ) -> Option<(Vec<CharacterChunk>, Option<String>)> {
+    ) -> Result<Option<(Vec<CharacterChunk>, Option<String>)>> {
         // FIXME: This is a hack that assumes all fixed-size panes are borderless. This
         // will eventually need fixing!
-        if self.frame && !(self.geom.rows.is_fixed() || self.geom.cols.is_fixed()) {
+        let res = if self.frame && !(self.geom.rows.is_fixed() || self.geom.cols.is_fixed()) {
             let pane_title = if self.pane_name.is_empty()
                 && input_mode == InputMode::RenamePane
                 && frame_params.is_main_client
@@ -251,10 +251,15 @@ impl Pane for PluginPane {
                 pane_title,
                 frame_params,
             );
-            Some(frame.render())
+            Some(
+                frame
+                    .render()
+                    .with_context(|| format!("failed to render frame form client {_client_id}"))?,
+            )
         } else {
             None
-        }
+        };
+        Ok(res)
     }
     fn render_fake_cursor(
         &mut self,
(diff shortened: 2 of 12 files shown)


zellij-server/src/panes/floating_panes/mod.rs
```diff
@@ -234,16 +235,25 @@ impl FloatingPanes {
...
-                .cmp(&self.z_indices.iter().position(|id| id == *b_id).unwrap())
+                .with_context(err_context)
+                .fatal()
+                .cmp(
+                    &self
+                        .z_indices
+                        .iter()
+                        .position(|id| id == *b_id)
+                        .with_context(err_context)
+                        .fatal(),
```
har7an, line 255: Why are you using `.fatal()` instead of `?` here?

zellij-server/src/panes/plugin_pane.rs
```diff
@@ -251,10 +251,15 @@ impl Pane for PluginPane {
                 pane_title,
                 frame_params,
             );
-            Some(frame.render())
+            Some(
+                frame
+                    .render()
+                    .with_context(|| format!("failed to render frame form client {_client_id}"))?,
```
har7an, line 257: There's a little typo here:
```suggestion
                    .with_context(|| format!("failed to render frame for client {_client_id}"))?,
```

zellij-server/src/panes/terminal_pane.rs
```diff
@@ -398,12 +398,13 @@ impl Pane for TerminalPane {
             frame.add_exit_status(exit_status.as_ref().copied());
         }
 
-        match self.frame.get(&client_id) {
+        let err_context = || format!("failed to render frame for client {client_id}");
```
har7an, line 401: Could you please move this to the very top of the function? It makes spotting whether someone already added an `err_context` much easier.

naosense:
Hi @har7an , could you please review this pr?

zellij-server/src/panes/tiled_panes/mod.rs
```diff
@@ -378,7 +379,8 @@ impl TiledPanes {
     pub fn has_panes(&self) -> bool {
         !self.panes.is_empty()
     }
-    pub fn render(&mut self, output: &mut Output, floating_panes_are_visible: bool) {
+    pub fn render(&mut self, output: &mut Output, floating_panes_are_visible: bool) -> Result<()> {
+        let err_context = || "failed to render output based on visible";
```
har7an, line 383: I get your intention here, but I don't think we should mention the "visible" condition in the error. What about: "failed to render tiled panes"?

(Same goes for the err_context further down)

zellij-server/src/ui/boundaries.rs
```diff
@@ -47,18 +48,24 @@ impl BoundarySymbol {
...
+        let tc = if self.invisible {
             EMPTY_TERMINAL_CHARACTER
         } else {
-            let character = self.boundary_type.chars().next().unwrap();
+            let character = self.boundary_type.chars().next().with_context(|| {
+                format!(
+                    "failed to as terminal character for boundary type {}",
+                    self.boundary_type
+                )
+            })?;
```
har7an, line 60: I think we should expand this a little. The `unwrap` here originally unwrapped on an `Option`. A `None`, however, doesn't tell us what "Error" occured here, because that is context specific. I would suggest something like this:

```rust
                .context("no boundary symbols defined")  // <-- Tells us why "None" here is actually an error for us
                .with_context(|| format!("failed to convert boundary symbol {} into terminal character", ...))?;  // <-- Gives us the usual error context
```

zellij-server/src/ui/overlay/mod.rs
```diff
@@ -28,9 +29,11 @@ pub enum OverlayType {
 }
 
 impl Overlayable for OverlayType {
-    fn generate_overlay(&self, size: Size) -> String {
+    fn generate_overlay(&self, size: Size) -> Result<String> {
         match &self {
-            OverlayType::Prompt(prompt) => prompt.generate_overlay(size),
+            OverlayType::Prompt(prompt) => prompt
+                .generate_overlay(size)
+                .context("failed to generate overlay for overlay type"),
```
har7an, line 36: Since we're implementing this trait below an a real type called `Overlay` we should word the context a little different here. Given the docstring for `Overlayable`, what do you think about the following?

```suggestion
                .context("failed to generate VTE output from overlay type"),
```

zellij-server/src/ui/overlay/mod.rs
```diff
@@ -44,15 +47,17 @@ pub struct OverlayWindow {
...
-    fn generate_overlay(&self, size: Size) -> String {
+    fn generate_overlay(&self, size: Size) -> Result<String> {
         let mut output = String::new();
         //let clear_display = "\u{1b}[2J";
         //output.push_str(&clear_display);
         for overlay in &self.overlay_stack {
-            let vte_output = overlay.generate_overlay(size);
+            let vte_output = overlay
+                .generate_overlay(size)
+                .context("failed to generate overlay for overlay window")?;
```
har7an, line 57:
```suggestion
                .context("failed to generate VTE output from overlay window")?;
```

zellij-server/src/ui/overlay/mod.rs
```diff
@@ -70,8 +75,10 @@ impl Overlay {
 }
 
 impl Overlayable for Overlay {
-    fn generate_overlay(&self, size: Size) -> String {
-        self.overlay_type.generate_overlay(size)
+    fn generate_overlay(&self, size: Size) -> Result<String> {
+        self.overlay_type
+            .generate_overlay(size)
+            .context("failed to generate overlay for overlay")
```
har7an, line 81:
```suggestion
            .context("failed to generate VTE output from overlay")
```

zellij-server/src/ui/overlay/prompt.rs
```diff
@@ -46,9 +47,9 @@ impl Overlayable for Prompt {
                 x + 1,
                 h,
             )
-            .unwrap();
+            .context("failed to generate overlay for prompt")?;
```
har7an, line 50:
```suggestion
            .context("failed to generate VTE output for prompt")?;
```

zellij-server/src/ui/pane_boundaries_frame.rs
```diff
@@ -600,24 +601,26 @@ impl PaneFrame {
...
     }
-    fn render_title(&self) -> Vec<TerminalCharacter> {
+    fn render_title(&self) -> Result<Vec<TerminalCharacter>> {
         let total_title_length = self.geom.cols.saturating_sub(2); // 2 for the left and right corners
 
         self.render_title_middle(total_title_length)
             .map(|(middle, middle_length)| self.title_line_with_middle(middle, &middle_length))
             .or_else(|| Some(self.title_line_without_middle()))
-            .unwrap()
+            .with_context(|| format!("failed to render title {}", self.title))
```
har7an, line 610: Personally, I like to enclose strings inside print statements into `'` quotes, because the string we print (`self.title`) may be empty, which can be confusing depending on what to print. But that's my personal preference, you can leave it as is, if you prefer!

zellij-server/src/ui/pane_boundaries_frame.rs
```diff
@@ -600,24 +601,26 @@ impl PaneFrame {
...
     }
-    fn render_held_undertitle(&self) -> Vec<TerminalCharacter> {
+    fn render_held_undertitle(&self) -> Result<Vec<TerminalCharacter>> {
         let max_undertitle_length = self.geom.cols.saturating_sub(2); // 2 for the left and right corners
-        let exit_status = self.exit_status.unwrap(); // unwrap is safe because we only call this if
+        let exit_status = self
+            .exit_status
+            .with_context(|| format!("failed to render held under title {}", self.title))?; // unwrap is safe because we only call this if
```
har7an, line 616: The function name here unfortunately doesn't really hint towards what the function tries to accomplish. Something like this may fit better:

```suggestion
            .with_context(|| format!("failed to render command pane status '{}'", self.title))?;
```

zellij-server/src/ui/pane_boundaries_frame.rs
```diff
@@ -665,14 +668,16 @@ impl PaneFrame {
             } else {
                 self.empty_undertitle(max_undertitle_length)
             }
-        }
+        };
+        Ok(res)
     }
-    pub fn render(&self) -> (Vec<CharacterChunk>, Option<String>) {
+    pub fn render(&self) -> Result<(Vec<CharacterChunk>, Option<String>)> {
+        let err_context = || "failed to render";
```
har7an, line 675: We have many functions called `render` in the code, so we can add some additional information:

```suggestion
        let err_context = || "failed to render pane frame";
```

har7an (changes requested):
Thanks for your continued help with the error handling!

I've left a few small remarks on your code. Overall this looks good and it helps us a lot.

zellij-server/src/panes/floating_panes/mod.rs
naosense: Because returning result is not allowed in a closure

har7an:
And on a sidenote, it seems you beat me to the change in `ui/boundaries` by a few minutes: #1871. :-)

But that's okay, since your PR touches many more things I'll close mine and we'll go with yours instead.

zellij-server/src/panes/floating_panes/mod.rs
har7an: Ah, I missed the closure above, my bad!

naosense:
Thank you for your patient review, I learned a lot from your comments

har7an (approved):
Well done, thanks for your contribution! And thank you for being so open to my remarks.

PR #1871, opened 2022-10-28
errors: Don't unwrap in `server::ui::boundaries`

har7an:
Remove calls to `unwrap` and return a `Result` type instead, enriched with error context messages.


har7an:
Resolved in #1870.

Issue #1864, opened 2022-10-27
Status-bar - Hide tip line when plugin is configurer with `size=1`


welcoMattic:
Actually, with a layout like this:

```kdl
pane size=1 borderless=true {
    plugin location="zellij:status-bar"
}
```

It renders only the "tip line"
![image](https://user-images.githubusercontent.com/773875/198249784-7872419d-5436-44bc-ac4a-e4d1c5aea6cf.png)

Could it be possible to change it to render only the "controls line"?

Or better, add an option to configure the plugin from the layout like:

```kdl
pane size=2 borderless=true {
    plugin location="zellij:status-bar" options="lines:tip,controls"
}
```

```kdl
pane size=1 borderless=true {
    plugin location="zellij:status-bar" options="lines:controls"
}
```
Where we can choose the rendered line when size=1?

Thanks

imsnif:
Hey, we can definitely add this as an option - but have you maybe considered the `compact-bar` instead?
https://github.com/zellij-org/zellij/blob/main/zellij-utils/assets/layouts/compact.kdl#L4

welcoMattic:
compact bar is about displaying tabs, right? It adds the display of the current mode, but not the controls list?

imsnif:
True - I guess you still want the mode+shortcuts display?
In that case we can make the status bar "do the right thing" depending on its size. Shouldn't be super difficult to change

welcoMattic:
Yes, exactly. I'm super newbie in Rust. I guess I found the right place to do the trick, but not sure.

I guess we have to add some condition right there https://github.com/zellij-org/zellij/blob/main/default-plugins/status-bar/src/main.rs#L225-L233 ?

imsnif:
> I guess we have to add some condition right there https://github.com/zellij-org/zellij/blob/main/default-plugins/status-bar/src/main.rs#L225-L233 ?

That seems right, yes! Would you like to give it a shot? If so, be sure to follow the build instructions in CONTRIBUTING.md - otherwise the plugins won't build and it could be very frustrating to figure out why your changes aren't taking effect.

welcoMattic:
@imsnif Pr is open ;)

imsnif:
Thanks again @welcoMattic !

PR #1875, opened 2022-10-28
Display controls line of status-bar plugin if pane size=1

welcoMattic:
Ref https://github.com/zellij-org/zellij/issues/1864

This PR will display only the controls line of status bar plugin if its pane has `size=1` instead of tip line.

The tip line is still displayed if `size=2`.

image

(FYI: it's my 1st PR on a Rust project, if it miss something let me know)

### DIFF
--- a/default-plugins/status-bar/src/main.rs
+++ b/default-plugins/status-bar/src/main.rs
@@ -204,7 +204,7 @@ impl ZellijPlugin for State {
         }
     }
 
-    fn render(&mut self, _rows: usize, cols: usize) {
+    fn render(&mut self, rows: usize, cols: usize) {
         let supports_arrow_fonts = !self.mode_info.capabilities.arrow_fonts;
         let separator = if supports_arrow_fonts {
             ARROW_SEPARATOR
@@ -230,7 +230,10 @@ impl ZellijPlugin for State {
                 println!("{}\u{1b}[48;5;{}m\u{1b}[0K", first_line, color);
             },
         }
-        println!("\u{1b}[m{}\u{1b}[0K", second_line);
+
+        if rows > 1 {
+            println!("\u{1b}[m{}\u{1b}[0K", second_line);
+        }
     }
 }
 


default-plugins/status-bar/src/main.rs
```diff
@@ -230,7 +230,10 @@ impl ZellijPlugin for State {
                 println!("{}\u{1b}[48;5;{}m\u{1b}[0K", first_line, color);
             },
         }
-        println!("\u{1b}[m{}\u{1b}[0K", second_line);
+
+        if _rows > 1 {
```
imsnif, line 234: Let's change this to `rows` - it was previously prefixed with an underscore to denote that it's an unused variable in this function. Now we use it.

imsnif (changes requested):
Hey @welcoMattic - looking great! Thanks for choosing us for your first Rust PR :)

I left a minor comment in the code. Otherwise, I think I'd rather not add another default layout for this so they won't get too crowded. We've now changed the plugin so it "does the right thing" if someone chooses to create such a layout.

After these changes, I'd be happy to merge.

welcoMattic:
@imsnif Thanks for your review. I addressed your comments, I hope it's now ok for you ;)

imsnif:
Thanks @welcoMattic !

panki27:
@welcoMattic does this still work for you?

I've tried this today with `zellij 0.34.4`. 
I'd love to only see the keybind hints, not the tips, but if I set `size=1` I don't get a status-bar at all anymore. With `size=2`, it works as normal.

Here is my layout:

```
layout {
    pane size=1 borderless=true {
    }
    pane {
        pane 
    }
    pane size=1 borderless=true {
    }
}
```

welcoMattic:
@panki27 you are right, since last version it's broken. Can you open a new issue please?

panki27:
@welcoMattic thank you for confirming!
I've created a new issue: #2077

PR #1877, opened 2022-10-28
docs(example): update the format of the themes for the example directory

jaeheonji:


a-kenji:
@jaeheonji 
This is not documenation for themes, 
this is documentation on the format of the themes for the example directory itself.

jaeheonji:
@a-kenji Thank you! Actually, I was confused on how to classify it.

example/themes/README.md
```diff
@@ -1,7 +1,19 @@
...
-Please make sure that the theme name and the file name are the same (+`.kdl`).
+
+Themes can contain different flavors in one file, or can be created as individual files.
 
 Example:
 
-- theme: gruvbox
-- filename: `gruvbox.kdl`
+```
+gruvbox.kdl
```
a-kenji, line 8: This is not a format we want in the example directory.

jaeheonji: I apologize for doing the merge first.

At this point, I just wanted to make it clear that a single file can contain multiple themes, and I didn't think much more than that. Because there is a detailed explanation in the official documentation.

Do you think we need a more detailed explanation here?

PR #1884, opened 2022-10-31
fix(keybinds): remove space key from shared_except

### LINKED ISSUE #1862: [0.32.0] Unable to rename tabs and panes with spaces
**Basic information**

- `zellij --version`: 0.32.0
- `stty size`: 48 211
- `uname -av`: Linux mybox 6.0.2-76060002-generic #202210150739\~1666289067\~22.04\~fe0ce53 SMP PREEMPT_DYNAMIC Thu O x86_64 x86_64 x86_64 GNU/Linux

**Programs**

- `alacritty --version`: 0.11.0 [8dbaa0b](https://github.com/alacritty/alacritty/commit/8dbaa0bbd308a72075910ff32d8ffcca09ae3529)
- `nu --version`: 0.70.0
- `starship --version`: 1.11.0

**Further information**

With Zellij 0.32.0, I am unable to rename my tabs and panes with spaces in their name such as "audio sharing". When I press the spacebar after typing "audio", it confirms the entry as "audio" and exits out of the rename mode.
I installed Zellij using `cargo install --locked zellij` using Rust 1.64.0.

I downgraded to Zellij 0.31.0 with `cargo install --locked zellij --version 0.31.0` and the renaming behaves as expected there. After confirming this, I upgraded to 0.32.0 again and I see the renaming issue again.

I don't have any configuration for Zellij because I'm happy with the defaults.

jaeheonji:
fix #1862 

In Rename mode, the space key is registered in the shared except, so it seems to be a problem that the key cannot be used.

@imsnif `shared_except` seems to be a new feature during kdl conversion.
Is this the correct solution? Or should I see this as an unintended bug?

### DIFF
--- a/zellij-utils/assets/config/default.kdl
+++ b/zellij-utils/assets/config/default.kdl
@@ -139,7 +139,7 @@ keybinds {
         bind "Alt -" { Resize "Decrease"; }
     }
     shared_except "normal" "locked" {
-        bind "Enter" "Space" "Esc" { SwitchToMode "Normal"; }
+        bind "Enter" "Esc" { SwitchToMode "Normal"; }
     }
     shared_except "pane" "locked" {
         bind "Ctrl p" { SwitchToMode "Pane"; }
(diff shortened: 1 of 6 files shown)


imsnif:
Good catch @jaeheonji ! Yes, I think this is the right way to do it. I just don't remember why I put it there :/

jaeheonji:
@imsnif Thank you for your answer! It happen :)

PR #1833, opened 2022-10-21
feat: bump wasmer to v2.3.0

xJonathanLEI:
This PR bumps `wamser` from v1.0.0 to v2.3.0 and resolve errors from the bump.

`wasmer` is one of the dependencies that break when building on `aarch64` Android. The version currently used (v1.0.0) fails to compile on Android but the updated version (v2.3.0) can build without issue. Thus this PR will help with #1201.

Issue #1021, opened 2022-01-23
Support termux


dead10ck:
It would be awesome if you could run Zellij in Termux on Android. Currently it fails to build because wasmer doesn't support Android. There's an upstream issue at https://github.com/wasmerio/wasmer/issues/2099 but I thought it would be good to track it here too.

I also wonder if it would be feasible to add a Cargo feature to make the whole plugin system optional. I'm not super familiar with zellij yet, so I'm not sure how much critical functionality comes from the plugins.

jaeheonji:
> I'm not super familiar with zellij yet, so I'm not sure how much critical functionality comes from the plugins.

Currently, plugins only provide optional functions. Most are only related to UI, so they are not involved in the overall function. 

> I also wonder if it would be feasible to add a Cargo feature to make the whole plugin system optional.

It is theoretically possible. However, this is not the main goal at this point, and it seems that this work will require a lot of verification and modification.

Hope it helps 😄

xJonathanLEI:
Unfortunately it looks like the upstream issue has been set to low priority. Would love to daily drive with Zellij, but this forces me to use `tmux` locally and only use Zellij on the remote. The context switch is a bit painful tbh.

xJonathanLEI:
Sorry for spamming _again_. It looks like `wasmer` 2.3.0 works on Termux. I just tried building it and it works:

```console
$ rustc- vV
rustc 1.64.0
binary: rustc
commit-hash: unknown
commit-date: unknown
host: aarch64-linux-android
release: 1.64.0
LLVM version: 15.0.1
```

In fact, `wasmer` v2.3.0 is in the Termux package registry with no patch needed.

Zellij uses `wasmer` v1 though. Maybe we can migrate to v2 and then it will automatically work on Termux?

xJonathanLEI:
Actually, I just upgraded `wasmer` to v2.3.0 in #1833.

xJonathanLEI:
As mentioned in https://github.com/zellij-org/zellij/pull/1833#issuecomment-1292919450, 2 dependencies are broken for building on Android:

- `wasmer`
- `termwiz`

After tackling `wasmer` in #1833 (though it's not merging anytime soon), I just submitted https://github.com/wez/wezterm/pull/2694 for fixing the `termwiz` error too. After this we should have resolved all the build errors, and can focus on resolving runtime errors instead :)

xJonathanLEI:
Guys I made it.

[...]

PR #1896, opened 2022-11-02
chore(deps): bump termwiz to 0.19.0

xJonathanLEI:
Thanks to the swift action from Wez (not using @ here to avoid spamming his inbox), https://github.com/wez/wezterm/pull/2694 has been merged and `termwiz` v0.19.0 has been [released](https://crates.io/crates/termwiz/0.19.0) with the fix. This PR bumps the dependency version to v0.19.0 to take advantage of it.

This is part of the effort towards fully resolving #1021, and it should be _much_ less disruptive compared to #1833.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -632,11 +632,12 @@ dependencies = [
 
 [[package]]
 name = "csscolorparser"
-version = "0.5.0"
+version = "0.6.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "b2fb3bd93ef32553e3d5b9f8020028f41ac64ff8a230033d5d548b8222d21fbe"
+checksum = "eb2a7d3066da2de787b7f032c736763eb7ae5d355f81a68bab2675a96008b0bf"
 dependencies = [
- "phf",
+ "lab",
+ "phf 0.11.1",
 ]
 
 [[package]]
@@ -692,6 +693,12 @@ dependencies = [
  "quote",
  "syn",
 ]
+
+[[package]]
+name = "deltae"
+version = "0.3.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "e412cd91a4ec62fcc739ea50c40babe21e3de60d69f36393cce377c7c04ead5a"
 
 [[package]]
 name = "derivative"
@@ -860,6 +867,18 @@ dependencies = [
  "thiserror",
  "winapi",
 ]
+
+[[package]]
+name = "finl_unicode"
+version = "1.2.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "8fcfdc7a0362c9f4444381a9e697c79d435fe65b52a37466fc2c1184cee9edc6"
+
+[[package]]
+name = "fixedbitset"
+version = "0.4.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "0ce7134b9999ecaf8bcd65542e436736ef32ddca1b3e06094cb6ec5755203b80"
 
 [[package]]
 name = "fnv"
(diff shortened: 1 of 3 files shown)


imsnif:
@tlinford - could you do me a favor and check this out on mac? Specifically to make sure all the default `Ctrl` keybindings work. I think it should be good but there was a breaking change and I want to make sure we won't cause trouble.

xJonathanLEI:
Sure! Will test this one on my macbook. Not sure about what caused the CI failure though.

imsnif:
@a-kenji - any ideas why the nix build msrv is failing?

a-kenji:
@imsnif,
The bump to `0.19.0` seems to be incompatible with our current advertised MSRV.
The `termwiz` release seems to need a more recent rust version. A minor version bump is probably sufficient.

imsnif:
Thanks @a-kenji - can you help me understand how this mismatch affects us?

tlinford:
Looks good to me - all keybindings in the default config seem fine.

xJonathanLEI:
The current advertised MSRV is `1.59`. I just tested with [cargo msrv](https://github.com/foresterre/cargo-msrv) and it looks like:

- the actual MSRV on `main` right now is `1.59.0`
- after this PR the actual MSRV is `1.60.0`

I guess the question then becomes: do we want to bump MSRV? For your reference:

[Packaging status](https://repology.org/project/rust/versions)

xJonathanLEI:
For comparison, [Helix](https://github.com/helix-editor/helix) has a MSRV of `1.61.0` on `master`. Personally I don't think MSRV matters too much...

imsnif:
@xJonathanLEI - what do you mean by this PR needing the MSRV bump after all?

Honestly, my preference in these things is not to make changes unless they give us some benefit. If our actual MSRV is higher than what we advertise due to the dependencies we use... why does it matter? How does it affect us or our users? If there's an effect I'm not aware of - for sure. Let's bump it. But otherwise I'd opt to keep things as they are and remove this CI test.

xJonathanLEI:
@imsnif MSRV indicates the minimum Rust compiler version that our code can be compiled with. It makes zero difference to users who always use the latest Rust tool chain version.

It only matters for users who are forced to use an old version of Rust:

- those who don't have permission to install stuff
- another project depending on us that maintains a lower MSRV (we are that "another project" right now)

This is when we changed our MSRV last time: https://github.com/zellij-org/zellij/pull/1177

This is where we "advertise" our MSRV:

https://github.com/zellij-org/zellij/blob/5975af6e42f512700101b3ca4ad6cbe5ca416117/Cargo.toml#L11

The CI is there to check the code can actually be compiled with the declared minimum version.  If it fails, it means our advertised MSRV needs a bump to match the actual MSRV.

imsnif:
> The CI is there to check the code can actually be compiled with the declared minimum version. If it fails, it means our advertised MSRV needs a bump to match the actual MSRV.

I totally understand and am aware of MSRV - but what I don't understand is this part. 

Why? Look how much time we're spending on this. If a user tries to compile with our MSRV version, they will get an error because of the MSRV in our dependencies. Same error they would get if we bump our own. What does it matter what we advertise? Why do we need to waste CI cycles on this? Is there something I'm missing?

xJonathanLEI:
Well the CI is there to make sure we're following [the standard](https://doc.rust-lang.org/cargo/reference/manifest.html#the-rust-version-field), where it requires the `rust-version` field to be the MSRV. This helps save user time if they attempt to compile with an older version: `cargo` will just exit directly without trying to build it.

So, if the advertised version here is lower than the actual MSRV, like we're having here before the second commit, users of these Rust versions in between will have to compile and encounter an error to find out - not a big deal IMO. But this is technically the benefit of advertising the accurate MSRV (asserted by the CI) now that you asked :)

In fact, the benefit is so marginal that many projects I came across don't take it seriously. See here Helix is [reporting a MSRV of `1.59`](https://github.com/helix-editor/helix/blob/master/helix-term/Cargo.toml#L13) where it actually is `1.61`.

imsnif:
Thanks for the explanation @xJonathanLEI ! I'm good with merging this as is with the new version because we've already done the work. But this is honestly not something I want in our CI because I think - as you say - the benefit is far too marginal and it makes it harder to merge features and bug fixes. I'll remove it in another PR.

Anything else we need before merging? IMO we're good.

xJonathanLEI:
I think we're good! I've been using this dependency bump for a while and didn't encounter any issue.

PR #1898, opened 2022-11-02
fix: use `temp_dir` for getting temp folder

xJonathanLEI:
The temp folder for storing logs etc. is currently hard-coded to `/tmp`, which is not very portable as this directory doesn't exist on some platforms (e.g. Android). This PR changes to use `temp_dir` from `std::env`, which uses the exact same path as we do now on non-Android Unix systems, but a special Termux path on Android.

This helps with #1021.

### DIFF
--- a/zellij-utils/src/consts.rs
+++ b/zellij-utils/src/consts.rs
@@ -45,7 +45,7 @@ mod unix_only {
     use crate::shared::set_permissions;
     use lazy_static::lazy_static;
     use nix::unistd::Uid;
-    use std::fs;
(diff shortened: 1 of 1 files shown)


xJonathanLEI:
Note that the default path used in `temp_dir` won't work for Termux - it likely only works for rooted Android devices. However, Termux has a [patch](https://github.com/termux/termux-packages/blob/cdbe704d8484a56e6128609a64041b21987f8f3c/packages/rust/os-tmpdir.patch) for this so we're all good.

PR #1910, opened 2022-11-05
convert ALT-centered config to kdl and smoothify

matu3ba:
- Alt-hjkl for pane movements everywhere
- faster pane creation in pane mode with hjkl
- add space to exit mode, if it does not break workflow
- dense: only 127 LOC keybindings, 7 LOC general config
- comment-heavy (for potential useful things): 61 LOC comments


imsnif:
Thanks!

PR #1917, opened 2022-11-09
fix(compact-bar): mouse-click in simplified-ui

imsuck:
Same changes as #1658 but for the compact-bar plugin.

### DIFF
--- a/default-plugins/compact-bar/src/line.rs
+++ b/default-plugins/compact-bar/src/line.rs
@@ -26,8 +26,23 @@ fn populate_tabs_in_tab_line(
     loop {
         let left_count = tabs_before_active.len();
         let right_count = tabs_after_active.len();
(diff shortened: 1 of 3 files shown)


imsuck:
How did the test case `cannot_split_terminals_vertically_when_active_terminal_is_too_small` fail when I'm just editing the source code of a plugin?

jaeheonji:
@imsuck Hey, Sorry for the late reply.

Do you want to find the source code that leads to the failure of that test case?
Do I understand that correctly?

imsuck:
@jaeheonji  I was wondering how did it fail in my PR but not in #1658

jaeheonji:
@imsuck 

> I was wondering how did it fail in my PR but not in https://github.com/zellij-org/zellij/pull/1658

That's weird. Could you please create an issue if the problem persists?

imsuck:
Well all the checks have passed now. That is pretty weird.

PR #1924, opened 2022-11-10
plugins: rework plugin loading

har7an:
~~This PR attempts to recover from a plugin version mismatch for all zellij builtin plugins. We do this with some future modifications in mind that will include making certain parts of the loadout initialization non-blocking, which will improve application performance but prevent us from interacting with the user via stdout/stdin.~~

~~The goal is to "preload" plugins and store them in a cache, which will allow us to greatly improve loading speed for subsequent plugin instantiations.~~

This PR entirely reworks the way plugins are loaded.

## Current State

Previously, when starting a plugin the following actions happened (shortened for readability):

- Check if a plugin exists at the given location
- Load the plugin
- Store the plugin in a cache folder on disk for faster subsequent access
- Check the plugin version
- Start the plugin

This approach had a number of drawbacks:

1. **All plugins are always loaded from disk**: Regardless whether the same plugin has been loaded before, it will always require a disk access. This can lead to all sorts of errors, most recently displayed in #1925.
2. **All plugins are serialized to a cache folder**: In Linux, this is usually found in `~/.cache/zellij` and can grow quite large as users either develop the application and frequently run dev versions of the plugins, or use zellij for a large number of releases
3. **Loading layouts is slow**: This is because loading a layout includes synchronously loading all the plugins inside the layout, too. Since the plugins are loaded from the cache folder on disk, this takes a significant amount of time. Note that e.g. opening a new tab also includes loading a layout.
4. **Internal plugins can grow out-of-sync with the application**: This mostly relates to people developing zellij. This was first handled by placing a `VERSION` file in the plugin folder, which didn't solve issues as expected. It was subsequently improved by introducing plugin version checks, which give the user a clear error description. However, it is still a suboptimal experience.

## Improving the situation

This PR tries to improve the situation in a number of ways.

### Do not store internal plugins to disk

All builtin, core plugins (status-bar, tab-bar, etc.) are now loaded directly from the application binary, where they have been stored before, too. The difference is that previously, during startup, the plugins were dumped to the `PLUGIN_DIR` folder on disk if a difference in the `VERSION` files content made this necessary.

### Introduce a plugin cache

The cache maps a plugins path to the plugins binary (bytes) and is held in memory. When zellij starts, the cache is empty. As plugins are loaded the cache is consulted and if, for a given path, no plugin binary is found, it is loaded from disk instead. This has a number of benefits:

- We do not dump ephemeral artifacts to a cache folder any longer
- Subsequent loading of the same plugin doesn't involve any new disk access anymore
- Plugin loading times are greatly increased

## Breakage introduced in this PR

Since we do no longer dump the builtin plugins to disk (previously termed "automatic asset installation"), the feature flag `disable_automatic_asset_installation` has been removed entirely. Also, we no longer use a `PLUGIN_DIR` to load plugins from and consequently no longer check the contents of this directory.

This will most likely bother application distributors, because the only way to publish the builtin plugins alongside zellij now is to ship a custom default layout that explicitly loads the plugins from a location on the disk. Note, however, that even in this case the builtin plugins are, as of now, still present and part of the application binary.


har7an:
@imsnif I'm sure that we'll have to discuss some of the design decisions I made for developing this one. I'll give you a TL;DR:

- We're no longer reading data from the `PLUGIN_DIR`, it's pretty much ignored now, even for custom plugins. Should I revert that?
- I removed the `disable_automatic_asset_installation` feature, because we're not installing assets to disk any longer. However, I imagine that for distributors it might be interesting to still provide the plugins from some other location. Maybe we can introduce a new feature `bundle_builtin_plugins` which reflects the actual feature better?

Also, other open questions:

- Should we clear leftover default plugins from `PLUGIN_DIR`?
- Should we clear the contents of `~/.cache/zellij`?

And you mentioned you had an idea how to test this feature. Would you mind sharing it with me?

imsnif:
Before I relate to the rest, just because I imagine this will require some work: the distributions must have a way to compile their own default plugins, place them in the PLUGIN_DIR and have them loaded in place of anything else. If they are not there, they should get an error. Ideally this will not require any change from anyone's side. Can we maybe re-use the flag (which they're already using anyway) to make this happen?

imsnif:
Also btw - if we're doing the above, we might want a manual flag to populate (and override) that folder. Will help us troubleshoot with users and solve environment issues. Something like `zellij setup --populate-default-plugin-dir` or some such?

imsnif:
Hell is production code my friend - what can I say?

We *have* to keep the flag, because it would be extremely hard for us to find out who is using it and make sure they fix it. What they are using it for right now is "let me compile my own plugins and error if I didn't do it right". While I do agree with you that the naming here is unfortunate, we're sadly stuck with it. 

I think your solution is good, but do note.
Flag set => no automatic asset installation, meaning we expect plugins to be "manually" installed, load them and explode if they are not there.
Flag not set => we load from within the binary.

I was thinking of the manual populate as an escape hatch solution in case a user accidentally deleted the plugins in a distribution-installed binary. What other options do they have? (reinstalling from their distro is a bad option here, we can't be sure it will fix it and I don't want to have to troubleshoot installation methods I do not own)

har7an:
> I think your solution is good, but do note.

See? I always get it wrong. :)

Here's the current state of affairs. With the feature flag set, but without the plugins in the plugin dir:

```
  ╰─▶ Program terminates: a fatal error occured
      
      Caused by:
          0: failed to start plugin PluginConfig {
                 path: "tab-bar",
                 run: Pane(
                     None,
                 ),
                 _allow_exec_host_cmd: false,
                 location: Zellij(
                     PluginTag(
                         "tab-bar",
                     ),
                 ),
             } for client 1
          1: failed to load a plugin
          2: This version of zellij was built to load the core plugins from
             the globally configured plugin directory. However, a plugin wasn't found:
      
                 Plugin name: 'tab-bar.wasm'
                 Plugin directory: '/var/home/andi/.local/share/zellij/plugins/'
      
             If you're a user:
                 Please report this error to the distributor of your current zellij version
      
             If you're a developer:
                 Either make sure to include the plugins with the application (See feature
                 'disable_automatic_asset_installation'), or make them available in the
                 plugin directory.
      
          3: No such file or directory (os error 2): 'tab-bar.wasm'
          4: No such file or directory (os error 2): 'tab-bar'
          5: No such file or directory (os error 2): '/var/home/andi/.local/share/zellij/plugins/tab-bar.wasm'
          6: failed to load plugin from disk
```

When requesting a plugin via `zellij:` that isn't part of the asset map:

```
  ╰─▶ Program terminates: a fatal error occured
      
      Caused by:
          0: failed to start plugin PluginConfig {
                 path: "tab-bar",
                 run: Pane(
                     None,
                 ),
                 _allow_exec_host_cmd: false,
                 location: Zellij(
                     PluginTag(
                         "tab-bar",
                     ),
                 ),
             } for client 1
          1: failed to load a plugin
          2: It seems you tried to load the following builtin plugin:
      
                 Plugin name: 'tab-bar.wasm'
      
             This is not a builtin plugin known to this version of zellij. If you were using
             a custom layout, please refer to the layout documentation at:
      
                 https://zellij.dev/documentation/creating-a-layout.html#plugin
      
             If you think this is a bug and the plugin is indeed an internal plugin, please
             open an issue on GitHub:
      
                 https://github.com/zellij-org/zellij/issues
          3: No such file or directory (os error 2): 'tab-bar.wasm'
          4: No such file or directory (os error 2): 'tab-bar'
          5: No such file or directory (os error 2): '/var/home/andi/.local/share/zellij/plugins/tab-bar.wasm'
          6: failed to load plugin from disk
```


> I was thinking of the manual populate as an escape hatch solution

I see. Actually this gave me a good idea with respect to restructuring the infamous `asset_map` we carry around in the code, so thanks for that! I'll go with `zellij setup --populate-default-plugin-dir` for now, we can discuss this tomorrow.

har7an:
And is there a way to just kill whatever pipe is used to exchange data with the PTY? Because a plugin error such as that causes a flood of ANSI codes in the terminal.

har7an:
# Current problems:

- When zellij terminates because of this, the error is drowned out by the wall of ANSI sequences sent from the TTY (color codes)
- This makes development difficult, because debug builds of zellij are built with the plugins from `assets/plugins`, which will be out of date...

har7an:
## Current state

Here's what changes with this PR:

### When the `disable_automatic_asset_installation` feature IS NOT set (default)

- Builtin plugins are now loaded directly from the binary and never dumped to disk
- If a builtin plugin is read from the binary but exists on disk, an info message is printed to the log:
```
Plugin 'tab-bar.wasm' exists in the 'PLUGIN DIR' at '/var/home/andi/.local/share/zellij/plugins/' but is being ignored
```
- The output of `zellij setup --check` was extended to hint towards this:
```
[PLUGIN DIR]: "/var/home/andi/.local/share/zellij/plugins"
 Builtin, default plugins will not be loaded from disk.
 Create a custom layout if you require this behavior.
```


### When the `disable_automatic_asset_installation` feature IS set

- Builtin plugins are never loaded from the binary, but only from disk
- If a builtin plugin is not found on disk but present inside the binary, an error is displayed:
```
  ╰─▶ Program terminates: a fatal error occured
      
      Caused by:
          0: failed to start plugin PluginConfig {
                 path: "tab-bar",
                 run: Pane(
                     None,
                 ),
                 _allow_exec_host_cmd: false,
                 location: Zellij(
                     PluginTag(
                         "tab-bar",
                     ),
                 ),
             } for client 1
          1: failed to load a plugin
          2: This version of zellij was built to load the core plugins from
             the globally configured plugin directory. However, a plugin wasn't found:
      
                 Plugin name: 'tab-bar.wasm'
                 Plugin directory: '/var/home/andi/.local/share/zellij/plugins/'
      
             If you're a user:
                 Please report this error to the distributor of your current zellij version
      
             If you're a developer:
                 Either make sure to include the plugins with the application (See feature
                 'disable_automatic_asset_installation'), or make them available in the
                 plugin directory.
      
             Possible fix for your problem:
                 Run `zellij setup --dump-plugins`, and optionally point it to your
                 'DATA DIR', visible in e.g. the output of `zellij setup --check`. Without
                 further arguments, it will use the default 'DATA DIR'.
      
          3: No such file or directory (os error 2): '/var/home/andi/.local/share/zellij/plugins/tab-bar.wasm'
          4: No such file or directory (os error 2): 'tab-bar.wasm'
          5: No such file or directory (os error 2): 'tab-bar'
          6: failed to load plugin from disk
```


### Regardless of the feature setting

- `zellij setup` now has a new option `--dump-plugins`:
```
        --dump-plugins [<DIR>...]
            Dump the builtin plugins to DIR or "DATA DIR" if unspecified
```
- This command behaves as follows:
```bash
❯ ./target/debug/zellij setup --dump-plugins /tmp/pluginzz
Dumping plugins to '/tmp/pluginzz'
# OR
❯ ./target/debug/zellij setup --dump-plugins              
Dumping plugins to '/var/home/har7an/.local/share/zellij'
```
- A hint towards this can also be seen in the error message shown above
- Loading a layout that declares a "builtin" plugin (with the `zellij:` specifier) which doesn't exist in the asset map will now cause a pretty error to appear, although it seems that this is already caugt by the config file handling
- Any plugin is now only read from disk once, when it is first loaded. Version mismatch checks are also only performed once, when initially loading a plugin. All plugins are cached inside the application, and subsequent accesses to the same plugin take the `wasmer::module::Module` directly from this cache.


## Things to discuss

- The `plugin_global_data_dir` (DATA_DIR/plugins/data) was never used, as far as I can tell. It certainly isn't made available to the plugins by means of mounting it inside the wasmer module, either. I think we can remove this, or should we prefer to keep it?


## Final thoughts

The `ZELLIJ_CACHE_DIR` was previously used for storing serialized wasmer modules, as well as allowing per-plugin data storage. With this PR, storing serialized wasmer modules is no longer needed. I'll leave this as is for now, but I'd like to open a separate PR to revise this. In there I would like to:

- Pave the way for allowing per-plugin configurations
- Deal with the cache folder for good, and even delete it entirely if we find no other purpose for it.
- 

@imsnif Sorry to bother you so much, but you know better how to deal with production code. Could you help me out with the points under "Things to discuss"?

har7an:
@imsnif Here is the binary from `cargo make build-e2e`: [zellij.gz](https://github.com/zellij-org/zellij/files/10019880/zellij.gz)

Don't know if that has all the plugins, though... ? If there's something missing please tell me, I'll add it!

imsnif:
@har7an - this is what I get when I run the executable:
```
Error occurred in server:

  × Thread 'wasm' panicked.
  ├─▶ Originating Thread(s)
  │     1. ipc_server: NewClient
  │     2. pty_thread: NewTab
  │     3. screen_thread: NewTab
  │     4. plugin_thread: Load
  │   
  ├─▶ At zellij-server/src/lib.rs:718:92
  ╰─▶ Program terminates: a fatal error occured
      
      Caused by:
          0: failed to load plugin for client 1
          1: failed to start plugin PluginConfig {
                 path: "tab-bar",
                 run: Pane(
                     None,
                 ),
                 _allow_exec_host_cmd: false,
                 location: Zellij(
                     PluginTag(
                         "tab-bar",
                     ),
                 ),
             } for client 1
          2: It seems your version of zellij was built with outdated core plugins.
             Detected versions:
      
             - Plugin version: 0.33.0
             - Zellij version: 0.34.0
             - Offending plugin: tab-bar
      
             If you're a user:
                 to them.
      
             If you're a developer:
                 Please run zellij with updated plugins. The easiest way to achieve this

```

imsnif:
@har7an - I'm doing some bug fixes now and waiting for this to be merged before continuing work on the plugin system. Is there anything I can do to help move this along?

har7an:
> Is there anything I can do to help move this along?

Do you want to have a look at why the e2e tests fail? Or, given your error message, I assume that's the exact reason. I have no idea how the e2e binary is built and how it differs from the regular `cargo make install`. Any pointers there are much appreciated.

imsnif:
Yeah, looks like it. Does the binary you sent me work for you? As far as I understand it, it should be loading plugins from within the binary itself right? If it's getting a mismatch, maybe it's not loading them from the executable after all?

har7an:
> If it's getting a mismatch, maybe it's not loading them from the executable after all?

That might be. But even if I point it to some empty `--data-dir`, I still get the mismatch. It seems like its including outdated assets. Is there a way to increase the log-level when running? I tried `--debug`, but that didn't seem to show `log::debug!` messages. I explicitly added debug logs in the code that tell exactly where a plugin was loaded from, then we'll know for sure.

imsnif:
Maybe I missed things entirely - but isn't the data dir supposed to be irrelevant in this case?

har7an:
> but isn't the data dir supposed to be irrelevant in this case?

It is, that was a test to exclude the possibility that it's loading the outdated plugins from disk where it clearly shouldn't.

imsnif:
Yes, the last executable works for me! What did you change?

har7an:
> Yes, the last executable works for me! What did you change?

Lol, honestly I don't know. I'm doing a cargo clean and rerunning to see if it's reproducible...

imsnif:
Alright... Maybe let's take some time to follow the flow of exactly where the plugins are taken from so we're calmer about this. :)

har7an:
Rebuilding from scratch fails, again. I guess the issue is that I build the asset map with `include_bytes!` on the `CARGO_MANIFEST_DIR`, which is essentially the `assets/` folder in the repo.
Hmmm... Where does the e2e binary take the plugins from? How does it work at the moment? The function that installed the assets to disk used the same `include_bytes!` code, so that can't be it, right? Does zellij rely to read the plugins from disk in e2e tests?

har7an:
Seems to me that `build-e2e` is dependent on `build-plugins-release`, which only compiles the plugins in release mode. `cargo make install` depends on `wasm-opt-plugins`, which in turn is dependent on `build-plugins-release`. Was this made on purpose?

The flow is:

- `make build-e2e --> build-plugins-release`
- `make install --> wasm-opt-plugins --> build-plugins-release`

Clearly the wasm-opt step is missing.

imsnif:
> Clearly the wasm-opt step is missing.

do you think that is the issue?

har7an:
Done, should pass now. You can merge if it does, otherwise I'll do tomorrow. :)

imsnif:
Hey @har7an - I've been poking around this PR and found a bit of a fundamental issue here. I'm hoping we can figure out a solution for it together.

So basically, the asset map is built here: https://github.com/zellij-org/zellij/pull/1924/files#diff-99319e67e4162f376e0b337828087cfedc994493cf71ddd78fdd7d33e4eb438dR33

This is a compile time inclusion. Meaning that we won't even compile without these wasm assets present. I guess this was the original reason we decided to keep the `assets` folder in the git repo.

Up until now, we've only been using this for our release. Essentially, we've been using this every time we compile, but for debug builds run with `cargo make run`, we added the `--data-dir` flag that would point to the precompiled plugins somewhere in the `target` folder. But now, we don't do that anymore and only use the internal assets in the binary.

So two problems:
1. We can't develop plugins - we will always be using the release plugins in the root `assets` folder
2. We can't use the release plugins in the assets folder because they're always 1 version behind and we get a mismatch

A few solutions come to mine, but none of them is good:
1. We keep the assets folder and expand it to have `release` and `dev` subfolders. Then we add some conditional compilation to the `include_bytes` part to take the plugins from the right place. When we release/compile the dev plugins, we get a new set and load them into the binary. In one word: yuck.
2. We move the asset folder completely to the git ignored `target` folder, having `target/release-plugins` and `target/dev-plugins`, then we have a separate crate inside the project that can compile without relying on the `include_bytes` in the first place and is in charge of populating these folders. Very error prone and a bit cumbersome.

I'm hoping you have more ideas :)

imsnif:
Also possibly:
3. We might be able to duckscript our way out of it by having a task inside the makefile. It would go to each plugin, build it and place it in the relevant target folder from point 2. I'm a little tired now to think about the downsides of this though - so let's pick it up tomorrow.

har7an:
> This is a compile time inclusion.

This is true, but it has been before this PR, too. I haven't changed anything fundamental about how the asset map is built, I merely moved it somewhere new.


> 1. We can't develop plugins - we will always be using the release plugins in the root `assets` folder

There are still ways to develop plugins:

1. Build zellij with the `disable_automatic_...` feature: In this case it will load the plugins from the plugins folder. The issue is we don't populate that automatically any longer, but when it isn't present you'll get an error that tells you to run `setup --dump-plugins`, which fixes this. Or, even better if you ask me:
2. Use a custom layout

In fact, 2. is how I've been developing zellij ever since that status bar PR. My layout looks like this:

```kdl
layout {
    pane size=1 borderless=true {
        plugin location="file:///var/home/har7an/repos/3_extras/zellij/target/wasm32-wasi/debug/tab-bar.wasm"
    }
    pane
    pane size=2 borderless=true {
        plugin location="file:///var/home/har7an/repos/3_extras/zellij/target/wasm32-wasi/debug/status-bar.wasm"
    }
}
```

I've called this layout "devel", and IMO it's the most comfortable way to develop zellij in any case. It allows me to run `./target/debug/zellij -l devel`, and doesn't have to bother with any release builds or wasm-opt stuff.




> 2\. We can't use the release plugins in the assets folder because they're always 1 version behind and we get a mismatch

This is true only if we do not rebuild the plugins into the assets folder. `cargo make install` does just that, which is the handywork of the `wasm-opt-plugins` step in `Makefile.toml`. The e2e tests in the CI only fail because the `wasm-opt` binary isn't present. I'm 90% confident that if we add that, it will work.


I like the direction your suggestions are going, but I favor a different solution: We include the plugins from somewhere else depending on whether we're a debug or a release build. Ideally we don't even use the `assets` folder, maybe even the `target/wasm32-wasi/*/` folder will do. I haven't had time to work on this one yet, but I think it's a good idea. What are your thoughts?

har7an:
## Update

Since our last discussion, the following things have changed:

- Reintroduced the `ZELLIJ_CACHE_DIR` to load serialized plugins from. The first load now takes a while, but all subsequent loads and plugin-related interactions are blazing fast.
- Build asset map with plugins from different locations:
    - From `target/wasm32-wasi/debug/*.wasm` in debug builds
    - From `assets/plugins/*.wasm` in release builds (This includes `make install` and `build-release` or `cargo build --release`)
- Modify the e2e github action to install wasm-opt and build the e2e binary like a "regular" release build (w.r.t. plugins)
- Modify `Makefile.toml` to always build plugins first in all runs
    - This has the side-effect that now the plugins are built once for every workspace member.
    - However, apart from generating a few more lines of compile output, it has barely an effect, because the plugins will be compiled after the first time, making subsequent "compiles" very quick.
    - Also, anyone used to building zellij with `cargo build` for devel must now run `cargo make plugins` first


I guess that some of these changes require an update to `CONTRIBUTING.md`, I'll do that soon.

@imsnif Have you gotten round to thinking about all the things/workflows we must test yet?

imsnif:
> @imsnif Have you gotten round to thinking about all the things/workflows we must test yet?

Just briefly so you're not stuck, I sent them on Matrix, but will also paste them here:

# Paths to test
* cargo make run (full compile including plugins)
* cargo run (should still work if we previously installed plugins)
* distribution install (should work by loading plugins from the HD cache, should crash if they are not there)
* normal install (crates.io) - test with dry-install, should get the plugins from the executable
* cargo make install - should get the plugins from within the executable
* release install (eg. generic musl binary) - should get the plugins from within the executable
* e2e tests - should get the plugins from within the executable (this should be the same as the musl install, we just need to make sure it still works)

har7an:
I wrote a handy shell script that performs all the test cases mentioned (except for the "normal install (crates.io)", because that requires all sub-crates to be present in the current version on crates.io already). It's here: [test_plugins.sh.gz](https://github.com/zellij-org/zellij/files/10067923/test_plugins.sh.gz)

I'm very confident the changes work now. I tried running the script, and I also moved some of the compiled binaries to "bare" containers and tried them there. Maybe I'll manage to try on another, different PC later today.

Issue #1922, opened 2022-11-09
Feature: support clipboard to text input


jaeheonji:
Related #1915

The following actions receive text input from the user.
* `PaneNameInput`
* `TabNameInput`
* `SearchInput`

These inputs currently only accept string inputs. And currently, the clipboard input is not considered at all.

However, the internal implementation of the these actions is very similar.
So if clipboard input is implemented, all of them can be applied.

naosense:
seems easy, I would like give it a try if no one take it

jaeheonji:
@naosense Thank you! You can take it if you want :)

naosense:
Excuse me, is your log still works fine? I can't print logs after updating the latest code. Before I can see logs print out in `/tmp/zellij-uid/zellij-log/zellij.log` after I type `cargo make run`

jaeheonji:
@naosense 

I haven't checked yet, but it seems to be related to this [PR](https://github.com/zellij-org/zellij/pull/1898).
Perhaps the location has been changed from the absolute path to another temp_dir.

naosense:
Thank you @jaeheonji .Confirmed, it is indeed caused by this PR

Issue #1915, opened 2022-11-08
Paste text into search bar does noting


Integralist:
**Basic information**

`zellij --version`: 0.32.0
`stty size`: 57 118
`uname -av`: Darwin D3496LK9V7 21.6.0 Darwin Kernel Version 21.6.0: Mon Aug 22 20:19:52 PDT 2022; root:xnu-8020.140.49~2/RELEASE_ARM64_T6000 arm64

List of programs you interact with:
`nvim --version`:  NVIM v0.8.0
`alacritty --version`: 0.10.1

**Further information**
I can search the terminal output using `<Ctrl-s>` and typing in a search phrase, but when prompted to enter the search phrase ("Enter search...") and I try to paste whats in my clipboard using `<Command-v>` (I'm on macOS), nothing gets pasted and I have to manually type it out.

jaeheonji:
fixed #1926

PR #1926, opened 2022-11-12
Support text input from clipboard

naosense:
According to #1922 and #1915  , this pr is mainly to support text input from clipboard. 

Hi @jaeheonji , i'm not sure if my implementation is correct, could you review it?

### DIFF
--- a/zellij-client/src/input_handler.rs
+++ b/zellij-client/src/input_handler.rs
@@ -114,6 +114,24 @@ impl InputHandler {
                                     None,
                                 );
                             }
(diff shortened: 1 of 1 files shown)


jaeheonji (approved):

Integralist:
Amazing! 

When is the next release that will contain this? I see 0.33.0 was just released the other day so I'm not sure if you have a scheduled release cycle or if it's ad-hoc. 

Thanks for fixing this issue ♥️

naosense:
@Integralist 

Yes, the next release will include it, and AFAIK there is no regular release cycle

PR #1927, opened 2022-11-12
Update log path in contributing doc

naosense:
As pr #1898 changes log path from the hardcoded `/temp` to `std::env::temp_dir()`, update contributing.md to avoid unnecessary misleading.


naosense:
Updated documentation about the log path by the way. Could you please review it? @jaeheonji

jaeheonji (approved):
Cool! Thank you 😄

PR #1928, opened 2022-11-12
Fix: cargo clippy lints

har7an:
Fixes all clippy lints introduced in the KDL PR and afterwards. Reenables strict clippy lint checking.

@imsnif Did we already reach a consensus on this and if so, what was it again? I can disable the strict checking, too.

### DIFF
--- a/Makefile.toml
+++ b/Makefile.toml
@@ -83,7 +83,8 @@ args = [
 
 # Simple clippy tweak
 [tasks.clippy]
-args = ["clippy", "--all-targets", "--all-features", "@@split(CARGO_MAKE_TASK_ARGS,;)"]
+args = ["clippy", "--all-targets", "--all-features", "--", "--deny",
+"warnings", "@@split(CARGO_MAKE_TASK_ARGS,;)"]
 
 # Release building and installing Zellij
 [tasks.install]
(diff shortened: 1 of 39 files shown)


imsnif:
Right now we agreed not to do this. Also, please don't merge this - it will cause havoc with unmerged branches.

bheylin:
> Right now we agreed not to do this. Also, please don't merge this - it will cause havoc with unmerged branches.

What is the plan to fix the clippy warnings?
I would imagine that fixing the clippy warnings should be done in each PR to be merge from this point forward.
If a given PR makes changes in four files, then ask the writer to fix the clippy warnings in those files before merging?

This would allow for a gradual improvement of the code and not result is many painful rebases.

imsnif:
Clippy is not a priority for this project at the time being. I would not want to spend time on it that can be spent fixing bugs or developing features. Most of its warnings are far too pedantic for the current state of the project.

So - a hard no from me. :)

bheylin:
Fair enough. I agree that clippy is less of a prio than removing `unwrap()` calls for instance.

imsnif:
Thanks for understanding @bheylin 

About `unwrap`s - sometimes they're okay, sometimes they're bad. I think it's best if we look at this in a higher resolution, more as "Fix crash when <x> - here's an integration test for <x>" than "remove unwrap in this theoretical situation". Hope this makes sense.

I'm closing this for now, anyway.

PR #1934, opened 2022-11-14
refactor(plugins): change the data flow

imsnif:
This is a much needed and long-called-for change to our plugin system infrastructure.

The goal was to change the data flow from:
1. Update Plugin with state
2. Render Zellij 
3. Tell each plugin to render and block until they render
4. Print to STDOUT

To:
1. Update Plugin with state
2. Render Plugin (as part of the update but using the same plugin API to avoid too many breaking changes)
3. Have 2 be interpreted internally by Zellij in `PluginPane` and stored as an internal state (just like we do with `TerminalPane`s)
4. Mark plugin to be re-rendered in the next render-loop

This has a few benefits, including improved performance (the render loop is now completely synchronous - it does not have to block while waiting for each plugin to render itself), reduced writes to STDOUT (since we now interpret the plugin STDOUT rather than just throwing it to the screen in the proper location, we can use the same `OutputBuffer` mechanism we use for terminals in order not to write sequences to the screen that are already there as well as increased functionality for plugins (they can now use the full ANSI range of terminals, opening the door to incorporate more blingy components from terminal UI libraries as plugins).

## Breaking Changes for Plugins
* If plugins exceed their scrollback, the lines will be deleted from the top and not the bottom as they were until now - this is to make the behaviour more consistent with "actual" terminals (lines are pushed into the scrollback rather than cut from below).
    - Probably the most significant change here is that if you did `\n` at the last line of the plugin, you'll need to remove, otherwise you'll get an empty line at the end of the plugin pane.
* Whenever plugins (re-)render, they should completely re-render themselves without considering the previous cursor position or artifacts on screen

## Other changes
* I upgraded wasmer to 2.3.0 (@xJonathanLEI - I hope you'll forgive me doing it in my own PR, I was already in the neighborhood)


xJonathanLEI:
No problem!

zellij-server/src/panes/plugin_pane.rs
```diff
@@ -139,127 +213,86 @@ impl Pane for PluginPane {
...
+                    if !self.borderless {
+                        let frame_output = frame.render().with_context(err_context)?;
+                        self.frame.insert(client_id, frame);
+                        Some(frame_output)
+                    } else {
+                        None
+                    }
+                },
+            };
+            Ok(res)
```
har7an, line 292: I've had a little fun with this one, hope you don't mind. This should be equivalent:

```rust
            self.frame
                .get(&client_id)
                .with_context(|| format!("failed to acquire last frame for client {}", client_id))
                .map_or_else(
                    |_err| !self.borderless,
                    |last_frame| (&frame != last_frame) && !self.borderless,
                )
                .then(|| {
                    let frame_output = frame.render().with_context(err_context)?;
                    self.frame.insert(client_id, frame);
                    Ok(frame_output)
                })
                .transpose()

```

imsnif: cool! Haven't looked at this specifically, but let's do it in a different PR. This bit was copy/pasted from elsewhere in the code - so we will have to change it there too (in terminal_pane.rs iirc)

imsnif: Also, on a side note... are you 100% sure this does the same thing? From a cursory read it seems not to, but I haven't gone deeply into this so maybe I'm wrong.

PR #1956, opened 2022-11-18
Fix crashes in status-bar plugin 

### LINKED ISSUE #1944: Corrupted Statusbar tips cache causes panic
If the file `/tmp/zellij-<uid>/status-bar-tips.cache` is corrupted, zellij won't start up or existing sessions crash when a new tab is opened:
```
Error occurred in server:

  × Thread 'wasm' panicked.
  ├─▶ Originating Thread(s)
  │       1. ipc_server: NewClient
  │       2. pty_thread: NewTab
  │       3. screen_thread: NewTab
  │       4. plugin_thread: Load
  │
  ├─▶ At zellij-server/src/wasm_vm.rs:670:5
  ╰─▶
A panic occured in a plugin
PanicInfo {
    payload: Any { .. },
    message: Some(
        called Result::unwrap() on an Err value: Serde(Error("trailing characters", line: 1, column: 238)),
    ),
    location: Location {
        file: "default-plugins/status-bar/src/tip/utils.rs",
        line: 40,
        col: 83,
    },
    can_unwind: true,
}
```
I experienced the corruption sometimes when start multiple zellij sessions in parallel.

rdeaton:
Turned a number of panics into logged errors that otherwise continue when interacting with the status-bar plugin's tips cache, and made writes to the cache atomic to avoid clobbering seen when running multiple zellij instances.

Fixes #1944

### DIFF
--- a/default-plugins/status-bar/src/tip/cache.rs
+++ b/default-plugins/status-bar/src/tip/cache.rs
@@ -1,9 +1,10 @@
 use std::collections::{HashMap, HashSet};
-use std::fs::{File, OpenOptions};
+use std::fs::OpenOptions;
 use std::io::{self, Read, Write};
-use std::path::PathBuf;
+use std::path::{self, PathBuf};
 
 use serde::{Deserialize, Serialize};
+use tempfile::{NamedTempFile, PersistError};
 use thiserror::Error;
 
 use zellij_tile::prelude::get_zellij_version;
@@ -33,6 +34,8 @@ pub enum LocalCacheError {
     // Deserialization error
     #[error("Deserialization error: {0}")]
     Serde(#[from] serde_json::Error),
+    #[error("PersistError: {0}")]
+    Persist(#[from] PersistError),
 }
 
 impl LocalCache {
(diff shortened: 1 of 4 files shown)


default-plugins/status-bar/src/tip/cache.rs
```diff
@@ -62,29 +65,79 @@ impl LocalCache {
...
+                                "Could not get a parent path for tips cache",
+                            ),
+                            self.path.clone(),
+                        )
+                    })?,
+                )
+                .map_err(LocalCacheError::Io)?;
                 file.write_all(json_cache.as_bytes())
                     .map_err(LocalCacheError::Io)?;
+                file.persist(self.path.as_path())
```
jaeheonji, line 127: Hi @rdeaton, Thank you for contributing!

When I run your code in my environment, I keep getting the following error.
```
DEBUG  |/home/meow/Projects/zelli| 2022-12-20 21:27:49.223 [id: 1     ] Error flushing local cache to disk: Persist(PersistError { error: Os { code: 20, kind: AlreadyExists, message: "File exists" }, file: NamedTempFile("/tmp/.tmpKk1Sqm") }) 
```

Could you check re-again?

PR #1978, opened 2022-11-30
Attempt to minimize search input lag in huge scrolling area

Voronar:
Hello.
When scrolling area is huge (5K+ lines) in case of searched string mismatch the search input is laggy (on 10K it is almost not responsive). According to my brief code investigation it is expensive to do actual scroll&search viewport movements in large scrolling area.
I tried to do some presearch across stringifyed versions of grid rows to prevent actual scroll&search in case of search mismatch.
I know my solution isn't good enough but I gave the idea.

### DIFF
--- a/zellij-server/src/panes/grid.rs
+++ b/zellij-server/src/panes/grid.rs
@@ -326,8 +326,10 @@ fn utf8_mouse_coordinates(column: usize, line: isize) -> Vec<u8> {
 #[derive(Clone)]
 pub struct Grid {
     pub(crate) lines_above: VecDeque<Row>,
+    pub(crate) lines_above_str: VecDeque<String>,
     pub(crate) viewport: Vec<Row>,
     pub(crate) lines_below: Vec<Row>,
+    pub(crate) lines_below_str: Vec<String>,
     horizontal_tabstops: BTreeSet<usize>,
     alternate_screen_state: Option<AlternateScreenState>,
     cursor: Cursor,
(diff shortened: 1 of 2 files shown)


imsnif:
Hey @Voronar - thanks for giving this a look.

I must admit I am not reproducing this output at all. My search input is super fast with the maximum scroll buffer and varied input (my test is to do `dmesg` and search for the word "error"). I'm also working on bare xterm, so not even a GPU accelerated terminal.

Could this be an environmental issue?

Voronar:
@imsnif
I tried on every terminal emulator on my Ubuntu 22.04 (gnome-terminal, alacritty, framebuffer).
Try to execute `dmesg` until scroll buffer reaches its maximum size (~11K). Then enter search mode and continue typing (fastly) random characters after `error` is typed.

P.S. My hardware configuration is i7 1255P, 16GB RAM, NVMe disk.

imsnif:
Aha, now I see what you mean!

Looks like this only affects misses though, right? Seems like we keep searching the whole scrollback for every keystroke. Maybe we can solve it with some caching instead? wdyt?

Voronar:
What kind of caching do you mean?

In my approach I do some kind of caching to make presearch (on stringified version of `Row`s) to prevent actual heavy scroll & search.
I stringify every `Row` during viewport change or scrolling, but I can do it only one time entering search mode.

imsnif:
> What kind of caching do you mean?

So, this only happens on cache misses, right? This means that we have some sort of caching for cache hits and we need to replicate the same behaviour for misses. Namely - if you searched for "errorj" and didn't find it, there's no reason to keep searching if you add another `j`, yeah? I think that's the general idea. I'd read over the search code and see how this is handled.

PR #1990, opened 2022-12-05
Reimplement resize code

har7an:
This PR reimplements the code for resizing tiled and floating panes and adds new resizing methods along the way.

## Scope of this PR

It started out removing more `unwrap` statements from the code in `server::panes`. I realized that much of the code for resizing panes is duplicated and changes only in minor aspects (using x or y coordinates, etc.). I decided to rewrite/merge the individual functions to reduce duplicate code and get a better grasp on how resizing works. While doing so I introduced a new data type (the new `Resize` action) to facilitate this work.

## New features

This PR adds the ability to actively *decrease* pane sizes. Previously, resizing a pane would always increase that panes area, unless the border in the direction being resized hit the viewport boundary, in which case the opposite border was moved (reducing pane area).

Now, a "reducing" resize can be bound to keys, like so:

```kdl
        bind "h" "Left" { Resize "Increase Left"; }
        bind "j" "Down" { Resize "Increase Down"; }
        bind "k" "Up" { Resize "Increase Up"; }
        bind "l" "Right" { Resize "Increase Right"; }
        bind "H" { Resize "Decrease Left"; }
        bind "J" { Resize "Decrease Down"; }
        bind "K" { Resize "Decrease Up"; }
        bind "L" { Resize "Decrease Right"; }
        bind "+" { Resize "Increase"; }
        bind "-" { Resize "Decrease"; }
```

and is also accessible from the CLI:

```
$ zellij action resize increase left
```

Given the keybinding above and e.g. a floating pane, one can now:

- press `h` to move the left border outwards (increase area)
- press `H` to move the left border inwards (decrease area)
- etc.

### UI changes

From the users perspective almost nothing changed. The hints in the status bar have been renamed to reflect the new behavior, the documentation has been modified where appropriate. Care has been taken to **maintain compatibility with existing configs**. Hence, a keybinding:

```kdl
        bind "h" "Left" { Resize "Increase Left"; }
```

does exactly the same as

```kdl
        bind "h" "Left" { Resize "Left"; }
```

because "Increase" is the default which is assumed when nothing else is specified. Previous resize behavior, where an "increasing" resize towards a viewport boundary decreases the panes area from the other side, is maintained, too.

**Breaking change**: The generic increase/decrease (without a direction specified) will now increase/decrease tiled panes into **every** direction unlike previously, where a number of preferred directions was tried in order until the first one worked. I'm not sure whether this is a bad thing, personally I prefer the new resize behavior. I'm open to discuss this, however.

### From a developers perspective

The `Resize` type has been extended to allow for greater flexibility and extensibility in the future. The resizing code has been streamlined and was compacted into fewer functions.

## Current quirks

Some resize operations shift neighboring panes positions around slightly. I assume this is a side-effect from the code for "solving" pane layouts but haven't extensively tested this yet. This is likely the reason why some of the server tests fail, too.


imsnif:
Hey @har7an - very happy to see this!

I haven't looked at the code, but a quick note: not sure if you were planning on doing this, but can we please also add this to the default config?

har7an:
> can we please also add this to the default config?

Totally, I just wanted to discuss sensible options for the default keybindings before doing so. I like `hjkl`/`HJKL`, because I do not use the arrow keys at all. I don't know how other users feel about this. Iirc it's not possible to combine the arrow keys with modifiers, right? Do you (or anyone reading this) have a different default keybinding in mind?

imsnif:
How about `<arrow> | Alt+<arrows>`? We know they all work (except for the poor mac users whom we need to find a solution for anyway)

har7an:
Here's the current state: [zellij.zip](https://github.com/zellij-org/zellij/files/10173836/zellij.zip)

Best enjoyed with `./zellij setup --clean` to make it load a new default config that includes the new behavior. Regarding the keybindings: I can't think of anything to make with the arrow keys. Binding to `Alt + arrows` is possible, but prevents moving to different panes without leaving resize mode. That's a pretty severe penalty in my personal opinion, but we can still ship it that way, of course.

har7an:
@imsnif I'm going with the following keybindings now:

- hjkl/Arrows increase size in the given direction (as before)
- HJKL decrease size from the given direction

I found no keybinding to bind decreasing resizes to that includes the arrow keys. Is that ok?

imsnif:
@har7an why not use `Alt`? We already know it works both with hjkl and arrow keys

har7an:
> why not use `Alt`?

Because it conflicts with the Alt+Arrow keybindings for movement. I don't think this is a good idea for a number of reasons:

1. It will likely break peoples workflows. Resizing multiple panes requires entering resize mode, resizing, leaving resize mode, moving focus, entering resize mode, ....
2. It makes the config more complicated in the process

Regarding point 2: This is the current config:

```kdl
    shared_except "locked" {
        // ...
        bind "Alt l" "Alt Right" { MoveFocusOrTab "Right"; }
        bind "Alt j" "Alt Down" { MoveFocus "Down"; }
        bind "Alt k" "Alt Up" { MoveFocus "Up"; }
    }
```

We'll have to change it to:

```kdl
    shared_except "locked" "resize" {
        // ...
    }
```

and then add the "basic" keybindings back into resize mode manually. Or am I missing something? In this case I'd prefer to announce this new feature in the changelog, maybe add a comment with the `Alt+Arrow` bindings into the config and explain the situation there. I suggest we keep the HJKL bindings regardless, but don't do anything with the arrow keys.

imsnif:
Well argued @har7an - I'm convinced. :)

PR #1997, opened 2022-12-07
feat(ux): show loading screen

imsnif:
When Zellij loads, it queries the terminal emulator for its default colors and its pixel to cell ratio. Since it gets the responses over STDIN, we must drop all user input in the time that this query happens - otherwise it will get mixed up. For that reason the first 0.5 seconds after Zellij loads, input gets lots.

We can't do anything about that, but this PR communicates this to the user, showing a brief loading prompt while this happens so they know not to start typing.

img-2022-12-07-141118

### DIFF
--- a/zellij-client/src/input_handler.rs
+++ b/zellij-client/src/input_handler.rs
@@ -147,6 +147,20 @@ impl InputHandler {
                         self.handle_stdin_ansi_instruction(ansi_instruction);
                     }
                 },
+                Ok((InputInstruction::StartedParsing, _error_context)) => {
+                    log::info!("sending done loading");
+                    self.send_client_instructions
+                        .send(ClientInstruction::StartedParsingStdinQuery)
+                        .unwrap();
+                    log::info!("done sent done loading");
+                }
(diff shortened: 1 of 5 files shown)


necauqua:
Hello!
Can't you like.. cache this? Or add an option to do so, again (also maybe reset the cache if TERM changes or something..)

0.5 seconds on _every_ load is way too much, and I know that opening a bunch of terminal emulators is a flow that zellij is supposed to solve, but people (including myself) do that anyway - and swallowing 0.5 seconds of user input is also something that I kind of noticed previously, but as #2024 points out the flash of the screen (on every new terminal) is way more jarring, the snappinness - even if it was apparent because of the lost input - is gone

imsnif:
Hey @necauqua - the idea of caching also came up in a separate discussion on Matrix. The issue there was that it's hard to know when to cache-bust. But in the end we decided that the UX of leaving cache busting to the user (eg. delete this file and reload Zellij) is far better than the delay we have now and the issues with snappiness as you say.

I'm going to try to hack a quick solution together for the next version. Thank you for your patience!

Issue #1929, opened 2022-11-12
`cli_send_*` tests in `zellij_server` fail when running on too many threads


har7an:
I noticed that the tests in `zellij_server`, when run from `cargo make`, started failing on my machine with one of the recent changes in `main`. I haven't yet bothered to track down the exact change that introduced this.

The symptom for me is that the following 16 tests inside `zellij_server` fail:

```
screen::screen_tests::send_cli_close_pane_action
screen::screen_tests::send_cli_close_tab_action
screen::screen_tests::send_cli_edit_action_with_default_parameters
screen::screen_tests::send_cli_edit_action_with_line_number
screen::screen_tests::send_cli_edit_action_with_split_direction
screen::screen_tests::send_cli_edit_scrollback_action
screen::screen_tests::send_cli_focus_next_pane_action
screen::screen_tests::send_cli_focus_previous_pane_action
screen::screen_tests::send_cli_goto_tab_action
screen::screen_tests::send_cli_half_page_scroll_down_action
screen::screen_tests::send_cli_half_page_scroll_up_action
screen::screen_tests::send_cli_move_focus_or_tab_pane_action
screen::screen_tests::send_cli_move_focus_pane_action
screen::screen_tests::send_cli_move_pane_action
screen::screen_tests::send_cli_new_pane_action_with_command_and_cwd
screen::screen_tests::send_cli_new_pane_action_with_default_parameters
```

When limiting the number of CPU threads to use with e.g. `taskset`, less tests will fail. When giving 6 threads to `cargo make`, all tests pass on my machine. This also explains why the CI doesn't fail, because IIRC that only has 3 threads available for x86 Ubuntu workers. The command I have to use is:

```bash
taskset -c 0-5 cargo make ...
```

As far as I could determine, this doesn't seem to be an error from shared global state. However, all these tests use `insta` to compare snapshots, and I don't know how insta works, so it may well be caused by races in global states of some sort. I tried the following to alleviate race conditions:

- Use `rusty_fork` to fork out the tests into a more isolated environment
- Add a global `Mutex` to the tests so simulating the sending from CLI doesn't happen in parallel any longer

None of these made the problems go away. One solution would be to limit the number of test threads in `cargo make`, but I'm wondering whether I'm the only person experiencing this? I also tried `cargo clean` repeatedly in case it was caused by some old artifacts messing with the tests, but that hasn't changed the situation for me.

imsnif:
What are the failures you're getting?

[...]

PR #2012, opened 2022-12-11
WIP: Use `xtask` as build system

har7an:
Moves the build system away from `cargo make` to `cargo xtask`.

## Rationale

`cargo make` isn't an ideal buildsystem for zellij for a number of reasons:

- Tasks by default run on each member of a workspace, member by member. Because we need to have the plugins built first (with the recent rework of the internal asset map in `utils`) this means that in fact we build the plugins about 10 times in a single `cargo build`
- `cargo make` produces a lot of `cargo make`-specific output for little program output. This often makes it hard to follow what task exactly failed and on which workspace member
- It is comparatively slow
- It doesn't give a sensible overview of the available tasks to run

Also, the recent release showed that clearly we hit some boundary with `cargo make` and the way we're using it. So it is probably time to think about an alternative.

## Why `xtask`

`cargo xtask` isn't so much a tool on its own as a member of the project workspace. It is a subcrate inside the project, which runs project-specific tasks. This has a great benefit: **It needs no additional dependencies**. `cargo xtask` is compiled as a rust binary and then parses all other arguments.

It is neither a recommended workflow, not a de-facto standard. The "website" can be found here: https://github.com/matklad/cargo-xtask

It is used by other influential rust projects, such as [rust-analyzer](https://github.com/rust-lang/rust-analyzer). The benefit for zellij is that we can customize it more easily and adapt it to our changing requirements. This is especially important since with the plugin system we need to handle some added complexity otherwise missing in "regular" rust projects.

## Current state

There are still a few things compared to `cargo make` that it isn't capable of (yet):

- Accept arbitrary arguments: This is mostly interesting for `cargo make run` or `cargo make test`, where developers can currently pass any additional CLI args at the end, which are passed through to the application. I'm still looking for a way to do this...
- The `cargo make publish` flow

Also I'd like to add another cargo alias/deprecation warning for users still using `cargo make` to inform them of the new build system.


har7an:
@imsnif Here's my xtask proposal PR. It's not entirely done yet (see description above). If you like, you can try and test it: It's already capable of most things (I think) we need. The available commands can be found with `cargo xtask --help`. Also, unlike `cargo make`, you can call it from *anywhere* in the repository. ;)

If you think there's something missing I haven't covered above, please let me know!

Curious to hear if you spot the "bonus" I added in terms of telling which command/task is currently being run...

imsnif:
@har7an - I've got a lot on my plate atm. Would be happy to give some specific feedback if you have any questions or are unsure of something? Otherwise I'd rather give it all a thorough look when you're positive it's good, finished and does everything it needs to.

har7an:
@imsnif It's done now. As far as I can tell it has full feature-parity with previous `cargo make`. I even added the `publish` pipeline already with some added features/benefits. I hope you like it and I'm looking forward to your feedback!

imsnif:
Hey @har7an - this already looks great! Looking forward to using this. A few comments:

1. It's important that we be able to pass arguments to commands. Eg. I should be able to do `cargo xtask run --layout /path/to/layout/file.kdl` or even `cargo xtask run -- --layout /path/to/layout/file.kdl`
2. Would be cool if we have some sort of "migration document" that people can use to quickly find out how to move from the old commands to the new ones. Better yet, alias all the old commands to error messages that will explain the migration and tell the user exactly what to do: install xtask <link>, run the following command: cargo xtask <your stuff here>
3. With `cargo make` I've always had the problem that `cargo make run` is super slow. Even when there's no change to any crate, it still goes through all the crates one by one, tries to compile them and then (for some reason) sleeps for almost 2 seconds before moving to the next crate. `cargo xtask run` does the exact same thing. With the plugins it can't be helped, but I wonder why this happens with all the workspace crates... In any case, if I'm not working on the plugins I usually do `cargo run -- --data-dir target/dev-data` - would be cool if we had this as an xtask task... maybe `cargo xtask quickrun`?

har7an:
> 1. It's important that we be able to pass arguments to commands.

Good catch, I forgot a `--` in the code. Thanks!


> 2\. Would be cool if we have some sort of "migration document" that people can use to quickly find out how to move from the old commands to the new ones.

Have you tried running `cargo make` with this build?


> 3\. would be cool if we had this as an xtask task...

I see. I added a `--data-dir` argument to `xtask run`, which, when present, will run zellij with the `disable_automatic_asset_installation` feature and pass `--data-dir` to the binary being run, too. Is that what you had in mind?



Oh btw, great "feature" about xtask (although irrelevant for you): #1929 doesn't occur any longer. :)

PR #2013, opened 2022-12-12
Fix: Make sure to get full path from SHELL env

### LINKED ISSUE #2006: Thread 'pty' panicked: Cannot find shell zsh
Please attach the files that were created in `/tmp/zellij-1000/zellij-log/` to the extent you are comfortable with.
```
INFO   |zellij_client            | 2022-12-10 03:13:45.446 [main      ] [zellij-client/src/lib.rs:140]: Starting Zellij client! 
INFO   |zellij_server            | 2022-12-10 03:13:45.447 [main      ] [zellij-server/src/lib.rs:213]: Starting Zellij server! 
INFO   |zellij_server::plugins   | 2022-12-10 03:13:45.499 [wasm      ] [zellij-server/src/plugins/mod.rs:61]: Wasm main thread starts 
INFO   |zellij_client::input_hand| 2022-12-10 03:13:45.499 [input_handler] [zellij-client/src/input_handler.rs:151]: sending done loading 
INFO   |zellij_client::input_hand| 2022-12-10 03:13:45.499 [input_handler] [zellij-client/src/input_handler.rs:155]: done sent done loading 
INFO   |zellij_utils::input::plug| 2022-12-10 03:13:45.500 [wasm      ] [zellij-utils/src/input/plugins.rs:139]: Plugin 'compact-bar.wasm' exists in the 'PLUGIN DIR' at '/home/ndavd/.local/share/zellij/plugins/' but is being ignored 
ERROR  |zellij_utils::errors::not| 2022-12-10 03:13:45.503 [pty       ] [zellij-utils/src/errors.rs:545]: Panic occured:
             thread: pty
             location: At zellij-server/src/pty.rs:452:13
             message: Cannot find shell zsh 
INFO   |zellij_server::plugins   | 2022-12-10 03:13:45.503 [wasm      ] [zellij-server/src/plugins/mod.rs:123]: wasm main thread exits 
INFO   |zellij_client::input_hand| 2022-12-10 03:13:46.000 [input_handler] [zellij-client/src/input_handler.rs:158]: sending done loading 
INFO   |zellij_client::input_hand| 2022-12-10 03:13:46.000 [input_handler] [zellij-client/src/input_handler.rs:162]: done sent done loading 
```

**Basic information**

*zellij --version*: `zellij 0.34.3`
*stty size*: `26 106`
*OS*: `Arch Linux 6.0.11-arch1-1`

**Further information**
Issue occurs just by running `zellij`. It's fixed by running `zellij options --default-shell zsh`.

image

ndavd:
Fixes #2006 by using my proposed solution https://github.com/zellij-org/zellij/issues/2006#issuecomment-1345299570

> Upon inspecting https://github.com/zellij-org/zellij/blame/main/zellij-server/src/pty.rs#L451 I noticed that the reason why this is happening is because it would need to take the absolute path of the shell. My `$SHELL` equals `zsh`, when checking if a path from that exists it will return false. Now, we could just check it and set it to the fallback `/bin/sh` instead.. But it's not only my system that does this, a far better solution would be to perhaps use the `which` command to get the full path of the `$SHELL` output and return that instead.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3752,13 +3752,13 @@ dependencies = [
 
 [[package]]
 name = "which"
-version = "4.2.5"
+version = "4.3.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "5c4fb54e6113b6a8772ee41c3404fb0301ac79604489467e0a9ce1f3e97c24ae"
+checksum = "1c831fbbee9e129a8cf93e7747a82da9d95ba8e16621cae60ec2cdc849bacb7b"
 dependencies = [
  "either",
- "lazy_static",
  "libc",
+ "once_cell",
 ]
 
 [[package]]
@@ -3946,6 +3946,7 @@ dependencies = [
  "uuid",
  "wasmer",
  "wasmer-wasi",
+ "which",
  "zellij-utils",
 ]
 
--- a/zellij-server/Cargo.toml
+++ b/zellij-server/Cargo.toml
@@ -32,6 +32,7 @@ sixel-image = "0.1.0"
 arrayvec = "0.7.2"
 uuid = { version = "0.8.2", features = ["serde", "v4"] }
 semver = "0.11.0"
+which = "4.3.0"
 
 [dev-dependencies]
 insta = "1.6.0"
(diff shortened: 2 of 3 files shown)


imsnif:
Hey @ndavd - while I appreciate the work you've put into this, as I mentioned in my comment in the original issue, I think the best way would be to remove this conditional. We do essentially this same logic you added a little later down the line when we actually open the pane, and if there's a problem we recover from it gracefully and show an error to the user.

If you remove the if entirely (the one linked in my comment), does it work for you?

ndavd:
Oh I see, that does work for me, yes. I updated the code.

ndavd:
Yup, done!

PR #2020, opened 2022-12-13
server/pty: Crash if shell doesn't exist

har7an:
Supplement to #2013, this makes sure that when zellij is being spawned but there is not valid shell found, the application crashes with an error message.


har7an:
@imsnif I know you're busy, but this needs attention before the release. Since #2013 reverts the fix implemented #1769, spawning an instance of zellij with a non-existent shell (however it was configured) now makes the application non-responsive again. I think I found the location where the `CommandNotFound` wasn't handled correctly, but I'm pretty new to this part of the code. Could you give it a quick look (in `pty.rs`) to make sure I understood this correctly? Essentially my assumption is that the `None` branch is equivalent to "Just spawn a new shell here please". Is this correct, or are there other cases where this code is being run?

imsnif:
Hey @har7an - what's the status here?

har7an:
@imsnif Well I tried to rebase onto main but the portion of code where I applied this doesn't exist where I expect it any longer, and honestly I couldn't find out where it went either. So I'll close this PR since whoever does this will have to start from scratch anyway. :)

PR #2049, opened 2022-12-26
feat(#2040): rudimentary background session creation

y-mx-b:
Currently, the screen will be cleared upon opening a session with the `--background` flag, but attaching the session and then detaching it will result everything but the attach and detach commands being shown again; the terminal essentially "forgets" about those two and then redraws the previously cleared screen. 

I'm not very familiar with ANSI escape codes or other terminal shenanigans, so I'd have to take a closer look at the `start_client_impl()` function in order to figure this out. If anyone could help me with this, that'd be great.


sploders101:
I haven't taken a lot of time to understand the problem, but your description sounds like this could be it. Check out the "Normal" and "Alternative" screen buffers in these docs. I think what may be happening is you're entering the "Alternative" screen buffer, so you'll need to keep Zellij from outputting those sequences when using the `--background` flag on startup

http://www.xfree86.org/4.7.0/ctlseqs.html

y-mx-b:
Alright, that sounds promising. Thanks for the docs, I'll take a look at this later.

imsnif:
Hey, what's the status here?

y-mx-b:
> Hey, what's the status here?

I fixed the bug I mentioned before and haven't run into any obvious issues since.

y-mx-b:
Huh, that's odd. I see all tests passing on my M1 MacBook.

imsnif:
Hey @y-mx-b - thanks for your patience on this, and for picking up this issue.

We currently already have a facility to bypass all the screen clearing so that we won't have to do all the hard-work manually. We call is the `cli_client` (https://github.com/zellij-org/zellij/blob/main/zellij-client/src/cli_client.rs) and use it for interpreting cli actions. Maybe we can also use it in this case?

y-mx-b:
No problem @imsnif, it's for my own benefit anyway. 

`cli_client` looks pretty helpful, is it being used in `start_client` in the current master branch? Because if it is, I could just try and update my fork with the current master.

imsnif:
Hey @y-mx-b - what's the status here? (no pressure, I just want to make sure I'm not blocking this through some misunderstanding)

y-mx-b:
@imsnif I haven't really taken a look at `cli_client`, since I've been focusing on other things instead. The feature (background session creation) itself was complete, but it looks like there are merge conflicts I need to resolve now, as I haven't kept my fork up to date.

y-mx-b:
I've pulled from the main repo on my fork and resolved merge conflicts. Formatting and testing has been done, and I've tested out the `--background` flag locally to make sure it still works.

imsnif:
Hey @y-mx-b - I'm a little confused here - would you like to give this a go through the CLI client as mentioned, or?

y-mx-b:
Forgot I had this PR. Closing because this feature was implemented in https://github.com/zellij-org/zellij/pull/3257.

Issue #2063, opened 2023-01-03
Scrolling (too quickly) causes zellij session to be detached and it must be killed to continue.


rob-p:
First, thank you for `zellij`; it's great!

The problem I am having is that when I have a terminal with a long history (i.e. a lot of text), and I scroll up too quickly, I get booted out of my zellij session.  After this, I am unable to reattach it, and I can't even start a new session until I kill the existing sessions.  I can see the session out of which I have been booted, and it shows up in `ps` with the status `Sl`.  I tried e.g. sending it a `CONT` signal, but to no avail.

In the cases where this has happened to me (it's occurred several times), I was scrolling up using the mouse, and each time I was scrolling quickly. It doesn't seem to happen when I scroll slowly, or when I scroll back using the keyboard bindings.

Also, perhaps strangely, processes attached to the unconnectable session are still going (I had a jupyter notebook in one tab, and I was still able to interact with the kernel, even though I couldn't reattach the zellij session from which it was launched).

**Basic information**

`zellij --version`: zellij 0.34.3
`stty size`: 72 258
`uname -av` or `ver`(Windows): Linux newton 5.4.0-132-generic #148-Ubuntu SMP Mon Oct 17 16:02:06 UTC 2022 x86_64 x86_64 x86_64 GNU/Linux

imsnif:
As mentioned on Discord, sorry for the experience and thanks for looking into this! Did you get some sort of error message when it happened? Did you see something in the logs (by default they're in /tmp/zellij-<uid>/zellij-log/zellij.log)?

rob-p:
Thanks for the super-quick response. I did not get any message when this happened, I was just booted back to the terminal from which I launched `Zellij`.  My `/tmp` was cleaned since this last happened, so I don't have that log, but if it occurs again, I'll capture any log output ASAP.

rob-p:
Ok, it happened again!  I am attaching the tail of the log below.

[zellij_tail.log](https://github.com/zellij-org/zellij/files/10338127/zellij_tail.log)

imsnif:
Hum... nothing useful in the log, I'm afraid. Is there a way you can consistently reproduce this? Even semi-consistently? If so I can try to give you a version with some extra logging or some such.

Also, when you got booted to your terminal, did you see any message? Even just the "Bye from Zellij" prompt?

rob-p:
Hi @imsnif,

[...]

Issue #2048, opened 2022-12-26
Sixel images distorted; sixel images have too much bottom padding; sixel causes buffer full error


jessebot:
<details>
  <summary>💌 Thank you note (that has nothing to do with the bug)</summary>
  
Thank you for working on zellij!  💙 I have been using it daily for only a week and it's already totally replacing tmux for me (and I'd been using screen/tmux since 2012!). I do a lot of screen sharing from my mac to a linux machine using a projector while I'm coding, so that my partner and I can hack on cool projects together in vim/neovim. They can then also jump into the same session when we're done rubberducking.

This tool has totally changed how we work together and I promise when I have time, I will post some pictures of just how cool and hacker aesthetic this looks when we have two tiny screens and one major screen running zellij sessions to manage everything from spotify to email to small python projects to active development of larger scale kubernetes projects :) Being able to have beautiful tabbing, floating windows, and fast splits, really means a lot to our productivity, so thank you so much. I will try to sponsor when I am not also a poor dev.

</details>

I'm having some issues with sixel support, mostly happening when scrolling after displaying an image somewhere. The image prints okay, but with a lot of padding on the bottom of empty space, and then you scroll, it breaks up, sometimes over writing the rest of the screen buffer and making it unreadable. If it gets too bad, zellij will crash. This happens on iterm2 (macOS) and wezterm (Ubuntu). The problem is particularly pronounced when using w3m, but also happens just using img2sixel directly on the command line. (in the screenshots/video below, `sixel` is an alias to `img2sixel`)

<img width="1422" alt="Screenshot 2022-12-26 at 11 39 16" src="https://user-images.githubusercontent.com/2389292/209539727-f818602a-7099-4ae2-b129-4390592cbcb3.png">

<img width="1440" alt="Screenshot 2022-12-26 at 11 36 48" src="https://user-images.githubusercontent.com/2389292/209539309-cfb50b63-33d5-4d9c-b12f-425a040d224a.png">

**In Case of Graphical or Performance Issues**

1. _Delete the contents of `/tmp/zellij-1000/zellij-log`, ie with `cd /tmp/zellij-1000/` and `rm -fr zellij-log/`
2. Run `zellij --debug`
3. Run `stty size`, copy the result and attach it in the bug report
4. Recreate your issue.
5. Quit Zellij immediately with ctrl-q (your bug should ideally still be visible on screen)

[...]

Issue #2041, opened 2022-12-22
Server: Buffer full


ovove:
Staring with Zellij 0.34.4 I get occasional crashes. I am unfortunately not able to come up with a reproducible way recreate it.
Whenever I get the crash, the only thing I see is this message

Error occurred in server:
Buffer full

**Basic information**

`zellij --version`: 0.34.4
`stty size`: 46 178
`uname -av`: Darwin oMac 22.2.0 Darwin Kernel Version 22.2.0: Fri Nov 11 02:08:47 PST 2022; root:xnu-8792.61.2~4/RELEASE_X86_64 x86_64

List of programs you interact with as, `PROGRAM --version`: output cropped meaningful, for example:
`zsh --version`: zsh 5.8.1 (x86_64-apple-darwin22.0)
`emacs --version`: GNU Emacs 28.2
`clang --version`: clang version 15.0.6

**Further information**
n/a

raphCode:
This usually happens when the client stalls and can't consume all the messages the server sent in a timely matter. Can this be related to high system load in your case?

Luckily you can reattach to your session afterwards.

ovove:
Not so mure about the load at the time of the crash. I have an 8 core macbook air, that I mostly use for some simple development during my free time. Nothing serious though. I'd be surprised if the load average goes much higher than 8-ish. But for sure, I can try to keep a close I to the load average and see if I can correlate it to any zellij crashes,

imsnif:
Also, which terminal emulator are you using?

ovove:
I'm using iTerm2

danillos:
It is happening to me too. `Iterm2` + `Helix`

aniketd:
This potentially causes loss of done work. I have seen this happen twice within the last 2 days.
```
uname -av: Linux nixos 5.15.85 #1-NixOS SMP Wed Dec 21 16:36:38 UTC 2022 x86_64 GNU/Linux
 terminal: Gnome Console
    panes:
      - 1 helix-editor
      - 4 other zsh sessions running programs that can generate a lot of output very quickly.
```

imsnif:
Until I issue a fix for this (need to investigate a little further) I can recommend as a workaround to use a GPU accelerated terminal. This likely happens because the terminal emulator doesn't read STDOUT fast enough, Using something like Alacritty should help a lot.

fbernier:
I've also been experiencing this on my ryzen desktop since the last release.

using:
- fully updated arch linux
[...]

Issue #2023, opened 2022-12-14
exa tree command crush


digika-ding:
Thank you for taking the time to file this issue! Please follow the instructions and fill in the missing parts below the instructions, if it is meaningful. Try to be brief and concise.

**In Case of Graphical or Performance Issues**

1. Delete the contents of `/tmp/zellij-1000/zellij-log`, ie with `cd /tmp/zellij-1000/` and `rm -fr zellij-log/`
2. Run `zellij --debug`
3. Run `stty size`, copy the result and attach it in the bug report
4. Recreate your issue.
5. Quit Zellij immediately with ctrl-q (your bug should ideally still be visible on screen)

Please attach the files that were created in `/tmp/zellij-1000/zellij-log/` to the extent you are comfortable with.

**Basic information**

`zellij --version`: zellij 0.34.3
`stty size`: 94 203
`uname -av` or `ver`(Windows): Darwin WeibingacStudio 22.1.0 Darwin Kernel Version 22.1.0: Sun Oct  9 20:15:09 PDT 2022; root:xnu-8792.41.9~2/RELEASE_ARM64_T6000 arm64

List of programs you interact with as, `PROGRAM --version`: output cropped meaningful, for example:
`exa --version`: v0.10.1 [+git]
`kitty --version`: kitty 0.26.5 created by Kovid Goyal

**Further information**
Reproduction steps, noticeable behavior, related issues, etc
`exa --long --header --icons --sort=modified --group-directories-first --time-style=long-iso --group --created --modified --tree` create a large output

In my case zellij-log were not created.
The output was **Error occurred in server: Buffer full**

imsnif:
This happens when the terminal emulator (Kitty, in this case) takes too long to read from STDOUT and Zellij's buffer gets filled. Note that the session itself isn't lost and you can manually attach back to it. I want to add a feature that does this automatically, but there's not a lot more we can do about it.

If this happens often, I'd recommend trying another terminal emulator - Alacritty is very fast.

PR #2068, opened 2023-01-08
Improve client disconnect handling

har7an:
This PR includes a few changes to the behavior when the session server disconnects a client as reaction to the client not handling messages quick enough. This happens for example when the system is under high load or the terminal application is blocked by some sort of user interaction (hence, keeping it from handling server messages). The server will continue sending `ServerToClientMsg` messages to all connected clients via an IPC channel. Currently, if the channel fills up the client is disconnected and a "Buffer full" error message is displayed to the user.

Now, we have a dynamically sized retry queue that stores messages that do not fit into the IPC channel. At the moment the queue is limited to 1000 messages, with the IPC channel itself holding at most 50 messages. This should drastically decrease disconnections caused by temporary flooding of STDOUT or high system load.

The retry queue is implemented as vector and allocated dynamically. Hence, on clients that are quick enough and never saw a "Buffer full" error before, it will not take any extra space. Slower clients will allocate the retry queue as needed.

In addition, when the IPC channel is full and messages are stored in the retry queue for the first time, a warning is now logged:

```
WARN   |zellij_server::os_input_o| 2023-01-08 08:33:48.696 [main      ] [zellij-server/src/os_input_output.rs:382]: client 1 is processing server messages too slow
```

If the client manages to catch up, a corresponding message is logged, too:

```
INFO   |zellij_server::os_input_o| 2023-01-08 08:43:23.394 [main      ] [zellij-server/src/os_input_output.rs:407]: client 1 caught up processing server messages
```

If the retry queue fills up the client is still forcefully disconnected, but logging on the server has now been improved:

```
ERROR  |???                      | 2023-01-08 08:34:04.220 [main      ] [zellij-server/src/lib.rs:572]: a non-fatal error occured

Caused by:
    0: client 1 is processing server messages too slow
    1: failed to send message to client 1
    2: failed to send or buffer message for client 1
    3: Client 1 is too slow to handle incoming messages
```

And instead of "Buffer full", the client will now see the following message on their terminal:

```
Your zellij client lost connection to the zellij server.

As a safety measure, you have been disconnected from the current zellij session.
However, the session should still exist and none of your data should be lost.

This usually means that your terminal didn't process server messages quick
enough. Maybe your system is currently under high load, or your terminal
isn't performant enough.

There are a few things you can try now:
    - Reattach to your previous session and see if it works out better this
      time: `zellij attach unarmed-snails`
    - Try using a faster terminal. GPU-accelerated terminals such as Kitty
      or Alacritty are cross-platform and known to work well with zellij.

```

This improves the UX on #2023, #2041, #2048 and #2063, and may help to mitigate client crashes in some of these cases. It does not, however, explain why sometimes the server bevomes unresponsive and locks up. I couldn't find an explanation for this.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 * errors: Remove more `unwrwap`s from server code (https://github.com/zellij-org/zellij/pull/2069)
 * fix: support UTF-8 character in tab name and pane name (https://github.com/zellij-org/zellij/pull/2102)
 * fix: handle missing/inaccessible cache directory (https://github.com/zellij-org/zellij/pull/2093)
+* errors: Improve client disconnect handling (https://github.com/zellij-org/zellij/pull/2068)
 
 ## [0.34.4] - 2022-12-13
 
--- a/xtask/src/pipelines.rs
+++ b/xtask/src/pipelines.rs
@@ -104,6 +104,7 @@ pub fn run(sh: &Shell, flags: flags::Run) -> anyhow::Result<()> {
                     .arg("--no-default-features")
                     .args(["--features", "disable_automatic_asset_installation"])
                     .args(["--", "--data-dir", &format!("{}", data_dir.display())])
+                    .args(&flags.args)
                     .run()
                     .map_err(anyhow::Error::new)
             })
--- a/zellij-server/src/lib.rs
+++ b/zellij-server/src/lib.rs
@@ -40,7 +40,7 @@ use route::route_thread_main;
 use zellij_utils::{
     channels::{self, ChannelWithContext, SenderWithContext},
     cli::CliArgs,
-    consts::{DEBUG_MODE, DEFAULT_SCROLL_BUFFER_SIZE, SCROLL_BUFFER_SIZE},
+    consts::{DEFAULT_SCROLL_BUFFER_SIZE, SCROLL_BUFFER_SIZE},
     data::{Event, PluginCapabilities},
     errors::{prelude::*, ContextType, ErrorInstruction, FatalError, ServerContext},
     input::{
@@ -152,28 +152,20 @@ macro_rules! send_to_client {
         let send_to_client_res = $os_input.send_to_client($client_id, $msg);
         if let Err(e) = send_to_client_res {
             // Try to recover the message
-            use zellij_utils::channels::TrySendError;
-            let message = match e.downcast_ref::<TrySendError<ServerToClientMsg>>() {
-                Some(TrySendError::Full(msg)) => Some(msg),
-                Some(TrySendError::Disconnected(msg)) => Some(msg),
-                _ => None,
-            };
-            // failed to send to client, remove it
-            let context = if *DEBUG_MODE.get().unwrap_or(&true) {
-                format!(
-                    "failed to route server message to client {}\nMessage content: {:?}",
-                    $client_id,
-                    if let Some(msg) = message {
-                        format!("{:?}", msg)
-                    } else {
-                        "unknown".to_string()
-                    }
-                )
-            } else {
-                format!("failed to route server message to client {}", $client_id)
+            let context = match e.downcast_ref::<ZellijError>() {
+                Some(ZellijError::ClientTooSlow { .. }) => {
+                    format!(
+                        "client {} is processing server messages too slow",
+                        $client_id
+                    )
+                },
+                _ => {
+                    format!("failed to route server message to client {}", $client_id)
+                },
             };
             // Log it so it isn't lost
             Err::<(), _>(e).context(context).non_fatal();
+            // failed to send to client, remove it
             remove_client!($client_id, $os_input, $session_state);
         }
     };
--- a/zellij-server/src/lib.rs
+++ b/zellij-server/src/lib.rs
@@ -40,9 +40,9 @@ use route::route_thread_main;
 use zellij_utils::{
     channels::{self, ChannelWithContext, SenderWithContext},
     cli::CliArgs,
-    consts::{DEFAULT_SCROLL_BUFFER_SIZE, SCROLL_BUFFER_SIZE},
+    consts::{DEBUG_MODE, DEFAULT_SCROLL_BUFFER_SIZE, SCROLL_BUFFER_SIZE},
     data::{Event, PluginCapabilities},
-    errors::{ContextType, ErrorInstruction, FatalError, ServerContext},
+    errors::{prelude::*, ContextType, ErrorInstruction, FatalError, ServerContext},
     input::{
         command::{RunCommand, TerminalAction},
         get_mode_info,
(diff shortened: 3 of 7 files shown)


raphCode:
Very nice work!

A question of curiosity: The updates the server sends are incremental, so we can't afford skipping some?

My line of thinking was that maybe clients which are too slow could be supplied with "absolute" updates, that redraw everything, just like the initial frame after attaching to a session?
These absolutes updates are not time-critical and don't make sense to pile up, so they can be send out in whatever pace the client can process.

raphCode:
I just realized that some of the messages sent to client are not renders, but also UI state changes etc. My idea might not work with these.

imsnif:
@raphCode - render messages are ANSI instructions which are stateful. We cannot skip any of them, otherwise things will likely get corrupted for the user.

We could implement redraws, which would essentially mean to have the client reattach every time they receive a buffer full message. As I mentioned in the original PR that exposed this issue, it's not super trivial to do - and arguably is a patch that doesn't address the underlying problem. The more I think about this, the more I feel an unbounded queue is the solution we're looking for here (possibly with the addition of timing out unresponsive clients after a certain amount of time).

raphCode:
Yeah, I hoped redraws are easy to implement because that is what happens anyways on attaches.

Buffering between server and client helps to bridge short times where the client falls behind, but eventually buffers are full and then there are three options:
1. **backpressure, stall the server**
this is what happened before: noone noticed because stalls usually were short, until fansaris weird ssh connection came along and everything blocked forever
2. **drop the client**
this is what we are doing now
3. **complete redraw when the client comes back**
tmux does this

I believe, ideally we should do 3. and maybe after a long timeout (not measured in buffered bytes, but in minutes or hours) do 2.

I tested what tmux does with `seq 1 10000000 | dd status=progress`, one can pause the tmux client (`Ctrl-B`, `Ctrl-Z`), resume it after a while and the session is instantly back without some sort of intermediate updates.
Also, to be completely sure I analyzed a `script` transscript of the tmux session: First half of the numbers is continous, afterwards there are gaps with full-screen redraws.
Also, sometimes when executing the command inside tmux, one can even see the redraws as visual lagging of the numbers updating, think of it like the output framerate is reduced.

imsnif:
Honestly, I thought an unbounded queue would fit here because I didn't realize these stalls can be measured in more than microseconds. You're right then that it's not ideal.

I've seen this tmux behavior and honestly I don't like it. At least not the way it's implemented in tmux. It's a bad UX IMO and for us even now it's much much smoother.

I'd probably opt for some more robust buffering (increasing the number of messages significantly to account for about 10-20 seconds of lag) and then disconnecting the client and having it reattach to get a redraw.

But since @har7an is the one working on it now, let's not engage in too much armchair quarterbacking and allow him to choose the best path forward.

raphCode:
Another view: Redraws are comparable to ubiquitous frame dropping during video playback or realtime rendering.

---

I just noticed zellij already seems to have the necessary infrastructure for redraws?:
I tried my `dd` command on the linux framebuffer console (`Ctrl-Alt-F2`), where it displays me a whopping 120 kb/s throughput.
Now with zellij running in that same slow terminal, I reach the normal speed of several MB/s.
That means zellij cannot possibly draw every single line in the client.

> I've seen this tmux behavior and honestly I don't like it. At least not the way it's implemented in tmux. It's a bad UX IMO

What do you not like about it?

imsnif:
@raphCode - I'm going to defer the rest of this conversation to @har7an who is the person implementing this. I feel further theoretical discussions will not do us much good here, and will mostly serve to clutter up this PR and make it harder for the implementor to catch up.

raphCode:
Alright.
I am just going to leave a test case here that simulates slow clients - think ssh connections over dialup:

```bash
mkfifo /tmp/pipe
pv --rate-limit 50K /tmp/pipe > /dev/null &
script /tmp/pipe
# start / attach zellij here and start command that spams output
```

har7an:
*Implementor catches up*

Hello you two, happy to see I've drawn your attention. :)


> Could we not do away with the retry-queue implemented in this PR by increasing that queue from 50 to 1000 as you suggested?

Yes we could, but then:

1. We wouldn't know when the client becomes slow in the first place (because we can't tell how full a crossbeam queue is). This isn't so bad I guess, because we will notice when the connection drops, but I thought it's a good hint to users to see this logged by the server.
2. I'm not sure about the implications w.r.t. memory allocations. With the vector I can tell for sure that it isn't allocated as long as the client is quick enough, which I think is good.

I mean we could throw away that boundary and just make the queue unbounded, fair enough. But I'm not sure what happens when the blocking client has successfully drained all free RAM. I guess it will kill the whole application with it, and I'm totally against that sort of behavior. Like: It's not the other clients fault.

We can bump the queue size to some ridiculously huge amount and just disconnect the client if it's full, but before I do that I want to read up on how crossbeam allocates memory for its' queues.

Generally I'm quite happy with the current behavior: Disconnect the client, display a message explaining what happened, and then have the user handle it themselves. If they know their server instance is lost beyond hope (maybe they ran @raphCode "billion cuddles attack") they can at least kill it without hassle.

The idea of a full redraw is nice, but as far as I know `Render` isn't the only instruction we  send to the client. What do we do with the others? Can we just ignore/drop them?

har7an:
Okay, so according to the crossbeam sources a bounded channel is preallocated (as an array), whereas an unbounded channel isn't (linked list). In addition, we can query the length of any channel type with the `len` method (missed that before).

I assume the bounded channel is faster, which is beneficial for us. I mean: if the client is slow already, I see no point in having it use a *slower* queue for IPC. So I guess using a bounded channel with a large length is totally fine. I'll try to find out just how big one of these `ServerToClientMsg` structs is so we get an idea of how many we can allocate without anyone noticing... :)

imsnif:
> > account for about 10-20 seconds of lag
> 
> On my main PC, I get zellij to sink terminal text output at around 6 MB/s. Would 10 seconds worth buffering mean 60MB, plus struct overhead?

There might be some more overhead given that Zellij is not a lossless pipe. We've gotta do all sorts of manipulations to the raw ANSI data - namely interpreting it and sending it to your screen in a way that will place it in the relevant pane(s) (as well as other small stuff such as moving the cursor between panes, hiding it during rendering so it won't jump around, etc.)

But all in all, it shouldn't be significantly larger. I'd be interested in seeing how @har7an 's experiment goes and how much memory usage 20-30 seconds of buffering continuous heavy traffic causes - at least, what's the scale we're talking about.

har7an:
So according to my measurements bumping the IPC queue from 50 messages to 5000 has no noticable impact on memory usage at all. I used the following commandline to check memory usage for both cases:

```
$ ps -e -o drs,rss,trs,vsz,cmd | grep target/debug
```

after I had started zellij using `cargo x run`. So I'm dumping the `Vec`-based solution I implemented and raise the queuesize to 5000 entries. If the client stalls entirely, that should be good for many seconds worth of buffering before we disconnect them. IF we do end up disconnecting the client, the error is now logged and the client display a pretty message to the user (See above). Case closed for now I would say, unless either of you has any objections. :)

PR #2072, opened 2023-01-08
feat: add args to new-tab action

### LINKED ISSUE #1967: Add possibility to define a layout to use when opening a new tab with a keybinidng
Hi,

it would be nice to have a way to pass the path to a layout file to the `NewTab` keybinding action. In that layout the layout of the new tab can be defined. 

Not sure how to handle it if the layout config file defines multiple tabs or if you can prohibit that.

jaeheonji:
fix #1967

I added args to NewTab action in keybind.
This can be used like this:
```
tab {
        ...
        bind "n" {
            NewTab {
                cwd "/tmp"
                layout "example2.kdl"
                name "Hello"
            };
            SwitchToMode "Normal"; 
        }
        ...
}
```

It can assign various layouts to key combinations.


imsnif:
@jaeheonji - this looks great!

Only issue I found is that `cwd` and `name` don't work unless a layout is also provided. That's okay - but if that's the case, maybe we should give an error in this case so the user knows about this?

jaeheonji:
@imsnif Thank you for reviewing!

I missed that point, I think it would be good to add it. For example, if only name is set, it will only change the tab name of the default layout.

I'll update that point and add test cases.

jaeheonji:
@imsnif 

I updated that issue. but, it still have a issue about `cwd`.
The simplest way to apply `cwd` is to load the basic layout `(default.kdl)` file and apply cwd when there is no layout.

However, this means that the tab is opened by default, and if zellij is started with another layout, it will be different from what the user intended.

So, `cwd` is applied only when layout is specified.

> but if that's the case, maybe we should give an error in this case so the user knows about this?

Then, coming back to this issue, **It looks good that it doesn't show any errors. at least for now.**
If it prints an error, I don't know yet how to print it. If an error is displayed like `CLIAction`, zellij is terminated unintentionally.

And while writing e2e tests, I found one issue.
Currently, I have confirmed several times that this feature is working well. But, it does not work in `RemoteRunner` of e2e test.
I don't know the exact reason, so it's difficult to write a test. Do you have any ideas about this?

zolrath:
As a user, I would expect to be able to provide any of these options independently and have them default to whatever `zellij action new-tab` with no options would produce when not supplied.

## No Layout
No layout? Current behavior is the layout this instance of zellij was initially run with.

This is actually a bit odd as I assumed it would simply load `default.kdl` when making a new tab, and this was backed up with the strider layout. When you make a new tab with strider, there is no strider pane on the left on the new tabs.

After making a custom layout, I can see that new tabs duplicate the custom layout.
This is a bit unexpected by me/a bit undesirable as the custom layout I made I really only want for my first tab and when creating new tabs I get a bunch of extra panes that I do not want and have to close every time.

I'd expect the default new-tab to be the `default` layout, and I'd then expect to just supply `new-tab --layout rust` to create a new tab with a custom layout. Also, as an aside the `new-tab --layout` command doesn't know the default layout folder and the user has to supply the full path.


## No cwd
No cwd? Current directory.

## No name
No name? Name is Tab #.

jaeheonji:
@zolrath Thank you for your comments.
This is what I'm trying to implement.

@imsnif 

> But, In this process, the default_layout set in config is not applied.
> For example, If the user specified default_layout as compact in the config file, when NewTab is executed, compact should be imported by default. But, to apply cwd this doesn't work.
> 
> Becuase, Layout::stringified_from_path_or_default function basically returns the default layout.

To solve this problem, I parsed config first before parsing KDL of keybinds, and modified the code so that config options can be used in keybinds.

So, it now works exactly as intended for `cwd` and `name`.
Can you review it again?

default-plugins/status-bar/src/main.rs
```diff
@@ -335,7 +335,18 @@ pub fn action_key(keymap: &[(Key, Vec<Action>)], action: &[Action]) -> Vec<Key>
     keymap
         .iter()
         .filter_map(|(key, acvec)| {
-            if acvec.as_slice() == action {
+            if acvec.len() == action.len() {
+                // TODO: is there be a better way?
```
imsnif, line 339: How about if we implement something like `Action::is_comparable(&other_action)`?

zellij-utils/src/snapshots/zellij_utils__setup__setup_test__default_config_with_no_cli_arguments.snap
```diff
@@ -974,7 +974,79 @@ Config {
                 'n',
             ): [
                 NewTab(
-                    None,
+                    Some(
```
imsnif, line 977: I'm not thrilled about this... :)

Can we maybe get around this by always returning `NewTab(None)` from the KDL parser unless there's a layout, cwd or name specified explicitly?

imsnif (changes requested):
Hey @jaeheonji - great work on this!

I added some comments that (I think) address your concerns. Let me know what you think.

zellij-utils/src/snapshots/zellij_utils__setup__setup_test__default_config_with_no_cli_arguments.snap
jaeheonji: Yes. It's possible. And I understand your concerns too. Thanks!

default-plugins/status-bar/src/main.rs
jaeheonji: That's a great idea! I'll try to implement it!

imsnif:
@jaeheonji - I'm about to merge a pretty big PR today/tomorrow that will cause lots and lots of conflicts here. I don't know what the status here is, but if you're able to merge this before - it will save a lot of work :)

jaeheonji:
@imsnif Thanks for letting me know!

I'll be updating it today. then if you think my update needs more fixing, I'm fine with putting off this PR for a later time.

imsnif:
I trust your judgement, @jaeheonji ! If you think it's good, please merge.

PR #2082, opened 2023-01-11
fix 3 bugs

wlsnx:
1. https://github.com/zellij-org/zellij/blob/3da1cbf95c34283091186fd54926673fe2ec10f6/zellij-server/src/panes/tiled_panes/tiled_pane_grid.rs#L1168-L1178

   first_rect.cols.inner == 1.
   Between `split_vertically()` and `relayout()`, there is a `change_size()` call. New pane usually has a frame,so `get_content_columns()` returns 0, `change_size()` is skipped.
   However if you set `pane_frames false` and vertically split the right most pane, `get_content_columns()` returns 1.`change_size()` calls `drain_until(1)` for each row.
   Finally, `drain_until(1)` discard first widechar and all characters behind it.
2. `ZELLIJ_SOCK_DIR.join(name)` should be a valid filename in directory `ZELLIJ_SOCK_DIR`,so name cannot equals to "." or "..", as well as contains '/'. 
3. `zellij -s ''` Session name cannot be empty. Please provide a specific session name.
   `zellij attach --create ''` Panic occured: called `Result::unwrap()` on an `Err` value: Os { code: 98, kind: AddrInUse, message: "Address already in use" }

### DIFF
--- a/src/commands.rs
+++ b/src/commands.rs
@@ -324,6 +324,10 @@ pub(crate) fn start_client(opts: CliArgs) {
     };
     let os_input = get_os_input(get_client_os_input);
 
+    let start_client_plan = |session_name: std::string::String| {
+        assert_session_ne(&session_name);
+    };
+
     if let Some(Command::Sessions(Sessions::Attach {
         session_name,
         create,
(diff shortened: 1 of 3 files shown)


imsnif:
Hey @wlsnx - thanks for looking into these. I do not understand the user facing impact of 1. Can you please explain it from the point of view of the user? What bug is this solving?

wlsnx:
https://www.youtube.com/watch?v=E95MuYAxYnM
After split vertically, wide characters (Chinese, emojis, etc.) disappeared.

wlsnx:
This happens only if you set `pane_frames false` and vertically split the right most pane.

imsnif:
Great work on all of these, @wlsnx ! I added a comment to the `drain_until` function to explain the new conditional.

PR #2091, opened 2023-01-13
Fix status bar not displaying w/ size=1 and show shortcuts

### LINKED ISSUE #2077: status-bar broken with size=1
I found [this PR](https://github.com/zellij-org/zellij/pull/1875), which is supposed to only display the shortcut bar and no tips with `size=1`.

This seems to be broken in `zellij 0.34.4`. I am using the following layout:

```
layout {
    pane size=1 borderless=true {
    }
    pane {
        pane 
    }
    pane size=1 borderless=true {
    }
}
```

But the status-bar does not appear at all. If I configure it with `size=2`, it works.

Also: If this `size=1` would work, what happens if I hit e.g. CTRL+ T? Usually, further hints (`<n> New`) are displayed on line 2.
Is it possible to still get these, but with `size=1`, basically overwriting the first line of the status bar?

Thanks in advance ❤

panki27:
Closes #2077 
This changes behaviour if `status-bar` is used with `size=1`.

The issue of the status bar not showing with `size=1` arises from using `println!` when reaching the last line.
In this case, `print!` will be used instead.

This will also show the "second layer" of hints when not in normal mode, changing the first line.

Be careful reviewing, it's my first Rust PR 🚀

### DIFF
--- a/default-plugins/status-bar/src/main.rs
+++ b/default-plugins/status-bar/src/main.rs
@@ -238,7 +238,7 @@ impl ZellijPlugin for State {
 
         let first_line = first_line(&self.mode_info, cols, separator);
         let second_line = self.second_line(cols);
-        
+
         let background = match self.mode_info.style.colors.theme_hue {
             ThemeHue::Dark => self.mode_info.style.colors.black,
             ThemeHue::Light => self.mode_info.style.colors.white,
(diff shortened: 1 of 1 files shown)


imsnif:
Looks great. Thank you for your patience in having this reviewed, and congrats on your first Rust PR!

PR #2110, opened 2023-01-22
feat: add ScrollToTop action

### LINKED ISSUE #1940: Add ScrollToTop and assign default key bindings
Hi, thanks for your great work! I'm wondering why there is not a `ScrollToTop` action paired with `ScrollToBottom`. If that's just because no one has requested, I would like to request this feature. I usually opens a new panel to run a new command, and it's convenient if I can jump to the top with one key press to see the original command.

Besides, I would also like to propose adding two default key bindings: `Home` key in scroll mode for `ScrollToTop`, `End` key in scroll mode for `ScrollToBottom`.

Thanks!

yves-biener:
Closes #1940
Add new action `ScrollToTop`, which scrolls up to the top of the scroll log. The implementation is very similar to the already existing action `ScrollToBottom`. This also applies to the new test case for the new action.

I did not change the default keymap, as #1940 wants, but there is no documentation of the new action. Where can I add these, such that users find the new action?


jaeheonji:
> Where can I add these, such that users find the new action?

If you can and want it, please update it to [Zellij User Guide](https://zellij.dev/documentation)?
you can update docs to `cli-commands.md` and `keybindings-possible-actions.md` in this repo [zellij-org.github.io](https://github.com/zellij-org/zellij-org.github.io/tree/main/docs/src)

PR #2117, opened 2023-01-27
Fix fullscreen navigation

### LINKED ISSUE #2106: Alt-h, Alt-l only go in one direction in fullscreen panes
**Basic information**

`zellij --version`: zellij 0.34.4 and [b274fc5](https://github.com/zellij-org/zellij/commit/b274fc5ab19abe8b3ef7f7afe38897f526263414)
`stty size`: 42 168
`uname -av` or `ver`(Windows): Linux endeavouros 6.1.6-arch1-3 #1 SMP PREEMPT_DYNAMIC Mon, 16 Jan 2023 12:51:23 +0000 x86_64 GNU/Linux

**Further information**

Reproduction steps:
1. Open `zellij`.
2. Open 3 or more panes in 1 tab.
3. Enable fullscreen.
4. Press `Alt-h`/`Alt-l`, they will both switch to the next pane but not the previous one.

https://user-images.githubusercontent.com/49095435/213832357-f4a28ebd-4001-40ce-913d-0f21e10798cc.mp4

yves-biener:
Closes #2106
Well it started to bother we as well, so I implemented a fix. Instead of focusing the next pane in fullscreen mode, now the corresponding direction is respected.

During the implementation I also saw a small error (the prev focus method used the next focus function instead of the prev focus function;) which I also fixed, but put into a separate commit.

I haven't bothered to add additional tests for these changes, do you want me to add tests as well?

### DIFF
--- a/zellij-server/src/panes/tiled_panes/mod.rs
+++ b/zellij-server/src/panes/tiled_panes/mod.rs
@@ -1181,6 +1181,30 @@ impl TiledPanes {
         }
     }
 
+    pub fn focus_pane_left_fullscreen(&mut self, client_id: ClientId) {
+        self.unset_fullscreen();
+        self.move_focus_left(client_id);
+        self.toggle_active_pane_fullscreen(client_id);
+    }
+
+    pub fn focus_pane_right_fullscreen(&mut self, client_id: ClientId) {
+        self.unset_fullscreen();
+        self.move_focus_right(client_id);
+        self.toggle_active_pane_fullscreen(client_id);
+    }
+
+    pub fn focus_pane_up_fullscreen(&mut self, client_id: ClientId) {
+        self.unset_fullscreen();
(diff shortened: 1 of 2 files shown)


imsnif:
Hey @yves-biener - I'll take a look at this a little later this week. Thanks for the fix!

imsnif:
This is great, thank-you! Sorry for taking longer to get to this, it got lost in my shuffle.

wviana:
Hi there, would be nice if we have a full-screen mode, just as pane and tab. So I could remap my keys. I would change to h and l always jump tabs and j and k to cycle panes when in full-screen mode.

imsnif:
@wviana - you can do this sort of thing with plugins. I think (especially once we stabilize them and add things like headless plugins) this is the right place for other flavors of our UX.

PR #2119, opened 2023-01-28
Extend display char in tab

### LINKED ISSUE #2108: CJK in Tab name
Using Fedora 37

The issue is not important, but still maybe a bug. 

When using layouts one is able to add CJK (Chinese, Japanese, Korean) Text as Tab/pane title. While the pane title shows this correctly, the Tab title stays blank. 

Example: 

tab name="Hello"      -> shows as  'Hello' correctly 
tab name="Helloば"  -> shows as '' ... blank
tab name="ば”　       -> is also blank 
Same with Kanji and Hangul. 

This shouldn't be a font issue, since pane name="" works flawlessly and the fonts are the same I assume.  
I have not tested using them in the status bar yet, but for users of non alphanumeric scripts this might be Interesting to look at.

naosense:
To fix #2108

### DIFF
--- a/zellij-server/src/screen.rs
+++ b/zellij-server/src/screen.rs
@@ -1120,7 +1120,7 @@ impl Screen {
                             },
                             c => {
                                 // It only allows printable unicode
-                                if buf.iter().all(|u| matches!(u, 0x20..=0x7E | 0xA0..=0xFF)) {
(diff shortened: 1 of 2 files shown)


naosense:
@imsnif Hey could you please review it?

imsnif:
Thanks @naosense, thanks for this ! I will take a look before releasing the next version (will do a bug-squashing spree before :) )

PR #2120, opened 2023-01-28
Add `GoToTabName` action to switch tab by name

### LINKED ISSUE #2099: Switch to tab by name command
In the same flow as #2098 it would be nice to have a command that allows a user to switch to a tab by supplying a name.
Having a `--create `option to make the tab (like with session attaching) if it doesn't exist would be nice as well.

naosense:
To resolve #2099 , add new action ` GoToTabName` to switch tab by name


imsnif:
Hey @naosense - looks great! I haven't tested this yet (hopefully will get some time later this week) - but for now, could you add a test for this? Maybe something like this? https://github.com/zellij-org/zellij/blob/main/zellij-server/src/unit/screen_tests.rs#L511

naosense:
Hey @imsnif , test case added! Please check again.

zellij-server/src/screen.rs
```diff
@@ -1981,6 +1999,39 @@ pub(crate) fn screen_thread_main(
...
+                } else if screen
+                    .active_tab_indices
+                    .contains_key(&client_id.expect("This is checked above"))
+                {
+                    client_id
+                } else {
+                    screen.active_tab_indices.keys().next().copied()
+                };
+                if let Some(client_id) = client_id {
+                    if let Ok(is_exist) = screen.go_to_tab_name(tab_name.clone(), client_id) {
```
imsnif, line 2014: Let's change `is_exist` to `tab_exists`, ok?

imsnif (changes requested):
The code looks great - I left one minor comment.

Creating a new tab doesn't work for me though... I think this is because we need to also include the new tab layout - as in, we should probably do the whole `ScreenInstruction::NewTab` thing. I'm not sure this will be trivial here, do if you want to give it a try - that would be great. But I think it's also cool to not implement this flag for now. Your choice.

naosense:
@imsnif it's weird, it worked for me
image

Can you tell me about your steps, i'll reproduce it and see what's going on

zellij-server/src/screen.rs
naosense: Sure

imsnif:
My apologies @naosense - I tested it again and it works. I think I might have forgotten the flag :sweat_smile:

imsnif (approved):
Great work on this!

Issue #2114, opened 2023-01-26
Unfortunate random session name


greysonp:
Very unfortunate session name :sweat_smile: 
I'd probably just remove the word "racial" from the word pool.
![image](https://user-images.githubusercontent.com/1358731/214890865-6aa396b0-72d5-44b4-b051-dce51505c0ea.png)

```
$ zellij --version
zellij 0.34.4
```

eblume:
I just got "itchy-balls". I mean, zellij ... yeah, you get it.

My personal opinion, itchy-balls is hilarious, but I do agree that pulling "racial" out of the list is a good idea, and it looks like Issue #2122 could handle my itchy-balls problem.

gingsi:
The first words that zellij greeted me with was "old-wound". I am now using an alias with a random 3 letter name

`alias jn='zellij --session $(tr -dc a-z0-9 </dev/urandom | head -c 3 ; echo "")'`

Connecting this similar issue https://github.com/zellij-org/zellij/issues/2617

Issue #1907, opened 2022-11-04
Feature request: ability to change type of auto-generated session name


linuscm:
Been using Zellij at work and set it auto starts with my terminal, however been having an issue with the auto generated session names. Kinda inappropriate ones like "longing-children" and "ruthless-grandfather" were the highlights of the last few days. Funny at home, hard to explain to your coworkers.

jaeheonji:
Sorry @linuscm 😭 
Sometimes, `names` auto-generate weird names.

Instead you can specify the session name like this:
```bash
$ zellij -s <SESSION_NAME> # Specify name of a new session
```

I hope this was the feature you wanted.

raphCode:
I once got insulted as `jobless-pig` :(

adigitoleo:
Is there a config option for that or does it have to be a CLI arg?

PR #2122, opened 2023-01-29
Create custom lists of adjectives and nouns for generating session names

khs26:
### Description
Possibly addresses #1907 and #2114 by changing from using the `names` crate's default dictionary to a custom (and hopefully exclusively positive/neutral) set of adjectives and nouns.

Some things to consider, that I'm not sure about and happy to attempt to address:
- Do we want this to be behind a CLI flag?
- Do we need bigger lists to reduce the likelihood of clashes? Currently at 1% for 10 sessions.

### Tests and formatting
All passed with `cargo xtask test` on Ubuntu (5.15.0-48 kernel) except for `tests::e2e` which were all ignored.

Formatted with `cargo fmt`.


naosense:
Thank you for your efforts, but I don't see the way to customize the words, which are just written in the code

imsnif:
Hey @khs26 - I'm happy to see a code contribution from you!

I have a suggestion: what do you think of approaching this as a black-list instead? We could black list specific words (racial slurs, curses, etc.) or specific combinations that might be problematic (as demonstrated in one of the linked issues).

I suggest this for several reasons:
1. I think it will be cleaner rather than including a hard-coded list as @naosense mentioned
2. I'm a little worried about licensing. I guess this was copied from the names package? If so we might need to take a look at what license they have and also copy it here if we use this approach.

What say you?

Otherwise, to your questions:
1. I don't think we need a CLI flag for this. We might want to add a separate feature now or in the future to allow users to configure their own name pool, but I'm happy to start by creating a non-offensive default for now.
2. What's the current collision likelihood? Does this make it much worse? What do we currently do if there's a collision?

khs26:
> Hey @khs26 - I'm happy to see a code contribution from you!
> 
> I have a suggestion: what do you think of approaching this as a black-list instead? We could black list specific words (racial slurs, curses, etc.) or specific combinations that might be problematic (as demonstrated in one of the linked issues).
> 
> I suggest this for several reasons:
> 
> 1. I think it will be cleaner rather than including a hard-coded list as @naosense mentioned
> 2. I'm a little worried about licensing. I guess this was copied from the names package? If so we might need to take a look at what license they have and also copy it here if we use this approach.
No worries about licensing - I just chose a bunch of words that I thought were either strictly positive or neutral in tone. However, I do agree that it would be nicer to block certain combinations that we don't want - I'll have to take a look at the `names` package to see if that's a feasible task or if the dictionaries are too large.

> 
> What say you?
> 
> Otherwise, to your questions:
> 
> 1. I don't think we need a CLI flag for this. We might want to add a separate feature now or in the future to allow users to configure their own name pool, but I'm happy to start by creating a non-offensive default for now.
Ack.

> 2. What's the current collision likelihood? Does this make it much worse? What do we currently do if there's a collision?
Unsure - when I look at `names` I'll work out what the chances of a collision are. I think it might be a reasonable extension (to include in this PR) to handle collisions gracefully. I expect that nothing is done to prevent them, but that they might be much less likely with `names`'s defaults.

raphCode:
> What's the current collision likelihood? Does this make it much worse? What do we currently do if there's a collision?

Currently, there are 1124 adjectives, 981 nouns in the `names` crate:
https://github.com/fnichol/names/tree/main/data
That gives D = 1102644 unique names.
Using the birthday paradox formula, with N = 100 sessions, that gives a collision chance of 0.45%.
Play around here for different numbers: https://www.bdayprob.com/

pergamomo:
Why not just remove the names and leave it to the user?
I'd prefer they were not there at all.

imsnif:
> Why not just remove the names and leave it to the user? I'd prefer they were not there at all.

Because "annoying elevator" is much easier for Humans to remember than `<16 bit hex string>`.

But you can totally remove this behavior with something like `alias zellij=zellij -s (tr -dc A-Za-z0-9 </dev/urandom | head -c 13 ; echo '')`

pergamomo:
@imsnif  Thanks for the suggestion! I normally set my session name to the complete work path so it can become quite long. I guess what I'm really looking for is to not have the session name show at all.

imsnif:
@pergamomo - alright. You can open a separate issue for that though if you like (if there isn't one already?) - but let's not derail this PR please :)

souserge:
Great to see this issue being raised, I hope this will make it to the next release. Sometimes the auto-generated names as of version 0.35.1 are very disturbing:
image

imsnif:
Hey @khs26 - no pressure, but just for transparency - would you still like to see through the blacklist implementation we spoke about above?

tlinford:
> Just wondering if there are anything blocking this PR from being merged? If the issue is word list length, the names package is MIT-licensed so I reckon we can just prune negative words from that package's list and have a credit on top?

Also wondering the same, and bumping this since this has come up a few times again recently. In my opinion maintaining a blacklist would be more complicated.

imsnif:
I think at this point unless @khs26 will quickly disagree, it is safe to assume this is up for grabs. I think in one of the issues someone linked a 3rd party sanitizer, maybe that would be the best course of action?

Honestly though, let's get this done. @tlinford - could you see to this getting into main in one way or another?

JakeEhrlich:
Just adding that I too would like for this to be landed.

JakeEhrlich:
But I would also be happy to just have the CMU denylist from here as well: https://www.cs.cmu.edu/~biglou/resources/

Issue #1396, opened 2022-05-10
Improve plugin developer experience


imsnif:
- [ ] Better stacktraces when a plugin crashes
- [ ] A plugin crash should not crash the whole tool
- [ ] Utility libraries in other languages (starting with javascript/typescript, python and Go)
- [ ] Some sort of easy-to-install one-click compilation solution

scabala:
Hi @imsnif 
I would like to ask about utility libraries for other languages point. 
Does it mean it will be possible to create plugin for Zellij entirely in, for example, Python? Or Lua?
What is the approach here? Is it:
- exposing internal mechanisms of Zellij via IPC endpoint like in i3/Sway? or
- have WASM plugin with Python interpreter which will execute script provided by user?
Or maybe something entirely else?

I'm asking out of curiosity - one thing that kept me away from creating plugin for Zellij is need to use WASM (which I'm not familiar with) to have something simple like custom status bar.

imsnif:
The idea is to provide a similar utility library for every language like we have with Rust - so that you can write plugins in python more easily for example.

Truthfully though, the more time passes - for various reasons (only some of them technological) I'm thinking more and more that we should switch out the whole WASM, write it off as a bad idea and move to lua or some similar plugin configuration language. I know @a-kenji has thoughts on this.

denismaxim0v:
Regarding this.
I'm not sure if you have heard about [rhai](https://github.com/rhaiscript/rhai), but it looks like an interesting idea to explore.
@imsnif

ItsHarper:
I'm curious if you're still considering ditching WASM / if you still have the same negative feelings about the decision to use it

PR #2130, opened 2023-02-01
Do not assume what is in scope inside macro

m-lima:
We should not assume what the user brought into scope using `use` when writing macros.
This PR removes the assumption that the user called `use zellij_tile::predule::*`

### DIFF
--- a/zellij-tile/src/lib.rs
+++ b/zellij-tile/src/lib.rs
@@ -34,27 +34,34 @@ macro_rules! register_plugin {
         fn main() {
             // Register custom panic handler
             std::panic::set_hook(Box::new(|info| {
-                report_panic(info);
+                $crate::shim::report_panic(info);
             }));
 
             STATE.with(|state| {
+                use $crate::ZellijPlugin;
                 state.borrow_mut().load();
             });
         }
 
         #[no_mangle]
         pub fn update() -> bool {
+            use $crate::prelude::Context;
+            use $crate::prelude::LoggableError;
             let object = $crate::shim::object_from_stdin()
                 .context($crate::PLUGIN_MISMATCH)
                 .to_stdout()
                 .unwrap();
 
-            STATE.with(|state| state.borrow_mut().update(object))
+            STATE.with(|state| {
+                use $crate::ZellijPlugin;
+                state.borrow_mut().update(object)
+            })
         }
 
         #[no_mangle]
         pub fn render(rows: i32, cols: i32) {
             STATE.with(|state| {
+                use $crate::ZellijPlugin;
                 state.borrow_mut().render(rows as usize, cols as usize);
             });
         }


imsnif:
Hey - thanks for this. While I understand the issue here, I'd rather hold off on any unnecessary changes to the plugin system for the time being. As mentioned at the top of the plugin documentation, it's going through an overhaul right now. If this is still an issue in the new system, we can re-examine this.

m-lima:
> Hey - thanks for this. While I understand the issue here, I'd rather hold off on any unnecessary changes to the plugin system for the time being. As mentioned at the top of the plugin documentation, it's going through an overhaul right now. If this is still an issue in the new system, we can re-examine this.

Fair enough 😄 Sorry for the noise!

I've been looking at the branches and issues to see the direction that Zellij is going for and to contribute with this effort.
I'm finding it hard to find where I can help, though.
The [plugin system](https://github.com/zellij-org/zellij/labels/plugin%20system) tag is a lot broader than just the plugin revamp.
In the [Milestones](https://github.com/zellij-org/zellij/milestone/2) I found this [issue](https://github.com/zellij-org/zellij/issues/1396) that I think is what covers this change (??), but not really actionable coming from the outside.
E.g. the "A plugin crash should not crash the whole tool" action item sounds like a good self-contained task, covering just the interface towards the plugin. Though it depends on how you have decided to approach plugins. If you are dropping wasm and exposing an FFI interface or embedding a scripting language or some other approach.

I'm really looking forward to helping. Can you point me in the direction of an issue that I can help with?

imsnif:
Hey @m-lima - I'm happy you're interested in joining the efforts! Honestly, most of the stuff you linked to is outdated - and the updated versions are only in my head. I'll admit I don't spend a lot of time publicizing this because I'd rather spend it developing - and the realities of OSS means that even the most well-intentioned contributors often do not find the time to contribute beyond an initial burst, and so the ROI on said publicizing is sadly often negative.

So here's a brief run-down of my near-to-mid future plans for the plugin system:
1. Allow applying layouts at runtime - while not directly a plugin system feature, this is a must for any sort of plugin-to-zellij API, so that plugins can change stuff in the user interface. I'm working on this now.
2. Improve the plugin developer experience: namely allow loading plugins at runtime not through a layout and have this happen fast (currently in development it's very slow).
3. Develop an additional default plugin in Rust (probably a session switcher - since it's very much in demand) and through this add capabilities to plugins.
4. Add capabilities to 3 (I really want to have a sort of built-in shell for Zellij to launch terminals, plugins, commands, layouts, etc.)
5. Add interfaces/templates in other languages
6. (there's more, but this is a good start :) )

If you'd like to participate, I could really use some help with 2. Right now, loading plugins in a developer environment is extremely slow - likely due to the wasm compilation. This is the first thing I'd address (since shortening the development feedback loop as a first step will make everything else faster to do). Are you interested in giving this a go? Fair warning though: this is a blocking task, so there is a chance I'll sweep in in a few weeks once I finish with 1 and some other stragglers.

What do you think?

m-lima:
I can give **2** a go. I'm assuming here that the wasm interface remains for this and that the focus is mostly on loading the plugin just like a new pane can be loaded at runtime.

Since the focus is on loading speed, I'll start by taking a look at `start_plugin()` in **wasm_bridge.rs**
It seems to me, though, that the main hit is from `Module::new()`. This is [tied](https://medium.com/wasmer/wasmer-1-0-3f86ca18c043) to the Engine being used. If I'm not mistaken, Zellij doesn't set a default engine feature flag, which [means](https://github.com/wasmerio/wasmer/blob/545a12ac93185f634844faa7e8e94049f475295b/lib/api/src/sys/store.rs#L159) that Cranelift is used.

Is this the kind of change you were hinting at?
I'll join discord so that it is easier to communicate :)

PR #2136, opened 2023-02-05
cleanups around "buffer full" error

har7an:
As discussed in the last team meeting, this PR performs two minor cleanups:

1. Add a `FIXME` note to the bounded queue between server and client that explains why it has its' current size, and in order to make it easier to find and change if the need arises (again)
2. Don't recommend terminal emulator applications inside error messages. It's fine for users/maintainers to recommend their personal favorite applications and suchlike, but zellij itself remains neutral in this regard


har7an:
@imsnif Implemented the changes discussed in the last meeting. I hope that's okay now?

imsnif:
Looks great @har7an - I added the bit about the "ideal" solution to the comment. Thanks!

PR #2141, opened 2023-02-07
fix: set sticky bit on socket

### LINKED ISSUE #2080: zellij sessions sockets may get deleted
zellij places its server sockets in XDG_RUNTIME_DIR.
Per specification*, this directory may be cleaned of old files periodically, and I suspect at least one user on discord lost access to his session this way.
The problem can be solved in two ways, ideally we do both:
- set the sticky bit on the socket file to avoid automatic cleanup *
- provide a method to recreate the socket file, e.g. on SIGUSR1

*Sources:
https://wiki.archlinux.org/title/XDG_Base_Directory#User_directories
https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html

tlinford:
Hopefully resolve #2080.

@imsnif I'm not sure if this actually solves the problem, but I don't really see any downside to this either, what do you think?

### DIFF
--- a/zellij-server/src/lib.rs
+++ b/zellij-server/src/lib.rs
@@ -275,7 +275,10 @@ pub fn start_server(mut os_input: Box<dyn ServerOsApi>, socket_path: PathBuf) {
             move || {
                 drop(std::fs::remove_file(&socket_path));
                 let listener = LocalSocketListener::bind(&*socket_path).unwrap();
-                set_permissions(&socket_path, 0o700).unwrap();
+                // set the sticky bit to avoid the socket file being potentially cleaned up
+                // https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html states that for XDG_RUNTIME_DIR:
+                // "To ensure that your files are not removed, they should have their access time timestamp modified at least once every 6 hours of monotonic time or the 'sticky' bit should be set on the file. "
+                set_permissions(&socket_path, 0o1700).unwrap();
                 for stream in listener.incoming() {
                     match stream {
                         Ok(stream) => {


imsnif:
I don't have a lot of context on this... what does @raphCode think?

raphCode:
as per the freedesktop spec, this should prevent automatic cleanup.
However, socket files can still be deleted manually or by accident - maybe we can provide a way of recreating the socket file of a running server? This way "lost" sessions may be recovered.

imsnif:
> maybe we can provide a way of recreating the socket file of a running server? This way "lost" sessions may be recovered.

This is a cool feature to have, but very much not trivial to implement. I vote for not compounding it with this one.

imsnif:
Hey, what's the status here?

imsnif:
@tlinford (for when you get back) - afaict, the sticky bit should only apply to directories and not files... are we sure we're doing the right thing here?

tlinford:
Forgot to answer here sorry, the freedesktop spec says: 'To ensure that your files are not removed, they should have their access time timestamp modified at least once every 6 hours of monotonic time or the 'sticky' bit should be set on the file.'

I also found that wezterm does this (notice the T attr):
```
❯ ll /run/user/1000/wezterm/
srwx-----T thomas thomas  0 B Wed Mar 29 10:11:49 2023  gui-sock-35301=
lrwxrwxrwx thomas thomas 37 B Wed Mar 29 10:11:49 2023  wayland-wayland-1-org.wezfurlong.wezterm@ ⇒ /run/user/1000/wezterm/gui-sock-35301
```

imsnif:
Cool, thanks for looking into this @tlinford . Anything else we need to do here or are waiting for?

PR #2145, opened 2023-02-08
Add action `QueryTabNames` to list all tab names

### LINKED ISSUE #2098: List opened tabs in a session?
Unless I'm very much mistaken, there doesn't seem to be any way to list the opened tabs, just like you can list opened sessions. Can we please have this feature? Absence of that, the `action` subcomamnd doesn't give that much power.

Thanks.

naosense:
To resolve #2098 , as shown below
image


zellij-server/src/lib.rs
```diff
@@ -609,6 +611,15 @@ pub fn start_server(mut os_input: Box<dyn ServerOsApi>, socket_path: PathBuf) {
                     session_state
                 );
             },
+            ServerInstruction::TabNameList(tab_names, client_id) => {
+                log::info!("tab name is {tab_names:?} client id {client_id}");
```
naosense, line 615: it can print tab names and client id here

zellij-client/src/lib.rs
```diff
@@ -416,6 +419,14 @@ pub fn start_client(
                     .send(InputInstruction::SwitchToMode(input_mode))
                     .unwrap();
             },
+            ClientInstruction::TabNameList(tab_names) => {
```
naosense, line 422: but here can't receive `ClientInstruction::TabNameList`,  @imsnif am i doing right?

imsnif: I didn't look at this very deeply, but I think this should be handled in the `cli_client` if you are sending this query from the CLI

naosense: Got it

zellij-client/src/cli_client.rs
```diff
@@ -28,6 +28,10 @@ pub fn start_cli_client(os_input: Box<dyn ClientOsApi>, session_name: &str, acti
                 os_input.send_to_server(ClientToServerMsg::ClientExited);
                 process::exit(0);
             },
+            Some((ServerToClientMsg::TabNameList(tab_names), _)) => {
```
imsnif, line 31: Hey @naosense - let's change `ServerToClientMsg::TabNameList` to `ServerToClientMsg::Log`. In `cli_client` it will do exactly what it does here, and in the normal client it will log to file (with `log::info`).

naosense: ok, one more question, what does a normal client refer to?

imsnif: Yeah, let's make things consistent.

The "normal" client is the UI, eg. the client that sends your keyboard clicks and actions to the server. This is as opposed to the cli client which does this from the command line. I want us to handle this instruction in both places properly, so that users will be able to eg. use `QueryTabList` as a keyboard binding too if they want, and then the tab list will appear in the log file.

You can find it here: https://github.com/zellij-org/zellij/blob/f8a078c1a959ca7bb243bd9a86b3ddc392605c9c/zellij-client/src/lib.rs#L391

Makes sense?

naosense: yeah, thanks for your reply

imsnif:
Hey @naosense - what's the status of this PR?

naosense:
I think it's ready for review

zellij-server/src/lib.rs
```diff
@@ -623,6 +625,14 @@ pub fn start_server(mut os_input: Box<dyn ServerOsApi>, socket_path: PathBuf) {
                     session_state
                 );
             },
+            ServerInstruction::Log(tab_names, client_id) => {
```
imsnif, line 628: Also here

naosense:
> I left some comments about naming, and otherwise - can we maybe add a test for this? Preferably one that tests the entire flow with the cli action. I think there are a few examples in `tab_integration_tests`.

after i read `tab_integration_tests`, i didnt find similar test. i want to run a command `zellij action query-tab-names`, then test the snapshot, but i cant figure out how to run cmd. Am i in the right direction?

imsnif:
Sorry @naosense - I meant this file, my mistake: https://github.com/zellij-org/zellij/blob/main/zellij-server/src/unit/screen_tests.rs#L989

Just be aware that you'll need to adjust your test a little bit, since the normal snapshot probably won't help you here (as it's taking a snapshot of the render and not the new log).

zellij-server/src/unit/snapshots/zellij_server__screen__screen_tests__send_cli_query_tab_names_action.snap
```diff
@@ -0,0 +1,30 @@
+---
```
naosense, line 1: test case added! i've no idea if this is ok @imsnif

imsnif: Hey, good start with this. I adjusted the test so that it tests for the `Log` action that we added rather than the `Render` (because that won't really tell us anything...)

I also merged from main and made some minor naming changes.

Thanks for your work on this! I'll merge once the CI is green.

naosense: thanks for your explanation, happy to learn that all

Issue #2144, opened 2023-02-08
Error occurs when `cargo xtask run` in macOS


naosense:
After updated the latest code 0255b56
```
> cargo xtask run

debug/zellij`

Error occurred in server:

  × Thread 'wasm' panicked.
  ├─▶ Originating Thread(s)
  │     1. ipc_server: NewClient
  │     2. screen_thread: NewTab
  │     3. plugin_thread: NewTab
  │
  ├─▶ At zellij-server/src/lib.rs:762:18
  ╰─▶ Program terminates: a fatal error occured

      Caused by:
          0: failed to load plugin for client 1
          1: failed to load plugin from instance Instance {
                 exports: {
                     (
                         "memory",
                         Memory(...),
                     ),
                     (
                         "__heap_base",
                         Global(...),
                     ),
                     (
                         "__data_end",
                         Global(...),
                     ),
                     (
                         "_start",
                         Function(...),
                     ),
                     (
                         "update",
                         Function(...),
                     ),
                     (
                         "render",
                         Function(...),
                     ),
                     (
                         "plugin_version",
                         Function(...),
                     ),
                     (
                         "main",
                         Function(...),
                     ),
                 },
             }
          2: RuntimeError: out of bounds memory access
          3: heap_get_oob
  help: If you are seeing this message, it means that something went wrong.

        -> To get additional information, check the log at: /var/folders/dr/y_q_0j3s4pj28s3rg91bjtgw0000gp/T/zellij-502/zellij-log/zellij.log
        -> To see a backtrace next time, reproduce the error with: RUST_BACKTRACE=1 zellij [...]
        -> To help us fix this, please open an issue: https://github.com/zellij-org/zellij/issues



Error: failed to run pipeline 'run' with args Run { args: [], data_dir: None, cranelift: false }

Caused by:
    command exited with non-zero code `/Users/xxx/.rustup/toolchains/stable-aarch64-apple-darwin/bin/cargo run --`: 1
```

imsnif:
Can you try clearing the plugin cache? By default I think it should be in ~/.local/cache/zellij/plugins
Also deleting the `target` folder might do the trick

naosense:
Deleting cache and target doesn't work for me, I'm using macOS if anyone who is also a Mac user can run it and see if it's my problem alone

tlinford:
Can confirm, same error running on macos.

imsnif:
@m-lima - any ideas?

m-lima:
[...]

PR #2146, opened 2023-02-08
Make the singlepass compiler opt-in through flags

m-lima:
Fixes https://github.com/zellij-org/zellij/issues/2144

Makes the simplepass compiler opt-in instead of the default compiler for debug mode.
This may be desired due to consistent failure when running the default `cargo x run` command on arm64.


imsnif:
Thankyou!

PR #2153, opened 2023-02-11
Refator and consolidate code around splitting panes

mdavezac:
My goal is to add commands to split panes left, right, up and down rather than just left and down. I think that would involve less cut-and-paste if the code and commands for splitting is consolidated. This PR does just the refactoring.

### DIFF
--- a/zellij-server/src/panes/tiled_panes/mod.rs
+++ b/zellij-server/src/panes/tiled_panes/mod.rs
@@ -314,24 +314,7 @@ impl TiledPanes {
             false
         }
     }
-    pub fn split_pane_horizontally(
-        &mut self,
-        pid: PaneId,
-        mut new_pane: Box<dyn Pane>,
-        client_id: ClientId,
-    ) {
-        let active_pane_id = &self.active_panes.get(&client_id).unwrap();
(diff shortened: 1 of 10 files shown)


imsnif:
Hey @mdavezac - while I really appreciate all the work you put into this and can understand your ultimate goal here, I do not want to make these changes at this time.

There are a few reasons for this:
1. I'm about to merge a very large PR in the next week or so that changes these parts of the code quite a bit. Namely adding stacked panes - which would likely collide heavily with these changes, since it adds many new requirements to all the tiled pane geometry handling.
2. I'm not 100% certain about ultimately having a true split-<direction> functionality. The aforementioned PR also adds `swap_layouts`, which would allow you to (re)arrange the layouts in a tab in a very precise manner. Please correct me if I'm wrong, but I think if we take a step back - this is really what you want, right? Define how the layouts should be rearranged rather than have to script it carefully with directional splits?
3. Generally, I realize there's a lot of repetition in that part of the code, but that's because it's very much still in flux. New features and behaviours are added to it frequently, and bunching stuff in generic structures will (in my view) make these changes more difficult.

mdavezac:
Sure, that makes sense. Thanks for detailed explanation.

PR #2156, opened 2023-02-12
Add Helix editor (`hx`) to editors that support prefix positions

bheylin:
This is dependant on Helix accepting the PR:

https://github.com/helix-editor/helix/pull/5945

### DIFF
--- a/zellij-server/src/os_input_output.rs
+++ b/zellij-server/src/os_input_output.rs
@@ -300,14 +300,18 @@ fn spawn_terminal(
                     || command.ends_with("emacs")
                     || command.ends_with("nano")
                     || command.ends_with("kak")
-                    || command.ends_with("hx")
-                    || command.ends_with("helix")
                 {
                     failover_cmd_args = Some(vec![file_to_open.clone()]);
                     args.push(format!("+{}", line_number));
+                    args.push(file_to_open);
+                } else if command.ends_with("hx") || command.ends_with("helix") {
+                    args.push(format!("{}:{}", file_to_open, line_number));
+                } else {
+                    args.push(file_to_open);
                 }
+            } else {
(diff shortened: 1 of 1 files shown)


bheylin:
This change assumes that all Helix users update regularly.
We may want to consider a fallback to postfix position args for Helix (`hx Cargo.toml:10`).
But that would require checking the Helix version in use `hx -V`.

zellij-server/src/os_input_output.rs
```diff
@@ -300,6 +300,7 @@ fn spawn_terminal(
                     || command.ends_with("emacs")
                     || command.ends_with("nano")
                     || command.ends_with("kak")
+                    || command.ends_with("hx")
```
uncomfyhalomacro, line 303: Could you also add `helix`? Some distros use the `helix` command instead of the `hx` command e.g. ArchLinux and openSUSE

bheylin: Added in https://github.com/zellij-org/zellij/pull/2156/commits/9acf464e566643863f13fcc294db032349395888

imsnif:
Hey @bheylin - thanks for this!

For better compatibility, how about - specifically for helix - we use the old notation? I think an additional `if` statement or similar would be worth the total transparency for users. What do you think?

imsnif:
Sorry, I added a comment and angered the `rustfmt` gods. I'll fix it, give this a small test run just to be sure and then merge if all is well.

bheylin:
> Sorry, I added a comment and angered the `rustfmt` gods. I'll fix it, give this a small test run just to be sure and then merge if all is well.

No worries, you take the PR from here

imsnif:
Thank you very much!

PR #2157, opened 2023-02-12
Add pencil theme

aral:
This adds the pencil theme. 

Only a light variant (meant to be used alongside the pencil light theme in your terminal) is available at the moment.


imsnif:
Beautiful <3

PR #2166, opened 2023-02-15
Apply correct color on 'more tabs' message

pedromfedricci:
Fix the color of 'more tabs' message for tab-bar and compact-bar plugins.

Currently: 
tab-pre-fix

Fix:
tab-post-fix

Thank you.


jaeheonji (approved):
Good Catch! LGTM!

PR #2181, opened 2023-02-20
Update bug_report.md template with OSX log directory

KaiSpencer:
Logs on OSX are located in `$TMPDIR` not `/tmp`, this hints to OSX users raising bug reports of where to find the log file


imsnif:
Thank you!

PR #2190, opened 2023-02-25
Fix plugin location path decoding from Url

tarkah:
Using URL file schema for plugin location is currently broken in a couple different ways:

- Naively converting URL path to a `PathBuf` doesn't decode the percent encoded URL. This makes any path with "spaces" or percent encoded characters break.
- Using `path` or now w/ `to_file_path` always returns an "absolute" `PathBuf`. The root prefix is now stripped when adding it to the config directory for relative lookup.

I've tested both relative lookup `file:plugin.wasm` as well as absolute path w/ spaces and both now work.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4095,6 +4095,7 @@ dependencies = [
  "miette 3.3.0",
  "nix 0.23.1",
  "once_cell",
+ "percent-encoding",
  "regex",
  "rmp-serde",
  "serde",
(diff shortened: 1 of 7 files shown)


imsnif:
Hey @tarkah - thanks for looking into this!

Could you please add some tests (even if just unit tests for the changed functions) to clarify what's going on here? Barring that (if it's too complex to add a test suite here) maybe add very detailed comments with examples?

imsnif (approved):
Thanks for this, the test looks great!

PR #2194, opened 2023-02-26
Feature: simulate publishing

har7an:
## Rationale

Zellij is a complex application, consisting of multiple subcrates that need to be released/published separately, along with a few application assets such as plugins, which are usually bundled with the application. For this reasons, it has depended for a long time now on external build-tools such as `cargo-make`/`makers`.

Recent contributions to the project have necessitated a different handling of these application assets (in particular: The bundled plugin binaries, see #1924) and introduced a new build system, `xtask`, to replace the previous `cargo-make` for a plethora of reasons (See #2012). All of these changes, although not obvious at first glance, introduce risk and uncertainty to the release process.

This is especially problematic because releasing a zellij version is not an atomic operation and can fail in the process, as it has a few times previously. Since there is no "undo"-button, a half-finished release sticks around and is visible for the public, although it is entirely unusable. Unfortunately, up until now, it is only possible to tell whether a release was successful after the release is performed. In turn, this means that a failure during release isn't uncovered until we're in the release process. This usually necessitates a decent amount of live-patching during the release process, as can be understood by looking at the release "history" between 0.34.0 and 0.34.3 (See 66593ec41a1c11ed134d729e9dea291e4c4cd38f..67ac1190a7039d99f4803e95a4689238d52eaa7b).

This is both cumbersome to users and stressful to the maintainers responsible for the release. In order to alleviate this, I'm proposing this PR as a means to simulate a zellij release ahead of time.

## Changes in this PR

This PR adds two new CLI flags to the `cargo xtask publish` command:

- `--git-remote`: Allows specifying a git remote (other than the default `origin`) to push release commits to. In conjunction with a private `zellij` fork, this allows us to safely test release commits before we publish them to the main repository.
- `--cargo-registry`: Allows specifying a different cargo registry to publish the zellij subcrates to. Just like `--git-remote`, this allows us to push/publish crates somewhere else than into the main `crates.io` registry.

In combination with a private registry, these allow us to simulate a full release ahead of time. Because the procedure requires a lot of manual setup and isn't exactly trivial (yet?), there is now a new piece of documentation in `RELEASE.md`, which explains how to set this up correctly.

I hope that in the future, this will make releases frictionless, even in the face of significant changes to the build/plugin system.

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -36,8 +36,6 @@ jobs:
       run: sudo apt-get install -y --no-install-recommends musl-tools
     - name: Add musl target
       run: rustup target add x86_64-unknown-linux-musl
-    - name: Install wasm-opt
-      run: sudo apt-get install -y --no-install-recommends binaryen
       #run: cargo install --debug cargo-make
     - name: Build asset
       run: cargo xtask ci e2e --build
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -63,14 +63,6 @@ jobs:
         if: matrix.os == 'ubuntu-latest'
         run: sudo apt-get install -y --no-install-recommends musl-tools
         
-      - name: Install wasm-opt (macos)
-        if: runner.os == 'macos'
-        run: brew install binaryen
-
-      - name: Install wasm-opt (ubuntu)
-        if: matrix.os == 'ubuntu-latest'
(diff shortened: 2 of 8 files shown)


har7an:
@imsnif I'd be very happy if you could go through the steps in `RELEASE.md`, run them locally and provide feedback to me in case anything doesn't work or is unclear. I've tried, to the best of my knowledge, to include all the changes/actions I needed to set this up to a working state.

PR #2223, opened 2023-03-05
feat: add self-provided themes

jaeheonji:
Now zellij provides its own variety of themes.
The types of themes provided at this point include those that existed in the `example` folder.

And the example folder will contain only examples in accordance with the intent.


imsnif:
This is grat @jaeheonji !! I'd like to release a version tomorrow morning. Do you think this will be ready in time?

jaeheonji:
@imsnif 
I think I'm almost ready. But only one weird thing is, the current e2e test doesn't pass. Exactly, the timeout occurs, but I don't know why.

naosense:
last time I also encountered this, the e2e failure disappeared after I recommit the code

jaeheonji:
Thanks for letting me know. I'll try to make a PR after re-commit.

Issue #2071, opened 2023-01-08
Feature: Allow MoveFocus to wrap


jemag:
For example, if I am in this pane:
![image](https://user-images.githubusercontent.com/7985687/211207400-25ba6388-2958-4651-855e-a4a59c804925.png)

Wrapping would allow me to `MoveFocus "right" and still go to the pane on the left:
![image](https://user-images.githubusercontent.com/7985687/211207455-1bfa0fbb-0bab-48c9-886b-a5a2e0b0004b.png)

This is particularly useful in situations where there might be a high amount of panes such as this:
![image](https://user-images.githubusercontent.com/7985687/211207483-5342a156-79c7-45f7-9e0b-bec745bee68f.png)
Now moving from pane#5 to pane#1 is as simple as moving the focus right.

---
This could be implemented in a number of ways, some possible options:
- global config option (enable wrapping globally or not)
- giving `MoveFocus` another parameterslike `wrap`
- having a completely different action `MoveFocusWrapped` (this is probably overkill)

raphCode:
related / duplicate: https://github.com/zellij-org/zellij/issues/1444

s1syph0s:
I would like to work on this issue!

Issue #1444, opened 2022-05-31
Pane traversal with one tab open


qballer:
When traversing panes within one tab in the multiplexer there is a case where I would like to see a small behaviour change. 
If we have one tab open and we jump to the right until there is no more panes I would expect that jumping right would place the cursor at the left most pane. As in :

|          |            | 
|.         |         ▮ | 
|          |.           |
should be 

|          |            | 
| ▮       |            | 
|          |.           |

This works well when we have multiple tabs open, it's just this edge case.  Currently the cursor just stays in place and ignores the keyboard

raphCode:
There seems to be another inconsistency, caused by the behavior which pane is active (usually the first one) upon switching to a tab:
- Open two tabs with two panes (vertical split) each
- Alt + L switches tabs on every second keypress
- Alt + H switches tabs on every keypress

While I see why this is so, the two key combinations should ideally be the inverse for each other.
The behavior even changes when a second user is connected, since now by default the pane that user is connected to is focused.

imsnif:
> There seems to be another inconsistency, caused by the behavior which pane is active (usually the first one) upon switching to a tab:
> 
>     * Open two tabs with two panes (vertical split) each
> 
>     * Alt + L switches tabs on every second keypress
> 
>     * Alt + H switches tabs on every keypress
> 
> 
> While I see why this is so, the two key combinations should ideally be the inverse for each other. 

This is a bug - it didn't use to be this way. When focusing a new tab, the pane that was last focused should be focused and then they should be consistent. We record this, I guess we just don't pick it up anymore in this case.

raphCode:
Oh well, this must be broken since forever / I started using zellij, I thought this is the intended behavior :D 

I saw occasional e2e test failures where the cursor ended up in the wrong pane, this might be related?

imsnif:
Anything is possible :) To be fair, when this feature was created the only way we had to test it were the e2e tests - so we were pretty reluctant to add lots of test cases. If this is fixed now, it should be relatively easy to add some unit/integration tests.

raphCode:
Yes, feel free to work on this, I haven't got much time.

To get started, I would advise to look at `move_focus_left_or_previous_tab()` in `zellij-server/src/screen.rs`.

Currently, we call into the current tab and try to move the pane focus left (`move_focus_left()` in `zellij-server/src/tab/mod.rs`). The function returns a boolean indicating if the focus change was successful because two things can happen:
[...]

PR #2242, opened 2023-03-07
[DO NOT REVIEW YET - DRAFT]wrap around focus

Kangaxx-0:
> please do not review this pr yet, some clean up and further tweak is not completed

Try to address #2071 and #1444 

Wrap around pane focus if there is only one tab

How the fix works:

wraparound

- [x] Run `cargo xtask test`

### DIFF
--- a/zellij-server/src/panes/tiled_panes/mod.rs
+++ b/zellij-server/src/panes/tiled_panes/mod.rs
@@ -863,7 +863,7 @@ impl TiledPanes {
             (size_in_pixels.height as f64 / size_in_pixels.width as f64).round() as usize
         })
     }
-    pub fn move_focus_left(&mut self, client_id: ClientId) -> bool {
+    pub fn move_focus_left(&mut self, client_id: ClientId, wrap_around: bool) -> bool {
         match self.get_active_pane_id(client_id) {
             Some(active_pane_id) => {
                 let pane_grid = TiledPaneGrid::new(
@@ -872,7 +872,11 @@ impl TiledPanes {
                     *self.display_area.borrow(),
                     *self.viewport.borrow(),
                 );
-                let next_index = pane_grid.next_selectable_pane_id_to_the_left(&active_pane_id);
+                let mut next_index = pane_grid.next_selectable_pane_id_to_the_left(&active_pane_id);
+                if next_index.is_none() && wrap_around {
+                    let index = self.panes.len() - 3;
+                    next_index = Some(PaneId::Terminal(index as u32));
+                }
                 match next_index {
                     Some(p) => {
                         // render previously active pane so that its frame does not remain actively
@@ -1021,14 +1025,12 @@ impl TiledPanes {
                     *self.display_area.borrow(),
                     *self.viewport.borrow(),
                 );
-                info!("active_pane_id: {:?}", active_pane_id);
                 let mut next_index =
                     pane_grid.next_selectable_pane_id_to_the_right(&active_pane_id);
 
                 if next_index.is_none() && single_tab_wraparoud {
                     next_index = Some(PaneId::Terminal(0));
                 }
-                info!("next_index: {:?}", next_index);
                 match next_index {
                     Some(p) => {
                         // render previously active pane so that its frame does not remain actively
@@ -1542,9 +1544,9 @@ impl TiledPanes {
         }
     }
 
-    pub fn focus_pane_left_fullscreen(&mut self, client_id: ClientId) {
+    pub fn focus_pane_left_fullscreen(&mut self, client_id: ClientId, wrap_around: bool) {
         self.unset_fullscreen();
-        self.move_focus_left(client_id);
+        self.move_focus_left(client_id, wrap_around);
         self.toggle_active_pane_fullscreen(client_id);
     }
 
--- a/zellij-server/src/panes/tiled_panes/mod.rs
+++ b/zellij-server/src/panes/tiled_panes/mod.rs
@@ -3,6 +3,7 @@ mod stacked_panes;
 mod tiled_pane_grid;
 
 use crate::resize_pty;
+use log::info;
 use tiled_pane_grid::{split, TiledPaneGrid, RESIZE_PERCENT};
 
 use crate::{
@@ -1011,7 +1012,7 @@ impl TiledPanes {
             None => false,
         }
     }
-    pub fn move_focus_right(&mut self, client_id: ClientId) -> bool {
+    pub fn move_focus_right(&mut self, client_id: ClientId, single_tab_wraparoud: bool) -> bool {
         match self.get_active_pane_id(client_id) {
             Some(active_pane_id) => {
                 let pane_grid = TiledPaneGrid::new(
@@ -1020,7 +1021,14 @@ impl TiledPanes {
                     *self.display_area.borrow(),
                     *self.viewport.borrow(),
                 );
-                let next_index = pane_grid.next_selectable_pane_id_to_the_right(&active_pane_id);
+                info!("active_pane_id: {:?}", active_pane_id);
+                let mut next_index =
+                    pane_grid.next_selectable_pane_id_to_the_right(&active_pane_id);
+
+                if next_index.is_none() && single_tab_wraparoud {
+                    next_index = Some(PaneId::Terminal(0));
+                }
+                info!("next_index: {:?}", next_index);
                 match next_index {
                     Some(p) => {
                         // render previously active pane so that its frame does not remain actively
@@ -1540,9 +1548,9 @@ impl TiledPanes {
         self.toggle_active_pane_fullscreen(client_id);
     }
 
-    pub fn focus_pane_right_fullscreen(&mut self, client_id: ClientId) {
+    pub fn focus_pane_right_fullscreen(&mut self, client_id: ClientId, single_tab_wraparoud: bool) {
         self.unset_fullscreen();
-        self.move_focus_right(client_id);
+        self.move_focus_right(client_id, single_tab_wraparoud);
         self.toggle_active_pane_fullscreen(client_id);
     }
 
--- a/zellij-server/src/screen.rs
+++ b/zellij-server/src/screen.rs
@@ -5,6 +5,7 @@ use std::collections::{BTreeMap, HashMap, HashSet};
 use std::rc::Rc;
 use std::str;
 
+use log::info;
 use zellij_utils::data::{Direction, Resize, ResizeStrategy};
 use zellij_utils::errors::prelude::*;
 use zellij_utils::input::command::RunCommand;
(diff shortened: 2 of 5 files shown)


zellij-server/src/panes/tiled_panes/mod.rs
```diff
@@ -871,7 +881,13 @@ impl TiledPanes {
                     *self.display_area.borrow(),
                     *self.viewport.borrow(),
                 );
-                let next_index = pane_grid.next_selectable_pane_id_to_the_left(&active_pane_id);
+                let mut next_index = pane_grid.next_selectable_pane_id_to_the_left(&active_pane_id);
+
+                if next_index.is_none() && wrap_around {
+                    let plugin_count = self.plugin_pane_count();
+                    let index = self.panes.len() - plugin_count - 1;
+                    next_index = Some(PaneId::Terminal(index as u32));
```
imsnif, line 889: I don't think this will always work. Maybe we can use `TiledPaneGrid` and `FloatingPaneGrid` to find the leftmost/rightmost pane in such cases?

zellij-server/src/panes/tiled_panes/mod.rs
```diff
@@ -1020,7 +1036,12 @@ impl TiledPanes {
                     *self.display_area.borrow(),
                     *self.viewport.borrow(),
                 );
-                let next_index = pane_grid.next_selectable_pane_id_to_the_right(&active_pane_id);
+                let mut next_index =
+                    pane_grid.next_selectable_pane_id_to_the_right(&active_pane_id);
+
+                if next_index.is_none() && single_tab_wraparoud {
+                    next_index = Some(PaneId::Terminal(0));
```
imsnif, line 1043: Same here as above - I don't think this will always work. IMO it would be best to explicitly search for the leftmost/rightmost pane with `TiledPaneGrid` and `FloatingPaneGrid`

zellij-server/src/screen.rs
```diff
@@ -877,10 +876,12 @@ impl Screen {
...
-            Some(tab) => self
-                .tabs
-                .get(tab)
-                .ok_or_else(|| anyhow!("active tab {} does not exist", tab)),
+            Some(tab) => {
+                return self
+                    .tabs
+                    .get(tab)
+                    .ok_or_else(|| anyhow!("active tab {} does not exist", tab));
+            },
```
imsnif, line 884: Maybe I'm missing something, but... why this change?

zellij-server/src/screen.rs
```diff
@@ -1313,11 +1314,12 @@ impl Screen {
         } else {
             self.get_first_client_id()
         };
+        let wrap_around_flag = if self.tabs.len() == 1 { true } else { false };
```
imsnif, line 1317: Nice

zellij-server/src/screen.rs
```diff
@@ -1654,7 +1658,7 @@ pub(crate) fn screen_thread_main(
                 active_tab_and_connected_client_id!(
                     screen,
                     client_id,
-                    |tab: &mut Tab, client_id: ClientId| tab.move_focus_left(client_id),
+                    |tab: &mut Tab, client_id: ClientId| tab.move_focus_left(client_id, true), // set wrap around flag to true does not cause problem since the action does not change tab
```
imsnif, line 1661: Let's make it `false` then... we can't know what this will be (mis)used for in the future :)

zellij-server/src/screen.rs
```diff
@@ -1679,7 +1683,7 @@ pub(crate) fn screen_thread_main(
                 active_tab_and_connected_client_id!(
                     screen,
                     client_id,
-                    |tab: &mut Tab, client_id: ClientId| tab.move_focus_right(client_id),
+                    |tab: &mut Tab, client_id: ClientId| tab.move_focus_right(client_id, true), // set wrap around flag to true does not cause problem since the action does not change tab
```
imsnif, line 1686: Same as above, let's make it `false`

zellij-server/src/tab/unit/tab_integration_tests.rs
```diff
@@ -3361,6 +3361,61 @@ fn move_focus_down_with_stacked_panes() {
...
+
+    let active_pane_id = if let Some(active_pane) = tab.get_active_pane_id(client_id) {
+        active_pane
+    } else {
+        PaneId::Terminal(u32::MAX)
+    };
+
+    assert_eq!(PaneId::Terminal(0), active_pane_id);
+}
+
```
imsnif, line 3418: Thanks for adding tests!

imsnif (changes requested):
Hey, good job on this! I left some comments on the code.
Also, if you could please merge from main?

zellij-server/src/panes/tiled_panes/mod.rs
Kangaxx-0: Repeat `next_selectable_pane_id_to_the_right` to update `Id` until Option -> None? Does this sound right ?

Kangaxx-0:
> Hey, good job on this! I left some comments on the code. Also, if you could please merge from main?

I merged latest `main`

Kangaxx-0:
@imsnif  I do not why the local build did not see the break, I've fixed it, could you trigger the CI again, and help to merge it

imsnif:
No worries @Kangaxx-0 - I wish my availability was better, but there's lots to do in the project, so I appreciate your patience!

I think my main issue right now is with the `while` loop. I get where you're coming from about wanting to re-use existing logic in order to achieve this feature, but I'm not enthusiastic about overloading that function in this way. This can open us up to quite a few logic issues (an example that comes to mind is: what happens if there's a "wall" of unselectable panes in the middle of the screen?)

We keep track of the pane coordinates, as well as their size. We also know the screen size (display area and viewport) and can use them as well. Could we maybe use these in order to find the edges instead?

Kangaxx-0:
> No worries @Kangaxx-0 - I wish my availability was better, but there's lots to do in the project, so I appreciate your patience!
> 
> I think my main issue right now is with the `while` loop. I get where you're coming from about wanting to re-use existing logic in order to achieve this feature, but I'm not enthusiastic about overloading that function in this way. This can open us up to quite a few logic issues (an example that comes to mind is: what happens if there's a "wall" of unselectable panes in the middle of the screen?)
> 
> We keep track of the pane coordinates, as well as their size. We also know the screen size (display area and viewport) and can use them as well. Could we maybe use these in order to find the edges instead?

Hi @imsnif , I send a new iteration and change how to locate the edge pane based on your feedback, would you take a look

zellij-server/src/panes/tiled_panes/mod.rs
```diff
@@ -917,7 +918,35 @@ impl TiledPanes {
...
+                        .iter()
+                        .filter(|(_, pane)| pane.selectable())
+                        .max_by_key(|(_, pane)| pane.x())
+                        .map(|(_, pane)| pane.x());
+
+                    next_index = if let Some(max_x) = max_x {
+                        let candidates = self
+                            .panes
+                            .values()
+                            .filter(|pane| pane.x() == max_x && pane.selectable())
```
imsnif, line 935: I think this is not accurate. Shouldn't it also include the pane width?

zellij-server/src/panes/tiled_panes/mod.rs
```diff
@@ -917,7 +918,35 @@ impl TiledPanes {
...
+                            .collect::<Vec<_>>();
+
+                        let rightmost = candidates.iter().min_by_key(|pane| {
+                            (pane.y() as isize - self.panes[&active_pane_id].y() as isize).abs()
```
imsnif, line 939: Better to use a `get` here on `self.panes`, even if we unwrap it afterwards, otherwise this is a very invisible crash waiting to happen if someone refactors this wrong.

zellij-server/src/panes/tiled_panes/mod.rs
```diff
@@ -917,7 +918,35 @@ impl TiledPanes {
...
+                        });
+                        Some(
+                            rightmost
+                                .expect("Rightmost candidate should not be empty")
```
imsnif, line 943: Any chance we can somehow merge all of these options to a `map` chain?
Something like (pseudocode): `next_to_the_right().or_else(|| leftmost()).map(|| /* more logic here */)`?

imsnif (changes requested):
Looking much better! I left some comments about the implementation. Also - is there any reason we're only doing right+left and not also up+down?

zellij-server/src/panes/tiled_panes/mod.rs
Kangaxx-0: Updated

zellij-server/src/panes/tiled_panes/mod.rs
Kangaxx-0: I made some changes to merge all into one big iterator call

zellij-server/src/panes/tiled_panes/mod.rs
Kangaxx-0: IMHO, the rightmost selectable pane  = pane.x is the larger number, the width does not really matter here 

I've tested a lot of times, and this solution works fine.

e.g - 
A(current focus) | B | C
-- | -- | --
D | E | F

When I `alt + <-` to move left, I just need to locate the pane with maximum `x` && `selectable = true`, if there are multiple, then the second part of logic comes for the help which is to calculate the `abs` of y -> Pane `C`

Did I miss a scenario that must consider the width?

Kangaxx-0: @imsnif  Sorry for the late reply, i was sick in last a few days. I've made a few changes, now the logic follows:

1. It filters the panes to find the ones with the maximum value of the x-coordinate plus the pane's width.
2. If there are multiple candidates, it further filters the candidates to find the panes with the maximum x-coordinate.
3. If there are still multiple candidates, it selects the one with the minimum gap in the y-coordinate.

Kangaxx-0: > Looking much better! I left some comments about the implementation. Also - is there any reason we're only doing right+left and not also up+down?

I added up + down as well

Kangaxx-0:
> 



> Hey @Kangaxx-0 - I'm really sorry for taking so long to get to this and other PRs. It is definitely still on my TODO, but realistically I'm not sure when I'll manage to get to it. I'm doing considerably more than 100% and I absolutely have to triage and prioritize. I know this sucks, but things that are not on the roadmap (and these days, specifically have to do with my near+medium term for the plugin system) tend to get pushed back.

I totally understood ! Please take as much as time you need

noahsc0tt:
I would still love this! As far as I can tell I don't think a solution has been found/developed yet.

joshdavenport:
Would be lovely to have this looked at again, time allowing of course

PR #2264, opened 2023-03-10
fix: zellij attach when the -c / --create option is passed

dsully:
`assert_session_ne` is called before the value of `-c` is known to create a session when attaching if it doesn't exist.

Move the call (and duplicate) to perform the session name validation, passing a boolean along to denote creation.

### DIFF
--- a/src/commands.rs
+++ b/src/commands.rs
@@ -333,10 +333,6 @@ pub(crate) fn start_client(opts: CliArgs) {
     };
     let os_input = get_os_input(get_client_os_input);
 
-    let start_client_plan = |session_name: std::string::String| {
-        assert_session_ne(&session_name);
-    };
-
     if let Some(Command::Sessions(Sessions::Attach {
         session_name,
         create,
@@ -344,6 +340,10 @@ pub(crate) fn start_client(opts: CliArgs) {
         options,
     })) = opts.command.clone()
     {
+        let start_client_plan = |session_name: std::string::String| {
+            assert_session_ne(&session_name, create);
+        };
+
         let config_options = match options.as_deref() {
             Some(SessionCommand::Options(o)) => config_options.merge_from_cli(o.to_owned().into()),
             None => config_options,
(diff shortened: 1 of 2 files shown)


src/commands.rs
```diff
@@ -333,17 +333,17 @@ pub(crate) fn start_client(opts: CliArgs) {
     };
     let os_input = get_os_input(get_client_os_input);
 
-    let start_client_plan = |session_name: std::string::String| {
-        assert_session_ne(&session_name);
-    };
-
```
pitoniak32, line 339: I am not a maintainer but it seems that the `start_client_plan` is an arrow function definition. It is not actually called here. It is just defined in this scope to avoid needing to define it more than once. So I don't think this is a needed change

dsully: I'm not sure I understand. This is the only place that `assert_session_ne` is called.

imsnif:
I really appreciate your efforts, but... https://github.com/zellij-org/zellij/pull/2247

dsully:
Great, I searched but didn't find that one.

src/commands.rs
pitoniak32: I know the issue was closed but it might be helpful for an explanation on what I mean. so the function `assert_session_ne` is called inside the following function. I was also using the wrong term for what they are considered in rust, arrow function is from javascript but rust they are closures you can check this article out
https://www.educative.io/answers/what-are-closures-in-rust

```rust
let start_client_plan = |session_name: std::string::String| {
    assert_session_ne(&session_name);
};
```
but the `start_client_plan` is just a variable that holds the closure that will call the `assert_session_plan`
```rust
`|session_name: std::string::String| assert_session_ne(&session_name);`
```
so this `assert_session_ne` is actually called [here](https://github.com/zellij-org/zellij/pull/2264/files#diff-b86d43fce6a9c57160aa925719550d40a6ed3f5e5a0117679905f76e68acc520R360) in the `map` function that will apply the function to the session name.

dsully: Thanks, “arrow function” threw me off. I’m very familiar with closures in other languages. 😎

PR #2273, opened 2023-03-11
Fix yanked cpufeatures crate warning

arnabanimesh:
This warning shows up when running install with --locked


imsnif:
I think manually changing the lockfile is not a good idea... do you think maybe we can get to the bottom of this warning otherwise?

Issue #2205, opened 2023-02-28
Support cwd as a zellij option


elijahmanor:
I make a [zellij session creator/switcher](https://gist.github.com/elijahmanor/a8af9a19e7b12b4bde7fa9b9d4b7037d) and a zellij feature that would be nice is having a way to tell zellij which directory to use when a session is created.

Something like the following would be helpful. Thank you for the great project!

`zellij options --cwd <path>`

pitoniak32:
I really like this idea, being able to say what dir to start in for a new session would be very helpful. In particular 
```
zellij -s <session-name> --create --layout <layout> --cwd <dir>
```

imsnif:
This has been implemented (thanks [@Kangaxx-0](https://github.com/Kangaxx-0)!) and is in `main`. Will be released when we next release in the near future.

elijahmanor:
YAY thanks @Kangaxx-0 and @imsnif I look forward to the release!

cblessing24:
Hey I noticed that `~` is not expanded when using it with `--default-cwd`. Is that intentional?

Kangaxx-0:
> Hey I noticed that `~` is not expanded when using it with `--default-cwd`. Is that intentional?

Can you provide more context, please?

cblessing24:
Running `zellij options --default-cwd=~/some/path` seems to ignore the option. I always end up in the home directory. Using `--default-cwd=/home/user/some/path` works as expected.

PR #2290, opened 2023-03-15
Fix 2205 - Support cwd

Kangaxx-0:
#2205 - supports cwd as option

- [x] `Zellij options --default-cwd <path>` without config
- [x]  `Zellij` with `default_cwd` = <path> in zellij config file → set path to `usr`
- [x]   `cargo xtask test`

Screenshot

### DIFF
--- a/zellij-server/src/lib.rs
+++ b/zellij-server/src/lib.rs
@@ -349,6 +349,7 @@ pub fn start_server(mut os_input: Box<dyn ServerOsApi>, socket_path: PathBuf) {
                         ..Default::default()
                     })
                 });
+                let cwd = config_options.default_cwd;
 
                 let spawn_tabs = |tab_layout, floating_panes_layout, tab_name, swap_layouts| {
                     session_data
@@ -358,6 +359,7 @@ pub fn start_server(mut os_input: Box<dyn ServerOsApi>, socket_path: PathBuf) {
                         .unwrap()
                         .senders
                         .send_to_screen(ScreenInstruction::NewTab(
+                            cwd.clone(),
                             default_shell.clone(),
                             tab_layout,
                             floating_panes_layout,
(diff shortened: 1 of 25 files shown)


imsnif (changes requested):
Hey @Kangaxx-0 - good job with this! The code looks great and the solution very native.

Only issue I found is that when the default_cwd is set, whenever we open a new pane it is set to it instead of the current folder. This to me is more of an `override_cwd` than a `default_cwd`.

If we adjust this, I think we're good to merge.

Kangaxx-0:
> Only issue I found is that when the default_cwd is set, whenever we open a new pane it is set to it instead of the current folder. This to me is more of an `override_cwd` than a `default_cwd`.
> 
> If we adjust this, I think we're good to merge.

I changed it to `override_cwd` :)

imsnif:
Hey @Kangaxx-0 - I think we need a `default_cwd` rather than an `override_cwd` (that's also what the issue was about). Would you like to work on it in that direction?

Kangaxx-0:
Hi @imsnif - Sorry, I am bit confused here.  In my initial commit, I used `default_cwd`, and I thought you would prefer `override_cwd` in your first comment, are you saying you feel `default_cwd` is better, and want me to change it back?

Besides the name change, any other change you feel we might need ?

Kangaxx-0:
@imsnif Thanks for your feedback, I've made some small changes, and I believe it works as you expected.

I have not sent new iteration, this is just my local changes, please let me know when I will update the new commit

defaultcwd

imsnif:
Hey @Kangaxx-0 - very elegant solution here! Only problem is that this now only works if a `default_shell` is defined... do you think we can find a solution that will work in any case?

Kangaxx-0:
> Hey @Kangaxx-0 - very elegant solution here! Only problem is that this now only works if a `default_shell` is defined... do you think we can find a solution that will work in any case?

Hello @imsnif, thanks for your comments, in order to make default cwd works without `default_shell`, I made a few changes to pass down a new param.

I've tested all below scenarios and they all worked fine

- [x]  no `default_shell` and  no `default_cwd`
- [x] no `default_shell` and  `default_cwd`
- [x] `default_shell` and  no `default_cwd`
- [x] `default_shell` and `default_cwd`

Please take a look when you get a chance, thanks

PR #2296, opened 2023-03-16
zellij-utils: Check the data dir for themes

cryptomilk:
In a distribution you want to provide themes via the data directory, so a user can specify one in the config file. However it should be possible by the user to proide their own. This patch will allow a distro to provide themes in /usr/share/zellij/themes.

See also https://build.opensuse.org/request/show/1071167


cryptomilk:
Closing in favor of #2307

Issue #2172, opened 2023-02-17
feature / ability to remove "Zellij (session-name)" from top bar


pergamomo:
When using work path as session name (see: https://github.com/zellij-org/zellij/issues/2171 ) the top bar becomes cluttered.

![image](https://user-images.githubusercontent.com/82210878/219704400-68b1d5f4-261a-486a-99b8-f2f32664a333.png)

I would prefer to not have it at all and only show the tabs.

Like this:
![image](https://user-images.githubusercontent.com/82210878/219705861-f687873a-5175-4932-a620-7b340a842faa.png)

Thanks!

mirql:
here is the plugin from zellij docs also without branding:

https://github.com/dj95/zjstatus

survirtual:
> I'm pasting my response to the relevant PR here to make this information easier to find:
> 
> Zellij is an open-source project without any significant funding. This means that in addition to developing and maintaining the software, us maintainers also need to take care of outreach. Help others find out about Zellij in various ways. This is not an easy task when you do not have money to put into it like some other VC-funded tools in this space.
> 
> One of the major ways people find out about Zellij is from screenshots, streams or even just by looking over a friend's shoulder. If we include a config parameter that will allow people to remove our name from the UI - we would be making it more difficult for new users to find out about us, and in turn more difficult for us maintainers to perform outreach (I can testify that doing so as is takes away a significant amount of my time from development).
> 
> If anyone feels strongly about this, they are always welcome to create a version of the UI without our tool's name and maintain it on their own. This is something that will become easier in the coming months as we improve our plugin system.
> 
> So for the time being, until our funding and/or outreach situation changes - I'm going to have to leave this as part of our UI to help more users find out about us. Apologies if this inconveniences anyone - I hope you can see this as a small price to pay for a tool you love.

This is not the way to do it and is very off-putting.  You're making users install a third-party plugin just so they can have a clean experience on a terminal multiplexer?

I'm going to go do some breathing exercises now because I shouldn't be getting this upset over such nonsense.  When you make a useful utility, keep your ego out of it please.  I got my hopes up and invested time to using this tool and have wasted hours chasing for a non-invasive way to remove this branding, as it is a non-starter for me and totally disrupts my mental flow.  This is something that should easily be configurable using a single-line configuration.

[...]

PR #2300, opened 2023-03-17
Add option to hide prefix in tab-bar (#2172)

marcel-baur:
Summary
---
This patch adds a `config.ui.pane_frames.hide_tab_bar_prefix` parameter to hide the `Zellij (<name>)` prefix, as requested in issue #2172 
The test for parsing the UI config has been adjusted to include the new parameter. I'm open to add tests for this functionality, I would however appreciate guidance on where and how to add those specific tests. Thanks!


imsnif:
Hi @marcel-baur - Zellij is an open-source project without any significant funding. This means that in addition to developing and maintaining the software, us maintainers also need to take care of outreach. Help others find out about Zellij in various ways. This is not an easy task when you do not have money to put into it like some other VC-funded tools in this space.

One of the major ways people find out about Zellij is from screenshots, streams or even just by looking over a friend's shoulder. If we include a config parameter that will allow people to remove our name from the UI - we would be making it more difficult for new users to find out about us, and in turn more difficult for us maintainers to perform outreach (I can testify that doing so as is takes away a significant amount of my time from development).

If anyone feels strongly about this, they are always welcome to create a version of the UI without our tool's name and maintain it on their own. This is something that will become easier in the coming months as we improve our plugin system.

So while I very much appreciate the work you put into this, I'm going to have to reject this change.

imsnif:
Yeah, sure. You can change it to `hide_session_name` and I'm totally game for that.

marcel-baur:
Alright, I'll get it done after work 👍🏻

PR #2301, opened 2023-03-17
Add option to hide session name in tab-bar (deriving from #2172)

marcel-baur:
Summary
---
This patch adds a `config.ui.pane_frames.hide_session_name` parameter to hide session name from the the Zellij (<session_name>) prefix, as requested in issue https://github.com/zellij-org/zellij/issues/2172

As discussed in https://github.com/zellij-org/zellij/pull/2300, I have adjusted the issue scope to only give the option hide the session name while keeping the tools name in the tab-bar prefix.

I'm open to add tests for this functionality, I would however appreciate guidance on where and how to add those specific tests. Thanks!

### DIFF
--- a/default-plugins/tab-bar/src/line.rs
+++ b/default-plugins/tab-bar/src/line.rs
@@ -223,7 +223,7 @@ pub fn tab_line(
     cols: usize,
     palette: Palette,
     capabilities: PluginCapabilities,
-    hide_prefix: bool,
+    hide_session_name: bool,
 ) -> Vec<LinePart> {
     let mut tabs_after_active = all_tabs.split_off(active_tab_index);
     let mut tabs_before_active = all_tabs;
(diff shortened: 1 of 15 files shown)


jaeheonji (approved):

jaeheonji:
Hi, @marcel-baur !
LGTM. and If you want to add test case for this feature, you can add e2e test case in `src/tests/e2e/cases.rs`.
After adding the test cases and running the e2e tests locally, you can create a snapshot file.

FYI:
* Instructions for running e2e tests are in the `CONTRIBUTING.md` file.
* You can use [`cargo-insta`](https://github.com/mitsuhiko/insta) to create snapshot files.

marcel-baur:
I've had a quick look through the e2e tests. It doesn't seem like the `rounded-corners` property has a e2e test case either, so I'd leave it be for the moment to be consistent.
Merging will be done by you I guess? Since it requires write permissions on the repo

Bhanukamax:
I tried this on the main branch locally, doesn't seem to work for me, rounded-corners does work but not the hide-session-name, not sure if I'm doing something wrong though, this is what I have in config:
```
ui {
    pane_frames {
        rounded_corners true
        hide_session_name true
    }
}
```

jaeheonji:
When I tested both this branch and the main branch locally for double checking, it worked fine.
If you think something is wrong, could you please create an issue with the current situation(logs, full config, system info and so on)?

pedromfedricci:
My best guess here is that @Bhanukamax is using the compact-bar plugin instead, could you please confirm?
If that is the case, there is now a PR to fix that: #2368.
Ty!

Bhanukamax:
> My best guess here is that @Bhanukamax is using the compact-bar plugin instead, could you please confirm? If that is the case, there is now a PR to fix that: #2368. Ty!

Yes, I'm using the compact-bar

PR #2307, opened 2023-03-20
feat: provide default themes

jaeheonji:
Refer #2223 #2229
Related #2296 

# History

Currently, zellij does not provide themes by default. If the user wants to apply the theme, they must configure the theme file directly or set it in the config file.

A theme is a very good component for organizing an application's community. In fact, many users add various themes to the `example` folder, but it is not possible to use them directly, and it is inconvenient for users to find them.

Therefore, as with other programs, providing themes by default will be very convenient for users.

I previously implemented providing a default theme by including all themes in the zellij binary. However, due to some issues and bugs, it was not included in the current version, and I thought of a better approach instead.

# Proposal

First, basic themes are managed in `zellij-utils/assets`.

As a way to provide basic theme assets to the system, first, include them in zellij release assets. For this we will need to update our CI build. Other than that, it seems that the task of installing the theme to the system from each package manager should be added. (like #2296)

All zellij has to do is add the lookup to the local folder where the default theme is installed. The location of this local folder is considered the [system location](https://zellij.dev/documentation/configuration.html#where-does-zellij-look-for-the-config-file).

# Conclusion

As a result, the space where all the theme settings are located is as follows:
- `config.kdl`
- `$HOME/.config/zellij/themes`
- system location (Perhaps, `/etc/zellij/themes` or `/usr/share/zellij/themes`. I'm not sure yet)


jaeheonji:
Hi, @imsnif! If you have time, could you give me your opinion on this proposal?

imsnif:
Hi @jaeheonji - sorry it took me so long to get to this!

I think this is a good solution, but I want to suggest another thing:
Maybe in addition to what you wrote above, we can do with themes what we do with the config file?
We have a default config file included in the Zellij binary, so when Zellij runs it looks for a config file on the hard drive (eg. in ~/.config/zellij), if it finds one it uses it, if not it uses the one included in the binary.

I think you did something very similar in the previous PR and I liked it. Maybe we just need to adjust it a little so that it behaves like the config? I think what you did here with moving the example themes to the asset folder is a good start. What do you think?

jaeheonji:
Thank you, @imsnif! 

 > I think you did something very similar in the previous PR and I liked it. Maybe we just need to adjust it a little so that it behaves like the config? I think what you did here with moving the example themes to the asset folder is a good start. What do you think?

Yes. The previous way also has many advantages. So, I'll include the theme in the binary as before, but with a bit more stability.

Issue #1948, opened 2022-11-16
Problems with repeatedly attaching to session - leaks memory and finally crashes


raphCode:
Just some stress testing with rapid attaching and detaching yielded this possible server crash locations:

```
Error occurred in server:

  × Thread 'server_listener' panicked.
  ├─▶ At .../zellij-utils/src/ipc.rs:213:37
  ╰─▶ called `Result::unwrap()` on an `Err` value: EMFILE
```

My build is at dc922907eb3aed6fd7a1ae6a2e23f03042801b28

matu3ba:
My gut feeling tells me that there is a fork without CLOEXEC used (either no OCLOEXEC set or no execv being used leading to file descriptor leaking into the child process), but this does not explain the memory leaks.

mfasold:
I ran into this issue in practice. We had a number of users running zellij on a single machine until all instances crashed at some point of time. The logs showed the same `EMFILE` error. 

Later I checked the number of open files, and found that zellij was using about 50k open file descriptors! The majority (85%) of them were Unix socket files for the server-client communication. About half of them were used by the "async-std" thread. 

Any ideas on how to get to the root of this issue? Would it be of any merit to build a server-client toy example that uses a LocalSocketListener in a separate thread?

gbrigandi:
Just submitted a PR that fixes this issue, see #2322 .

raphCode:
Nice! It fixes the crashes.

I fear we might still have a small memory leak tho:
- zellij, default layout with 1 tab, 1 pane, statusbar and tabbar
- directly after start + detach: 631MB resident set size
- 1000x `zellij ls`: 655MB

imsnif:
@raphCode - these might not strictly be leaks, I think @tlinford has done some digging into the rust allocator and might have a relevant link handy?

tlinford:
> @raphCode - these might not strictly be leaks, I think @tlinford has done some digging into the rust allocator and might have a relevant link handy?

Here is what I found about allocators in another issue: https://github.com/zellij-org/zellij/issues/1625#issuecomment-1411735893

raphCode:
I am aware of the allocator playing a role, but this case looks like a leak to me:
Memory rises linearly with attach count, no anomalies. I performed 5000 attach repetitions. Or 5 repetitions of the '1000x attach' case, if you want.

Tell me if any of these would be helpful:
- rechecking with jemalloc branch
- running it under valgrind
- letting the test run longer

PR #2322, opened 2023-03-24
Fixed issue #1948

gbrigandi:
When a client connects to the zellij server, a Unix socket is created. However, a bug in version 1.1.1 and earlier of the interprocess crate used for interprocess communication (IPC) prevented the socket from being properly disposed of after use, which generated a descriptor leak. The bug was fixed in version 1.2.0 of the interprocess crate. To ensure that the zellij client and server use the latest interprocess crate version and properly dispose of the Unix socket upon a client's disconnection, it's necessary to upgrade the zellij-utils crate to use interprocess library version 1.2.1.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -997,28 +997,12 @@ dependencies = [
 ]
 
 [[package]]
-name = "futures"
-version = "0.3.21"
-source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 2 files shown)


imsnif:
Thank you for this!
@raphCode - what do you think?

raphCode:
Thanks, very nice work!
Built the branch and could not crash it even with > 2500 attaches.

---

Hey look @imsnif, a dependency upgrade that actually makes zellij crash less often :P

We can merge this from my side. Is there anything more to do than adding it to the changelog? No experience with dep upgrades.

PR #2323, opened 2023-03-24
Added launch in xonsh

anki-code:
Hi! Thank you for [zellij](https://github.com/zellij-org/zellij)!
Added launch in xonsh.

### DIFF
--- a/README.md
+++ b/README.md
@@ -52,12 +52,16 @@ cargo install --locked zellij
 #### Try Zellij without installing
 
 bash/zsh:
-```
+```bash
 bash <(curl -L zellij.dev/launch)
 ```
 fish:
-```
+```fish
 bash (curl -L zellij.dev/launch | psub)
+```
+xonsh:
+```xsh
+bash -c $(curl -sL zellij.dev/launch)
 ```
 
 ## How do I get involved?
--- a/README.md
+++ b/README.md
@@ -55,13 +55,9 @@ bash/zsh:
 ```bash
 bash <(curl -L zellij.dev/launch)
 ```
-fish:
-```fish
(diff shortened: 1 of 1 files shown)


raphCode:
The context is that xonsh is another shell with slightly differing syntax for process substitution.

I tested it, seems to work.

Is there any possibility of finding a command line that works in bash and xonsh?
I tried `bash -c "$(curl -L zellij.dev/launch)"` , but that fails in xonsh.

raphCode:
> ```shell
> bash -c 'bash -c "$(curl -L zellij.dev/launch)"'
> ```

Yeah, I also thought of using bash twice. I personally prefer this alternative with less quoting:
```bash
bash -c 'bash <(curl -L zellij.dev/launch)'
```
But that is quite a mouthful.

@imsnif What do you think, should we have:
- 3 concise shell commands (bash, fish, xonsh)?
- a short one for bash and the long command for all other shells?
- only the long universal command for all shells?

I am thinking about the website at this point.

Shorter commands are easier to type, but maybe everybody is copy-pasting this anyways (keep in mind that it is not possible in every situation).



> Oh, I will find and add! If you can point me it will be cool. Thanks!

https://github.com/zellij-org/zellij-org.github.io/blob/d780c724990200025075dafca477ef1c1cc58b14/themes/hello-friend-ng/layouts/index.html#L26

But maybe wait until we decided what command(s) we want to have.

imsnif:
I think a short one for bash and a long one for other shells would be good :+1:

raphCode:
@anki-code Would you like to update your PR accordingly and open one for the website too?

raphCode:
Thank you!
We could even write "other shells" on the website to cover all future possibilities, what do you think?

PR #2355, opened 2023-04-11
fix(grid): reap dropped links

tlinford:
This fix solves solves a memory leak by ensuring that when `TerminalCharacters` that hold osc 8 links are removed from the scroll buffer, the related data is cleaned up.

Another fix is to clear all osc8 link data when a pane is reset.

### DIFF
--- a/zellij-server/src/panes/grid.rs
+++ b/zellij-server/src/panes/grid.rs
@@ -1716,6 +1716,7 @@ impl Grid {
         if let Some(images_to_reap) = self.sixel_grid.clear() {
             self.sixel_grid.reap_images(images_to_reap);
         }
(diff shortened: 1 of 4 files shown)


tlinford:
@imsnif I'm a bit fuzzy on the alternate screen handling, i see that the sixel grid gets cleared [here](https://github.com/zellij-org/zellij/blob/00af395bb7745c065643f3204fcecce333ff22f2/zellij-server/src/panes/grid.rs#L2477-L2496), should I do the same for the `link_handler`?

imsnif:
Yes!

tlinford:
Need to figure out how to clear all links for a pane since the link_handler is a singleton

PR #2362, opened 2023-04-13
Introduce `match_error!` for better error-type handling

har7an:
## Rationale

This PR introduce the `match_error!` macro for convenient error handling when using multiple different error types. The errors can be wrapped into higher-level error types such as `anyhow` or `miette` errors.

The macro works by using the `std::error::Error` traits `downcast_ref` method and exposing its' functionality in a `match`-like fashion. This reduces both the amount of boilerplate-code needed for this task, and makes handling of errors, which may contain multiple different low-level error types, easier.


har7an:
@imsnif Here's a draft for the improved error handling, suporting multiple underlying low-level error types. It uses the same `downcast_ref` approach as currently implemented e.g. in `server/pty`, but hides it behind a `match`-like structure. Check out the diff in `server/pty` to see it in action.

I'll add more elaborate docs once I know it's in a state that we will use for the foreseeable future. If you have questions or suggestions, please let me know. Do you think this works better than what we have now?

PR #2368, opened 2023-04-15
fix(plugin): respect hide session option on compact-bar

pedromfedricci:
This PR extends the "hide session name" option  from #2301 for the compact-bar plugin.


jaeheonji (approved):
Thank you! LGTM 👍

PR #2378, opened 2023-04-20
fix: pass cwd to new tab

onichandame:
### Problem

When creating new tab using `Ctrl-T N` key binding, the panes in the new tab should inherit the cwd of the current pane. However not all panes respect that.

<details>
<summary>
My layout
</summary>

```
layout {
    pane size=1 borderless=true{
    }
    pane split_direction="vertical"{
        pane size="70%"{
            pane size="80%" name="editor" focus=true command="bash"
            pane size="20%" name="console"
        }
        pane size="30%" name="side"{
            pane{
                plugin location="zellij:strider"
            }
            pane command="htop"
        }
    }
    pane size=2 borderless=true{
    }
}
```
</details>

<details>
<summary>
reproduction steps
</summary>

image

`Ctrl-T N`

image

Here you can see that the panes in the new tab are in different cwds. the focused pane is in the old cwd while the non-focused pane is in the new(correct) cwd.

My experiment tells that the issue is not about focus, instead it is the `command="bash"` option that prevents the tab from knowing the right cwd.
</details>

### Solution

I am new to zellij and haven't looked into every corner, so I still do not fully understand the issue. For example, I have no clue why the pane without command is in the right cwd. Please point out the missing parts if any.

Here I pass the cwd option to the `NewTab` command and the problem is fixed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 * fix(plugin): respect hide session option on compact-bar (https://github.com/zellij-org/zellij/pull/2368)
 * feat: allow excluding tabs from tab sync in layouts (https://github.com/zellij-org/zellij/pull/2314)
 * feat: support default cwd (https://github.com/zellij-org/zellij/pull/2290)
+* feat: cli action to reload plugins at runtime for easier plugin development (https://github.com/zellij-org/zellij/pull/2372)
 
 ## [0.36.0] - 2023-04-13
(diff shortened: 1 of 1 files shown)


jaeheonji (commented):
Hi, @onichandame 
Sorry for late review.

First, I understood and tested that this change is trying to change the cwd of the pane with the command.

When I tested this in my environment, the cwd still doesn't change in the pane with the command.
I think simply passing cwd to NewTab doesn't seem to solve this problem.

Following the actual NewTab creation process, the actual pane creation logic is handled in `apply_tiled_panes_layout` of `src/tab/layout_applier.rs`.

Pane with RunCommand here seems to not change cwd even when Tab is regenerated.

I think you need to look more into this to solve this problem.

onichandame:
Thanks for you time @jaeheonji . I'll look deeper to find a fix

PR #2396, opened 2023-04-22
improve build/ci times

tlinford:
- avoid building all workspace crates with `cargo x build` (only plugins and main binary)
- only set the target triple in tests for plugins
- add new profile for `cargo x run` to build with optimized dependencies => FAST plugins when developing (thanks [Bevy Book](https://bevyengine.org/learn/book/getting-started/setup/#compile-with-performance-optimizations) for the idea)
- use https://github.com/Swatinem/rust-cache to avoid rebuilding dependencies every time in ci
- split `Build & Test` job into two so they run in parallel
- hopefully improve the flaky tests situation, this also makes the e2e tests run much faster (some tests produced correct snapshots but had some logic errors causing them to loop for much longer than necessary). Add some output to the tests so it is easier to see if something goes wrong.
- remove verbose build output from e2e test build


xtask/src/ci.rs
```diff
@@ -102,25 +102,16 @@ fn e2e_test(sh: &Shell, args: Vec<OsString>) -> anyhow::Result<()> {
...
-                .args(args)
-                .run()
-                .map_err(anyhow::Error::new)
+            cmd!(
+                sh,
+                "{cargo} test --no-default-features -- --ignored --nocapture --test-threads 1"
+            )
+            .args(args)
+            .run()
+            .map_err(anyhow::Error::new)
```
har7an, line 114: This makes zellij take whatever plugins lie in the `zellij-utils/assets/plugin/` folder. Why did you change this? Of course it speeds up the e2e tests, but we do not guarantee that the content of `zellij-utils/assets/plugins` is up to date and matches the current zellij version. That's only the case when building the plugins in release mode first, because then they're copied into the `assets/plugins` folder. Or does the e2e binary take the plugins from somewhere else entirely? Afaik it did a while ago, but iirc I changed this to make it bundle the plugins into the e2e binary.

har7an (commented):
The rest of the changes look alright to me! Haven't tested the speedup myself, though. Do you have a few numbers to share maybe?

xtask/src/ci.rs
tlinford: This had me confused for a while too. But as far as I understand it, we just need any plugins to be available so that the test binary can be built (because it depends on zellij_utils i think), so we might as well just use the already built ones.  And we already have them because we build the in `e2e_build`. 

The e2e test binary in the end is just running the release built zellij binary (built in `e2e_build`) through ssh machinery, and setting it up to use a custom 'e2e-data' dir so that it finds the release optimized plugins (built in `e2e_build`).

Probably I should change it to ensure that e2e_test always calls e2e_build, and have it as a single step in the github flow. That way we calling the command locally everything should always be up to date.

Hope the explanation makes sense, it's all a bit convoluted :)

tlinford:
> The rest of the changes look alright to me! Haven't tested the speedup myself, though. Do you have a few numbers to share maybe?

Sure. A simple test on my machine done by running `cargo clean`, followed by each command, on main and on this pr's branch:

|command|main|branch|
|---------------|-------|----------|
|cargo x build|99 s|53 s|
|cargo x build --release|258 s|210 s|
|cargo x test|132 s|110 s|
|cargo x ci e2e --test|596 s|331 s|

CI runs on average seem to be 3 times faster when using the cache.

PR #2411, opened 2023-04-28
Fix error loading non-existant themes directory and use default themes as the base when merging

Imberflur:
When I cloned the repo and ran `cargo xtask run` I encountered an issue where the `zellij` was trying to load a themes directory at `~/.config/zellij/themes` that did not exist.

If looks like a recent PR (https://github.com/zellij-org/zellij/pull/2307), removed the `if theme_dir.is_dir() {` that was checking if the theme directory exists before trying to use it. I assume we do want to produce a hard error when the theme dir is specifically specified in the config, so I added an equivalent check that only occurs in the case where the theme dir is **not** specified in the config.

While fixing this, I also noticed that the default themes are applied over the top of the themes specified in the config. Presumably, the opposite is better since it allows overriding the defaults (and is more consistent since the themes from the themes directory are applied on top of everything else and so are able to override the defaults regardless).

One other thing I noticed (but haven't made any changes on), is that there is an option in `Command::Setup` called `clean` that indicates to only use the defaults for the config provided by zellij. I was wondering, if this should also apply to themes, since right now (afaict from the code) zellij will still load themes from the themes directory with this enabled.

P.S. I joined the discord server, feel free to contact me there as `@imbris`

### DIFF
--- a/zellij-utils/src/setup.rs
+++ b/zellij-utils/src/setup.rs
@@ -322,14 +322,16 @@ impl Setup {
 
         config.themes = get_default_themes().merge(config.themes);
 
-        let user_theme_dir = config_options.theme_dir.clone().or_else(|| {
-            get_theme_dir(cli_args.config_dir.clone().or_else(find_default_config_dir))
-                // If theme dir is not explicitly specified in config_options,
-                // only try to use it if it exists.
-                .filter(|dir| dir.exists())
-        });
-        if let Some(user_theme_dir) = user_theme_dir {
-            config.themes = config.themes.merge(Themes::from_dir(user_theme_dir)?);
+        if let Some(Command::Setup(Setup { clean: false, .. })) = &cli_args.command {
+            let user_theme_dir = config_options.theme_dir.clone().or_else(|| {
+                get_theme_dir(cli_args.config_dir.clone().or_else(find_default_config_dir))
+                    // If theme dir is not explicitly specified in config_options,
+                    // only try to use it if it exists.
+                    .filter(|dir| dir.exists())
+            });
+            if let Some(user_theme_dir) = user_theme_dir {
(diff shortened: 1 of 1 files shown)


imsnif:
Tagging @tlinford to take a look as he also encountered this problem (or a similar one?)

tlinford:
> Tagging @tlinford to take a look as he also encountered this problem (or a similar one?)

Problem is fixed with these changes.

imsnif:
Thanks @tlinford - now that we know this is the same issue, I'm going to tag @jaeheonji for this. What do you think about these changes?

jaeheonji:
@imsnif @Imberflur 
I like this change!

But, The only question is, what are the advantages of not providing a default theme via `clean`?

If the same theme exists, there will be no problem because the user's theme will be prioritized.
Even if we use `clean`, there are still themes inside the zellij binary.
(If what I understand is correct, it just doesn't merge the data of the default theme in the config.)

If there is a case where `clean` can provide a better UX, I have no reason not to agree to apply it.

Imberflur:
(FWIW I'm not familiar at all with the usage of `zellij setup --clean` so this idea may be completely irrelevant.)

> But, The only question is, what are the advantages of not providing a default theme via clean?

I was imaging that the default theme would be provided but the user's theme(s) (from the themes directory) would not be loaded if `--clean` is used.

Imberflur:
Is `zellij setup --clean` just used for overwriting the config file? So anything unrelated to this doesn't matter?

jaeheonji:
@Imberflur Oh I see, it make sense.

> So I think it is an improvement to not load from the themes directory if --clean is used. Does that sound correct?

Sure, I totally agree with this idea.

Imberflur:
@jaeheonji I pushed a commit that should implement this

jaeheonji (approved):
LGMT

PR #2412, opened 2023-04-28
Add file path context to all IO errors in ConfigError

Imberflur:
When I encountered the issue noted in https://github.com/zellij-org/zellij/pull/2411, it was a bit difficult to understand what was happening since the missing file path was not included in the displayed error.

This PR removes `ConfigError::Io` and replaces all usages of it with `ConfigError::IoPath`. Hopefully, this will improve the experience when debugging unexpected errors in the future (although there may be more room to add useful context, such as where in the code these errors are being produced).


imsnif:
Thank you!

PR #2422, opened 2023-05-04
CI: create custom source artifact with rendered manpages

tranzystorekk:
Rationale: this change is useful for packaging in distributions that build zellij from source. Generating a rendered `zellij.1` manpage eliminates a build dependency and reduces guesswork when determining how to build it.

I picked `pandoc` for CI as it is a reasonable and versatile standard.

I briefly tested the Release workflow on my fork with success.

Highlights:

- Adjusted `docs/MANPAGE.md` format to what pandoc accepts
- Added generating the manpage and compressing it into a custom source code artifact in CI

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -119,6 +119,55 @@ jobs:
           asset_name: zellij-${{matrix.target}}.sha256sum
           asset_content_type: text/plain
 
+  prepare-source-release:
+    runs-on: ubuntu-latest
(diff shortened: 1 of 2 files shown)


tranzystorekk:
After some discussion, the manpage is outdated and doesn't make much sense to be rendered in the current state.

Possibly a candidate for adding a deprecation notice in favor of `--help` and the online docs.

Issue #2297, opened 2023-03-16
[Proposal] New theme definition spec


cyberglot:
As pointed out in issue #2160, the theme definition spec doesn't follow any rules besides “green is the main colour for some reason”. When scrolling through the theme gallery page, I thought the same image was mistakenly being rendered for all themes. But no, they're all green-biased.

Here's some pictures of what I had to do to restore the dracula theme vibes:
![Screenshot 2023-03-16 at 11 26 17](https://user-images.githubusercontent.com/609862/225608343-871b5c6c-1c7e-4328-94ab-a83d85d41b01.png)
![Screenshot 2023-03-16 at 11 42 52](https://user-images.githubusercontent.com/609862/225608362-43b41188-1b52-4b69-9078-e72f86bde049.png)
(It's important to notice that the current dracula theme absolutely follows a coherent implementation given zellij's interface.)

While it works, it's highly confusing and frustrating, and I believe the dev experience here could be vastly improved. For single file configurations or scripts, I use helix, and I've just noticed that they have a pretty good spec for themes:

```toml
# User Interface
# --------------
"ui.background" = { fg = "text", bg = "base" }

"ui.linenr" = { fg = "surface1" }
"ui.linenr.selected" = { fg = "lavender" }

"ui.statusline" = { fg = "text", bg = "mantle" }
"ui.statusline.inactive" = { fg = "surface2", bg = "mantle" }
"ui.statusline.normal" = { fg = "base", bg = "lavender", modifiers = ["bold"] }
"ui.statusline.insert" = { fg = "base", bg = "green", modifiers = ["bold"]  }
"ui.statusline.select" = { fg = "base", bg = "flamingo", modifiers = ["bold"]  }

"ui.popup" = { fg = "text", bg = "surface0" }
"ui.window" = { fg = "crust" }
"ui.help" = { fg = "overlay2", bg = "surface0" }

"ui.bufferline" = { fg = "surface1", bg = "mantle" }
"ui.bufferline.active" = { fg = "text", bg = "base", modifiers = ["bold", "italic"] }
"ui.bufferline.background" = { bg = "surface0" }

"ui.text" = "text" 
"ui.text.focus" = { fg = "text", bg = "surface0", modifiers = ["bold"] }

"ui.virtual" = "overlay0"
"ui.virtual.ruler" = { bg = "surface0" }
"ui.virtual.indent-guide" = "surface0"

"ui.selection" = { bg = "surface1" }

[...]

PR #2429, opened 2023-05-06
fix: add calculation of theme hue based on bg

### LINKED ISSUE #2428: Theme Hue no longer calulcated
It seems like the theme hue is no longer calculated. This causes light themes such as [catppuccin-latte](https://github.com/catppuccin/zellij) to be nearly impossible to use, as the contrast of some text is too low. 

After some investigation I figured out that `detect_theme_hue` [1] is no longer being called when deserializing themes.
This causes all themes to be initialized with the default value of ThemeHue, which is Dark. 

[1] https://github.com/zellij-org/zellij/blob/ecd63a100b9d927d78c64cda73bae5ec16cac419/zellij-utils/src/shared.rs#L115

**Basic information**
`zellij --version`: zellij 0.36.0
`stty size`: 26 127
`uname -av` or `ver`(Windows): Linux patrick-laptop 6.3.1-arch1-1 #1 SMP PREEMPT_DYNAMIC Mon, 01 May 2023 17:42:39 +0000 x86_64 GNU/Linux
`alacritty --version`: alacritty 0.12.0 (5a728195)

**Further information**
Install catppuccin-latte theme, run zellij

patrickpichler:
It seems like during the transition from yaml based themes to kdl based
config, the theme hue calculation was somehow dropped. It has been added
again.

This fixes the issue that zellij light themes have bad contrast.

Additionally, the catppuccin-light theme has been fixed, since in order to counter 
the missing hue calculation, it tried to swap black and white colors. This still caused
some low contrast text.

fixes #2428

### DIFF
--- a/zellij-utils/assets/themes/catppuccin.kdl
+++ b/zellij-utils/assets/themes/catppuccin.kdl
@@ -12,8 +12,8 @@ themes {
     magenta "#ea76cb" // Pink
     orange "#fe640b" // Peach
     cyan "#04a5e5" // Sky
-    black "#dce0e8" // Crust
-    white "#4c4f69" // Text
+    white "#dce0e8" // Crust
+    black "#4c4f69" // Text
   }
 
   catppuccin-frappe {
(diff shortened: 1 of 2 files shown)


imsnif:
Hey, while I appreciate the proposal to fix this - this would be a breaking change for themes and I'd much rather skip it and go through https://github.com/zellij-org/zellij/issues/2297

Also - to my memory the theme detection never really worked and was dropped almost immediately after it was introduced.

patrickpichler:
The problem is, that zellij is currently pretty much unusable with a light theme. Since #2297 will anyway be 
a breaking change, what do you think of exposing the `theme_hue` field to the theme, so that at least people
can set it manually?

patrickpichler:
Fine with me 🙂 
Would you mind sharing the theme you tried?
Catppuccin latte is pretty hard to use as a beginner, since I am kinda depend on the status bar hints 😅

I will close the PR and the issue then 👍

Issue #2421, opened 2023-05-03
`zellij setup --check` returns wrong info


azinsharaf:
`zellij setup --check` confirms that `config.kdl` is in the correct location but when i luanch it it doesn't read the config file at all.
```
➜  ~ zellij setup --check
[Version]: "0.36.0"
[CONFIG DIR]: "/home/azin/.config/zellij"
[CONFIG FILE]: "/home/azin/.config/zellij/config.kdl"
[CONFIG FILE]: Well defined.
[DATA DIR]: "/home/azin/.local/share/zellij"
[PLUGIN DIR]: "/home/azin/.local/share/zellij/plugins"
 Builtin, default plugins will not be loaded from disk.
 Create a custom layout if you require this behavior.
[LAYOUT DIR]: "~/.config/zellij/layouts"
[SYSTEM DATA DIR]: "/usr/share/zellij"
[ARROW SEPARATOR]: 
 Is the [ARROW_SEPARATOR] displayed correctly?
 If not you may want to either start zellij with a compatible mode: 'zellij options --simplified-ui true'
 Or check the font that is in use:
 https://zellij.dev/documentation/compatibility.html#the-status-bar-fonts-dont-render-correctly
[MOUSE INTERACTION]:
 Can be temporarily disabled through pressing the [SHIFT] key.
 If that doesn't fix any issues consider to disable the mouse handling of zellij: 'zellij options --disable-mouse-mode'
[DEFAULT EDITOR]: nvim
[FEATURES]: []
[DOCUMENTATION]: zellij.dev/documentation
```
if I move the `config.kdl` to the `config` folder :
```
➜  zellij tree
.
├── config
│   └── config.kdl
└── layouts
    └── default.kdl
```
then running zellij DOES recegnize the config file but `zellij check --setup` retruns the following error:
```
➜  zellij zellij setup --check
[Version]: "0.36.0"
[CONFIG DIR]: "/home/azin/.config/zellij"
[CONFIG FILE]: "/home/azin/.config/zellij/config.kdl"
[CONFIG ERROR]: IoError: No such file or directory (os error 2), File: /home/azin/.config/zellij/config.kdl
[DATA DIR]: "/home/azin/.local/share/zellij"
[PLUGIN DIR]: "/home/azin/.local/share/zellij/plugins"
 Builtin, default plugins will not be loaded from disk.
 Create a custom layout if you require this behavior.
[LAYOUT DIR]: "/home/azin/.config/zellij/layouts"
[...]

PR #2443, opened 2023-05-11
Fix issue 2421 - Update config file output

Kangaxx-0:
#2421  change the output `CONFIG FILE` to `LOOKING FOR CONFIG FILE FROM`, hopefully, this change could avoid the confusion

### DIFF
--- a/zellij-utils/src/setup.rs
+++ b/zellij-utils/src/setup.rs
@@ -452,7 +452,12 @@ impl Setup {
             .unwrap();
             match Config::from_path(&config_file, None) {
                 Ok(_) => message.push_str("[CONFIG FILE]: Well defined.\n"),
-                Err(e) => writeln!(&mut message, "[CONFIG ERROR]: {}", e).unwrap(),
+                Err(e) => writeln!(
+                    &mut message,
+                    "[CONFIG ERROR]: {}. \n By default, zellij loads default configuration",
+                    e
+                )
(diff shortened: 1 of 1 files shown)


zellij-utils/src/setup.rs
```diff
@@ -444,8 +444,12 @@ impl Setup {
...
-            // match Config::new(&config_file) {
+            writeln!(
+                &mut message,
+                "[LOOKING FOR CONFIG FILE FROM]: {:?}",
+                config_file
+            )
+            .unwrap();
             match Config::from_path(&config_file, None) {
                 Ok(_) => message.push_str("[CONFIG FILE]: Well defined.\n"),
                 Err(e) => writeln!(&mut message, "[CONFIG ERROR]: {}", e).unwrap(),
```
jaeheonji, line 455: How about adding this line here like following?
I think that would reduce a bit more confusion.

```
...
[CONFIG ERROR]: IoError: No such file or directory (os error 2), File: ...
 By default zellij will load default configuration
...
```

jaeheonji (commented):
Thank you @Kangaxx-0 !

I think it is good.
And I have an idea so leave a comment.

zellij-utils/src/setup.rs
Kangaxx-0: Good suggestion!

jaeheonji (approved):

PR #2450, opened 2023-05-16
chore: add rust-analyzer to toolchain

onichandame:
For better DX as rust-analyzer has joined the rustup's components collection

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -776,11 +776,12 @@ checksum = "3c877555693c14d2f84191cfd3ad8582790fc52b5e2274b40b59cf5f5cea25c7"
 
 [[package]]
 name = "dialoguer"
-version = "0.10.1"
+version = "0.10.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "d8c8ae48e400addc32a8710c8d62d55cb84249a7d58ac4cd959daecfbaddc545"
+checksum = "59c6f2989294b9a498d3ad5491a79c6deb604617378e1cdc4bfc1c1361fe2f87"
 dependencies = [
  "console",
+ "shell-words",
  "tempfile",
  "zeroize",
 ]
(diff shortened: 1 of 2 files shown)


imsnif:
Hey @onichandame - thanks for this PR. Can you please help me understand why this is needed? Personally I prefer to maintain as few things as possible and am happy to let users install whatever they want for their environments...

onichandame:
@imsnif Hi, I am using helix as editor with the rust-analyzer installed globally with rust 1.69.0 . When editing zellij sources, the lsp throws an error saying that rust-analyzer for target 1.67.0-x86_64-unknown-linux-gnu(as required by rust-toolchain.toml) is not installed . After adding the `rust-analyzer` to the components section and run `cargo xtask`, the error is gone.

I am actually wondering how other people are handling this situation. Can it be approached differently?

jaeheonji:
@onichandame

I have already installed rust-analyzer globally and am using it with Helix. it works fine for me.
I think your problem is caused by installing a specific version of the Rust (1.69.0) globally.
For Rust development, I recommend toolchain management through `rustup`.

onichandame:
@jaeheonji Mind sharing your setup in details? I installed rustup and rust-analyzer the way recommended by the official docs. for rust-analyzer, I installed it by `rustup component add rust-analyzer` as per [the docs](https://rust-analyzer.github.io/manual.html#rustup). how did you install it?

jaeheonji:
@onichandame 

In official docs,

> However, in contrast to component add clippy or component add rustfmt, this does not actually place a rust-analyzer binary in ~/.cargo/bin, see [this issue](https://github.com/rust-lang/rustup/issues/2411). You can find the path to the binary using:

Because of this issue, I don't prefer to install rust-analyzer with rsutup. I installed it with package manager (nix, pacman apt and so on)

Also check that Helix successfully found the rust-analyzer binary with `hx --health`.

And this content seems to be a different topic from zellij. Is it okay if I close the issue?

onichandame:
@jaeheonji Thanks for the info. I would prefer to manage all the components using rustup instead. Just found out that by running `rustup toolchain install 1.67.0-x86_64-unknown-linux-gnu -c rust-analyzer`, the required rust-analyzer is installed.

PR #2455, opened 2023-05-17
fix: cwd of newtab action

onichandame:
context: #2378


jaeheonji (approved):
Sorry for late! LGTM👍

PR #2460, opened 2023-05-18
fix 2459 - the banner does not show

### LINKED ISSUE #2459: First run banner does not show in the first time
1. With a `Layout` like below

```
  layout {
    pane size=1 borderless=true {
    }
      pane command="dust"
    pane size=2 borderless=true {
    }
}
```
2. run zellij. In the first time, the banner should display 

image

NOTE - Since the on_hold_exit is the default behavior of command panes, so would it better to show the banner all the time?

Kangaxx-0:
Fix #2459, the reason the banner does not show is that the callback triggers the screen instruction of `HoldPane`, the first_run is set to `false` before rendering

The fix is to display the banner every time when the command reruns, this might not be a perfect fix, so suggestion is welcome

### DIFF
--- a/zellij-server/src/lib.rs
+++ b/zellij-server/src/lib.rs
@@ -524,14 +524,6 @@ pub fn start_server(mut os_input: Box<dyn ServerOsApi>, socket_path: PathBuf) {
                     .as_ref()
                     .unwrap()
                     .senders
-                    .send_to_screen(ScreenInstruction::RemoveClient(client_id))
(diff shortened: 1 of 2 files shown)


Kangaxx-0:
Its tough to work on the e2e test....everything was passed on my local, but got failed in CI

Kangaxx-0:
hmm...I've spent a lot of time, but still could not figure out why my local worked, and CI is still failing...

image

PR #2464, opened 2023-05-20
Enable sending ctrl-g

coughingmouse:
Solves https://github.com/zellij-org/zellij/issues/418

### DIFF
--- a/zellij-utils/assets/config/default.kdl
+++ b/zellij-utils/assets/config/default.kdl
@@ -5,7 +5,7 @@ keybinds {
         // bind "Alt c" { Copy; }
     }
     locked {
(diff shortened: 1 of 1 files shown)


jaeheonji:
Hi, @coughingmouse 

I saw an issue (#418) with this PR and I know some people want that key binding (`CTRL-G`), but I think it takes discussion and persuasion about that key to replace a new key value (`CTRL-F`).

Create an issue for the key you want to change, and when a consensus is reached about whether the new key is reasonable through discussion, you can create a PR again.

The reason for the default conservative approach to this issue is that zellij provides the ability to reset keymaps by default.

Issue #2249, opened 2023-03-08
Theme change detection during runtime


souserge:
**Basic information**

`zellij --version`: 0.35.1
`uname -av` or `ver`(Windows): Linux fedora 6.1.14-200.fc37.x86_64 #1 SMP PREEMPT_DYNAMIC Sun Feb 26 00:13:26 UTC 2023 x86_64 x86_64 x86_64 GNU/Linux

**Further information**
I switch my GNOME system theme between light and dark on a daily basis to avoid eye strain. With that, I'd like other tools to follow system settings. I set up my terminal (Alacritty) to switch themes automatically using [color-scheme-monitor developed by Martin Kühl](https://codeberg.org/mkhl/color-scheme-monitor). The tool simply edits the config file of a specified program when system settings change. When I tried the same hack with Zellij, it didn't immediately react to changes in config like Alacritty does. It would be nice to have this in Zellij as well. Detaching from a session doesn't seem to react to config changes either, so the only I found so far is starting a new session.

If the config cannot be reloaded automatically, is there a way to force a config reload using a CLI command?

I guess this issue is related to https://github.com/zellij-org/zellij/pull/1242 and the "Have UI reflect config changes" item on the roadmap.

jurica:
I would also love to see this!

0rphee:
Yes! this would be great! Maybe this could also be solved sending a signal to the process, like helix does:

[Helix docs](https://docs.helix-editor.com/master/configuration.html?highlight=usr1#configuration
)
> Additionally, you can reload the configuration file by sending the USR1 signal to the Helix process on Unix operating 
systems, such as by using the command `pkill -USR1 hx`.

0rphee:
@souserge glad it could help you! i also have some scripts for changing the alactritty theme and wezterm theme in case you need help for either! :)

masonmcelvain:
Sorry for resurrecting this thread, but it seems to be the closest open issue with respect to reloading the configuration file.

I notice that the roadmap indicates that the UI should now reflect config changes. It doesn't link to anything so hard to say when/where it was implemented.

![image](https://github.com/zellij-org/zellij/assets/52104630/2c035598-5aa6-4569-87c9-12ab3f05097e)

In practice I do not see the UI reflect config changes in active sessions (version `0.39.2`). I also do not see anything in the docs about reloading the configuration file. Maybe I'm missing something, because the roadmap says this is implemented.

[...]

PR #2469, opened 2023-05-23
only load colors once in default plugins

jurica:
This is a small improvement related to #2249 and #2297.
It's meant to be an intermediate step that immediately help out while waiting for the bigger improvements in this area.


imadnyc:
Would be awesome if someone could take a look at this!

PR #2480, opened 2023-05-25
feat(client): introduce ZELLIJ_TMP_LOG_FILE env variable

### LINKED ISSUE #2465: Env variable with a path to the zellij log file
My use case:
For debug purposes I'd like to view the logs in realtime, preferable in zellij itself :). Introducing env variable that has a path to the zellij log file will solve my issue.

I did this locally for myself in a hacky way and even created a layout that runs a script which does `tail -F $ZELLIJ_TMP_LOG_FILE`. 

Interested to know your thoughts if this is valuable in zellij.
Are there other ways to achieve what I need?

zummenix:
This introduces `ZELLIJ_TMP_LOG_FILE` env variable in zellij session. I believe this is useful is several ways. Developers can create some automation or even layouts that include a pane which runs a script with `tail -F $ZELLIJ_TMP_LOG_FILE` for debugging purposes. For users the docs can be updated with a note how to access the log file and share it.

Closes #2465

### DIFF
--- a/zellij-utils/src/envs.rs
+++ b/zellij-utils/src/envs.rs
@@ -31,6 +31,11 @@ pub fn get_socket_dir() -> Result<String> {
     Ok(var(SOCKET_DIR_ENV_KEY)?)
 }
 
+pub const ZELLIJ_TMP_LOG_FILE_ENV_KEY: &str = "ZELLIJ_TMP_LOG_FILE";
+pub fn set_tmp_log_file(v: String) {
+    set_var(ZELLIJ_TMP_LOG_FILE_ENV_KEY, v);
+}
+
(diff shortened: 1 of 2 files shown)


zellij-utils/src/envs.rs
```diff
@@ -31,6 +31,11 @@ pub fn get_socket_dir() -> Result<String> {
     Ok(var(SOCKET_DIR_ENV_KEY)?)
 }
 
+pub const ZELLIJ_TMP_LOG_FILE_ENV_KEY: &str = "ZELLIJ_TMP_LOG_FILE";
```
zummenix, line 34: note: I was following the architecture of the module and made the const public but I'm not sure why thees keys need to be public.

PR #2488, opened 2023-05-27
Add option to invert selection colors

### LINKED ISSUE #2395: feature request: invert bg and fg colors in selection
It may be useful to invert the colors with the following settings.

```kdl
ui {
    selection {
        invert_colors true
    }
}
```

In current behavior, zellij uses `bg` color as selection background color.
if terminal emulator has similar color scheme to zellij, this behavior reduces visibility.
I tried invert bg and fg colors in selection, visibility looks better.

image

tarkah:
Resolves #2395 

This adds a new option that inverts the fg / bg colors of the selection instead of coloring w/ palette.black.

I've just passed along a `bool` with `chunk_selection_and_colors` but I can refactor into a better defined enum type if desired.

### DIFF
--- a/zellij-client/src/lib.rs
+++ b/zellij-client/src/lib.rs
@@ -168,6 +168,7 @@ pub fn start_client(
             colors: palette,
             rounded_corners: config.ui.pane_frames.rounded_corners,
             hide_session_name: config.ui.pane_frames.hide_session_name,
+            invert_selection_colors: config_options.invert_selection_colors.unwrap_or_default(),
         },
         keybinds: config.keybinds.clone(),
     };
--- a/zellij-server/src/panes/grid.rs
+++ b/zellij-server/src/panes/grid.rs
@@ -1060,6 +1060,7 @@ impl Grid {
                     self.selection,
                     background_color,
                     None,
+                    style.invert_selection_colors,
                     content_x,
                     content_y,
                 );
@@ -1084,6 +1085,7 @@ impl Grid {
                             *res,
                             background_color,
                             Some(foreground_color),
+                            false,
                             content_x,
                             content_y,
                         );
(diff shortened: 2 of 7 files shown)


zellij-client/src/lib.rs
```diff
@@ -168,6 +168,7 @@ pub fn start_client(
             colors: palette,
             rounded_corners: config.ui.pane_frames.rounded_corners,
             hide_session_name: config.ui.pane_frames.hide_session_name,
+            invert_selection_colors: config_options.invert_selection_colors.unwrap_or_default(),
```
tarkah, line 171: This was initially confusing for me... why do we pass in `config` and `config_options` separately? I originally accessed this via `config.options` but apparently the CLI options flags aren't sync'd to this. I spent quite a bit of time scratching my head. Maybe some newtype safety can help here if this separation is intentional.

zellij-server/src/panes/grid.rs
```diff
@@ -1052,20 +1052,21 @@ impl Grid {
                 .selection
                 .contains_row(character_chunk.y.saturating_sub(content_y))
             {
-                let background_color = match style.colors.bg {
-                    PaletteColor::Rgb(rgb) => AnsiCode::RgbCode(rgb),
-                    PaletteColor::EightBit(col) => AnsiCode::ColorIndex(col),
+                let selection = self.selection.offset(content_x, content_y);
```
tarkah, line 1055: I've moved the selection offsetting here as depending on `offset_*` in `add_selection_and_colors` or now `SelectionColors::invert / discrete` seems like a layer violation. Those methods expect a `Selection` for determining which color styles to apply, but we shouldn't leak another implementation detail into them. 

However, this doesn't eliminate the risk of forgetting to call `Selection::offset`. Maybe returning an `OffsetSelection` newtype from `Selection::offset` and depending on that in `SelectionColors` would ensure this guarantee.

tarkah: I've added this guarantee in  54c107af. Now the compiler will complain if you try to pass a `Selection` to `SelectionColors`, so we _must_ call `offset` first.

jgayfer:
Thanks for taking this on. Hopefully gets some traction!

PR #2492, opened 2023-05-29
Initial implementation of --only-unattached

catlee:
This PR adds a new flag to `zellij attach` which will allow attaching only to unattached sessions.

It also updates the output of `zellij ls` to indicate which sessions currently are attached to a client.

The goal is to enable behaviour described in https://github.com/zellij-org/zellij/discussions/2461


catlee:
Is it possible to get this PR reviewed?

PR #2535, opened 2023-06-13
Feature/go to tab name cwd

pitoniak32:
when using --create flag, allow cwd to be specified if tab needs to be created.

This will be very useful when using in a script to open a new tab per project.

Example:
```bash
zellij action go-to-tab-name --create test --cwd ~/.local/bin
```
when a tab named `test` does not exist, should use `~/.local/bin` as the start directory.

This is my first PR, and first time attempting to contribute to OSS. I really enjoy this project, and would love any recommendations / comments / concerns!

Thank you for your time!

### DIFF
--- a/zellij-client/src/input_handler.rs
+++ b/zellij-client/src/input_handler.rs
@@ -308,7 +308,7 @@ impl InputHandler {
             | Action::GoToPreviousTab
             | Action::CloseTab
             | Action::GoToTab(_)
-            | Action::GoToTabName(_, _)
+            | Action::GoToTabName(_, _, _)
             | Action::ToggleTab
             | Action::MoveFocusOrTab(_) => {
                 self.command_is_executing.blocking_input_thread();
--- a/zellij-server/src/plugins/zellij_exports.rs
+++ b/zellij-server/src/plugins/zellij_exports.rs
@@ -921,7 +921,7 @@ fn host_go_to_tab_name(env: &ForeignFunctionEnv) {
     wasi_read_string(&env.plugin_env.wasi_env)
         .and_then(|tab_name| {
             let create = false;
-            let action = Action::GoToTabName(tab_name, create);
+            let action = Action::GoToTabName(tab_name, create, None);
             apply_action!(action, error_msg, env);
             Ok(())
         })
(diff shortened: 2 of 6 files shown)


silicakes:
Great initiative! But why not make it available for all entities - i.e panes and windows too.
We can maybe pass the`-cwd` option after the main `zellij` command and set it globally so that every entity will be able to consume it if it wants to:

`zellij --cwd /dev/null action ...`

The action command will then make it available for everyone to consume.

pitoniak32:
> 

It would make sense to have it available in other cases too like for creating panes and such, but I am not sure about the globally, that would seem to indicate that all commands have a use for the flag. But I do see your point, once I have a second I will think about how that might work 👍🏼!

silicakes:
> > Great initiative! But why not make it available for all entities - i.e panes and windows too.
> > We can maybe pass the-cwd option after the main zellij command and set it globally so that every entity will be able to consume it if it wants to:
> > zellij --cwd /dev/null action ...
> > The action command will then make it available for everyone to consume.
> 
> @silicakes It would make sense to have it available in other cases too like for creating panes and such, but I am not sure about the globally, that would seem to indicate that all commands have a use for the flag. But I do see your point, once I have a second I will think about how that might work 👍🏼!

Oh this should totally be optional and just use the sessions root as default (like today), the main point is about having the other members doing it too.
Thanks!

pitoniak32:
@silicakes After looking at the existing commands I think that only setting this flag for the one action makes sense. Adding it as a base flag that is optional might be confusing since it would be a no-op on commands that are not using the value.

existing ex:
- [NewPane](https://github.com/zellij-org/zellij/blob/317bdfc90252058d51b3f2a1847b2074baf5e80e/zellij-utils/src/cli.rs#L265)
- [NewTab](https://github.com/zellij-org/zellij/blob/317bdfc90252058d51b3f2a1847b2074baf5e80e/zellij-utils/src/cli.rs#L371)

silicakes:
@pitoniak32 

I get your point.
I wish the api would look something like:

```sh
> zellij new pane/tab/window/session -c your/path/to/stuff 
```
check this out: https://github.com/tmux/tmux/wiki/Advanced-Use#working-directories

Let me know if it clarifies things for you or whether it can be bound to all three (four?) options

pitoniak32:
No longer needed, sessions manager covers this use case :)

PR #2548, opened 2023-06-17
performance(server): greatly improve dump_screen! performance

zynaxsoft:
This PR dramatically improves the `dump_screen!` macro performance. 

### Motivation
First of all, I love the editing scrollback feature!
I was playing around with `scroll_buffer_size` setting. I tried setting it to 1,000,000.
I executed `yes` to fill up the screen buffer and then tried to edit the scrollback.
There was a big delay from issuing the command to editing in my editor.
So I tried to figure out what causing this delay. "1 million lines shouldn't be too big?", I thought.

### Method
* Pre-allocates `buf` string.
* Uses `trim_end_matches()` instead of using regex.

I wrote a simple benchmark [here](https://gist.github.com/zynaxsoft/c26f9e2e08be2d5315203e172596d5ab), simulating what I experienced.
```
Result
Old method: 7.55s
New method: 16.72ms
```

### DIFF
--- a/zellij-server/src/panes/grid.rs
+++ b/zellij-server/src/panes/grid.rs
@@ -284,7 +284,7 @@ fn subtract_isize_from_usize(u: usize, i: isize) -> usize {
 macro_rules! dump_screen {
     ($lines:expr) => {{
         let mut is_first = true;
-        let mut buf = "".to_owned();
+        let mut buf = String::with_capacity($lines.iter().map(|l| l.len()).sum());
 
         for line in &$lines {
             if line.is_canonical && !is_first {
@@ -293,8 +293,7 @@ macro_rules! dump_screen {
             let s: String = (&line.columns).into_iter().map(|x| x.character).collect();
             // Replace the spaces at the end of the line. Sometimes, the lines are
             // collected with spaces until the end of the panel.
-            let re = Regex::new("([^ ])[ ]*$").unwrap();
-            buf.push_str(&(re.replace(&s, "${1}")));
+            buf.push_str(&s.trim_end_matches(' '));
             is_first = false;
         }
         buf


koalazub:
has anything happened with this? Or has there been anything to address it yet? I'm getting similar issues when dealing with large dumps

insidewhy:
This also improves performance in general outside of dump screen, I tried rebasing this branch against `main` to see if scroll behaviour would improve and it's so much better.

See my original report at https://github.com/zellij-org/zellij/issues/3793

insidewhy:
@imsnif Could you consider merging this, it improves scroll performance in general from something I can't really tolerate to a level where it's close to native.

Lockszmith-GH:
@insidewhy - Please read #3794

imsnif:
@insidewhy - unless you're dumping your screen on every render (which I would recommend you do not do) this change should not have any performance impact as this is a completely unrelated code-path. It's far more likely that any issue you're experiencing has to do with the environment and not the code (eg. compiling without `--release`) and that differences with this branch are a red herring.

Personally, I'm a vim user and I use Zellij all the time without noticing any issue. Not saying the performance is 60fps (this is a pre `1.0.0` tool after all) but it is definitely something me and a considerably large user base work with happily every day.

Otherwise, what @Lockszmith-GH said. I'm not going to reply here further in order not to encourage this practice. All the best to you.

For the creators and followers of this PR: I'll point to the project's policy in `CONTRIBUTING.md` (in case you are wondering and don't already know) and otherwise thank you for your work and patience.

insidewhy:
> @insidewhy - unless you're dumping your screen on every render (which I would recommend you do not do) this change should not have any performance impact as this is a completely unrelated code-path. It's far more likely that any issue you're experiencing has to do with the environment and not the code (eg. compiling without `--release`) and that differences with this branch are a red herring.

I compiled the code from `main` both with and without the commit from this PR merged to test, so I'm 98% sure that it fixes the jerkiness in neovim, but why I do not know.

koalazub:
> @insidewhy - unless you're dumping your screen on every render (which I would recommend you do not do) this change should not have any performance impact as this is a completely unrelated code-path. It's far more likely that any issue you're experiencing has to do with the environment and not the code (eg. compiling without `--release`) and that differences with this branch are a red herring.

Unrelated how? If this is a red herring - then what's your line of thought for where this performance hit is occuring? Even intuitively we might be able to investigate why this is occuring.

> Personally, I'm a vim user and I use Zellij all the time without noticing any issue. Not saying the performance is 60fps (this is a pre `1.0.0` tool after all) but it is definitely something me and a considerably large user base work with happily every day.

I'm also working on a large codebase, with a large dump of text and I can verify that the delay is substantial enough that I've had to use Alacritty separately to render the text without the entire terminal crashing due to zellij. 

> Otherwise, what @Lockszmith-GH said. I'm not going to reply here further in order not to encourage this practice. All the best to you.

I don't understand this perspective. You've come in and essentially said 'well it works fine on my machine', dropped breadcrumbs, and then concluded by saying you're not going to respond further because of a - justified - etiquette with the maintainer. Yet you can respond, and subsequently deny the response?

There are now 2 people saying that there's a significant enough performance issue in the code base when it comes to rendering larger lines of text. I've linked to this PR #3793 due to it now having an impact. If you genuinely don't think it's due to this, then I'd appreciate some deliberate thought on where/why this could be happening and what we can do to resolve this. 

> I did notice that it happens to an extreme level during scrollback via pane as well. Maybe this isn't related. But worth noting nonetheless

insidewhy:
> > I compiled the code from `main` both with and without the commit from this PR merged to test, so I'm 98% sure that it fixes the jerkiness in neovim, but why I do not know.
> 
> Yes, same exerience for me in helix. This fixes the scrolling.

The author was convinced that `dump_screen` is not in the code path called by scrolling, but given we both found the scrolling performance noticeably better with the exact same code but with/without this commit merged, I think that assumption must be incorrect.

So I guess fixing `dump_screen` would be a band-aid (and a minor nice thing for those who frequently dump the screen), but ideally we need to find out how `dump_screen` is called during scroll operations and remove that if possible.

imsnif:
Thank you @zynaxsoft !!

Issue #1461, opened 2022-06-07
Add action to dump the whole tab


a-kenji:

marianomarciello:
Hi @a-kenji I would like to dig into this issue.
So just for explanation, the new action `DumpTab` should dump every pane in the current tab. Is this correct or I misunderstood your message ?
Thank you for your time.

PR #2588, opened 2023-06-29
Action dump tab

marianomarciello:
Requested in [#1461 ](https://github.com/zellij-org/zellij/issues/1461).
A new action is added which can be called using the following syntax:
zellij action dump-tab [OPTIONS] <PATH>.
This action will dump the Tiled panel and Floating Panel in this order.
Tiled panels content's starts with an header with the following format:
[Tiled pane: {pane_id}]
Floating panels content's starts with an header with the following format:
[Floating pane: {pane_id}]

### DIFF
--- a/zellij-server/src/route.rs
+++ b/zellij-server/src/route.rs
@@ -172,6 +172,11 @@ pub(crate) fn route_action(
                 .send_to_screen(ScreenInstruction::DumpScreen(val, client_id, full))
                 .with_context(err_context)?;
         },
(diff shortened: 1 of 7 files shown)


marianomarciello:
Hi @a-kenji any news on this ? :smile:

marianomarciello:
Hi @imsnif thank for your time. 
I understand your point . I will try to develop somethings that is in the current road map.
:smile:

Thank you again for your time.

PR #2599, opened 2023-07-01
Add documentation to crates

Lilja:
As a result of trying to familiarize myself with the project, I'd thought it would be good to start to make the crates more digestible.

There are no readmes for client and server, mainly because I don't understand what they are. Could use some help describing what they do and what they contain.

### DIFF
--- a/src/README.md
+++ b/src/README.md
@@ -1,0 +1,5 @@
+# zellij cli
+
+This directory is related to the CLI tool of zellij. It's a way to interact with the zellij sesson with arguments. 
+
+See `zellij --help` for what more information.
--- a/zellij-tile/README.md
+++ b/zellij-tile/README.md
@@ -1,0 +1,5 @@
+# zellij-tile
+
+Source code of the `ZellijPlugin` trait, a core detail for implementing your own plugin. With this crate you can interact with the zellij session through [shims](https://docs.rs/zellij-tile/latest/zellij_tile/shim/index.html)
+
+Generated documentation can be found at [crates.io](https://docs.rs/zellij-tile/latest/zellij_tile/)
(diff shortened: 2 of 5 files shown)


src/README.md
```diff
@@ -0,0 +1,5 @@
+# zellij cli
+
+This directory is related to the CLI tool of zellij. It's a way to interact with the zellij sesson with arguments.
```
Zykino, line 3: Typo:
```suggestion
This directory is related to the CLI tool of zellij. It's a way to interact with the zellij session with arguments. 
```

zellij-tile/README.md
```diff
@@ -0,0 +1,5 @@
+# zellij-tile
+
+Source code of the `ZellijPlugin` trait, a core detail for implementing your own plugin. With this crate you can interact with the zellij session through [shims](https://docs.rs/zellij-tile/latest/zellij_tile/shim/index.html)
```
Zykino, line 3: I think it is better to call it what it calls itself in the docs : "it is the Rust API for developing plugins". So I suppose "zellij-tile-utils" is linked to that.

zellij-utils/README.md
```diff
@@ -0,0 +1,6 @@
+# zellij-utils
+
+Utility macros for how to read KDL-files(layout & other config). Also holds types like [Event](https://docs.rs/zellij-utils/latest/zellij_utils/data/enum.Event.html) which is how a zellij plugin receives information from the zellij session.
+
+Generated documentation can be found at [crates.io](https://docs.rs/zellij-utils/latest/zellij_utils/)
+
```
Zykino, line 6: I don’t know how github render files but this one is the only one that as a final blank line. The POSIX definition of a line stipulate that it ends with a newline character. => Not sure if this file have 2 or a newline is missing in the other readmes ^^'

Zykino (commented):
I’m not a Zellij maintainer, but seeing the typo I went ahead and did my comments.

For you interrogation (and based on deduction of what I understand about zellij):
I think the server is like a daemon : running in the background keeping sessions alive.
A client should be the part living in the terminal that connect to the server. So the part that do the actual printing to screen and getting keybindings.

PR #2602, opened 2023-07-03
Add action QueryPaneNames to list pane names

thled:
This adds an action to Zellij called `query-pane-names`. It is inspired by `query-tab-names` (see https://github.com/zellij-org/zellij/pull/2145).

It shows a list of all panes in the current tab.

### DIFF
--- a/zellij-server/src/route.rs
+++ b/zellij-server/src/route.rs
@@ -602,6 +602,11 @@ pub(crate) fn route_action(
                 .send_to_screen(ScreenInstruction::QueryTabNames(client_id))
                 .with_context(err_context)?;
         },
+        Action::QueryPaneNames => {
+            senders
+                .send_to_screen(ScreenInstruction::QueryPaneNames(client_id))
+                .with_context(err_context)?;
+        },
         Action::NewTiledPluginPane(run_plugin, name) => {
             senders
                 .send_to_screen(ScreenInstruction::NewTiledPluginPane(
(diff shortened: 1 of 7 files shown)


serranomorante:
Looking forward to this feature!! :smile:

VasilisManol:
This would be a useful feature!

last-genius:
This issue (along with a version of `go-to-tab-name` for panes) would help out a lot for Zellij's support from third-party tools. Currently an integration with a debugger plugin I'm writing is stuck because of the inability to close a pane with a particular name (since Zellij, to my knowledge, is lacking any other unique identifiers for panes and tabs, and directions, or "previous pane" isn't enough of a guarantee ). `tmux` has both unique IDs and an ability to kill panes (among other things) by ID, without switch->close routine present in Zellij, for example.

#3314 has helped this somewhat (though it doesn't list non-focused panes), but I'm missing how pane IDs could be utilized in the rest of the CLI actions. Is this possibly planned?

I'm wondering why this PR hasn't been taken up? Is the project lacking hands for review? Could I help?

VasilisManol:
I had the same question, here is a good explanation I got from the author https://discord.com/channels/771367133715628073/1096025009731936327/1314894610358599680.

last-genius:
> I had the same question, here is a good explanation I got from the author https://discord.com/channels/771367133715628073/1096025009731936327/1314894610358599680.

Could you copy-paste the text from there please? I have no access via this link and also for long term preservation purposes discord is a terrible platform

PR #2619, opened 2023-07-12
feat(ui): new status bar mode

imsnif:
This is an effort to finally solve the "colliding keys" problem.

This adds a new mode to the status bar, basically defaulting to "locked" mode and treating any other non-normal mode as a "sticky" mode. So that whenever a user moves from "locked" to "normal", they will be placed in said sticky mode. When the user then returns from the sticky mode to normal, they will be placed back in locked mode.

This would effectively mean that when most times, Zellij will only take up 1 keybinding (by default `Ctrl g` to enter the sticky mode).

Right now, this is behind a feature flag, but after this I plan to do work to allow plugins to be configurable and users will be able to goggle this off/on.

### DIFF
--- a/default-plugins/status-bar/src/first_line.rs
+++ b/default-plugins/status-bar/src/first_line.rs
@@ -210,7 +210,11 @@ fn short_key_indicators(
     mode_info: &ModeInfo,
     no_super: bool,
 ) -> LinePart {
-    let mut line_part = if no_super { LinePart::default() } else { superkey(palette, separator, mode_info) };
+    let mut line_part = if no_super {
+        LinePart::default()
+    } else {
+        superkey(palette, separator, mode_info)
+    };
     let shared_super = if no_super { true } else { line_part.len > 0 };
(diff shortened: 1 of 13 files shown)


Lockszmith-GH:
Is this implementing a 'lead key' - similar to tmux/screen?

imsnif:
> Is this implementing a 'lead key' - similar to tmux/screen?

Kind of, yeah - but coupled with a sticky mode (so you don't have to go down two levels every time you want to do something), and some UI to make it clear.

It's still a WIP and will be clearer once I put some more work into it (and ofc it will always be possible to keep using the "classic" way we have now)

brunoherrick:
Would this mean that instead of e.g. laboriously pressing `Ctrl g`, then `Ctrl p` and say j/k to move the the upper/lower pane, I would need to just press something like `Ctrl g`, then `p` and finally j/k?

Furthermore, how can we test this feature by enabling the feature flag? Thank you for the tremendous work @imsnif

PR #2624, opened 2023-07-13
feat: add plugin permission system

### LINKED ISSUE #1395: Create a plugin permission system
As we add more and more capabilities to Zellij plugins, we should create a permission system that will give users transparency into which permission this plugin requires (eg. file system access, session management, scrollback reading and manipulation) and the ability to grant them or not.

Depends on: #317, #771

jaeheonji:
resolve #1395 

> This PR is currently a "very" draft. So sample and weird code still exist.

### Tasks
- [x] create API draft
- [x] create UI for permission screen
- [x] Add basic permission (such as `KeyboardInput`)
- [x] caching for granted permissions 
  - [x] memory
  - [x] file


default-plugins/strider/src/main.rs
```diff
@@ -19,6 +19,9 @@ register_worker!(
 
 impl ZellijPlugin for State {
     fn load(&mut self) {
+        // Example
+        request_permission(&[PluginPermission::KeyboardInput]);
```
jaeheonji, line 23: it's sample code. it will be changed.

zellij-server/src/panes/plugin_pane.rs
```diff
@@ -193,14 +196,34 @@ impl Pane for PluginPane {
             .vte_parsers
             .entry(client_id)
             .or_insert_with(|| vte::Parser::new());
-        for &byte in &bytes {
+
+        let mut vte_bytes = bytes;
+        if self.requesting_permissions.is_some() {
+            vte_bytes = "Example...".into();
```
jaeheonji, line 202: pseudocode. Display for PluginPermission will be implemented later.

zellij-server/src/tab/mod.rs
```diff
@@ -1552,17 +1554,27 @@ impl Tab {
...
+                    for key in parse_keys(&adjusted_input) {
+                        plugin_updates.push((Some(pid), client_id, Event::Key(key)));
+                    }
+                    self.senders
+                        .send_to_plugin(PluginInstruction::Update(plugin_updates))
+                        .with_context(err_context)?;
+                },
+                Some(AdjustedInput::Confirmed(_)) => {
+                    self.request_plugin_permissions(pid, None);
+                    // TODO: Maybe call render() in plugin instance?
```
jaeheonji, line 1573: The UI is not updated immediately, and an update event must be generated from the plugin to call a new render and change the screen.

I think I have to force render to be called once. Is that correct?

jaeheonji (commented):
Hi, @imsnif !

I uploaded the code by applying only a very basic flow related to the UI.
Does this flow make sense?

imsnif:
Hey @jaeheonji - this looks great!

I think this is a very good flow, and the trick you did here: https://github.com/zellij-org/zellij/pull/2624/files#diff-74ec31ae71096d7d0ec36db5be3abfd8f6ffdc920f36c2e1bc8a83ae553e115fR201 is actually more clever than what I had in mind. I like it (in the end maybe we will add some comments to explain it more, but for now it's good).

What I would add to this is: once the user confirms the permissions, we need to let the plugin know (because then the plugin can make decisions about it - for example, if the user did not give permissions, the plugin can try to work in a different way, or without a certain feature).

So what I suggest, is in this case to send a new `PluginInstruction`. Maybe something like `PluginInstruction::PermissionRequestResults`, which would go to the plugin thread and generate a new `Event` to send to the plugin. Maybe something like `Event::PermissionRequestResults`. When the plugin gets this event in its `update` method (it needs to make sure to `subscribe` to it in the beginning...) then it can decide to render itself again (by returning `true` from the `update` method where it got this `Event`) and then this will all be fixed, I think.

What do you think? Does this make sense?

jaeheonji:
Thank you! @imsnif 

> So what I suggest, is in this case to send a new PluginInstruction. Maybe something like PluginInstruction::PermissionRequestResults, which would go to the plugin thread and generate a new Event to send to the plugin. Maybe something like Event::PermissionRequestResults. When the plugin gets this event in its update method (it needs to make sure to subscribe to it in the beginning...) then it can decide to render itself again (by returning true from the update method where it got this Event) and then this will all be fixed, I think.
> 
> What do you think? Does this make sense?

Yes! it makes sense. I thought I needed something similar, and it's a very good way. I'll update the code that puts all this together as soon as possible.

jaeheonji:
Hi, @imsnif !

I would like to inform you that The plugin permissions system have been implemented 😄
Here are the currently implemented features:

- A cache file is created with the name `permissions.kdl` in the plugin data folder. (please comment if you have any other opinion about the cache file location)
- **If the cache file does not exist** or **if "N" was selected** for the previous permission request, the plugin permission request UI is displayed when the plugin is executed.

If there is no problem in the system, we can add additional permissions before merging.

zellij-server/src/panes/plugin_pane.rs
```diff
@@ -595,4 +634,25 @@ impl PluginPane {
...
+        let cyan = style!(self.terminal_emulator_colors.borrow().cyan).bold();
+
+        let mut messages = String::new();
+        let permissions: HashSet<PermissionType> =
+            plugin_permission.permissions.clone().into_iter().collect();
+
+        messages.push_str(&format!(
+            "Plugin {} would like permission to:",
+            cyan.paint(&plugin_permission.name),
+        ));
```
jaeheonji, line 647: I have a question about this part, I want to apply a color to the message, but the correct color is not applied. (only grayed out)

Is it a problem with the way I'm getting the color? Or is it another problem?

imsnif: Yeah, there is a bit of a mess here, sorry :/

To make this work you can change it to:
```rust
         let cyan = style!(self.style.colors.cyan).bold();
```

imsnif:
Hey @jaeheonji - I'll start by saying this is, as usual, fantastic work from you. You created a really great system in a short amount of time, meeting the deadline and even allowing time for review. Kudos!

I have a few comments (also left some in the code), but I think nothing major:
1. I think it would be a good idea to always read from `permissions.kdl` when we need permissions - meaning, never keep it in memory. This will prevent a lot of troubles such as "the user deleted the cache while the plugin is still running", "the user deleted the cache and reloaded the plugin", etc. I think it's not a big performance overhead at all and will save a lot of trouble. We can always revisit this decision later if we want.
2. Let's add some tests for this. I think there are some good examples here: https://github.com/zellij-org/zellij/blob/main/zellij-server/src/plugins/unit/plugin_tests.rs (these tests are run as part of our e2e tests, but not in the docker environment... they are only there because they can be a little long). I think tests such as "load plugin, try to do something that needs permission, see that we fail" and "load plugin, request permission, see that we are able to perform the action" and finally "load plugin, request *different* permission, see that we are not able to perform the action"
3. Will it be OK for you if I play around a little bit with the UI to add colors, newlines, maybe truncations and such? I've just done it 100 times and can do it quickly :)

Once again - great work here! Looking forward to getting this merged soon.

imsnif (changes requested):
Hey @jaeheonji - I'll start by saying this is, as usual, fantastic work from you. You created a really great system in a short amount of time, meeting the deadline and even allowing time for review. Kudos!

I have a few comments (also left some in the code), but I think nothing major:

1. I think it would be a good idea to always read from permissions.kdl when we need permissions - meaning, never keep it in memory. This will prevent a lot of troubles such as "the user deleted the cache while the plugin is still running", "the user deleted the cache and reloaded the plugin", etc. I think it's not a big performance overhead at all and will save a lot of trouble. We can always revisit this decision later if we want.
2. Let's add some tests for this. I think there are some good examples here: https://github.com/zellij-org/zellij/blob/main/zellij-server/src/plugins/unit/plugin_tests.rs (these tests are run as part of our e2e tests, but not in the docker environment... they are only there because they can be a little long). I think tests such as "load plugin, try to do something that needs permission, see that we fail" and "load plugin, request permission, see that we are able to perform the action" and finally "load plugin, request different permission, see that we are not able to perform the action"
3. Will it be OK for you if I play around a little bit with the UI to add colors, newlines, maybe truncations and such? I've just done it 100 times and can do it quickly :)

Once again - great work here! Looking forward to getting this merged soon.

EDIT: oops, added this as a comment and not a PR review, deleted and fixed now

zellij-server/src/plugins/zellij_exports.rs
```diff
@@ -212,6 +214,34 @@ fn host_set_selectable(env: &ForeignFunctionEnv, selectable: i32) {
     }
 }
 
+fn host_request_permission(env: &ForeignFunctionEnv) {
+    wasi_read_object::<Vec<PermissionType>>(&env.plugin_env.wasi_env)
+        .and_then(|permissions| {
+            if let Some(permissions) = &env.plugin_env.plugin_permissions {
+                if !permissions.is_empty() {
+                    return Ok(());
```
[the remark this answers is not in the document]
jaeheonji: Yes I can understand you Looking at this part alone, this code is hard to understand.

As of now, when the server starts, the server reads `permissions.kdl` and loads the cached data into `plugin_env`.
Therefore, if `plugin_env.plugin_permissions` is not empty, it means that the user has set permissions in the past.

In fact, I understood the purpose of the file cache by this. If this code doesn't exist, the user will see the UI every time.

Of course, as this code exists, we may need additional actions. For example, `Re-Granted Permission`?

jaeheonji:
Thank you so much @imsnif !
Your help is greatly appreciated in everything.

> I think it would be a good idea to always read from permissions.kdl when we need permissions - meaning, never keep it in memory. This will prevent a lot of troubles such as "the user deleted the cache while the plugin is still running", "the user deleted the cache and reloaded the plugin", etc. I think it's not a big performance overhead at all and will save a lot of trouble. We can always revisit this decision later if we want.

Yes. it makes sense. It seems we can always read data from the file cache whenever the plugin calls `RequestPluginPermissions`. We still have time, and I think I can apply this quickly. How about applying it right away?

> Let's add some tests for this. I think there are some good examples here: https://github.com/zellij-org/zellij/blob/main/zellij-server/src/plugins/unit/plugin_tests.rs (these tests are run as part of our e2e tests, but not in the docker environment... they are only there because they can be a little long). I think tests such as "load plugin, try to do something that needs permission, see that we fail" and "load plugin, request permission, see that we are able to perform the action" and finally "load plugin, request different permission, see that we are not able to perform the action"

Sure! I'll try to add it soon.

> Will it be OK for you if I play around a little bit with the UI to add colors, newlines, maybe truncations and such? I've just done it 100 times and can do it quickly :)

Welcome anytime. If you have something to add or update, feel free to apply right away!

imsnif:
> Yes. it makes sense. It seems we can always read data from the file cache whenever the plugin calls `RequestPluginPermissions`. We still have time, and I think I can apply this quickly. How about applying it right away?

Sounds good!

> Welcome anytime. If you have something to add or update, feel free to apply right away!

Great, I'll let you work and finish this up and then I'll fix up the UI before merging.

imsnif:
Hey @jaeheonji - I resolved the merge conflicts from main, hope it's okay. Going to try and find the time to work a little on the UI as promised today.

imsnif:
Alright - touched up the UI a little bit:
img-2023-07-25-153147

I hope you like it!

I'm also going to try to arrange the different permissions into categories before we merge. Not sure when I get to it, but it shouldn't block you. Please let me know if I messed up anything :)

zellij-server/src/screen.rs
```diff
@@ -2815,6 +2823,36 @@ pub(crate) fn screen_thread_main(
...
+                    Ok(granted_permission) => {
+                        if let Some(p) = granted_permission.get(&permissions.name) {
+                            // TODO: How to deal cached data?
+                            permissions
+                        } else {
+                            permissions
+                        }
+                    },
+                    Err(_) => permissions,
+                };
```
jaeheonji, line 2837: @imsnif 

Now, we don't save permission data in memory. Permission data is fetched from the cached file on every request.

But, There is still one dilemma here. If the cache data is fetched and applied, there is no way for the user to see the Request UI again. The only way is clearing the cache data.

Otherwise, if the cache data is not imported and used, the cache is meaningless.

Is there any good solution?

imsnif: Hmm... I see what you mean. Good catch.

Do you think we can:
1. If the user granted permissions, we remember it until they delete the cache explicitly
2. If the user did not grant permissions, we ask them again (if the plugin requests permissions) once per session

What do you think?

imsnif: Actually, it would be even better (if it's not too much work), to also do:

3. If the user presses `Ctrl c`, it is "no and never ask me again"

(I can add the UI for this)

But really, only if it's not a lot more work for you!

jaeheonji: Oh Thank you! it makes sense.

> If the user presses Ctrl c, it is "no and never ask me again"

Yes, this way would be better. But, for now, I will implement it based on 1 and 2. I think 3 would be better to improve later.

jaeheonji:
Hey @imsnif !

I finished all implementations and am currently trying to write a plugin test.
As an example, I made a test for `RequestPluginPermissions` (with reference to other tests).

However, when I try to run the test, it takes a very long time. (other tests too)
Is it right that it takes "extremely" this long? I haven't seen the ok sign, at least until now. 😭

imsnif:
Hi @jaeheonji - I think it's safe to say that if by the time you got this message the tests did not end, they are not going to end :)

I think I also see the tests here in CI taking some time... so maybe it can be that they are waiting for permission and so get stuck because they did not request it?

jaeheonji:
If only the tests I wrote do not run, I will judge that there is a problem with my tests. But for some reason, other (all) tests are not completing, for example, `send_configuration_to_plugins`.

I only see this phrase `test plugins::plugin_tests::send_configuration_to_plugins has been running for over 60 seconds`. 

Is it possible to check if the test is working normally in your local environment?

imsnif:
Hey @jaeheonji - how is it going?

jaeheonji:
@imsnif 

I just added some test cases and did some refactoring.

If there are no problems after CI is finished, I think we are ready to merge this feature. 😄

imsnif:
Amazing news @jaeheonji !! Before we merge I'll quickly arrange the permissions a little bit so that this won't block anything (eg. no need to get permission for receiving keypresses when the plugin is focused, but maybe we do need it for something else!)

Let me know when you're done so that I won't bother your work.

jaeheonji:
Hey, @imsnif 

I created 2 test cases, which work locally, but fail unexpectedly on e2e. Perhaps, I'm missing something, I'll have to do this again tomorrow. So, If you need to work on something, you can do it now.

imsnif:
No worries @jaeheonji - I have some other stuff to do. Let me know when you're done and I'll adjust some things before I merge. No rush, we're almost there!

imsnif:
Hey @jaeheonji - any updates?

jaeheonji:
@imsnif Sorry for late replying

Currently, I have only one problem left. The 2 plugin tests I added don't work only inside the workflow. (it works very well in local and e2e)

I'm still looking into the problem. This can be done independently, so if you need to fix or add something, you can 😄

jaeheonji:
> Only thing I can think of is maybe locally you're still using an old fixture plugin? Maybe try to build everything again? Maybe even completely delete the target folder and try again just in case?

I tried cleaning the project perfectly and running it again. Works fine locally, but still not in workflow. 😢

imsnif:
Hey @jaeheonji - I looked into this and it was indeed quite a tricky problem! The tests were failing because of a race condition, since the plugins are loaded asynchronously and in these cases nothing was waiting for a message from the screen thread.

I fixed these by conditionally sending a message from a "fake" screen thread when they receive a permission request from the plugins. This solved the problem locally for me.

I also merged from main and so transferred the permission request to use protobuffers for serializing across the wasm boundary (which we now do in main) and fixed up some other merge issues. I hope the e2e tests still pass :D 

I'm going to configure some permissions and change some stuff around before merging, hopefully I'll get it all done tomorrow. Great work on this!!

jaeheonji:
Thank you for your hard work @imsnif !

imsnif:
Hey @jaeheonji - so I think I'm finally done with my part in this (hopefully, if the CI passes :crossed_fingers: )

I configured the following permissions:
```rust
    ReadApplicationState,
    ChangeApplicationState,
    OpenFiles,
    RunCommands,
    OpenTerminalsOrPlugins,
    WriteToStdin
```

I removed the KeyboardInput permission for now, because I think plugins should be allowed to see their input without permission. I think we will add it again when we add the ability for plugins to reigster global keybindings (I hope to be adding this soon).

I also changed the permissions in `RunningPlugin` to be an `Arc<Mutex>`. The reason I did this was that when I also moved permissions to the commands, it was using a cloned plugin environment, and we need this to be the same struct.

So, all in all - this is a really good system. I think it works very well, I like the caching mechanism and it's quite simple to extend it. Good job on this and thanks for the quick and punctual implementation!

PR #2650, opened 2023-07-25
refactor: separate aliases from completions

DavSanchez:


norman-zon:
Is there a roadmap for when this will get merged? Would greatly appreciate it.

PR #2652, opened 2023-07-26
Added utility functions to find active pane and tab

Nacho114:
While porting harpoon to Zellij as a [plugin](https://github.com/Nacho114/harpoon), I had to know the currently active pane and tab via the `Vec<TabInfo>` and the `PaneManifest`. Since this might be a useful utility function I propose to add them in this PR. 

TL;DR

Given the `PaneManifest` and `Vec<TabInfo>`, I want to easily get the active pane and tab.

### DIFF
--- a/zellij-tile/src/shim.rs
+++ b/zellij-tile/src/shim.rs
@@ -372,6 +372,31 @@ pub fn rename_tab<S: AsRef<str>>(tab_position: i32, new_name: S) {
     unsafe { host_rename_tab() };
 }
 
+// Utility Functions
+
+/// Returns the `TabInfo` corresponding to the currently active tab
+fn get_focused_tab(tab_infos: &Vec<TabInfo>) -> Option<TabInfo> {
+    for tab_info in tab_infos {
+        if tab_info.active {
+            return Some(tab_info.clone());
+        }
+    }
+    return None;
+}
+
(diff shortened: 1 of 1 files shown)


Nacho114:
Actually, in `get_focused_pane()`, we don't need to pass the pane position since we can just iterate over every tab in the PaneManifest. Seems cleaner that way and the compute difference doesn't really matter. wdyt?

imsnif:
> Actually, in `get_focused_pane(position, pane_manifest)`, we don't need to pass the tab position since we can just iterate over every tab in the PaneManifest. Seems cleaner that way and the compute difference doesn't really matter. wdyt?

Sure, sounds good. Would you like to make this change before I merge?

Nacho114:
So I played around with the other version (without the `position`) and it's not as good. 

I've not investigated in detail so it's mostly based on observation:

If I go from tab 1 to tab 2, then:

First TabInfo gets updated, and then the manifest.

This means that, if run `get_focused_pane(pane_manifest)` quickly when I go to tab 2, I get the previously activated pane in tab 1, since only the TabInfo was updated. However, if I do: `get_focused_pane(position, pane_manifest)` with the newly fetched position I don't run into this issue. 

I would say merge what currently is there as it seems to be more consistent due to this.

imsnif:
Thank you!

Issue #1625, opened 2022-08-01
Memory leak


cosminadrianpopescu:
Thank you for taking the time to file this issue! Please follow the instructions and fill in the missing parts below the instructions, if it is meaningful. Try to be brief and concise.

**In Case of Graphical or Performance Issues**

1. Delete the contents of `/tmp/zellij-1000/zellij-log`, ie with `cd /tmp/zellij-1000/` and `rm -fr zellij-log/`
2. Run `zellij --debug`
3. Recreate your issue.
3. Quit Zellij immediately with ctrl-q (your bug should ideally still be visible on screen)

Please attach the files that were created in `/tmp/zellij-1000/zellij-log/` to the extent you are comfortable with.

**Basic information**

`zellij --version`: `zellij 0.30.0`
`stty size`: `46 197`
`uname -av` or `ver`(Windows): `Linux ip-### Wed Dec 16 22:44:04 UTC 2020 x86_64 x86_64 x86_64 GNU/Linux`

List of programs you interact with as, `PROGRAM --version`: output cropped meaningful, for example:
`nvim --version`: NVIM v0.5.0-dev+1299-g1c2e504d5  (used the appimage release)
`alacritty --version`: alacritty 0.7.2 (5ac8060b)

**Further information**
Reproduction steps, noticeable behavior, related issues, etc

I have a long running `zellij` instance to which I connect and disconnect often (more than once per day). After weeks of using the same session, I notice that more than 2GB or RAM are being used by the zellij server. See the attached screen shot.

cosminadrianpopescu:
![bug-zellij](https://user-images.githubusercontent.com/5187873/182101105-96fd7549-61b6-4ece-9e82-7abc45e732d5.jpg)

olanod:
I have a similar set up and was surprised to see zellij take 2.3Gb

har7an:
Thank you for reporting this!

Sorry we took so long to reply. I tried to reproduce this with the latest zellij version locally and I can confirm this behavior. That clearly shouldn't happen!

I have a suspicion where this may originate. I'll investigate and get back to you once I found something.

har7an:
So this turns out to be more subtle than I had imagined. But I have another question: Do you have a scrollback limit set? Or is your scrollback infinite?

cosminadrianpopescu:
No, I don't have by default any limit set. Should I set up a limit?

InnovativeInventor:
[...]

PR #2675, opened 2023-08-04
fix(performance): plug memory leak

imsnif:
Fixes https://github.com/zellij-org/zellij/issues/1625

For some reason (which I admittedly cannot fully explain) when we pre-allocate memory for the terminal viewport, scrollback and each line therein that memory is not deallocated when its container is dropped.

Removing this pre-allocation doesn't harm performance in a noticeable way and plugs this leak. If any Rustacean reading this has ideas for why this is happening, I'd be very curious to hear them. :)

### DIFF
--- a/zellij-server/src/output/mod.rs
+++ b/zellij-server/src/output/mod.rs
@@ -874,7 +874,7 @@ impl OutputBuffer {
         y_offset: usize,
     ) -> Vec<CharacterChunk> {
         if self.should_update_all_lines {
-            let mut changed_chunks = Vec::with_capacity(viewport.len());
+            let mut changed_chunks = Vec::new();
             for line_index in 0..viewport_height {
                 let terminal_characters =
                     self.extract_line_from_viewport(line_index, viewport, viewport_width);
@@ -888,7 +888,7 @@ impl OutputBuffer {
             let mut line_changes = self.changed_lines.to_vec();
             line_changes.sort_unstable();
             line_changes.dedup();
-            let mut changed_chunks = Vec::with_capacity(line_changes.len());
+            let mut changed_chunks = Vec::new();
             for line_index in line_changes {
                 let terminal_characters =
                     self.extract_line_from_viewport(line_index, viewport, viewport_width);
--- a/zellij-server/src/panes/grid.rs
+++ b/zellij-server/src/panes/grid.rs
@@ -135,8 +135,8 @@ fn transfer_rows_from_lines_above_to_viewport(
                     lines_added_to_viewport -= top_non_canonical_rows_in_dst.len() as isize;
                     next_lines.push(next_line);
                     next_lines.append(&mut top_non_canonical_rows_in_dst);
-                    next_lines = Row::from_rows(next_lines, max_viewport_width)
-                        .split_to_rows_of_length(max_viewport_width);
+                    next_lines =
+                        Row::from_rows(next_lines).split_to_rows_of_length(max_viewport_width);
                     if next_lines.is_empty() {
                         // no more lines at lines_above, the line we popped was probably empty
                         break;
@@ -149,7 +149,7 @@ fn transfer_rows_from_lines_above_to_viewport(
         lines_added_to_viewport += 1;
     }
     if !next_lines.is_empty() {
-        let excess_row = Row::from_rows(next_lines, 0);
+        let excess_row = Row::from_rows(next_lines);
         bounded_push(lines_above, sixel_grid, excess_row);
     }
     match usize::try_from(lines_added_to_viewport) {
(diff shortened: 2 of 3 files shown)


raphCode:
I would not consider myself a Rustacean, but it seems to me all this preallocation finally happens in the standard library, right?
Under the assumption that the stdlib knows how to clean up its (pre-allocated memory) when told to do so, I too do not understand how it might fix the leak.
My fear at this point is that the preallocation maybe just exaggerated any leaks to the point where they became obvious?

imsnif:
>  My fear at this point is that the preallocation maybe just exaggerated any leaks to the point where they became obvious?

That's what I thought as well, but allocating the same count of items by filling the buffer did not even make a dent in the resident memory utilization.

My best guess right now is that this is some combination of the elements being drained and the drain iterator leaking the memory (intentionally or unintentionally), or not being dropped itself or whatnot.

imsnif:
Also, cool to see you're still around @raphCode :)

raphCode:
> That's what I thought as well, but allocating the same count of items by filling the buffer did not even make a dent in the resident memory utilization.

Alright, Fingers crossed that it helps!
Nonetheless it might be worth keeping the discrepancy in mind if leaks resurface.

> Also, cool to see you're still around @raphCode :)

Of course :) Just too much to do lately sadly...
I got a ping on the memory leak issue that got closed by this PR, and I got curious how it was solved.

mo8it:
@imsnif This issue is not solved. I think that only a symptom was hidden when testing with `yes` because it prints only one character `y` by default. By replacing `with_capacity` with `new` and then pushing `y`, we allocate less memory which means that we leak less memory afterwards, but this does not mean that we don't leak memory at all.

If you run something like

`yes -- "-----------------------------------------------------------------------------------------------------------------------------------------------"`

in multiple panes on the `main` branch (I did it with `cargo xtask run`) and then create a new clean pane and close all others, you will see the memory leak. For me, the leak was about 3.5 GB for about 10 panes with a scroll back max of 20000 lines.

I would revert this PR and search for the real issue: Why are lines not dropped?

raphCode:
The whole preallocation discussion made me think... maybe it can be useful for hunting down which collection exactly is leaked and when?
Deliberately using a huge preallocation in specific places should reveal itself in the memory usage.

imsnif:
@mo8it - thanks for chiming in. For your test case with `yes`, please see the issue comment here: https://github.com/zellij-org/zellij/issues/1625#issuecomment-1411735893

In short - I don't think that's a memory leak, but just the default rust linux allocator doing its thing. Just to make sure, I tried reproducing this and did not see consistent memory climbs (please be sure to test this in only one pane, otherwise it can very easily be mistaken for the aforementioned allocator behavior).

I am happy to be wrong, and I'm happy to keep working on this, but I ask - nay, beg - of you: please be very careful what you write about this issue. I have personally spent a growing number of hours debugging it and chasing inconsistent ghosts only to find out they are expected behavior (unless I want to switch allocators which I would really rather not).

When people write stuff like in the comments above, others do a github search because they misinterpret some memory diagnostic, find these issues/PRs and after a cursory read conclude that "Zellij has a memory leak, I will not use it" - which in turn means more hours of chasing ghosts for me :) (mostly because I am a penchant for accuracy, less because I care if people use my software).

So again - I am very happy for the help, but please point to specific places in the code base if you think they leak memory, or to a consistent and significant climb in one pane (in this case) as was pointed out in the btop issue in the original thread.

As a hint to get you started, this is where the rows are `Drop`ed: https://github.com/zellij-org/zellij/blob/c3e140cb4b3c0897329bf07ee7f51e9fd402b3df/zellij-server/src/panes/grid.rs#L187

When rendering, they also go into our `Output` buffer and are ultimately dropped when it goes out of scope at the end of this function hree: https://github.com/zellij-org/zellij/blob/c3e140cb4b3c0897329bf07ee7f51e9fd402b3df/zellij-server/src/screen.rs#L961

Thanks again, and for those chiming in: for the time being - no additional leaks have been reliably reproduced.

imsnif:
Also also, regarding the original conundrum on the `with_capacity` causing the leak, I strongly suspect this iterator here: https://github.com/zellij-org/zellij/blob/c3e140cb4b3c0897329bf07ee7f51e9fd402b3df/zellij-server/src/output/mod.rs#L907

Maybe when doing this on a preallocated vec, some issue arises. Not sure. I'm moving on to other things, if someone wants to dive deeper into this and finds something I missed or an explanation for our riddle - I would be super happy for a ping.

mo8it:
I am very sorry @imsnif for the too quick, wrong conclusion. I did edit my comment above.
Thank you a lot, not only for working on this exhausting issue, but of course for Zellij as a whole ❤️

imsnif:
No worries, @mo8it ! Thanks for chiming in and keeping me on my toes :) If you do end up uncovering something I've missed I'd be very happy to know about it.

yshui:
Sorry for barging in, but could it be related to this? https://blog.polybdenum.com/2024/01/17/identifying-the-collect-vec-memory-leak-footgun.html

PR #2706, opened 2023-08-18
feat(plugins): upgrade wasmer to 3.1.1

tlinford:
This brings `wasmer` and `wasmer-wasi` to v3.1.1
The main breaking changes are `Store` no longer implementing `Clone`, and a `&mut Store` is required to call wasm functions.
To address this, instead of using a single `Store` instance, each plugin (or worker), gets its own Store.

More work is needed to upgrade to even newer versions, since `wasmer-wasi` was deprecated and replaced by the `wasmer-wasix` crate.


imsnif:
Just FYI: following this discussion: https://github.com/zellij-org/zellij/pull/2706 - we need this in order to package Zellij for Debian... @tlinford any chance this can be merged soon?

PR #2711, opened 2023-08-20
xtask/pipeline: Fix publish task

har7an:
which was previously stuck in an infinite loop after successfully publishing a crate. The error originated in the code only checking for error conditions but not breaking out of the inner infinite loop in case of success.

In addition, when an error occurs, the user now has more actions to choose from, allowing to either ignore the error and continue with the next crate, retry the current publish, or aborting execution altogether.


har7an:
# Why is this PR necessary?

While preparing for the v0.38.0 release of zellij we noticed that, unfortunately, there are some unknown issues with a build script for the zellij-utils crate that has been brought in during the current release cycle. This build script is responsible for compiling the protobuf definitions that have become part of the plugin API in hopes to eliminate plugin/zellij version mismatch issues. The expected outcome of this change is:

- Build the zellij-utils crate
- Before compiling zellij-utils, the build script is run
- Build script executes and compiles the protobuf definitions in a defined temporary directory
- zellij-utils builds and picks up the protobuf definitions from that directory

This works as expected while building and developing zellij, even before this PR. The issue we noticed is that, during a `cargo publish --dry-run` apparently the build script is never executed. This leads to the following issue:

- Build the zellij-utils crate
- *Build script is NOT executed*
- zellij-utils builds and cannot pick up protobuf definition files
- **Compilation fails**

We haven't been able to determine the root cause of this behavior yet. However, we decided not to postpone the release until this problem is fixed because, as stated, developing zellij with the build script included works fine.

This PR introduces a workaround to the problem, which is to compile the protobuf files once for this release and statically include them in a fixed location in the `assets/` subfolder of zellij-utils. The build script is temporarily disabled to prevent issues with unwanted modifications/changes.

PR #2726, opened 2023-08-25
fix(keybinds): add 'floating' and 'name' to the Run command keybinding

imsnif:
This fixes an oversight where it was impossible to start a floating pane or a pane with a title from a keybinding.

So now, we can do:

```kdl
keybinds {
// ...
    shared {
        Run "htop" "--tree" {
            floating true;
            name "my awesome floating pane";
        }
    }
//...
}
```


imsnif:
Good catch @robrecord, thanks! I'll add it when I do a go-over on them to add other stuff.

PR #2763, opened 2023-09-01
chore: bump suggest crate version to 0.5

ken-matsui:


imsnif:
Hey, is there a problem with 0.4.0?

ken-matsui:
@imsnif No, but 0.5 removed the dependency on clap. This does not matter since Zellij depends on clap as well, but the dependency resolution performance might benefit. Also, if you will move this suggest to another workspace that doesn't depend on clap, the dependencies get lighter with this upgrade.

PR #2765, opened 2023-09-01
fix(plugins): make auxiliary functions (get_focused_tab and get_focused_pane) public

Nacho114:
I forgot to make auxiliary functions public (`get_focused_tab` and `get_focused_pane`). This is fixed in this PR.


imsnif:
Thank you!

Issue #2785, opened 2023-09-13
neovim wrapped lines render different lines


k-cross:
Thank you for taking the time to file this issue! Please follow the instructions and fill in the missing parts below the instructions, if it is meaningful. Try to be brief and concise. 

Opened another issue here since this most noticeably happens with the latest neovim release https://github.com/neovim/neovim/issues/25139

**In Case of Graphical or Performance Issues**

1. Delete the contents of `/tmp/zellij-1000/zellij-log`, ie with `cd /tmp/zellij-1000/` and `rm -fr zellij-log/` (`/tmp/` is `$TMPDIR/` on OSX)
2. Run `zellij --debug`
3. Run `stty size`, copy the result and attach it in the bug report
4. Recreate your issue.
5. Quit Zellij immediately with ctrl-q (your bug should ideally still be visible on screen)

Please attach the files that were created in `/tmp/zellij-1000/zellij-log/` to the extent you are comfortable with.

**Basic information**

`zellij --version`: zellij 0.38.1
`stty size`: 49 118
`uname -av` or `ver`(Windows): Darwin 22.6.0 Darwin Kernel Version 22.6.0: root:xnu-8796.141.3~6/RELEASE_X86_64 x86_64

List of programs you interact with as, `PROGRAM --version`: output cropped meaningful, for example:
`nvim --version`: NVIM v0.9.2
`alacritty --version`: alacritty 0.12.2 (9d9982d)

This also happens w/ mac's native terminal while using w/ zellij.

**Further information**
Reproduction steps, noticeable behavior, related issues, etc

   1. Write a few lines that don't wrap.
   2. In the middle of those few lines, write a long line that produces a wrap
   3. Change to command mode and move to a line below and observe

k-cross:
I did not think of this as a graphical issue but if you need that output I will follow those steps too. If there is any information I left out, I'm sorry. I will provide it to help triage. Thank you for everything and anything you are able to do!

RichardVanLeeuwen:
Additional info: I have the same issue on linux. Outside of zellij the bug doesn't happen. It seems like neovim and zellij are miscommunicating about line wrapping. Neovim thinks lines are wrapped, thus they end where the line runs off the screen; zellij thinks the line isn't wrapped and displays the next line as if nothing is going on. This looks okay at first, but when you start moving around it starts glitching (parts of lines are unreachable, lines are displayed twice, able to move the cursor to where no text is displayed). Visual mode helps visualize this.
[...]

PR #2789, opened 2023-09-15
fix(terminal): wrap lines when adding characters in alternate screen

imsnif:
Fixes https://github.com/zellij-org/zellij/issues/2785

This was my bad, and happened when implementing https://github.com/zellij-org/zellij/pull/2654 - most changes in that PR are desired, except the one change where we actually *should* wrap lines while typing when in alternate screen mode.

Incidentally, this was only caught a week after this release because (as far as I understand), neovim 0.9.1 does not rely on this behavior while neovim 0.9.2 (which was released while I was on vacation) does. Anyhoo - we should totally support this, so this is the fix.

### DIFF
--- a/zellij-server/src/panes/grid.rs
+++ b/zellij-server/src/panes/grid.rs
@@ -1331,7 +1331,7 @@ impl Grid {
         if character_width == 0 {
             return;
         }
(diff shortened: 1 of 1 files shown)


brneor:
Ty for this! I was very confused and didn't know what was causing the issue. Zellij was the last place I looked at.

PR #2798, opened 2023-09-20
feat(client): terminal synchronized output

gmorer:
# Context
I noticed some screen flickering/tearing while using zellij and an code editor like neovim at the same time. Apparently it's possible to block the terminal emulator from updating the window: 
 https://gist.github.com/christianparpart/d8a62cc1ab659194337d73e399004036

# Tests
I tested the terminal feature detection on foot, kitty, alacrity and st, it does recognize successfully when the feature is available (foot and kitty).

**I did notice visual improvement on a low end laptop**

### DIFF
--- a/zellij-client/src/input_handler.rs
+++ b/zellij-client/src/input_handler.rs
@@ -197,6 +197,11 @@ impl InputHandler {
                 self.os_input
                     .send_to_server(ClientToServerMsg::ColorRegisters(color_registers));
             },
+            AnsiStdinInstruction::SynchronizedOutput(enabled) => {
+                self.send_client_instructions
+                    .send(ClientInstruction::SetSynchronizedOutput(enabled))
+                    .unwrap();
+            },
         }
     }
     fn handle_mouse_event(&mut self, mouse_event: &MouseEvent) {
--- a/zellij-client/src/lib.rs
+++ b/zellij-client/src/lib.rs
@@ -16,7 +16,7 @@ use std::sync::{Arc, Mutex};
 use std::thread;
 use zellij_utils::errors::FatalError;
 
-use crate::stdin_ansi_parser::{AnsiStdinInstruction, StdinAnsiParser};
+use crate::stdin_ansi_parser::{AnsiStdinInstruction, StdinAnsiParser, SyncOutput};
 use crate::{
     command_is_executing::CommandIsExecuting, input_handler::input_loop,
     os_input_output::ClientOsApi, stdin_handler::stdin_loop,
@@ -47,6 +47,7 @@ pub(crate) enum ClientInstruction {
     DoneParsingStdinQuery,
     Log(Vec<String>),
     SwitchSession(ConnectToSession),
+    SetSynchronizedOutput(Option<SyncOutput>),
 }
 
 impl From<ServerToClientMsg> for ClientInstruction {
@@ -82,6 +83,7 @@ impl From<&ClientInstruction> for ClientContext {
             ClientInstruction::StartedParsingStdinQuery => ClientContext::StartedParsingStdinQuery,
             ClientInstruction::DoneParsingStdinQuery => ClientContext::DoneParsingStdinQuery,
             ClientInstruction::SwitchSession(..) => ClientContext::SwitchSession,
+            ClientInstruction::SetSynchronizedOutput(..) => ClientContext::SetSynchronisedOutput,
         }
     }
 }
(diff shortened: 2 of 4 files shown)


imsnif:
Hey, thanks for putting this together and kudos for finding all the right places.

I'm a little concerned about support here, as there seem to be a few competing standards. Alacritty for example [supports the older DCS](https://github.com/alacritty/alacritty/issues/6507). Being a terminal multiplexer rather than a terminal emulator, our primary motivation in the case of escape sequences is compatibility.

Is there more widespread support for this standard rather than the older DCS one? Do you think it would be possible to implement both?

gmorer:
> Is there more widespread support for this standard rather than the older DCS one? Do you think it would be possible to implement both?

Look to me that the standard is the `SM/RM ? 2026`.
I will implement the old way of doing it for supporting older terminals emulator and Alacritty.

imsnif:
> Look to me that the standard is the `SM/RM ? 2026`. I will implement the old way of doing it for supporting older terminals emulator and Alacritty.

That would be amazing. I think Alacritty (not to mention gnome terminal and all the VTE family) are incredibly more wide-spread than foot and kitty combined.

gmorer:
Look like Alacritty is moving to CSI: https://github.com/alacritty/vte/commit/76b565ada7f9511b0bd718d25bfd2ce26ba78578

So not sure implementing DCS is relevant anymore.

Lockszmith-GH:
An alacritty user here, chiming in, while Alacritty will probably get updated, it will take a long time (months, upto 1-2 years) until all the stable repos of the different OSes will get these 'cutting edge' updates.
So improved support for 'older tech' would go a long way in driving adoption.

I'm an Alacritty user, and I've seen the issues you've described, just didn't know who to even describe them.

imsnif:
I tend to agree. Also I guess we probably have the same issue with the VTE family. Let's implement both.

roland-5:
Implementing both is the best approach. I'm not a programmer, but I work in system/network administration and the lion's share of my admin brothers don't care about the terminal they use - whatever is by default in the Linux distribution they use is what they will use. Those on Gnome will use gnome-terminal, those on KDE will use Konsole etc. When I showed Alacritty and said that it had GPU acceleration, they asked me why and who needed it, it's just a terminal. :P

gmorer:
I'm on holiday next week, i looked a bit into the DCS thing, might be tricky to find out if the terminal support it, but definitly dohable.

gmorer:
I pushed the changes, finding out if the terminal is using DCS is not possible.
Tmux do it by looking at the name of the terminal via the following seq: `'\x1b[>q'` to get terminal info (name, version) and know with which terminal to use it (only for iterm2, not relevant today since they use CSI). Alacritty does not implement the sequence, so i ended up by looking at the $TERM variable -__- .
If alacritty update and use CSI and if this is fix: https://github.com/alacritty/vte/issues/100 then the actual way of doing sync will be CSI since it will override the DCS value, if alacritty update without implementing the DECRQM request for sync, then **I dont know how to support both old versions and new version of alacritty**.

imsnif:
Thanks for your continued work on this and the explanations @gmorer - do you think it will be possible in the case of Alacritty to default to DCS and move to CSI if we get a proper response to the startup query?

Also, what about other terminals? I see we only do DCS with Alacritty in this case... I was led to believe this is the default with most terminals (eg. Konsole, gnome-terminal, etc.) - is it wrong?

gmorer:
> do you think it will be possible in the case of Alacritty to default to DCS and move to CSI if we get a proper response to the startup query?

With the state of the code RN in the merge request, if the user use alacritty and it's $TERM variables is correctly set to `alacritty`(should be the case) then it will default to DCS, and then if the terminal successfully answer to the startup query about sync, then the value is overwritted to use CSI instead. But this has to be implemented on alacritty, the issue I linked to the alacritty/vte repo earlier.

> Also, what about other terminals? I see we only do DCS with Alacritty in this case... I was led to believe this is the default with most terminals (eg. Konsole, gnome-terminal, etc.) - is it wrong?

From the testing i have done:
 - Gnome-terminal: Sucessfully answer that it doesnt implement the CSI based synchronisation, and doesnt implement the DCS one.
 - Konsole: Doesnt answer the CSI query, doesnt implement CSI nor DCS.
 
 From what i have found it started with iterm2 using the DCS one, then alacritty did the same, but then a more standard way has been decided, using CSI, so iterm2 moved to CSI (i dont have a mac, cant test but [`iTerm2 has now also adopted to use the new syntax instead of using DCS.`](https://gist.github.com/christianparpart/d8a62cc1ab659194337d73e399004036) ). Alacritty didnt moved (yet, incomming though).

no other terminal that i'm aware of use DCS, but [multiple already use CSI](https://gist.github.com/christianparpart/d8a62cc1ab659194337d73e399004036?permalink_comment_id=3946967#adoption-state) .

imsnif:
Great work on this! Thank you for your patience.

PR #2803, opened 2023-09-21
chore: rename file

AlixBernard:
Renamed the file `persistence.rs` to `session_serialization.rs` in order to appear as a contributor of the branch `session-resurrection`, to reflect my work on the serialization of `PaneGeom`s to the kdl format.


imsnif:
Just to be explicitly clear for the record: @AlixBernard 's work on the session resurrection feature was instrumental in making it happen. We were forced to do this because of an unfortunate git accident. Without @AlixBernard , this feature would not have come to be.

PR #2819, opened 2023-09-28
remove leftover Cargo.lock file from session-manager plugin

silwol:
It appears this file was checked into git accidentally or is no longer needed, none of the tooling regenerates it.


imsnif:
Thank you!

PR #2820, opened 2023-09-28
chore: migrate to directories from directories-next

silwol:
The `directories-next` crate used to be the go-to fork crate when `directories` was unmaintained for some time. This has changed, nowadays `directories-next` appears abandoned while `directories` has a more recent release.

fyi: I discovered this when looking into packaging zellij for Debian. We have `directories` packaged there, but don't have `directories-next`, and probably also won't, given the circumstances.


imsnif:
Thank you! Please keep them coming :)

PR #2830, opened 2023-10-02
fix(plugins): address potential security issue

tlinford:
- the upgrade from wasmer 2.3 to 3.1 brought in a version of cranelift-codegen with a security vulnerability: https://github.com/bytecodealliance/wasmtime/security/advisories/GHSA-ff4p-7xrq-q5r8
- for the removal of doubt: we caught this issue before releasing a version, so there is no version with this vulnerability
- we intend to move away from wasmer, and so decided on this temporary mitigation which does not affect our performance rather than another lengthy breaking upgrade

### DIFF
--- a/zellij-server/src/lib.rs
+++ b/zellij-server/src/lib.rs
@@ -910,8 +910,15 @@ fn init_session(
 
 #[cfg(not(feature = "singlepass"))]
 fn get_store() -> Store {
+    use wasmer::{BaseTunables, Cranelift, Engine, Pages, Target};
     log::info!("Compiling plugins using Cranelift");
-    Store::new(wasmer::Cranelift::default())
+    let mut tunables = BaseTunables::for_target(&Target::default());
+    tunables.static_memory_bound = Pages(0);
+    let compiler = Cranelift::default();
+    let mut engine: Engine = compiler.into();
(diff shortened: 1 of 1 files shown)


imsnif (approved):
Thanks for the excellent work on this!

syrusakbary:
> we intend to move away from wasmer

Oh no! Let me know if there's some feature that you guys are missing. I'd be happy to give high priority to this :)

imsnif:
Hey @syrusakbary - my apologies for taking so long to get to this and give constructive feedback.

Basically, the short of it is that we do not want to adopt wasix and as far as we can tell, there's no way around this when upgrading beyond our current version.

The reason we do not want to adopt it is that we don't need anything it offers (we really just need the filesystem interaction, everything else we gate behind our own APIs and permission system). When we briefly tried to upgrade, we got a hard crash when the laptop we were testing this on went to sleep. While I know bugs happen and these things can be fixed, this tells me that the surface area of wasix is far greater and unexpected than I am comfortable with. And since we don't need its benefits, I do not feel comfortable adopting it and exposing our users to its potential dangers (even if they can be quickly mitigated and fixed).

That being said - we really appreciate your work and this great runtime that has helped us build an amazing plugin system over the years.

PR #2837, opened 2023-10-05
Fix killing of process groups

printfn:
When running [Consul](https://www.consul.io/) in zellij, the Consul daemon is not terminated properly when killing the pane or exiting zellij entirely. This is because the consul daemon forks itself several times, and the `kill` syscall is only sent to one process instead of the entire [process group](https://en.wikipedia.org/wiki/Process_group). This means that Consul fails to restart due to the old daemon still being bound to the same port, requiring a manual `killall`.

To fix this, I've changed the `kill` function to always invoke `killpg` internally. This matches the standard job control behaviour that POSIX shells use, and ensures that processes like Consul can be terminated properly.

You can reproduce the issue as follows:
1. Install consul
2. Run zellij with the following layout file (you may need to adjust the bind IP address depending on network configuration):
    ```kdl
    layout {
        default_tab_template {
            pane size=1 borderless=true {
            }
            children
            pane size=2 borderless=true {
            }
        }
        tab {
            pane {
                command "consul"
                args "agent" "-bind" "172.20.192.145" "-data-dir" "/tmp/foo"
            }
        }
    }
    ```
3. Press Ctrl+Q to exit zellij, or Ctrl-P, X to kill the pane
4. Consul is still running

### DIFF
--- a/zellij-server/src/os_input_output.rs
+++ b/zellij-server/src/os_input_output.rs
@@ -5,7 +5,7 @@ use interprocess::local_socket::LocalSocketStream;
 use nix::{
     pty::{openpty, OpenptyResult, Winsize},
     sys::{
-        signal::{kill, Signal},
+        signal::{kill, killpg, Signal},
         termios,
     },
     unistd,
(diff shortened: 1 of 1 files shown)


imsnif:
Hey, thanks for this. I'm a little hesitant to make this change because it can have very far-reaching consequences that will be impossible for me as a single developer without access to many different platforms and setups to properly test.

What happens in your scenario if you open Consul in a terminal without Zellij and then close the terminal window?

PR #2842, opened 2023-10-07
Rename pid to terminal_id for pty 

Kangaxx-0:
The id returns from `spawn_terminal` is a unique identity for  identifying the pty master and slave FDs, thus, rename it to `terminal_id` to avoid confusion(like `pid` stands for `process id`)


imsnif:
Thanks for this. I appreciate your willingness to contribute, but in this case I think this will be confusing with different entities by the same name.

PR #2849, opened 2023-10-12
Buy some flexibility for Option

Kangaxx-0:
Minor change to use Option<& T>:

1: Flexibility: We can pass None or a reference to a T without having to reference an existing Option<T> value.
2: More Expressive Semantics: Using Option<&T> makes it clear that the function may or may not use a reference to T.

### DIFF
--- a/default-plugins/compact-bar/src/line.rs
+++ b/default-plugins/compact-bar/src/line.rs
@@ -257,7 +257,7 @@ pub fn tab_line(
     capabilities: PluginCapabilities,
     hide_session_name: bool,
     mode: InputMode,
-    active_swap_layout_name: &Option<String>,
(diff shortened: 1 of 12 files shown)


imsnif:
Thanks for this. I appreciate the thought and the time you spent on this, but the Zellij codebase is not at the stage where such changes would make sense. We're pre `1.0`, everything is in flux, this does not improve performance in a meaningful way and I think it would make sense to devote development resources (both of those developing, those reviewing and those who'll have to later deal with the merge conflicts) elsewhere.

Since this is one of a few PRs you opened, I can only once more implore you to coordinate what you'd like to work on beforehand so work will not be lost. Up to you though of course.

Issue #2852, opened 2023-10-13
Fish completion for `zellij attach` includes too many characters


loqusion:
Tab completion in fish includes ansi escape characters (used for color output) and resurrection info:

![zellij-fish-completion-issue](https://github.com/zellij-org/zellij/assets/38332081/232bd354-b69b-430d-8c37-06a77bd97506)

Instead, it should only include the session identifier (here `polite-newt`).

### Version

`zellij --version`: `zellij 0.39.0`
`fish --version`: `fish, version 3.6.1`

loqusion:
Amending the `__fish_complete_sessions` function in the completions file (`/usr/share/fish/vendor_completions.d/zellij.fish`) with `zellij list-sessions --no-formatting 2>/dev/null` fixes the ansi escape character problem on my system.

I assume this is the source:

https://github.com/zellij-org/zellij/blob/36237f04143219468395654e9d9e5e15a566d114/zellij-utils/assets/completions/comp.fish#L1-L3

After that, all that's left is the resurrection info. We could manually strip the info (e.g. using `cut`), but that relies on the output of `zellij list-sessions` being formatted a specific way. Instead, we could implement an additional option for the `list-sessions` subcommand that only includes the session identifiers.

dj95:
I implemented your suggested fix in a fork - also with the `--short` flag. If you and the maintainers are fine with this, I could create a pull request with the changes.

https://github.com/dj95/zellij/tree/feature/print-session-names-only

loqusion:
I've tested the fork and there are no issues on my end.

loqusion:
I'll go ahead and close this since #2857 has been merged.

PR #2857, opened 2023-10-14
fix(cli): session names only for attach in fish completion

dj95:
This pull request addresses https://github.com/zellij-org/zellij/issues/2852 . It implements a `--short` flag for `zellij ls`, that is then used in the fish autocompletion for completing session names without formatting and other information.

275173975-cd5f1e5c-ffa1-4465-9f7e-c427fc3da5a2

Screenshot 2023-10-14 at 23 13 20


imsnif:
Thank you!

PR #2915, opened 2023-11-07
Dropping to shell in command panes reuses CWD

### LINKED ISSUE #2910: FR: Retain cwd when drop to shell
When a session is restored after machine shutdown, all panes with a running command (e.g. nvim) would become a command pane where the user is prompted to run again, drop to shell or exit. Is it possible to configure so that the "drop to shell" option would retain the cwd of the pending command instead of the default $HOME folder?

d-alonso:
Fixes #2910 

First contribution, please let me know if I can make it better somehow (even with the conventional commits, or whatever else). I tested it locally and it seems to work correctly, and the changes are pretty uncontroversial IMO


imsnif:
Looks great, thank you!

PR #2918, opened 2023-11-07
Fix styled underlines in editors

mike-lloyd03:
Styled underlines fail to render in text editors like neovim and helix while in a Zellij session. 

This is because `RESET_STYLES` in `terminal_character.rs` sets `styled_underlines` to false. In order to address this, a new parameter is added on `Cursor::new` to read the `styled_underlines` config variable as it's passed down the stack from `Screen` > `Tab` > `TerminalPane` > `Grid`.

I've tried to make this as minimal of a change as possible but some significant boilerplate had to be added.

Finally, I was only able to test this in Neovim. Helix, for whatever reason, ws not rendering styled underlines inside or outside of Zellij. Probably an issue with my config. But if someone else can confirm this works, that would be greatly appreciated.

### DIFF
--- a/src/tests/e2e/remote_runner.rs
+++ b/src/tests/e2e/remote_runner.rs
@@ -236,6 +236,7 @@ fn read_from_channel(
                 let sixel_image_store = Rc::new(RefCell::new(SixelImageStore::default()));
                 let debug = false;
                 let arrow_fonts = true;
+                let styled_underlines = true;
                 let mut terminal_output = TerminalPane::new(
                     0,
                     pane_geom,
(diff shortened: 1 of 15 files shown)


zicklag:
I can confirm that this works with helix in alacritty. 🚀

mike-lloyd03:
@zicklag thank you. Would you mind test driving this branch for a few days to see if any other issues pop up? There may be edge cases where styled underlines are being disabled.

zicklag:
Sure thing! I already replaced my existing installation with this one. I'll let you know if I run into anything.

imsnif:
Great work! Thanks for the super quick fix.

danillos:
Working for me on Helix too.

alxpettit:
Strangely enough, this isn't working consistently for me in Wezterm.

rockboynton:
@crabdancing what was the issue with Helix and Wezterm.
I am seeing undercurls running Helix in Wezterm, but _not_ when running Helix in Zellij in Wezterm

PR #2920, opened 2023-11-08
optimize: update refresh prorcess for sysinfo to reduce cpu usage

xuanyuan300:
Use `refresh_processes_specifics`   will use more cpu. Use refresh_process_specifics instead of refresh_processes_specifics to reduce cpu usage.


xuanyuan300:
> Good catch!! I'd be happy to merge this once it's formatted (I tried pushing to your branch but did not have permissions). Maybe you'd like to do a `cargo fmt`?

Thanks feedback.
Okay, let me run `cargo fmt`.

PR #2935, opened 2023-11-13
fix: fix flakey scroll history

### LINKED ISSUE #2881: Flakey scroll history after scrolling up inside less
In https://github.com/zellij-org/zellij/issues/797 I mentioned I also sometimes see some artifacts when scrolling up in zellij.
The linked issue was resolved, but as of zellij 0.38.2, I still see some artifacts. This makes me believe there is another, different problem.

In the mean time, I found a way to reproduce my issue. It seems to involve the pager `less`, but with a specific configuration so that the last paged content stays on the screen when less is exited with `q` (not using alternate screen of the terminal that would clear the text on exit). I found that git does this when paging, but probably there is a way of invoking `less` directly with the correct flags.

The specific steps that reproduce the bug on my machine:
- enter a git repo with enough commits that you can fill your screen with
- `PAGER=less git log --oneline`
- scroll down inside `less` by pressing `Page-down` a couple of times
- scroll up inside `less` by pressing `Page-down` or multiple times `arrow-up`
- **optional**: Exit `less` by pressing `q` (the last paged content must stay on the screen), continue working in the terminal
- scroll up inside zellij with the mouse wheel
- Observe garbled scrollback somewhere in the `less` output

Screenshot before scrolling up:
image
After:
image

xuanyuan300:
Fixes #2881.
Regarding #2881, I found that if I set `$TERM=linux` or `vt100`, execute the less command, exit, and then scroll up, flakey scrolling history happens.
When the less command is executed, the b'M' instruction will be generated and the `move_cursor_up_with_scrolling` function will be called. At this time, the viewport should insert the `canonical row `(there is actually a comment in the code). After testing, it seems normal.
I'm not sure if my understanding is correct. Looking forward to your feedback.


imsnif:
Good catch, thank you!

PR #2955, opened 2023-11-20
DRAFT: fix: ctrl+t c should rename to match pane rename

qballer:
Suggestion 
==============

I would like to offer a small fix and change the default tab rename to `ctrl+t c`. This should match `ctrl+p c` which renames a pane. I understand that people might be used to this keybinding, which is a legitimate reason to reject this PR. 

The reason I didn't use `ctrl+p r`  for rename is because that is used for split right. I assumed that moving that for someone who is used to that would be extra annoying.


qballer:
I've closed this for now as we don't want to break current users.

PR #2957, opened 2023-11-21
docs(manpage): improved default config file & directory information

jstamant:
Modified some of the manpage content to be more accurate and complete for command-line use.

Originally, I was only going to change the references of `$XDG_HOME` to `$XDG_CONFIG_HOME`, but then I ended-up looking through the code to see what the actual order of loading config files was.

I tested the output by building the manpage, and it looks good.


antonkesy (approved):
LGTM

PR #3038, opened 2023-12-28
feat(cli): add run/edit in-place functions

cristiand391:
Adds new shell functions for `zellij <run|edit> --in-place`:

ref: https://discord.com/channels/771367133715628073/771367133715628076/1189219893875318834

I looked at the code that generates shell completion and _I think_ no other change is needed for this, but let me know if I missed something.

Also tried building zellij locally and run `zellij --setup --generate-completion zsh` but it always panics (even without my changes):
image

e2e also passed on my fork 👍🏼


imsnif:
Thanks for this, sorry for taking so long to take to it!

PR #3043, opened 2023-12-29
Improve reflow performance and general memory usage by reducing size of TerminalCharacter

aidanhs:
cc https://github.com/zellij-org/zellij/issues/2622
This PR probably moves zellij to 'good enough for me' that it won't completely lock up on long lines.

Third in a series: https://github.com/zellij-org/zellij/pull/3045, https://github.com/zellij-org/zellij/pull/3032, https://github.com/zellij-org/zellij/pull/3043 (this pr)
This PR builds on the first+second and contains all commits from them - it will be easier to review after merge of the of them both.

There is a refactor (enabled by the previous PR in the series) to merge the two separate branches (`new_columns != self.width` + `new_height != self.height`) for viewport transfer.

The primary optimisation pursued here is in reducing the size of TerminalCharacter so the memory copies performed when reflowing are much cheaper. This also has the nice side effect of reducing overall memory usage.

I also added a few minor cleanups I noticed, and made zellij-utils stop unconditionally rebuilding.

I *highly* recommend reviewing these changes commit-by-commit, as I've carefully tried to make each one standalone and self-consistent.

PR #3032, opened 2023-12-27
Fix some real/saved cursor bugs during resize

aidanhs:
cc https://github.com/zellij-org/zellij/issues/2622

Second in a series: https://github.com/zellij-org/zellij/pull/3045, https://github.com/zellij-org/zellij/pull/3032 (this pr), https://github.com/zellij-org/zellij/pull/3043
This PR builds on the first and contains all commits from it - it will be easier to review after merge of the first.

This PR contains small refactors and a fix for some cursor bugs, in preparation for the final PR in the series.

I *highly* recommend reviewing these changes commit-by-commit, as I've carefully tried to make each one standalone and self-consistent.

Issue #2622, opened 2023-07-13
Zellij hangs with 100% CPU when opening a new pane with a very long line on screen


aidanhs:
```
~ $ zellij --version
zellij 0.37.2
~ $ stty size
55 211
~ $ uname -av
Linux a-winthinkpad 5.19.0-45-generic #46~22.04.1-Ubuntu SMP PREEMPT_DYNAMIC Wed Jun 7 15:06:04 UTC 20 x86_64 x86_64 x86_64 GNU/Linux
~ $ gnome-terminal --version
# GNOME Terminal 3.44.0 using VTE 0.68.0 +BIDI +GNUTLS +ICU +SYSTEMD
```

Reproduction:
```
$ zellij
$ cat x.c
#include <stdio.h>
int main() { while (1) { printf("/this/is/a/quite/long/path/for/printing"); } }
$ gcc -O2 x.c
$ ./a.out
```
Wait for 5s then hit ctrl+c. Now attempt to open a new pane with ctrl+p, n. Zellij will be unresponsive and taking 100% of a CPU.

andreytkachenko:
Yes, very annoying bug.

aidanhs:
For watchers of this issue, my two PRs linked above improve the situation by at least 10x, more for longer lines (as some of the behavior was quadratic). Up to 10MB lines the performance is now tolerable.

If you're building for yourself, take the second PR - it includes all commits from the first.

aidanhs:
The three PRs have been merged and presumably will be in the next release. In the original test case, trying to split to a new pane might respond in about 5 mins (sometimes less, sometimes more). After the changes, it takes <10s for me - not fast, but at least you can slowly navigate back to the original pane and `clear` or `reset` to fix the issue.

I am ok with this performance for now. Not ideal, but good enough that I don't need to force close sessions - I don't mind if this issue is closed.

[...]

PR #3045, opened 2023-12-30
Improvements to reflow performance by removing O(n^2) behavior

aidanhs:
cc https://github.com/zellij-org/zellij/issues/2622

First in a series: #3045 (this pr), #3032, #3043

This PR contains targeted improvements to reflow performance. Gives about 3x improvement on a 1MB line, significantly more on a 10MB line (because of quadratic behavior).

I highly recommend reviewing these changes commit-by-commit, as I've carefully tried to make each one standalone and self-consistent.

### DIFF
--- a/zellij-server/src/panes/grid.rs
+++ b/zellij-server/src/panes/grid.rs
@@ -166,35 +166,26 @@ fn transfer_rows_from_viewport_to_lines_above(
     count: usize,
     max_viewport_width: usize,
 ) -> isize {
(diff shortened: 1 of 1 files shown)


imsnif:
This looks great. Thank you for your patience and for breaking this down into small chunks. This is very old code that has not seen as much attention as I would have liked.

Great work, I'm going to also take a look at the other PRs.

giyany:
Thank you, I have been waiting for this improvement for literally over a year.

PR #3112, opened 2024-02-06
feat(sessions): welcome screen

imsnif:
img-2024-02-01-174741

This PR adds an optional "welcome screen" to Zellij which includes a main menu functionality. When starting Zellij with the welcome screen, we'll have the option to:
1. Start a new session with an optional name and an optional layout (the layout list will include the built-in layouts such as "compact" as well as the user configured layouts that appear in the "layouts" configuration folder).
2. Attach to an existing session
3. Resurrect a dead session

Additional functionality includes:
1. Deleting one or more dead sessions (from the "Resurrect Session" screen)
2. Killing one or more active sessions (from thje "Attach" screen)
3. Disconnecting all other clients from the current session (mostly relevant when using the session-manager)

To use it, start Zellij with `zellij -l welcome`.

### How does it work?
The new welcome screen is a new built-in layout called "welcome" which includes an instance of the "session-manager" plugin with the `is_welcome_screen true` configuration. This configuration creates some UI niceties (such as rendering a welcome banner) and is geared toward being the only pane in the session. Once a new session is started (or once the user attaches to/resurrects an existing session) the original session is closed (and deleted).

### Intended use
The recommended use for the welcome screen would be to hook it up so that it launches every time we open a terminal (eg. instead of the default shell). This way, when we launch our terminal we'll be greeted with a menu allowing us to either start a new default session (by quickly pressing `ENTER` twice), start a specialized session with a custom layout (eg. some troubleshooting session, a project specific session, etc.), attach to an existing session (perhaps running in another terminal window we can't find at the moment), or resurrect an old session.

I believe this can be a very powerful new way of using the terminal and managing multiple sessions and contexts.

### `session-manager` Changes and Additions
This PR includes some nice UI changes as well as functionality additions to the `session-manager`, chief among which is the ability to disconnect all other clients in this session. This is especially useful when connecting to a single session from multiple computers and wanting to reset its size to the size of the current machine by disconnecting all other clients.

It's also now possible to kill one or multiple other sessions on the machine, thus cleaning up unused sessions (the upshot here is that if we do this with the "Intended use" described above, the terminal windows of the other sessions will be closed as well).

### Plugin API additions
Several new plugin API methods were added to allow the above functionality:
#### switch_session_with_layout
Same as `switch_session`, but allowing the user to specify one of the built-in or custom user-layouts by name. The list of these layouts is also now available as part of the `SessionInfo` event.

#### disconnect_other_clients
Disconnect all other users except the current one connected to this session.

#### kill_sessions
Kill all sessions in the list by their name.


dtomvan:
IDK if this should have a new issue, but here goes: maybe it could use a way for the (more experienced) user to get rid of dome of the visual clutter?

imsnif:
I'm sorry you don't like the "visual clutter". Please note that the UI decorations only appear if there are room for them on the screen, so they don't take away space from content.

Otherwise, you can start the session-manager with `is_welcome_screen false` to get the classical `session-manager` experience, or you can of course always fork the plugin and do whatever you like.

dtomvan:
Sounds good, thank you!

PR #3123, opened 2024-02-08
Working read and write to/from pty

ulyssesdotcodes:
Reading from the pty has to be blocking and threaded so that writing can still happen. I'm not entirely sure which of the pty's threads are blocked or locked or anything, but this seems to do the trick.

Just a heads up that I had an issue testing with my usual starship prompt - maybe there's funky utf-16 codes that aren't properly translated?


ulyssesdotcodes:
Sorry, accidentally opened on the wrong repo. It was a mistake!

PR #3125, opened 2024-02-09
fix(terminal): curly underlines mid-chunk

imsnif:
This is a small fix for a recent regression in main (that was not released yet) in which curly underlines were not rendered mid line.

@aidanhs - any chance you could briefly take a look to make sure I didn't ruin any of your improvements?


aidanhs:
Looks good to me

PR #3167, opened 2024-02-28
fix(session): dead sessions without layout shouldn't be resurrectable

pedromfedricci:
It is possible for a dead session to not have a committed **session-layout.kdl** file, either because serialization is turned off, or because the session was killed before being flushed into disc. Since there isn't a file to load from, the session shouldn't be marked as resurrectable (for example, by the session-manager plugin), because its state is lost. As an example, the session-manager plugin will simply load the default layout for the dead session if selected, so it is just better to not list it at all.


pedromfedricci:
Fixed by: #3264

PR #3179, opened 2024-03-02
Add support for Ctrl and Alt modifiers on function keys in keybindings.

hovinen:
Previously, it was not possible to define keybindings with a modifier and a function key. The `Key` enum only supported combinging the Ctrl and Alt modifiers with letters. This is somewhat limiting: I would like to make the Zellij keybindings more "distant" than those used in the programs I use from within Zelilj, so that Zellij does not intefere with those programs. Thus I would like to move some of the keybindings from Ctrl+_character_ to Ctrl+_function key_.

This change adds:

  * support for function keys with the Ctrl and Alt modifiers in the `Key` enum,
  * support for parsing such keybindings from the configuration file,
  * support for such keybindings in the protobuf which communicates the keybindings to plugins, and
  * support for these keybindings in the plugin API.

This is tested by modifying one of the e2e tests to include an example of such keybindings. This verifies that the configuration is correctly parsed, communicated with the plugin, and rendered.

### DIFF
--- a/default-plugins/status-bar/src/first_line.rs
+++ b/default-plugins/status-bar/src/first_line.rs
@@ -62,6 +62,8 @@ impl KeyShortcut {
         } else {
             match key {
                 Key::F(c) => format!("{}", c),
+                Key::CtrlF(n) => format!("F{}", n),
+                Key::AltF(n) => format!("F{}", n),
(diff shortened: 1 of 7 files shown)


imsnif:
Great work! Thank you for this.

PR #3182, opened 2024-03-05
Reuse allocations if available when swapping alternate state

tgross35:
This just does what I suggested in https://github.com/zellij-org/zellij/issues/1625#issuecomment-1970728633 - instead of dropping and recreating the collections, reset them so future use doesn't need to reallocate.

Tests pass; however, I need some help figuring out what action actually exercises this code path via the binary.

### DIFF
--- a/zellij-server/src/panes/grid.rs
+++ b/zellij-server/src/panes/grid.rs
@@ -2,6 +2,7 @@ use super::sixel::{PixelRect, SixelGrid, SixelImageStore};
 use std::borrow::Cow;
 use std::cell::RefCell;
 use std::collections::HashMap;
+use std::mem;
 use std::rc::Rc;
 use zellij_utils::data::Style;
 use zellij_utils::errors::prelude::*;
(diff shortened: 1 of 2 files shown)


imsnif:
Hey @tgross35 - thanks for looking into this! I'll try to make some time for a detailed review in the coming weeks, but for now - you can test the alternate screen either manually by sending the following ANSI codes from within a Zellij pane like so:
```
echo -e "\033[?1049h" # enter alternate screen
echo -e "\033[?1049l" # leave alternate screen
```
Or start some sort of TUI program that does this for you (eg. vim, htop, etc.)

imsnif:
Hey @tgross35 - so I'm finally getting to this, thank you for your patience!

So first, a bit of terminal lore just to make sure we're on the same page:
The alternate buffer of a terminal is often used by "full screen" terminal applications in order to render themselves on the full viewport while still preserving our terminal history when we exit them.

So when the application sends the terminal the `<CSI>?1049h` instruction, we save all the current terminal state, place it aside and give it a fresh new-one to render itself on (minus the scrollback since it only needs the viewport). When the application sends us the `<CSI>?1049l` instruction, we switch everything back - erasing the intermediate alternate buffer.

While it's technically possible to enter an alternate buffer on top of an already existing alternate buffer - I have never encountered an application that does this.

So now to these changes: it seems to me that the only occasion in which these changes will have an effect (essentially this branch being used: https://github.com/zellij-org/zellij/pull/3182/files#diff-758384aed3fe1bc47b267810b52b79e7377b5daf22cd72515710c8cd2a5f89ecR2639) is when entering an alternate buffer on top of an already existing alternate buffer. The reason being that when we leave an alternate buffer, we make sure (I hope) to drop it. This happens here: https://github.com/zellij-org/zellij/blob/47b1d0e7f07d0e0a44c64776a4977cae95247616/zellij-server/src/panes/grid.rs#L2532

So this leaves me with a question I hope you can help me out with: do we properly drop the alternate buffer state when leaving it?

PR #3193, opened 2024-03-10
Feature/format UI components

Zykino:
Add ways to serialize/format ui components instead of only proposing `print` options.

I can remove the first commit if you want, but it did not work on my PC without it:
1. `rust -vV` error when using the "previous" rust toolchain.
2. `protoc` complain that `optional` is an experimental feature of the syntax3.

I also changed the interface to borrow the `Text` elements instead of owning it, so it can be reused.

I had some reflection on the `ribbon`: plugin developers will often want to have multiples of them in a row, so accepting an `into_iter` sounded nicer.
I also wanted to have a possibility to color the full line with the background color, but I’m not sure how to get the bg color from the user’s theme.

Note: I changed the session plugin as needed and others, but I did not `run` everything plugin before/after. However I did not see obvious breakage in the "after" state.

### DIFF
--- a/zellij-tile/src/ui_components/nested_list.rs
+++ b/zellij-tile/src/ui_components/nested_list.rs
@@ -1,4 +1,5 @@
 use super::Text;
+use std::borrow::Borrow;
 use std::ops::RangeBounds;
 
 #[derive(Debug, Default, Clone)]
@@ -44,12 +45,7 @@ impl NestedListItem {
 
 /// render a nested list with arbitrary data
 pub fn print_nested_list(items: Vec<NestedListItem>) {
-    let items = items
-        .into_iter()
-        .map(|i| i.serialize())
-        .collect::<Vec<_>>()
-        .join(";");
-    print!("\u{1b}Pznested_list;{}\u{1b}\\", items)
+    print!("{}", serialize_nested_list(&items))
 }
 
 pub fn print_nested_list_with_coordinates(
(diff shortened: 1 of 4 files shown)


Zykino:
Thinking about it again, I kind of want to have an `SerializeAnsi` trait that requires the equivalent of `serialize_text_with_coordinates` and `serialize_text`. Maybe another equivalent trait for iterable (useful for ribbon’s line and nested list’s view, maybe even tables?).

But in the same time I’m not sure it will help format/print elements. It may be too much into the "gui framework" side, which is apparently already hard to do (at least in rust).

What do you think ?

imsnif:
Hey @Zykino - sorry for taking so long to get to this and thanks for your patience!

Any chance we can weed this down to just adding a simple serialization interface? I appreciate the thought you put into code reuse as well as improving existing build issues and such, but I have to minimize the surface of changes due to the size of this project and the burden of maintainership. Thanks for understanding.

Zykino:
* I removed the compilation commit (kept it in my fork so I can re-use it if needed).
* Transformed the `print_*` function to take `&references` so we don’t need to clone elements if we want to use print them multiples times
* Add `serialize_*` equivalent function that returns `String`s
* For the ribbon, I think it is interesting to give multiples items and print them in a line (like for the table, and the nested list), so I added `serialize_ribbon_line*`
* Fixed the compilation where it now needs reference

imsnif:
hi @Zykino - thanks for your work and efforts on this. Can we please change this to just add the serialization interface without changing anything else?

imsnif:
Thanks for this and for the changes!

PR #3196, opened 2024-03-12
Improve action new-tab

### LINKED ISSUE #2981: can't set working directory for new tab w/o layout
zellij 0.39.2

```
$ zellij action new-tab -n ok -c /                                                                                                                                                                                                                                
error: The following required arguments were not provided:                                                                                                                                                                                                                     
    --layout <LAYOUT>
```

however, without the `-c` option, the tab opens fine. it seems odd to require a layout

Workaround: `zellij action new-tab -n ok -c /  -l default`

abatkin:
Improvements to `zellij action new-tab`:
- setting `--cwd <dir>` no longer requires specifying a layout
- add `--run <command>` (and `--args <arg>`) to run a command
- add `--edit <filename>` to launch editor of the specified file

A few other cleanups along the way:
- Update `.gitignore` to add `.idea` project files (JetBrains tools like CLion/RustRover)
- Update `.gitignore` to add `docker-compose.override.yml` (for example to set additional SELinux settings when running tests)
- Quote "true" in `docker-compose.yml` since otherwise it could be treated as a boolean and rejected

Fixes #2981


abatkin:
A few questions:
1. Should an entry be added to `CHANGELOG.md`?
2. Should the `.wasm` files be checked in? I'm guessing no, so I can pull those back out (otherwise there will be conflicts everywhere, right?)
3. A few `.snap` files had to be updated, since it was necessary to change the `NewTab` interface, is that okay? I don't see another way around changing that, but I hope that doesn't break anything else that I haven't thought about
4. `TerminalAction` gets directly serialized now, and I'm not sure if that's appropriate, or if there should be an intermediate representation (that will be otherwise identical, at least for now) for the client-server datamodel

s1nistr4:
Man, zellij action new-tab --cmd "" needs to exist, this is keeping a lot of people on tmux because you can't do any scripting with zellij tabs, layouts don't really work for this either without doing a bunch of workarounds with sed. Really hope this PR gets approved.

devyntk:
Bumping this issue - currently this is blocking me from implementing some extremely useful scripting that I use all the time in tmux. The ability to create a new tab with a specified command in it seems like an extremely basic and necessary feature for any amount of scripting in Zellij, and is a major detriment to adoption of Zellij in my team.

abatkin:
A few updates:
- rebased from master, fixed conflicts
- removed `.wasm` files from PR since they will always conflict, presumably whoever handles releases will update those

Tests pass, and I've been playing around and it seems to work properly.

imsnif:
Hey @abatkin - first, thank you for your patience in waiting for a review and even rebasing the code to the latest changes.

While I'm generally happy to add this feature, I'm not thrilled about the implementation. Nothing to do with you or your skills of course (on the contrary, I'm very impressed with all the non-trivial threading you did here), but I think overriding the default shell in this case is not the best way forward.

The internal and external way Zellij communicates about commands and their placement is through layouts. Bypassing this can have present and future unforeseen issues. For example, checking out this branch - I do not get an error if I `--run` a non-existent command. I just get a blank pane opened. Whereas if I did this through a layout (eg. with `pane command="lkjdslkjdsf"`) I'd get a descriptive error inside the pane.

Another issue is that passing args to a command-line through a different command-line is... not ideal. Honestly, I personally could not find a way to get this to work:

```
$ target/dev-opt/zellij action new-tab --run htop --args --tree
error: Found argument '--tree' which wasn't expected, or isn't valid in this context

        If you tried to supply `--tree` as a value rather than a flag, use `-- --tree`

USAGE:
    zellij action new-tab --run <RUN> --args <ARGS>

For more information try --help
```

I do realize that having to provide a layout to `action new-tab` every time just to run one command is a bit cumbersome. So I'd be happy to provide users with some sort of shortcut, but I'd like this shortcut to be native to the way Zellij does things. Namely, have the command inserted in the user's default layout and then that layout run as usual. Implementing this might be a bit tricky, but I can provide pointers.

Let me know if you're interested in seeing this through.

PR #3206, opened 2024-03-19
feat: add *Nushell* autostart script

stevenxxiu:


stevenxxiu:
I've fixed the formatting error.

stevenxxiu:
Hopefully it works now. I just cherry picked the commit in my fork, and my run <https://github.com/stevenxxiu/zellij/actions/runs/8868986642/job/24349164906> passed.

PitiBouchon:
Is this MR still ongoing ?

stevenxxiu:
It's waiting for review.

zellij-utils/src/setup.rs
```diff
@@ -619,6 +625,13 @@ impl Setup {
     }
 
     fn generate_auto_start(shell: &str) {
+        let mut out = std::io::stdout();
+
+        if shell.to_lowercase() == "nu" {
```
KamWithK, line 630: Would be nicer to remove this and just add an entry to the match below instead imo

KamWithK: Okay maybe `Shell::` won't be possible since [Clamp has nushell support](https://github.com/clap-rs/clap/blob/master/clap_complete_nushell/src/lib.rs#L30), but [isn't integrated](https://github.com/clap-rs/clap/blob/master/clap_complete/src/aot/shells/shell.rs#L14) into the shells enum

Nonetheless should still be possible to add `"nu" => {...}`?

stevenxxiu: I need to do `shell.to_lowercase()` in this case, like the other shells. Since it's not an instance of `Shell`, this is the only way I could think of.

KamWithK: OK fair

Probably worth upstreaming to add Nushell as a supported shell too tho, not sure if there's something left to implement or whether it just didn't get added to their supported shells list from my quick look though

Issue #2760, opened 2023-08-31
compilation fails: `reference to packed field is unaligned`


selfagency:
i realize you don't specifically support windows, but i thought you might be interested to know that cargo installation fails under windows due to the following error:

```sh
~ > cargo install zellij                                                                          08/31/2023 10:07:28 AM    Updating crates.io index
  Installing zellij v0.38.1
    Updating crates.io index
   Compiling proc-macro2 v1.0.66
   Compiling unicode-ident v1.0.11
   Compiling autocfg v1.1.0
   Compiling cfg-if v1.0.0
   Compiling version_check v0.9.4
   Compiling serde v1.0.188
   Compiling syn v1.0.109
   Compiling winapi v0.3.9
   Compiling once_cell v1.18.0
   Compiling thiserror v1.0.47
   Compiling windows_x86_64_msvc v0.48.5
   Compiling libc v0.2.147
   Compiling getrandom v0.2.10
   Compiling memchr v2.6.2
   Compiling value-bag v1.4.1
   Compiling crossbeam-utils v0.8.16
   Compiling ahash v0.7.6
   Compiling cc v1.0.83
   Compiling indexmap v1.9.3
   Compiling scopeguard v1.2.0
   Compiling windows-targets v0.48.5
   Compiling rustc-demangle v0.1.23
   Compiling windows-sys v0.48.0
   Compiling smallvec v1.11.0
   Compiling bitflags v1.3.2
   Compiling pin-project-lite v0.2.13
   Compiling ident_case v1.0.1
   Compiling fnv v1.0.7
   Compiling quote v1.0.33
   Compiling hashbrown v0.12.3
   Compiling rustversion v1.0.14
   Compiling num-traits v0.2.16
   Compiling rkyv v0.7.42
   Compiling syn v2.0.29
   Compiling target-lexicon v0.12.11
   Compiling seahash v4.1.0
   Compiling memoffset v0.9.0
   Compiling lazy_static v1.4.0
   Compiling futures-core v0.3.28
   Compiling rand_core v0.6.4
   Compiling more-asserts v0.2.2
   Compiling siphasher v0.3.11
   Compiling rand v0.8.5
   Compiling crossbeam-epoch v0.9.15
   Compiling proc-macro-error-attr v1.0.4
   Compiling phf_shared v0.11.2
   Compiling backtrace v0.3.69
   Compiling concurrent-queue v2.2.0
   Compiling crossbeam-channel v0.5.8
[...]

PR #2926, opened 2023-11-10
Windows support

iXialumy:
I am starting to track things that need work at the top now for people who just want to see the status or that want to help and need a "thing" to work at:

- [x] IPC (basic with 2 pipes)
- [ ] IPC (with the "correct" duplex pipe approach)
- [x] IPC performance investigation
- [x] Reading from pty
- [x] Writing to pty
- [x] Recognizing control characters
- [ ] Listing sessions
- [ ] Reattaching to sessions

---
Original Message:

I have gotten it to compile under windows and (hopefully) IPC working.
I hope this did not break anything for non-windows systems, but I did not test it at the moment.

From here I hope the compiler will guide me to all the places that have yet to be implemented.

PR #3214, opened 2024-03-21
Update `mio` to the latest version

tgross35:
This is one package that previously depended on an outdated version of `ntapi`, which is causing build issues at
<https://github.com/zellij-org/zellij/issues/2760>.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2181,36 +2181,14 @@ dependencies = [
 
 [[package]]
 name = "mio"
-version = "0.7.14"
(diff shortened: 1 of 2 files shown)


imsnif:
I appreciate you putting this up, but I'm very conservative about dependency updates in this project. Since the only direct effect this has is on a still unsupported platform, I'd prefer to leave things as is. Maybe this can happen as part of https://github.com/zellij-org/zellij/pull/2926 ?

PR #3244, opened 2024-04-04
utils/cli: Add `move-tab` action

har7an:
which was previously only exposed through keybindings. This is a follow-up to #3047 which likely forgot to add this.


pencilcheck:
I can't find this command in the official documentation, I think this will help a lot of newcomers to find this if this is also on documentation with examples

PR #3255, opened 2024-04-11
fix(theme): this was a fast fix for gruvbox light and dark

uncomfyhalomacro:
the previous themes do not have the correct colors and aren't great for the eyes. when selecting e.g. to copy text, it uses red for light theme and the same bg for dark theme for highlighting text.

the previous light theme is not even a light theme so I also fixed that as well.

### Dark Theme Now
image

### Light Theme Now
image

### DIFF
--- a/zellij-utils/assets/themes/gruvbox.kdl
+++ b/zellij-utils/assets/themes/gruvbox.kdl
@@ -1,20 +1,20 @@
 themes {
     gruvbox-light {
-        fg 124 111 100
-        bg 251 82 75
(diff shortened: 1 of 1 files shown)


uncomfyhalomacro:
I was so confused at first. But I hope this helps everyone who uses gruvbox

NachoNievaG:
Highly agree with this theme.
Thanks, already using it in my dotfiles.

imsnif:
Amazing, thank you! And thanks for your patience.

Issue #3248, opened 2024-04-06
ui(compact-bar): mode indicator not centered


cristiand391:
# 2. Issues with the Zellij UI / behavior / crash

## Issue description

I noticed the mode indicator isn't correctly centered when the mode name length > 6, you can see in this video how when switching to renametab/renamepane  the text isn't padded (IIRC it's not noticeable on the left side because the `Zellij (session-name)` section includes a whitespace).
[Screencast from 04-06-2024 05:32:06 PM.webm](https://github.com/zellij-org/zellij/assets/6853656/40e8096b-d4db-4cd0-88c6-ffdb1f7079f9)

This is how I improved the padding in my fork of compact-bar, you can see the mode name is always centered (even short ones like `pane`):
[Screencast from 04-06-2024 05:44:54 PM.webm](https://github.com/zellij-org/zellij/assets/6853656/eeaf4a3d-b5bb-4773-8262-bf8854b447a0)

https://github.com/cristiand391/zj-status-bar/commit/2de3d60446dfe8fe9d9037122d77daf927be8d7d

If the current padding behavior in zellij's compact-bar isn't the intended and the maintainer agrees, I'm happy to contribute the fix upstream.

## Minimal reproduction

save this as layout
```
layout {
  pane size=1 borderless=true {
    plugin location="zellij:compact-bar"
  }
  pane
}
```
switch to renametab or any mode name > 6:
```
zellij action switch-mode renametab
```

![image](https://github.com/zellij-org/zellij/assets/6853656/4c7e742c-f915-4534-91b3-0524f2319373)

## Other relevant information
```
zellij --version
zellij 0.39.2

stty size
35 159
```

Zykino:
Love that fix, looks elegant to me. Not sure what maintainers would think of it.

cristiand391:
> Just thinking about it, another/easier fix would be: let mode_part_padded = format!(" {} ", mode_part); isn’t it?

ha, yes that should work too, good catch. I probably went with the aligment solution because I read that was a thing in the docs 😄 
https://doc.rust-lang.org/std/fmt/index.html#fillalignment

Zykino:
Yup, and the starting code you had was like this, so I think I would have done the same at first too ^^.

jaeheonji:
That's great, I love it too. If possible, can you make a PR?

cristiand391:
@jaeheonji 
Here it is: https://github.com/zellij-org/zellij/pull/3260
[...]

PR #3260, opened 2024-04-12
fix(compact-bar): properly pad mode indicator

cristiand391:
This PR updates the compact-bar plugin to make the mode indicator to be always centered.

### before

mode indicator section hard-coded to 8-chars length, on mode names > 6 there's no padding

compact-bar-old

### after

mode section gets printed as ` <mode-name> `, use a whitespace for padding.

compact-bar-new

Closes https://github.com/zellij-org/zellij/issues/3248

### DIFF
--- a/default-plugins/compact-bar/src/line.rs
+++ b/default-plugins/compact-bar/src/line.rs
@@ -199,7 +199,7 @@ fn tab_line_prefix(
         tab_index: None,
     }];
     if let Some(name) = session_name {
-        let name_part = format!("({}) ", name);
+        let name_part = format!("({})", name);
         let name_part_len = name_part.width();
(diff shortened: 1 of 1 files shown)


default-plugins/compact-bar/src/line.rs
```diff
@@ -199,7 +199,7 @@ fn tab_line_prefix(
         tab_index: None,
     }];
     if let Some(name) = session_name {
-        let name_part = format!("({}) ", name);
+        let name_part = format!("({})", name);
```
cristiand391, line 202: This is the session-name part, remove the whitespace on the right, the mode part will handle padding

image

jaeheonji (approved):
LGTM! Thanks👍

Issue #3269, opened 2024-04-16
action rename-tab not functioning


burgessa23:
## Issue description

`zellij action rename-tab` does not change tab name 

## Minimal reproduction

1. open a new tab
2. run `zellij action rename-tab "alice the cat"`
3. notice that the tab name has not changed

## Other relevant information
zellij version 0.4.0

tmerse:
For me `zellij action rename-tab` will change the tab name to empty string after upgrading to 4.0.

Anyhow congrats and thanks for the 4.0 release!

Segaja:
I can confirm this issue and this also affects `zellij action rename-pane "test"` which just doesn't do anything.

For neither of these commands is anything written to the log.

marmeladema:
Hi!
I am hitting this bug with 0.40.0, would it be possible to tag a new release?
Or is the appropriate thing to do is to use the main/master branch?

imsnif:
Please be patient, a patch will come when it's ready. You are more than welcome to use main if you like.

PR #3295, opened 2024-04-24
fix(cli): tab and pane name input

imsnif:
Fixes https://github.com/zellij-org/zellij/issues/3269

This was a regression caused by the `pipes` implementation and some refactoring in this area. Namely, only the first action created from a single cli command was used and the rest were ignored. This only affected renaming panes and tabs through the CLI.


tennox:
Confirmed working, thx :)

Issue #2679, opened 2023-08-07
Can't bind Ctrl-j


willothy:
`zellij --version`: 0.37.2
`stty size`:  50 254
`uname -av` or `ver`(Windows): Linux ... 6.4.8-arch1-1 #1 SMP PREEMPT_DYNAMIC ... x86_64 GNU/Linux

`wezterm --version`: wezterm 20230717-111633-94ab0496

I'm attempting to setup smooth pane changes between Neovim, Wezterm and Zellij. My intended bindings use `Ctrl h/j/k/l` to move focus, and `Alt h/j/k/l` to resize. All of those work, except for `Ctrl j`, and I can't figure out why. I've disabled default keybinds in Zellij, ensured that the key isn't bound to anything in Wezterm, and still can't get it to work as expected.

To reproduce: 
- Set a normal mode binding `bind "Ctrl j" { MoveFocus "Down"; }`

willothy:
Seems like a `termwiz` issue, see https://github.com/wez/wezterm/issues/2511

p00f:
I can't bind Ctrl m either

ssipos90:
I can't bind it either. Ctrl-j sends an enter. In tmux navigation is fine.

```
    normal clear-defaults=true {
        bind "Ctrl a" { SwitchToMode "Tmux"; }

        bind "Ctrl h" { MoveFocus "Left"; }
        bind "Ctrl j" { MoveFocus "Down"; }
        bind "Ctrl k" { MoveFocus "Up"; }
        bind "Ctrl l" { MoveFocus "Right"; }
    }
```

zellij 0.37.2 
alacritty 0.12.2 (9d9982df)
Manjaro Linux 23.0.0

flinesse:
It seems that zellij mimics the behavior of many line editor's response to control sequences. As a quick test in addition to https://github.com/zellij-org/zellij/issues/2846#issuecomment-1760310459, I tried unbinding `Ctrl j` by giving `bindkey -r '^J'` in zsh and using the following keybinds with `clear-defaults=true` at the top level:
```
    shared_except "locked" {
        bind "Ctrl g"          { SwitchToMode "Locked"; }
        bind "Ctrl q"          { Detach; }
        bind "Ctrl Q"          { Quit; }
        bind "Ctrl ;"          { NewPane "Down"; }
        bind "Ctrl '"          { NewPane "Right"; }
        bind "Ctrl j"          { MoveFocus "Down"; }
[...]

Issue #2954, opened 2023-11-20
ctrl+j is the same as pressing enter/return


APoniatowski:
**Basic information**

`zellij --version`: 0.39.1
`stty size`: full screen
`uname -av` or `ver`(Windows): Linux home-laptop 6.6.1-arch1-1

Hi there,

Yeah, it seems like there is an interesting bug (maybe on my side), where if I press ctrl+j,  it acts as enter/return in the active panel.

Also (maybe on my side), the clipboard does not transfer/paste over to my desktop (KDE), nor does it paste to itself, yet the highlighting disappears and it says "Text copied to system clipboard". I think there is an existing issue open for it, if not, let me know and I will create it

amiorin:
I have tested nvim, tmux, wezterm, and zellij.

I can confirm that tmux distinguish between ctrl+j and ctrl+m/return while zellij does not.
In nvim using ctrl+v I see the correct control codes.
```
ctrl+m ^M
ctrl+j ^@
```

Vinh-CHUC:
I can confirm I have the same issue, it works perfectly well in `tmux`

ssipos90:
It's been tracked at #2679

hiasr:
This is now fixed, the issue can be closed!

PR #3307, opened 2024-04-26
Fixed keybinding of Control J

hiasr:
Made it possible to bind Ctrl+j to a key, this would be very helpful for people using vim keybindings. 
Closes issue #2679 and #2954. 

Should be safe since it doesn't do anything as long as the key is not bound, this is the same behavior as tmux. I manually tested with Ctrl+j both bound and unbound in Kitty, Alacritty, Gnome Terminal and Konsole. 

One remark with this conditional parsing is that the keybindings are not available [here](https://github.com/zellij-org/zellij/blob/158260799b5112e31d86557af8a32b7ab0d77a26/zellij-server/src/tab/mod.rs#L1761C1-L1770C19). As a result, plugins still won't be able to use Ctrl+j, although I don't think that is so big of an issue.

### DIFF
--- a/zellij-client/src/input_handler.rs
+++ b/zellij-client/src/input_handler.rs
@@ -91,7 +91,11 @@ impl InputHandler {
                 Ok((InputInstruction::KeyEvent(input_event, raw_bytes), _error_context)) => {
                     match input_event {
                         InputEvent::Key(key_event) => {
-                            let key = cast_termwiz_key(key_event, &raw_bytes, Some((&self.config.keybinds, &self.mode)));
+                            let key = cast_termwiz_key(
+                                key_event,
+                                &raw_bytes,
+                                Some((&self.config.keybinds, &self.mode)),
+                            );
                             self.handle_key(&key, raw_bytes);
                         },
                         InputEvent::Mouse(mouse_event) => {
--- a/zellij-client/src/input_handler.rs
+++ b/zellij-client/src/input_handler.rs
@@ -91,7 +91,7 @@ impl InputHandler {
                 Ok((InputInstruction::KeyEvent(input_event, raw_bytes), _error_context)) => {
                     match input_event {
                         InputEvent::Key(key_event) => {
-                            let key = cast_termwiz_key(key_event, &raw_bytes);
+                            let key = cast_termwiz_key(key_event, &raw_bytes, Some((&self.config.keybinds, &self.mode)));
                             self.handle_key(&key, raw_bytes);
                         },
                         InputEvent::Mouse(mouse_event) => {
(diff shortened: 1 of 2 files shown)


imsnif:
Super cool! I want to test this in some more terminals to make sure none of them use Ctrl+j for linefeed.
@Zykino - if you find the time, would you be willing to test this out in Windows Terminal?
@tlinford - if you find the time, can you test this in iTerm2 and the default mac terminal?

To test this:
1. compile this fork
2. add this line to your keybinds config under "normal":
```kdl
        bind "Ctrl j" { NewPane; }
```
3. Start Zellij with this config and see if Ctrl j opens a new pane - make sure you can still press `ENTER` and it works as expected

Thanks!!

imsnif:
(I testes this in xterm and it works)

anopheles:
just tested on windows terminal and it works

`WSL-Version: 2.1.5.0                                                                                                                                        ││Kernelversion: 5.15.146.1-2                                                                                                                                 ││WSLg-Version: 1.0.60                                                                                                                                        ││MSRDC-Version: 1.2.5105                                                                                                                                     ││Direct3D-Version: 1.611.1-81528511                                                                                                                          ││DXCore-Version: 10.0.25131.1002-220531-1700.rs-onecore-base2-hyp                                                                                            ││Windows-Version: 10.0.19045.4355 `

imsnif:
Maybe also @jaeheonji can help us test this on the mac terminals? (instructions [here](https://github.com/zellij-org/zellij/pull/3307#issuecomment-2080163169))

jaeheonji:
@imsnif 

I completed the test in the following environment. and it works well :)

Machine: MacBook Pro
OS: macOS Sonoma 14.4.1
Terminal:
- Default macOS Terminal
- Alacritty 0.13.2
- iTerm2 3.4.23

imsnif:
Thanks to all the testers! Will merge now.

Thanks for this @hiasr !! I think the main point that was missing here was the research you did into how to overcome the CR(LF) collisions. This is a cool solution that not only seems not to bother modern terminals, but will also be backwards compatible to anything that relies on LF from STDIN (as long as they don't actively bind this key).

Viroide:
I would say that this bug is back. 

I'm using the last commit on main and `Ctrl+j `is not working. But in the last release it works.

LEI:
I expected locked mode to prevent this issue but I have similar symptoms (in normal mode too): `Ctrl j` seems to be correctly sent without any configuration change to programs running in alacritty but not in xfce4-terminal or wezterm.

`bind "Ctrl j" { NewPane; }` works everywhere but `unbind "Ctrl j"` seems to have no effect.
I could not figure out how to send `^j` with `WriteChars`, am I missing something?

```
alacritty 0.13.2 (bb8ea18e)
xfce4-terminal 1.1.3 (Xfce 4.18)
zellij 0.41.0
```

Zykino:
It sould be re-fixed here: #3746

PR #3308, opened 2024-04-26
Add basic ANSI theme

mikegreiling:
This adds a simple theme using only the base 16 ANSI colors.

Some terminal emulators (like the macOS Terminal.app) do not support truecolor, and one can get around this by applying custom values to the 16 ANSI colors through the terminal's "theme". This theme utilizes only those colors so that one's applied terminal theme will carry through to zellij.


imsnif:
Great addition, thanks!

PR #3313, opened 2024-04-30
theme: add onedark theme to the available themes

shonebinu:
Onedark is a popular theme from Atom text editor. 

I've used these sources as the reference for implementing the theme in zellij. https://github.com/joshdick/onedark.vim
https://www.figma.com/community/file/1137445418485757476/atom-one-dark-color-palette


imsnif:
Looks great, thanks!

Issue #2835, opened 2023-10-04
Feature: cli actions: query panes; go-to pane by name or id; query


fedoranvar:
Good day to you!

suggesting features for zellij action:

* query tabs: 
input: `zellij action query-tabs`
output:  `
{
'id': tabid,
'name':  'tabname',
'active': active_or_not
}
`
* query  panes: 
input: `zellij action query-panes`
output:  `
{
'id': paneid,
'name':  panename,
'active': active_or_not,
'tabid': tabid,
'command': currentcommand
}
`
* go-to pane: 
input: `zellij action go-to pane --name pane_in_currentab`
output:  focus named pane in current tab

usecase: make IDE by bonding zellij and Helix editor

azarmadr:
It would be great such api to query current pane and allow other apps/scripts to use that knowledge to perform actions like, on active pane do not do X

davxy:
Related: https://github.com/zellij-org/zellij/issues/2779 https://github.com/zellij-org/zellij/issues/2845

mrdgo:
Another use case: I integrate zjstatus with neovim. For that, the zjstatus module needs to identify the active pane to render the appropriate status line.

imsnif:
All of these can be done with plugins. This is the Zellij way.

I realize a lot of people are used to doing these things through the CLI - I get it and I totally empathize: this is not the Zellij way. We concentrate on plugins so that distinct pieces of logic will be cross-platform, require no installation, be aware of multiple users, be able to in the future render themselves in a text/ui agnostic way and be sandboxed to user permissions.

The CLI interface is a stop gap solution and I don't want to spend a lot of time expanding it. Please write plugins - I will do (and am doing) everything in my power to make this as easy as possible.

Thanks.

WizardUli:
> All of these can be done with plugins. This is the Zellij way.
> 
> I realize a lot of people are used to doing these things through the CLI - I get it and I totally empathize: this is not the Zellij way. We concentrate on plugins so that distinct pieces of logic will be cross-platform, require no installation, be aware of multiple users, be able to in the future render themselves in a text/ui agnostic way and be sandboxed to user permissions.
> 
> The CLI interface is a stop gap solution and I don't want to spend a lot of time expanding it. Please write plugins - I will do (and am doing) everything in my power to make this as easy as possible.
> 
> Thanks.

The last one absolutely cannot be done via plugins!
[...]

PR #3314, opened 2024-04-30
feat(cli): list clients, their focused pane_id and the running command

imsnif:
This adds the `zellij action list-clients` CLI command, which returns a list of the clients connected to the session, as well as their focused pane_id and (when available) the command running inside that pane..

Example output:
```
$ zellij action list-clients

CLIENT_ID ZELLIJ_PANE_ID RUNNING_COMMAND
1         plugin_2       zellij:session-manager
2         terminal_3     vim /tmp/my-file.txt
```

This is intended as a solution to those wanting to implemented integrated navigation with other multiplexers (eg. nvim).

### DIFF
--- a/zellij-server/src/plugins/mod.rs
+++ b/zellij-server/src/plugins/mod.rs
@@ -105,6 +105,7 @@ pub enum PluginInstruction {
         Option<PathBuf>,
     ),
     DumpLayout(SessionLayoutMetadata, ClientId),
+    ListClientsMetadata(SessionLayoutMetadata, ClientId),
     DumpLayoutToPlugin(SessionLayoutMetadata, PluginId),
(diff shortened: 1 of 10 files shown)


rockboynton:
Does this also close https://github.com/zellij-org/zellij/issues/2835 (and other linked issues in there)? You mentioned in a comment there that the functionality should be done with plugins, but with this cli addition it does seem possible unless I'm missing something?

imsnif:
If you can do what you want with the CLI commands - all power to you, of course! I meant it as more of a "I can't devote a lot of development time to the CLI interface now" sort of thing.

This PR was to get some people unstuck ASAP by offloading the "which client am I?" question to the various implementations.

PR #3315, opened 2024-05-01
Update soon-to-be incompatible dependencies

har7an:
We've had warnigns cropping up during builds for a while now, warning us about future incompatibilities in some of our dependencies. Earlier attempts at resolving those have failed, unfortunately, since these were transitive dependencies of packages that had no later releases at that time.

Meanwhile this has changed and this PR alleviates the issue by updating the packages `termwiz` and `daemonize`.


har7an:
Hey @imsnif,

here's a heads-up to let you know that I'm fiddling with the dependencies (again). This removes the "deprecation" warnings for our deps we've been receiving in builds for a while. This requires a little modification in `zellij_server::lib` for the MacOS builds (It's resolving a `umask` value as u16 on MacOS, where it's a u32 on Linux, apparently) and I can't test whether this is an actual problem due to a lack of MacOS PCs.

I'd suggest we merge this and in case any reports crop up by fellow developers on MacOS complaining that their builds stopped working, this might be the cause. Please let me know in case that happens, or if you object to any of this.

I'm afraid that this change will become necessary at least when we bump the rust toolchain version again, since the code will probably no longer work then. I would prefer to "debug" issues with these two processes separately, though.

imsnif:
Hey, thanks for doing this. I'd rather test these things before releasing them... Can we hold this off until after the next patch version (which should hopefully happen in the next few days)?

imsnif:
Right, version is out. So - maybe @tlinford or @jaeheonji can help with testing this on mac when they have the time?

Also - which dependencies did we end up updating? The diff is a little hard to read since I think things changed order a bit? Hope you can help me out.

har7an:
> Right, version is out.

Damn, "the next few days" went by pretty fast. :)

> which dependencies did we end up updating?

Right, I sorted the dependencies in the last commit here. We updated:

- `termwiz` from 0.20.0 to 0.22.0
- `daemonize` from 0.4.1 to 0.5.0

imsnif:
Hum... so - both of these dependency updates are breaking changes. That's okay, but I'd like for us to do some thorough manual testing in everything they touch.

For daemonize this is everything involving creating new sessions (namely creating a session, detaching from it and reattaching, attaching with multiple users, from multiple terminals, etc). For termwiz this is about STDIN parsing... I think making sure we can bind all common keys will do the trick.

Sorry, I know this is a lot - but we gotta be 100% here and these things are not trivial to test automatically.

imsnif:
Addendum: thinking about it, if all the keys in the default config work (plus some eccentricities such as the new `Ctrl j` and other things here: https://github.com/zellij-org/zellij/blob/main/zellij-utils/src/input/mod.rs#L77-L87 it should be fine)

har7an:
@imsnif No need to apologize, I much prefer we test these things before others do. But you definitely have a better overview about what these things touch and how we test that, so thanks for the plan. ;)

So let's try to formalize that:

## daemonize

- [ ] creating sessions works
- [ ] detaching and reattaching to sessions works
- [ ] attaching with multiple users works
- [ ] attaching from multiple terminals

## termwiz

- [ ] Bind all the keys from the default config
- [ ] Bind `Ctrl+j`
- [ ] Bind `Ctrl+h`

If you want something added, let me know. I'll start testing this weekend. Do you think different terminal emulators have an influence on the test results, or can I use an arbitrary terminal for that?

jaeheonji:
@imsnif  I'll also test all key combinations on my Mac device based on what @har7an has outlined until tomorrow. (Thanks for the summary.)

tlinford:
From a quick test on, it seems to run fine. Detached and reattached to a session, tried out the welcome layout, all ok as far as I can tell.

jaeheonji (approved):
I checked on my Mac that everything works fine for all keys. 🚀

har7an:
Hey, sorry for the long delay, I ran the tests mentioned above on Linux and they all completed fine.

One thing I did notice is that, apparently, Ctrl-based keybindings aren't really handled different from their "non-Ctrl" counterparts? As in: It appears that zellij can't really differ between the two (say: a and Ctrl+a). I'll have to investigate this some more, but that issues also appears on 0.40.0 and 0.40.1, so it's unrelated to this change. I'll try to write a good config to reproduce this and run a bisect when I find a PC big enough to compile zellij quicker than mine can. :P  

If nobody else has any objections to the changes here, I'll merge them on saturday.

Issue #2740, opened 2023-08-28
Feature: Improve session manager navigation


mo8it:
The session manager is awesome as I already said on Mastodon. Thank you very much for working on it ❤️

But the navigation could be improved though.

Currently, the search mode is activated by default and navigation is only possible using the arrow keys.
But because of my Vim/Helix nature, I don't personally like navigating using the arrow keys. Supporting navigation using **j,k,l,m** would be an awesome addition to the arrow keys (not replacement).

Should the search mode be activated by default? I would much rather prefer if you could start the search using something like `/`. The default should be plain navigation because I don't think that people would have many named sessions to navigate through with search. I could be wrong though.

I also can not stop expecting to be able to navigate down after the end of a nesting. For example, if I expand a session, then I can navigate its tabs with the up and down arrows, but I would expect the down arrow to bring me to the next session after the last tab. This should also work with going up.

imsnif:
Hey @mo8it - thanks for the great comments and I'm very happy you like the session manager!

One of the reasons behind implementing this as a plugin is so that people can roll their own behavior if they don't like the default one. Zellij is becoming large and hard to maintain, and while I want to provide sane defaults, allowing for particulars inside those defaults is a bit much for the current maintaining power we have. How about forking the plugin and making it behave as you will? Wdyt?

The rationale for the way things are btw is: most users are not `h/j/k/l` people. I'm one myself and this pains me, but that's the world we live in. The fuzzy finding is also for panes/tabs (using the titles provided by the applications running inside panes), and so is a very powerful way to navigate through the entire zellij-scape of the machine you're on. And so it is the default option.

imsnif:
Mate, I very much appreciate your offer of help - but I realistically do not see myself finding the time to review this. The main reason it was so important to me to devote time to the plugin system is so that people will be able to configure custom things like these for themselves - without me needing to review their code or maintain it later.

I think you don't need any special APIs to make this happen, right?

danielyrovas:
I think `Tab` and `S-Tab` are often used to select menu items in a context like this, and coming from vim (and nvim-cmp) I would also like `Ctrl n` and `Ctrl p` to cycle forward/back through the list.

PR #3321, opened 2024-05-03
feat(session-manager): allow Ctrl n and Ctrl p in attach to session to navigate items

anopheles:
addresses https://github.com/zellij-org/zellij/issues/2740

Unfortunately, ctrl+hjkl doesn't seem to work in this context.

### DIFF
--- a/default-plugins/session-manager/src/main.rs
+++ b/default-plugins/session-manager/src/main.rs
@@ -267,6 +267,12 @@ impl State {
         } else if let Key::Up = key {
             self.sessions.move_selection_up();
             should_render = true;
+        } else if let Key::Ctrl('n') = key {
+            self.sessions.move_selection_down();
+            should_render = true;
+        } else if let Key::Ctrl('p') = key {
+            self.sessions.move_selection_up();
+            should_render = true;
         } else if let Key::Char(character) = key {
             if character == '\n' {
                 self.handle_selection();


Zykino:
`Ctrl+p` is already bind to start the pane mode -> This is a conflicting default keybindings.

Also, since all you change is in the config, feel free to fork the session-manager. I’m not sure exactly how to do it, but it should be possible to give the keybindings you want to use as configuration at load.

For an example on how to get the config see:
https://github.com/zellij-org/zellij/blob/95ea75bfafada3666eba2b3ac3c9f74e777e756b/default-plugins/session-manager/src/main.rs#L54-L58

anopheles:
> `Ctrl+p` is already bind to start the pane mode -> This is a conflicting default keybindings.

Good observation. The arrow keys, which the session manager **currently** uses for navigation, is also already used in various contexts. I therefore don't see it as a problem. If a user uses the session-manager all keys are handled by the plugin, until it exits (as far as I can tell).

> Also, since all you change is in the config, feel free to fork the session-manager.

I hoped to merge this into the zellij main repository. As I expect the session-manager to evolve in the next couple of months, I am unable to invest capacity to keep my fork in sync.

> For an example on how to get the config see:

Thanks! I'll see what I can do.

Disclaimer: This is my first XP with rust :)

anopheles:
Does someone have any pointers to due something like this in the plugin section?
```
session-manager location="zellij:session-manager" {
        bind "Ctrl n" { MoveSelectionUp; }
        bind "Ctrl p" { MoveSelectionDown; }
    }
```

`configuration: BTreeMap<String, String>` only seems to store string values

towry:
@anopheles Does this pr solve this conflict issue? https://github.com/zellij-org/zellij/pull/3422

anopheles:
> @anopheles Does this pr solve this conflict issue? #3422

I don't think so. This PR just adds Ctrl + n/p navigation to the session navigator. I don't see how https://github.com/zellij-org/zellij/pull/3422 can solve this, but don't take my word for it.

PR #3349, opened 2024-05-14
Switch from Wasmer to Wasmtime

bjorn3:
Zellij is currently unable to update Wasmer as the currently used version is the last one without mandatory wasix support which Zellij doesn't want to adopt for various reasons. In the mean time a CVE has been found in the Cranelift version used by this Wasmer version. While it has been worked around in https://github.com/zellij-org/zellij/pull/2830, future CVE's may not be possible to workaround without updating Cranelift and thus Wasmer. This PR switches from Wasmer to Wasmtime to avoid all these problems.

Each individual commit builds. All commits except for the last two commits keep using Wasmer and thus could land without the actual switch to Wasmtime if preferred.


zellij-utils/src/logging.rs
```diff
@@ -68,15 +68,15 @@ pub fn configure_logger() {
...
     // For `zellij_server::logging_pipe`, we use custom format as we use logging macros to forward stderr output from plugins
     let config = Config::builder()
         .appender(Appender::builder().build("logFile", Box::new(log_file)))
         .appender(Appender::builder().build("logPlugin", Box::new(log_plugin)))
         .logger(
             Logger::builder()
-                .appender("logFile")
-                .build("wasmer_compiler_cranelift", LevelFilter::Warn),
+                .appender("logPlugin")
+                .build("wasmtime_wasi", LevelFilter::Warn),
```
bjorn3, line 79: Should this go to `logFile` or `logPlugin`. On the one hand these are logs in response to plugin actions, on the other hand they are not logs directly produced by the plugin itself.

bjorn3:
CI should be fixed now.

bjorn3:
(Rebased. Will likely keep doing this as soon as conflicts arise to avoid accumulating conflicts.)

syrusakbary:
Given that this PR can break many behaviors for plugins on Zellij, we can commit to update Cranelift in Wasmer very soon, so it doesn't become a problem for the project

imsnif:
> Given that this PR can break many behaviors for plugins on Zellij, we can commit to update Cranelift in Wasmer very soon, so it doesn't become a problem for the project

Much appreciated @syrusakbary !! But since I don't want you to perform extra work, I will emphasize something I mentioned elsewhere: we will not consider any upgrade of wasmer that includes wasix or any sort of extra dependency that comes with wasix.

Given that I understood this is a big ask for your project (totally legit ofc, every project and application has different requirements), I gave @bjorn3 the go-ahead for this work. If all checks out with it (I haven't gone over it yet) and we manage to iron out the differences with plugins you rightfully mentioned (some of them have already been brought up as points of order by @bjorn3 ) - we will go through with this change.

imsnif:
Thanks @syrusakbary ! As mentioned here and elsewhere, a feature flag will probably not suffice. We also need things like dependencies that are only used in wasix not to be included at all (for our packagers). This would likely have to be an entirely different crate (though I of course do not know about your internals).

But as I said - the work has already started here, so I intend to see it through. If we hit a road block that prevents us from migrating, or feel the migration is too risky - I'll be sure to check out the work in the linked issue to see if it would be a better path forward for us.

Thanks for your efforts and I am sorry things have to be this way, but we've sounded this alarm before a few times and we have to think of our application first.

syrusakbary:
@imsnif of course, no worries! Do what you think is best for the project.

Just a minor clarification: on the Wasmer side, we can turn off the WASIX syscalls, dependencies and code paths incredibly easily via a feature flag, so a new create shall not be needed if you would like to stick with Wasmer :)

src/tests/e2e/remote_runner.rs
```diff
@@ -84,7 +84,7 @@ fn start_zellij(channel: &mut ssh2::Channel) {
         )
         .unwrap();
     channel.flush().unwrap();
-    std::thread::sleep(std::time::Duration::from_secs(1)); // wait until Zellij stops parsing startup ANSI codes from the terminal STDIN
+    std::thread::sleep(std::time::Duration::from_secs(3)); // wait until Zellij stops parsing startup ANSI codes from the terminal STDIN
```
imsnif, line 87: Any idea why we need the increase here? I don't mind increasing it in in this case of the e2e tests if we need it, but I'd prefer to understand what happened.

zellij-server/Cargo.toml
```diff
@@ -14,15 +14,17 @@ ansi_term = "0.12.1"
...
 unicode-width = "0.1.8"
 url = "2.2.2"
-wasmer = "3.1.1"
-wasmer-wasi = "3.1.1"
+wasmtime = "20.0.2"
+wasmtime-wasi = "20.0.2"
 cassowary = "0.3.0"
 zellij-utils = { path = "../zellij-utils/", version = "0.41.0" }
 log = "0.4.17"
+tokio = "1.37.0"
```
imsnif, line 27: Why tokio? We're already using async_std internally, I'm not thrilled about adding another async runtime...

imsnif (changes requested):
Hey @bjorn3 - first, thanks so much for your thorough work on this. This is a major infrastructure change for us and I do not want to take it lightly - so I hope you don't mind being patient with my review cycle.

I'm going on a short vacation tomorrow and didn't want to leave this hanging without a response while I'm away. So I added some comments about things that immediately jumped out at me to get us started. I will give this a more thorough shake-up and review when I get back around the beginning of June.

> In Wasmer trying to create a file with a relative path creates it in the /host directory. In Wasmtime it will result in an error that there is no "pre-opened file descriptor" through which the file can be opened. As WASI doesn't have any such concept as a current working directory, the Wasmtime behavior is more natural. It does break the current released version of zjstatus however. https://github.com/dj95/zjstatus/commit/25e3a56f8be3732661a622352590886957c5a59a works around this. I tried to work around the difference, but while it fixed this specific issue, it broke a lot more things. As such I would prefer documenting the difference rather than making Wasmtime behave the same way as Wasmer.

I've been thinking about this specific issue and while I hear you about the difficulty with the behavior change, I'm not enthusiastic about it. `zjstatus` is just one example that we happened to catch, I'm sure there are more. Backwards compatibility of already-compiled plugin is a strong value for our ecosystem.

I'm not saying this is a deal-breaker, but I'd like to understand the trade-offs before going ahead with this. What other stuff breaks if we try to work around this problem? How big of a hack is it? Will it affect upgrades? What else?

zellij-server/Cargo.toml
bjorn3: wasmtime-wasi has a dependency on tokio. In any case this direct dependency here can be removed I think. I used it in an earlier version of the stdio pipe implementation, but didn't remove it when I changed the implementation.

src/tests/e2e/remote_runner.rs
bjorn3: Without it, the first snapshot for several tests seem to be taken while still compiling wasm, which suggests that compiling is slower. This is surprising to me as locally it actually seemed to be faster. Maybe the singlepass compiler of Wasmtime is slower than the one of Wasmer and only the optimizing compiler is faster?

bjorn3:
> I've been thinking about this specific issue and while I hear you about the difficulty with the behavior change, I'm not enthusiastic about it. zjstatus is just one example that we happened to catch, I'm sure there are more. Backwards compatibility of already-compiled plugins is a strong value for our ecosystem.

I just reproduced this behavior in standalone Wasmer. Turns out it is completely disregarding the intention of the WASI specification:

```rust
fn main() {
    println!("{:?}", std::fs::write("foo", ""));
    println!("{:?}", std::fs::read("foo"));
    println!("{:?}", std::fs::read_dir("/").map(|dir| dir.collect::<Vec<_>>()));
    println!("{:?}", std::fs::read_dir(".").map(|dir| dir.collect::<Vec<_>>()));
    println!("{:?}", std::fs::read_dir("/tmp/foo/x").map(|dir| dir.collect::<Vec<_>>()));
}
```

when run in wasmtime with a `/tmp/foo/x` directory (containing `my_file`) passed through to the guest will print:

```
Err(Custom { kind: Uncategorized, error: "failed to find a pre-opened file descriptor through which \"foo\" could be opened" })
Err(Custom { kind: Uncategorized, error: "failed to find a pre-opened file descriptor through which \"foo\" could be opened" })
Err(Custom { kind: Uncategorized, error: "failed to find a pre-opened file descriptor through which \"/\" could be opened" })
Err(Custom { kind: Uncategorized, error: "failed to find a pre-opened file descriptor through which \".\" could be opened" })
Ok([Ok(DirEntry("/tmp/foo/x/my_file"))])
```

as I would have expected while wasmer prints

```
Ok(())
Ok([])
Ok([Ok(DirEntry("/.app")), Ok(DirEntry("/.private")), Ok(DirEntry("/bin")), Ok(DirEntry("/dev")), Ok(DirEntry("/etc")), Ok(DirEntry("/foo")), Ok(DirEntry("/tmp"))])
Ok([Ok(DirEntry("./.app")), Ok(DirEntry("./.private")), Ok(DirEntry("./bin")), Ok(DirEntry("./dev")), Ok(DirEntry("./etc")), Ok(DirEntry("./foo")), Ok(DirEntry("./tmp"))])
Ok([Ok(DirEntry("/tmp/foo/x/my_file"))])
```

In other words rather than passing through the directories the user specified as pre-opened directories as intended, it creates an in-memory virtual filesystem whose contents are discarded upon exit and inside this VFS it mounts the user specified directories. The pre-opened directories it passes are `/` (fd 4), `/` (fd 5) and `.` (fd 6). Yes, it passes `/` twice for whatever reason. I can imagine that some guest languages will get upset about that. Wasmtime directly passes `/tmp/foo/x` as a pre-opened directory instead.

```
$ RUST_LOG=wasmer_wasix::syscalls::wasi::fd_prestat_dir_name=trace wasmer run target/wasm32-wasi/debug/foo.wasm
2024-05-27T17:00:01.380160Z TRACE ThreadId(01) fd_prestat_dir_name: wasmer_wasix::syscalls::wasi::fd_prestat_dir_name: return=Errno::success fd=3 path="/"
2024-05-27T17:00:01.380179Z TRACE ThreadId(01) fd_prestat_dir_name: wasmer_wasix::syscalls::wasi::fd_prestat_dir_name: close time.busy=31.2µs time.idle=8.81µs fd=3 path="/"
2024-05-27T17:00:01.380201Z TRACE ThreadId(01) fd_prestat_dir_name: wasmer_wasix::syscalls::wasi::fd_prestat_dir_name: return=Errno::success fd=4 path="/"
2024-05-27T17:00:01.380205Z TRACE ThreadId(01) fd_prestat_dir_name: wasmer_wasix::syscalls::wasi::fd_prestat_dir_name: close time.busy=4.83µs time.idle=448ns fd=4 path="/"
2024-05-27T17:00:01.380213Z TRACE ThreadId(01) fd_prestat_dir_name: wasmer_wasix::syscalls::wasi::fd_prestat_dir_name: return=Errno::success fd=5 path="."
2024-05-27T17:00:01.380217Z TRACE ThreadId(01) fd_prestat_dir_name: wasmer_wasix::syscalls::wasi::fd_prestat_dir_name: close time.busy=4.09µs time.idle=326ns fd=5 path="."
[...]
```

It would probably be possible to emulate this with wasmtime-wasi, but I think it would only hide bugs inside plugins. From what I can tell before https://github.com/dj95/zjstatus/commit/25e3a56f8be3732661a622352590886957c5a59a the `.zjstatus.log` file would have been discarded rather than saved on the host as was intended. If you do still want me to implement support for emulating the Wasmer behavior, I will do so. Maybe it could be a future compatibility warning when it seems like the plugin depends on the Wasmer behavior.

src/tests/e2e/remote_runner.rs
syrusakbary: Note: Winch is about 2-6 times slower compiling than Wasmer's Singlepass for non-trivial applications (>2Mb), you can check more examples such as time to compile Spidermonkey or ffmpeg on each runtime on this article:

https://wasmi-labs.github.io/blog/posts/wasmi-v0.32/

bjorn3:
I can't reproduce this behavior locally. For me the counts are preserved as I switch between tabs.

[Peek 2024-06-07 17-36.webm](https://github.com/zellij-org/zellij/assets/17426603/02c25ce8-0bb6-4a4b-888d-eba634e3d85c)

imsnif:
Also when the second plugin instance is initially launched? That's when it gets erased for me.

bjorn3:
Yes, opening a new tab preserves the state too:

[Peek 2024-06-07 17-36.webm](https://github.com/zellij-org/zellij/assets/17426603/5bc1ca99-90d9-4d13-848d-db87c8535065)

imsnif:
Alright, I'll do some troubleshooting next week and find out what's up.

imsnif:
Hey, so first - I did some testing and this was a configuration error on my part. My apologies!

Otherwise, I'm still testing this PR. I'm sorry for not getting to do a full review this week, it was on my schedule and got pushed off. I fully plan to do so next week. Thank you for your patience @bjorn3 !

imsnif (commented):
Hey @bjorn3 - I've given this a thorough shakedown and gone over the code and all in all this looks great.

Going over the code, I looked for areas that can be potentially problematic, are not explicitly covered by the e2e tests and might have been accidentally broken by someone who is not intimately familiar with the application. All of them work as expected, but I'd like to list what I've done here for the record:

## Manually tested (all pass)
* Rendering / logging from within a plugin (both STDOUT and STDERR work as expected)
* Plugin pipes
    - rendering/not rendering after a pipe
    - blocking/not blocking CLI pipe with/without render
    - multiple simultaneous concurrent CLI pipes
    - piping data between plugins
* Filesystem locations bound to plugins behave as they should
    - /host (CWD of last focused pane)
    - /data (specific cache folder, shared between plugin instances, deleted on plugin exit)
    - /tmp (temporary folder)

## Regarding the CWD question
Your tests have convinced me that reimplementing this behavior is likely more trouble than it's worth - especially in the long run. Only thing is: is it possible to make the error clearer? Instead of `failed to find a pre-opened file descriptor through which \"foo\" could be opened`, could we make it `Zellij Plugins can't use relative paths, please use an absolute path. For more info, see: https://zellij.dev/documentation/plugin-api-file-system` (bonus points if it can also include a suggestion based on the user's supplied relative path: `Zellij Plugins can't use relative paths, please use an absolute path, for example /host/<relative-path>. For more info, see: https://zellij.dev/documentation/plugin-api-file-system`.)

## Final thoughts
Otherwise, I've asked our Go SDK author to do a quick smoke test of this just to be 100% certain there isn't any implicit behavior the Go SDK is relying on that we're overriding here.

One last thing is the dependency count. This seems to add ~50 packages to Zellij. I don't know if this is possible, but I do want to ask: is it an option to somehow weed down the transient dependencies here? Not a deal breaker, but would certainly be nice.

All in all, this looks great and once we sort out these last few things I think we can merge this. Fantastic work @bjorn3 !! Thank you.

bjorn3:
> Only thing is: is it possible to make the error clearer? Instead of failed to find a pre-opened file descriptor through which \"foo\" could be opened, could we make it Zellij Plugins can't use relative paths, please use an absolute path. For more info, see: https://zellij.dev/documentation/plugin-api-file-system (bonus points if it can also include a suggestion based on the user's supplied relative path: Zellij Plugins can't use relative paths, please use an absolute path, for example /host/<relative-path>. For more info, see: https://zellij.dev/documentation/plugin-api-file-system.)

I don't think that is possible unfortunately. This error is produced by the copy of the rust standard library linked into the plugin itself (https://github.com/rust-lang/rust/blob/1aaab8b9f8dc488cadc4f083b3a11fb11b45cb77/library/std/src/sys/pal/wasi/fs.rs#L722-L726), not by wasmtime-wasi.

> One last thing is the dependency count. This seems to add ~50 packages to Zellij. I don't know if this is possible, but I do want to ask: is it an option to somehow weed down the transient dependencies here? Not a deal breaker, but would certainly be nice.

I can disable profiling support to cut the dependency count down a bit. It needs manual configuration to enable at runtime anyway, which this PR doesn't do. Other than that I think most new dependencies are coming from tokio, which wasmtime-wasi has a hard dependency on.

imsnif:
> I don't think that is possible unfortunately. This error is produced by the copy of the rust standard library linked into the plugin itself (https://github.com/rust-lang/rust/blob/1aaab8b9f8dc488cadc4f083b3a11fb11b45cb77/library/std/src/sys/pal/wasi/fs.rs#L722-L726), not by wasmtime-wasi.

Too bad. Will document this for the next release then.
  
> I can disable profiling support to cut the dependency count down a bit. It needs manual configuration to enable at runtime anyway, which this PR doesn't do. Other than that I think most new dependencies are coming from tokio, which wasmtime-wasi has a hard dependency on.

Alright, let's do that.

bjorn3:
> Alright, let's do that.

Done and disabled a couple more features too.

By the way since I opened this PR, Wasmtime 21.0 has been released. Do you want me to update to it before or after merging this PR?

imsnif:
> By the way since I opened this PR, Wasmtime 21.0 has been released. Do you want me to update to it before or after merging this PR?

Let's do it before if we're already here. Do you think there are significant changes that will affect us?

bjorn3:
Updated. No api changes that affects us. Only had to ensure the new std feature of wasmtime is enabled.

bjorn3:
> Done and disabled a couple more features too.

By the way, I forgot to mention that I disabled the "pooling allocator" too. It is a performance optimization for when you quickly drop and recreate instances of a wasm module. For example creating an instance for each request in an http server. Are plugin instances meant to be quickly destroyed again after invoking a command inside them or do they live for a while in Zellij?

imsnif:
> By the way, I forgot to mention that I disabled the "pooling allocator" too. It is a performance optimization for when you quickly drop and recreate instances of a wasm module. For example creating an instance for each request in an http server. Are plugin instances meant to be quickly destroyed again after invoking a command inside them or do they live for a while in Zellij?

They definitely live for a while. There can be cases where a plugin is spun up to deal with a request (eg. through a CLI pipe) and then kills itself, but I don't think there's a use-case for it then spinning itself up again multiple times.

imsnif:
So where things are now: I'm going to give stacab a few days to see if he can test this with some Go plugins, and if all is well or he doesn't have time, I'll give it one more shakeup with the new version and merge if all is well.

imsnif:
Alright. Things look good, multiple people have also tested this manually in various scenarios, so I'm giving this the go.

I would like to emphasize that this is not a criticism of wasmer in any way or form. Wasmer has been a great runtime for us over the years and has served us well. A heartfelt thank you to the maintainers.

Thank you very much for your work on this @bjorn3 !

imsnif:
@bjorn3 - I hope it's alright to consult you about this after we've already merged.

Working with `main`, I have been getting some of these hard crashes occasionally:
```
Error occurred in server:

  × Thread 'async-std/runtime' panicked.
  ├─▶ At /home/aram/.cargo/registry/src/index.crates.io-6f17d22bba15001f/wasmtime-21.0.1/src/runtime/type_registry.rs:740:48
  ╰─▶ called `Option::unwrap()` on a `None` value

        -> To get additional information, check the log at: /tmp/zellij-1000/zellij-log/zellij.log
        -> To see a backtrace next time, reproduce the error with: RUST_BACKTRACE=1 zellij [...]
        -> To help us fix this, please open an issue: https://github.com/zellij-org/zellij/issues
```

I'm using `--singlepass` (winch?) compiler because I'm developing - so could this just be a development thing? I'm not too sure how to reproduce these unfortunately.

Any ideas?

bjorn3:
> I hope it's alright to consult you about this after we've already merged.

Of course.

> Working with main, I have been getting some of these hard crashes occasionally:

I will look into it. Do you have a backtrace for the crash?

imsnif:
Thank you! I unfortunately do not have a backtrace, but I'll try to keep an eye on this and produce one if it happens again.

bjorn3:
Looks like the assertion in question is a debug assertion, so it would only be hit during development: https://github.com/bytecodealliance/wasmtime/blob/cedf9aa0f9620391cab465361b474b149654e88d/crates/wasmtime/src/runtime/type_registry.rs#L740 I will ask someone familiar with this code if they know what is wrong.

fitzgen:
Hi @imsnif,

Wasmtime maintainer here. @bjorn3 pinged me about this bug. Do you have a test case and steps I can follow to reproduce the assertion failure?

Thanks!

imsnif:
Hi @fitzgen - thanks for the very quick response! I unfortunately do not have a test case or steps to reproduce. Application-wise, this seems to happen pretty randomly (happened to me when I opened a new tab in Zellij which in this context causes new plugins to be instantiated).

It happens to me occasionally while developing, I'll try to keep an eye out for the trace next time.

Since this only happens in development though - on my side at least it's is only an extremely minor inconvenience and nothing to worry about.

imsnif:
Hey @bjorn3 and @fitzgen - this now happened to me in release code:

```
  × Thread 'async-std/runtime' panicked.
  ├─▶ At /home/aram/.cargo/registry/src/index.crates.io-6f17d22bba15001f/wasmtime-slab-21.0.1/src/lib.rs:434:35
  ╰─▶ attempt to deallocate an entry that is already vacant

        -> To get additional information, check the log at: /tmp/zellij-1000/zellij-log/zellij.log
        -> To see a backtrace next time, reproduce the error with: RUST_BACKTRACE=1 zellij [...]
        -> To help us fix this, please open an issue: https://github.com/zellij-org/zellij/issues
```

I don't customarily run with `RUST_BACKTRACE` and due to the nature of this application it's not trivial to reproduce this. I'm honestly starting to get a little worried... any thoughts?

fitzgen:
Unfortunately, it is hard to debug this issue with so little information.

Can you confirm whether you are using Wasm GC types (and specifically the `wasmtime::ManuallyRooted<T>` API type) and/or Wasm `externref`s?

I will do another review of some potentially relevant code paths.

fitzgen:
I think I've figured out what this is actually. Will link the PR when I have it up.

imsnif:
Much appreciated @fitzgen ! I can totally understand this is not trivial to troubleshoot with so little information. I'm working with the Zellij main branch as a matter of course and have now made sure to also have `RUST_BACKTRACE=1` up and will provide more information here as I see it.

As for specific questions, I did not do the integration myself - so I think @bjorn3 can answer them best. But hopefully as you say you figured it out and this will not be needed.

bjorn3:
No wasm GC types are used in the external interface as it is using wasip1 just like the Wasmer code. And as most (all) plugins are written in Rust, there is also no GC types usage expected internally within the plugins.

fitzgen:
Thanks for providing additional info.

Fix coming soon, I've managed to reproduce locally and verify the fix.

Aside: it might be neat to install a custom panic handler to Zellij that captures the stack trace and panic message and logs them to a file or something like that. Could make debugging this kind of bug (whether in Wasmtime, another dep, or Zellij itself) easier in the future.

Anyways, thanks for your patience while we figured this out!

imsnif:
> Aside: it might be neat to install a custom panic handler to Zellij that captures the stack trace and panic message and logs them to a file or something like that. Could make debugging this kind of bug (whether in Wasmtime, another dep, or Zellij itself) easier in the future.

I totally agree. We actually do have a panic hook that is tied to some other internal app logic. I very much agree we should get it to do the right thing in this case. Like everything, it's a matter of developer capacity which is something we're unfortunately in short supply of.

Thanks for the quick turnaround on this!

bjorn3:
@fitzgen which PR fixed this bug?

alexcrichton:
The bug was https://github.com/bytecodealliance/wasmtime/security/advisories/GHSA-7qmx-3fpx-r45m so it ended up not having an official PR for it but rather a [direct merge](https://github.com/bytecodealliance/wasmtime/commit/0ebe54d05f0e1f6c64b7c8bb48c9e9f6c95cacba)

bjorn3:
Opened https://github.com/zellij-org/zellij/pull/3685

Issue #3325, opened 2024-05-04
Modify watch_filesystem to take in a list of folders to watch


leakec:
It would be nice to modify `watch_filesystem` to take in a list of folders to watch. The reasons for doing this are:
1. Avoid doing unnecessary work watching files we don't care about.
2. Currently, this watches the `/host` folder and recurses through all child folders. As a result, if the user doesn't have access to a file in the `/host` folder or child folders, this results in an error. Restricting the folders watched to the minimum required should limit this happening.

To make this change, `watch_filesystem` could be modified from this function signature
```rust
pub fn watch_filesystem()
```
to 
```rust
pub fn watch_filesystem(files: &HashMap<PathBuf, bool>)
```

If `watch_filesystem` is called with an empty `HashMap`, we can fill it with `/host` and `true` to get the previous behavior.

The downside of this is the function signature of `watch_filesystem` will change, which means any plugins using it will need to update this call. This doesn't seem to be used much in the zellij core. There was only one instance of it in the core, and it was for a test. `multitask` is another plugin I know that uses this, and I will update said plugin if this change is made.

I've got this implemented and tested on a local fork. Would be happy to submit a PR if this is accepted (or modify and submit a PR based on any discussion here).

leakec:
The PR I have ready for this is located [here](https://github.com/leakec/zellij/tree/feature/files_watch) (should have added this when I first opened the issue).

leakec:
Closing this as the PR has been closed. There is a better way to fix this, which is with new plugin APIs available in 0.41! :)

PR #3377, opened 2024-05-25
Modify watch_filesystem to take in a list of folders to watch

leakec:
This pull request addresses issue #3325.


imsnif:
Hey @leakec - I first wanted to apologize for taking so embarrassingly long to get to this.

Otherwise - I'm a little concerned about the current watch_filesystem as I find it quite buggy on certain platforms and I'm not sure I want to keep giving plugins so much power over the application.

I guess you want this for `multitask`... Do you think it can be solved in a different way? Perhaps a UI that opens the user's `$EDITOR` and once it's closed reads the contents? There are new APIs now in the upcoming version that should accommodate this. I can point them out after the release if you like.

leakec:
@imsnif No need to apologize at all. I know you are super busy maintaining this awesome tool we all love and enjoy! 

You're correct in that this PR was targeting some improvements I was hoping to make on `multitask`. I agree that this could probably be solved (and maybe better to solve) by not using `watch_filesystem` at all. 

One option would be to just have `zellij` read from a user-specified file (or create one on `/tmp` if none are specified) and have the user use a keybind to signal to the plugin that they are ready to run their multitasks; rather than the current signaling method which is once the file is saved/closed. One reason I like a file rather than a temporary editor pane is it allows users to come back and re-use that file across multiple sessions, multiple days, etc. I find myself often creating a multitask file and then re-using it for weeks or months, so it would be nice to keep this option.

Sounds good r.e. the APIs in the upcoming version. I'll read the updates when it comes out and let you know if I have any questions. I'm happy to close this PR rather than merge if you think that's best.

leakec:
@imsnif This sounds great! I think it might be nice to add a second way in addition to closing the editor pane (maybe an optional, opt-in keybind or something) so users can run the multitask with the editor window still open. This is really nice if the tasks run fast and you want to run them, make a change, run them again, etc. 

Regardless, it sounds like all of this will fit really nicely within the new API :) 

I like this plan a lot better than the PR here, since then we don't have to worry about permission for anything except the multitask file and running the commands, so we don't have to even worry about watching any part of the file sytem. Therefore, I'm going to close the PR. Thanks for all the suggestions and work on the upcoming, new API options!

PR #3427, opened 2024-06-17
feat(keybindings): Match TMUX numbered tab switch behavior (ctrl + b + [tab number])

jknight:
ctrl + b + [1-9] switches to the tab number. 
This matches TMUX's default behavior.


mfbehrens:
@imsnif Please merge this. This makes the default tmux mode almost unusable for a lot of people.

PR #3440, opened 2024-06-21
fix: use ITU T.418 format for emitted CSI 58 sequences

tranzystorekk:
Addresses https://github.com/microsoft/terminal/issues/17426 on our side.

Tested both on nightly wezterm on Windows WSL2, and on a native linux wezterm for a cursory regression check.

### DIFF
--- a/zellij-server/src/panes/terminal_character.rs
+++ b/zellij-server/src/panes/terminal_character.rs
@@ -619,10 +619,10 @@ impl Display for CharacterStyles {
             if let Some(ansi_code) = self.underline_color {
                 match ansi_code {
                     AnsiCode::RgbCode((r, g, b)) => {
-                        write!(f, "\u{1b}[58;2;{};{};{}m", r, g, b)?;
+                        write!(f, "\u{1b}[58:2::{}:{}:{}m", r, g, b)?;
                     },
                     AnsiCode::ColorIndex(color_index) => {
-                        write!(f, "\u{1b}[58;5;{}m", color_index)?;
+                        write!(f, "\u{1b}[58:5:{}m", color_index)?;
                     },
                     AnsiCode::Reset => {
                         write!(f, "\u{1b}[59m")?;


zellij-server/src/panes/terminal_character.rs
```diff
@@ -619,10 +619,10 @@ impl Display for CharacterStyles {
             if let Some(ansi_code) = self.underline_color {
                 match ansi_code {
                     AnsiCode::RgbCode((r, g, b)) => {
-                        write!(f, "\u{1b}[58;2;{};{};{}m", r, g, b)?;
+                        write!(f, "\u{1b}[58:2::{}:{}:{}m", r, g, b)?;
```
imsnif, line 622: While I tested this and it works this way, I wonder - why the two colons? For me it works both ways.

tranzystorekk: According to https://en.wikipedia.org/wiki/ANSI_escape_code#24-bit the format is:

```
ESC[38:2:⟨Color-Space-ID⟩:⟨r⟩:⟨g⟩:⟨b⟩:⟨unused⟩:⟨[CS tolerance](https://en.wikipedia.org/wiki/Color_difference)⟩:⟨Color-Space associated with tolerance: 0 for "CIELUV"; 1 for "CIELAB"⟩ m Select RGB foreground color
ESC[48:2:⟨Color-Space-ID⟩:⟨r⟩:⟨g⟩:⟨b⟩:⟨unused⟩:⟨CS tolerance⟩:⟨Color-Space associated with tolerance: 0 for "CIELUV"; 1 for "CIELAB"⟩ m Select RGB background color
```

where the two colons is the empty `Color-Space-ID` parameter i.e. "unspecified"

imsnif: Gotcha, thanks

imsnif:
Thanks for this! I'm happy to see this issue resolved for the joint users of Zellij and windows terminal.

Just for the record, as specified in the linked issue: I see this as a workaround for an issue in windows terminal. I mean this in the most blameless way possible, but in case we find that this workaround causes issues in other terminals in a way that cannot be consolidated, we'll have to undo it and hope that the windows terminal folks fix the issue on their side in one way or another.

PR #3450, opened 2024-06-25
fix(themes): fix text selection color for nightfox

avinal:
## Changes

- bg is used as text background color for selection instead of
  the background color of the zellij itself
- this non-conventional naming confuses theme contributors and the user
  sees no highlight on selection
- this commit updates the selection color for nightfox theme and adds
  missing variations of the theme
- refer https://github.com/EdenEast/nightfox.nvim/pull/444

Signed-off-by: Avinal Kumar <avinal.xlvii@gmail.com>


giyany:
Changes to the theme definition are expected at some point, #3242 .

PR #3460, opened 2024-06-28
fix: copy and paste not working due to kitty mode

### LINKED ISSUE #3421: Cannot paste into Helix or Neovim or Kakoune when using git version of Zellij
This has happened to me on MacOS 14.4 running kitty, iterm2, and terminal.

I have tried this with homebrew installed helix and my own local compiled version. No difference.

883855587c4c963f3037f287e5239e8cd297ab69 is the top most commit in the log.

Steps to reproduce:
1. build zellij from git checkout -- cargo xtask dist is what I used.
2. run this zellij
3. start helix (you can run `brew install helix` if you don't have it)
4. enter insert mode with 'i', just like vim.
5. Press the usual Cmd+v to paste as normal on a Mac. In iterm2 I can see the Edit menu flash so it recognized the keys...

If I exit helix I can paste into the shell inside of zellij. If I start neovim installed from Homebrew I also cannot paste. Same thing, enter insert mode and Cmd+v does nothing. I just installed Kakoune from homebrew and again, cannot paste. However, if I use /usr/bin/vi I can paste. No problems. /usr/bin/vi is the older vim shipped with MacOS.

This all works just fine with the homebrew installed version of Zellij.

towry:
Fix #3421

Some plugin are using the `write` and `write_chars` methods but it doesn't take care of the kitty keyboard protocol, see https://github.com/zellij-org/zellij/blob/fa110515aa672ff47b61fd69090da46a3c28d103/zellij-server/src/plugins/zellij_exports.rs#L1065

### DIFF
--- a/zellij-client/src/input_handler.rs
+++ b/zellij-client/src/input_handler.rs
@@ -45,6 +45,7 @@ struct InputHandler {
     receive_input_instructions: Receiver<(InputInstruction, ErrorContext)>,
     holding_mouse: Option<HeldMouseButton>,
     mouse_mode_active: bool,
+    is_kitty_keyboard_protocol: bool,
 }
 
 impl InputHandler {
@@ -57,6 +58,7 @@ impl InputHandler {
         send_client_instructions: SenderWithContext<ClientInstruction>,
         mode: InputMode,
         receive_input_instructions: Receiver<(InputInstruction, ErrorContext)>,
+        is_kitty_keyboard_protocol: bool,
     ) -> Self {
         InputHandler {
             mode,
@@ -69,6 +71,7 @@ impl InputHandler {
             receive_input_instructions,
             holding_mouse: None,
             mouse_mode_active: false,
+            is_kitty_keyboard_protocol,
         }
     }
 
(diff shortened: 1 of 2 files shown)


imsnif:
Can you help me understand what you are trying to fix? This doesn't seem related to #3421 - but maybe I'm missing something?

towry:
I am trying to fix the paste not working in neovim and helix issue, the above changes works (build and tested), the changes is related to #3421.

The link in the above related to another issue with plugin, for example, in plugin `write_chars` not working any more if kitty keyboard protocol is enabled, unless I change `false` to `true` in https://github.com/zellij-org/zellij/blob/fa110515aa672ff47b61fd69090da46a3c28d103/zellij-server/src/plugins/zellij_exports.rs#L1065 , but as i said this is not related to #3421.

imsnif:
I appreciate this PR and thank you for it. I hear you that this fixes the issue (and also the issue with the plugin), but I think there might be a different underlying problem (eg. with bracketed paste or the way we serialize keys to the terminal pane when it's operating in a different keyboard mode than the controlling terminal).

I'd rather look more deeply into it and solve the root problem, because what's happening here might just be solving a symptom of it. I fully plan on doing this before the release and will remind everyone that it's better not to use git versions, but rather to use the latest release version.

Thanks for putting this together though.

PR #3472, opened 2024-07-04
fix(terminal): only refresh process specifics for given PID in get_cwd()

Hylian:
Update `get_cwd()` to only refresh process specifics for the given PID, matching the `get_cwds()` implementation.

This significantly reduces the latency of opening a new Pane or Tab.

On a 64-core Threadripper system, this reduces the latency of `get_cwd()` from 850ms to 1ms. On an 8-core Ryzen system, the improvement is smaller, but still perceptible.


imsnif:
Thank you!

PR #3480, opened 2024-07-07
cargo: Remove `rust-version` (msrv) field

### LINKED ISSUE #3476: Incorrect MSRV
Cargo.toml specifies `rust-version = 1.60`, which is wrong:
- rust-toolchain.toml has the toolchain pinned to 1.75
- dependencies like wasmtime has an MSRV of 1.75

There are a few ways to fix this discrepancy:
- remove `rust-version` from Cargo.toml
- remove `channel` from rust-toolchain.toml and set `rust-version = 1.75` in Cargo.toml
- update `rust-version` to 1.75, and cross our fingers that it will stay in sync with `channel` in the future  

By looking at the commit logs I couldn't figure out the reason for pinning it to one specific version in rust-toolchain.toml, so I recommend bumping `rust-version` in Cargo.toml to 1.75 and removing `channel` from rust-toolchain.toml.

@har7an

har7an:
from `Cargo.toml`. This has gone out of sync with the rust version specified in `rust-toolchain.toml` and appears to have little consequence to development, since it defines only the MSRV. We're using `rust-toolchain.toml` to strictly enforce a specific rust version for development, hence we stick with that.

Closes #3476

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 * fix(terminal): workaround for windows terminal not interpreting underline colors with semicolons (https://github.com/zellij-org/zellij/pull/3440)
 * dependencies: switch from wasmer to wasmtime (https://github.com/zellij-org/zellij/pull/3349)
 * feat(ui): status-bar redesign (https://github.com/zellij-org/zellij/pull/3475)
+* chore: Remove MSRV from Cargo.toml (https://github.com/zellij-org/zellij/pull/3480)
 
 ## [0.40.1] - 2024-05-02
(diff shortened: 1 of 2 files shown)


imsnif:
Just to make sure: will this version requirement still be picked up when installing from crates.io?

har7an:
> Just to make sure: will this version requirement still be picked up when installing from crates.io?

Uhm, that's a good hint... I see that according to the docs cargo will actually evaluate this field and error when the users rust version is too old: https://doc.rust-lang.org/cargo/reference/manifest.html#the-rust-version-field

Sorry, I should have read the docs completely before. My bad!

I'll add the field again along with a comment in `rust-toolchain.toml` to update `rust-version` in `Cargo.toml` as well.

imsnif:
All good and no worries! I only know this because I remembered it used to be an issue. Thanks!

PR #3488, opened 2024-07-10
fix(pane): handle ^M and ^[K codes properly

Hylian:
* ^M is the control code for carriage return ('\r')
* ^[K is equivalent to ^[0K


imsnif:
Hey @Hylian - thanks for this. The CI tells us this is failing some vttests, which might mean the fix should be a little more nuanced. On Discord we talked about providing the exact sequence that is causing trouble. Ideally a minimal reproduction that we could `cat` inside and outside of Zellij to compare and see the different behavior.

You could use the `--debug` flag as instructed in the ISSUE_TEMPLATE with the offending program if you like.

Could you provide this minimal reproduction so that I'll take a closer look? Thanks.

Hylian:
Yep, will do. I was having some trouble reproducing the same scenario on my personal machine (unfortunately, I can't directly share the failing build output.) I'll file an issue after investigating a bit more and checking my assumptions.

Hylian:
Had a brainfart and mixed up ^M and ESC M, since vim seems to display them the same. Fixed in updated patch.

Also created https://github.com/zellij-org/zellij/issues/3489 with repro for the issue I'm seeing. It doesn't seem directly related to this patch, so root cause may be different.

PR #3498, opened 2024-07-14
fix: (tabs) move to next tab if moving to next pane from fullscreen pane

### LINKED ISSUE #3434: MoveFocusOrTab does not switch tab when a pane is in fullscreen
## Issue description

When having two tabs, with the first having two panes of which one is in fullscreen, pressing <kbd>Alt</kbd><kbd>→</kbd> does not move to the second tab.

The following keybinding from the default bindings:

```kdl
    shared_except "locked" {
		// ...
        bind "Alt l" "Alt Right" { MoveFocusOrTab "Right"; }
        // ...
    }
```

seems to not work when a pane is in fullscreen, even though there is another tab to switch to. Going to fullscreen does not automatically enter locked mode so I had the expectation that this would work.

## Minimal reproduction

1. Create two tabs
2. In the first tab, create two panes
3. Make the first pane fullscreen
4. Press <kbd>Alt</kbd>+<kbd>Right</kbd>
	* **Expected:** The second tab is focused
	* **Actual:** nothing happens 

## Other relevant information

Zellij: 0.40.1

VasilisManol:
fixes #3434

### DIFF
--- a/zellij-server/src/panes/tiled_panes/mod.rs
+++ b/zellij-server/src/panes/tiled_panes/mod.rs
@@ -1665,16 +1665,18 @@ impl TiledPanes {
         }
     }
 
-    pub fn focus_pane_left_fullscreen(&mut self, client_id: ClientId) {
+    pub fn focus_pane_left_fullscreen(&mut self, client_id: ClientId) -> bool {
         self.unset_fullscreen();
-        self.move_focus_left(client_id);
+        let ret = self.move_focus_left(client_id);
         self.toggle_active_pane_fullscreen(client_id);
+        return ret;
     }
 
-    pub fn focus_pane_right_fullscreen(&mut self, client_id: ClientId) {
+    pub fn focus_pane_right_fullscreen(&mut self, client_id: ClientId) -> bool {
         self.unset_fullscreen();
-        self.move_focus_right(client_id);
+        let ret = self.move_focus_right(client_id);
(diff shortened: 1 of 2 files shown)


imsnif:
Hey, circling back to this (thank you for your patience!) - I tested this out and found a slight issue: this only works to change to the next tab if the original pane was on the side of the screen. Otherwise the first move action doesn't do anything (behind the scenes it switches to the next pane, but the user doesn't see anything because it's hidden) and only then it switches to the next tab.

If you want to fix this, I'll be happy to take another look and merge.

VasilisManol:
Hey, I can't reproduce this. I tried with a horizontal layout with 5 panes. 1 at the row on top and 4 at the row below. Even If I start at one pane in the second row in between and switch to right the correct pane comes to fullscreen and changes.
Did you mean such a case or did I got you wrong?

imsnif:
Oh, wait - sorry! This was a misunderstanding on my part. When in this situation as you describe, the focus first switches to the other pane in the same tab (which then turns full screen) and then when it reaches the edge, to the other tab. I wasn't seeing this because I tested this with all panes being identical. My bad!

I don't know why we don't see CI running in this PR (probably because it's quite old). I'm going to run the tests locally on this branch and if they pass I'll merge - otherwise we can troubleshoot things here.

imsnif:
Everything's green. Thanks for putting this together and once again, thanks for your patience!

PR #3504, opened 2024-07-18
fix(serialization): do not serialize junk sessions

imsnif:
Previously, sessions would be serialized (for resurrection later) every `serialization_interval` (default: 60 seconds) regardless of their content. This created a situation where lots of "junk sessions" were created in the cache and clogged it up. Most of these sessions would just be one-offs - not something that we'd actually want to serialize and later resurrect.

This fixes the problem by only serializing a session if it has more than 1 pane*, or if a command is running inside a pane that is not the default shell (eg. vim).

*more accurately, a different number of panes than the layout it was created with - by default this means more than one terminal pane.


luigir-it:
I feel like this should be documented, or even better be configurable.

One may work inside just one pane, run multiple "one shot" commands (`ls`,  `cat`, `docker ps`), then reboot expecting serialization (with `pane_viewport_serialization true`)  to have the whole terminal output recorded. This zellij session is instead considered a junk-session.

Issue #3372, opened 2024-05-23
Sixel support broken since v0.40.0


Zettexe:
**Basic information**

`zellij --version`: zellij 0.40.0, zellij 0.40.1

`stty size`: 76 157

`uname -av` or `ver`(Windows): Linux zettnet-desktop 6.1.87 #1-NixOS SMP PREEMPT_DYNAMIC Wed Apr 17 09:18:29 UTC 2024 x86_64 GNU/Linux

zellij.log:
```
INFO   |zellij_client            | 2024-05-23 14:05:30.463 [main      ] [zellij-client/src/lib.rs:178]: Starting Zellij client! 
INFO   |zellij_server            | 2024-05-23 14:05:30.469 [main      ] [zellij-server/src/lib.rs:281]: Starting Zellij server! 
WARN   |zellij_server::route     | 2024-05-23 14:05:30.472 [server_router] [zellij-server/src/route.rs:1025]: Server not ready, trying to place instruction in retry queue... 
WARN   |zellij_server::route     | 2024-05-23 14:05:30.472 [server_router] [zellij-server/src/route.rs:1126]: Server ready, retrying sending instruction. 
WARN   |zellij_server::route     | 2024-05-23 14:05:30.472 [server_router] [zellij-server/src/route.rs:1025]: Server not ready, trying to place instruction in retry queue... 
WARN   |zellij_server::route     | 2024-05-23 14:05:30.472 [server_router] [zellij-server/src/route.rs:1034]: Server not ready, trying to place instruction in retry queue... 
WARN   |zellij_server::route     | 2024-05-23 14:05:30.472 [server_router] [zellij-server/src/route.rs:1126]: Server ready, retrying sending instruction. 
WARN   |zellij_server::route     | 2024-05-23 14:05:30.472 [server_router] [zellij-server/src/route.rs:1034]: Server not ready, trying to place instruction in retry queue... 
WARN   |zellij_server::route     | 2024-05-23 14:05:30.472 [server_router] [zellij-server/src/route.rs:1045]: Server not ready, trying to place instruction in retry queue... 
WARN   |zellij_server::route     | 2024-05-23 14:05:30.472 [server_router] [zellij-server/src/route.rs:1126]: Server ready, retrying sending instruction. 
WARN   |zellij_server::route     | 2024-05-23 14:05:30.472 [server_router] [zellij-server/src/route.rs:1045]: Server not ready, trying to place instruction in retry queue... 
[...]

PR #3506, opened 2024-07-18
fix(terminal): repeat retry screen instruction sending during stdin cache replay (#3372)

lypanov:
The lack of re-retry on screen instruction playback from the stdin_cache ends up breaking sixel support as screen instructions such as pixel dimension detection response are unreliably replayed when the screen was not yet ready. This is a regression in sixel support introduced in 0.35.0.

Unsure if this is universal or something specific to our platform. However given it was seen by other users across multiple terminal emulators I started to look into it myself.
  
Existing tests pass though I'd much prefer to have written a test for this relatively complicated regression. Alas it'll be a while before I can invest the time so wanted to push the PR already and get some feedback / see how you feel about it.

### DIFF
--- a/zellij-server/src/route.rs
+++ b/zellij-server/src/route.rs
@@ -1180,13 +1180,17 @@ pub(crate) fn route_thread_main(
                     }
                     Ok(should_break)
                 };
+                let mut repeat_retries = VecDeque::new();
                 while let Some(instruction_to_retry) = retry_queue.pop_front() {
                     log::warn!("Server ready, retrying sending instruction.");
-                    let should_break = handle_instruction(instruction_to_retry, None)?;
+                    let should_break =
+                        handle_instruction(instruction_to_retry, Some(&mut repeat_retries))?;
                     if should_break {
                         break 'route_loop;
                     }
                 }
+                // retry on loop around
+                retry_queue.append(&mut repeat_retries);
                 let should_break = handle_instruction(instruction, Some(&mut retry_queue))?;
                 if should_break {
                     break 'route_loop;


lypanov:
Force pushed fix for formatting.

imsnif:
Great find! Correct me if I'm wrong though - doesn't this create an infinite loop with no delay and no exit condition? Or am I missing something?

lypanov:
> Great find! Correct me if I'm wrong though - doesn't this create an infinite loop with no delay and no exit condition? Or am I missing something?

If the needed session never gets to a ready state it could infinite loop. As it is already it *does* result in a burst of logging activity so I originally had a small sleep in-between retry attempts, but removed it in the end as not sure it's worth the extra complexity (never saw more than 15 retries). Any artificial (as in, not the normal path of session becomes ready) exit condition would probably mean keeping track of retry counts which I'm not convinced makes sense from an overhead perspective. Are there other conditions I'm missing? Could a single screen instruction failure result in this retrying infinitely? My Rust is pretty basic but my reading is that the only condition that would result in a retry is the screen not being ready.

To prevent a feasible "logging is inundated by retries" when unrelated development work results in screen never getting to a ready state the logging itself could be modified to make this clear, or an exponential backoff could be added.

Let me know your thoughts / preferences!

manueldeprada:
Thanks for the work @lypanov 
Tthe fix works as expected for me

imsnif:
@lypanov - I would like to add some sort of pause between retries. I think that would be our best option. Otherwise even breaking the process if the server becomes unresponsive would be challenging, the machine will probably go to 100% CPU and it would generally be an even more unpleasant experience when things crash.

imsnif:
No stress ofc, but any updates on this?

lypanov:
Sorry for the delay Aram postponed this for a week or two more as currently working on a personal project after a lay off.

manueldeprada:
@imsnif take a look here: 64a1cbc775e4df581e9764b9eeba0ed239992e88

Is adding a 100 millis sleep there enough? Or you were thinking in other kind of pause?

imsnif:
@manueldeprada - 100ms is a lot, but I think it might even work with 5-10ms. You can test this by creating an infinite loop in this case (eg. not handling the instructions and just pushing them over and over into the Vec) and seeing if you can kill the process without `-9`. (make sure to compile with `--release` or to install with `cargo x install <path>`).

lypanov:
Take into account the spam this will create in the logging. Some sort of counter anyway might be needed if taking this approach.

imsnif:
While I don't always follow [Zero One Infinity](https://en.wikipedia.org/wiki/Zero_one_infinity_rule) as can be seen in the code base, I think this is a good place to follow it. In other words: we have no way of knowing how many attempts will be the right amount (especially considering old hardware, temporary hiccups, etc.)

I hear you about the log spam, but we have spooling mechanisms in place so at worst it shouldn't take up more than a few hundred kilobytes if memory serves.

manueldeprada:
> @manueldeprada - 100ms is a lot, but I think it might even work with 5-10ms. You can test this by creating an infinite loop in this case (eg. not handling the instructions and just pushing them over and over into the Vec) and seeing if you can kill the process without `-9`. (make sure to compile with `--release` or to install with `cargo x install <path>`).

I just tested with 5ms, `--release` and a `while true` loop, and the impact on CPU seems totally negligible: 
image

I will create a new pull request with the patch.

manueldeprada:
This PR can be closed now :) 

Thanks for your efforts @imsnif 

and thanks so much for initially pushing this PR @lypanov !!!

PR #3510, opened 2024-07-19
feat(cli): Add `alive` option to `ls`

### LINKED ISSUE #3168: Add `--alive` option for `ls`
Currently the `ls` subcommand always prints all sessions, both active and exited. I suggest adding a `--alive` flag that only shows sessions that are alive. The name should probably be bikeshedded. Some other options: 

- `--only-alive`
- `--active`
- `--only-active`
- `--hide-dead`
- `--hide-exited`
- `--show-dead/exited=false`
- etc

This would add to the two existing flags: `--no-formatting` and `--short`. 

My specific use case is that I've set up ssh to show a list of active sessions, but I want it to be nice and short and I really don't care about dead sessions. In general it's a nice option that I think a lot of people would expect to have. 

I've looked through the code and it seems like a simple enough change. I would be more than glad to make a PR if it's something the maintainers agree with!

odilf:
Closes #3168 

In the end, I added the flag with the name `alive`, which I think is reasonable. Of course it can still be bikeshed, but since the issue hadn't had activity for a few months so I decided to go forward with what I think sounds best! 

I think it might make sense to only show alive session by default and have another `--exited` or `--show-exited` flag, but that would change the current behavior, so I figured I'd let a maintainer take that decision. 

As an FYI, I tried to run and test everything locally, but I couldn't get it to compile (the problem is some missing `target/../wasm32-wasi/...` plugin thingy) but the changes seem simple so I hope it just works 😅

### DIFF
--- a/src/commands.rs
+++ b/src/commands.rs
@@ -208,14 +208,14 @@ pub(crate) fn send_action_to_session(
                         "Session '{}' not found. The following sessions are active:",
                         session_name
                     );
-                    list_sessions(false, false, true);
+                    list_sessions(false, false, true, false);
                     std::process::exit(1);
                 }
             } else if let Ok(session_name) = envs::get_session_name() {
                 attach_with_cli_client(cli_action, &session_name, config);
             } else {
                 eprintln!("Please specify the session name to send actions to. The following sessions are active:");
-                list_sessions(false, false, true);
+                list_sessions(false, false, true, false);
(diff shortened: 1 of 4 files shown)


Rodrigodd:
Any expectations for eventually merging this?

I rebased this PR in my fork: https://github.com/Rodrigodd/zellij/tree/pr-3510

Issue #1679, opened 2022-08-19
Feature request: mouse any-event Tracking (1003)


AutumnMeowMeow:
Reference: https://www.invisible-island.net/xterm/ctlseqs/ctlseqs.html#h3-Any-event-tracking

This would round out mouse support, permitting 1) applications to know the mouse position always, and 2) implement focus-follows-mouse for panes.

imsnif:
I really like the "focus follows mouse" part, though I think we'll have to give some thoughts about the UX here. I'm not sure it should be on by default, at least... I guess we'll have to play around with it and see how it feels.

Another point that I'm pretty certain about: I think it would be best to only send mouse movements to the focused pane (assuming of course the movements are over said pane).

But regardless - I'd very much love to see this feature in Zellij. I guess we'll also have to place ourselves in any-event tracking mode. I wonder how wide-spread support for this is in terminals?

If you're interested in working on this @AutumnMeowMeow - I'd be happy to point you to the relevant parts of the code-base to get what you need (if you want me to, ofc).

AutumnMeowMeow:
> Another point that I'm pretty certain about: I think it would be best to only send mouse movements to the focused pane (assuming of course the movements are over said pane).

I believe the most common implementation for terminals and embedded terminal panes/widgets is:
* No button down - mouse motion events are only passed to the terminal if the mouse is over that pane.
  - Looking at a few terminals: xterm, mlterm, contour will report events even if the window is not focused; wezterm only reports if the window is focused.  I think it would make sense to either follow xterm's example and always send any-event if the mouse is over a pane whether focused or not; or make it configurable with whatever default you want.
* Button down - mouse motion events are passed to the focused terminal, with the coordinates clamped to the terminal dimensions.  (So no negative coordinates when click-dragging off screen.)  This is the click-and-drag scenario.

> I guess we'll also have to place ourselves in any-event tracking mode.

zellij can activate both button-event and any-event in that order on startup, like: "\x1b[?1002;1003h", and then zellij will get any-event if the host supports that, and button-event if the host doesn't support any-event but does support button-event.  Easy graceful fallback.  (And a good test for terminals that don't loop on all the DECSET parameters. :-) )

> I wonder how wide-spread support for this is in terminals?

[...]

PR #3538, opened 2024-08-03
Mouse AnyEvent tracking (1003) -- work in progress

AutumnMeowMeow:
PR for more detailed discussion, for #1679 

Current bit compiles, passes tests, fails clippy (not my fault, other parts of zellij-utils doing that).

Things to do:

- [x] Fix broken mouse Release events
- [x] Decide how Motion events will be exposed to zellij code, i.e. which Action:: items will it have?
- [x] Implement Motion events
- [ ] Fix broken mouse text selection / copy-paste
- [x] Add CSI ? 1003 support inside the terminal pane
- [ ] Test Normal tracking in vttest
- [ ] Test ButtonEvent tracking in vttest
- [ ] Test AnyEvent tracking in vttest
- [ ] Fixup other code style issues (I still suck at Rust)


AutumnMeowMeow:
@imsnif 

You can see my TODOs on top.  I will try to minimize commits to this PR so it doesn't constantly spin up the tests.

Initial questions:

* Does it make sense what I have here so far?  Moving away from a "single button down" --> "held" --> "release" model, instead to "button(s) down" --> "button(s) maybe released or dragged, no buttons, etc."
* What Action's would you like to see for Motion events?  (My vote would be a Action::MouseMotion(point, buttons).)

zellij-utils/src/input/mouse.rs
```diff
@@ -2,102 +2,135 @@ use serde::{Deserialize, Serialize};
...
+
+impl MouseEvent {
+    #[allow(unused)]
+    pub fn from_termwiz(mut old_buttons: MouseButtons, event: termwiz::input::MouseEvent) -> Self {
+        // We use the state of old_buttons vs new_buttons to determine
+        // if this event is a Press, Release, or Motion.  This is an
+        // unfortunate side effect of the pre-SGR-encoded X10 mouse
+        // protocol design in which release events don't carry
+        // information about WHICH button(s) were released, so we have
+        // to maintain a wee bit of state in between events.
```
imsnif, line 55: Honestly, even this is much better than what we've been doing up until now with the HeldMouse stuff.

imsnif:
Hey @AutumnMeowMeow !

The direction looks very good. We do not enforce clippy anywhere, so all good.
 
>     * Does it make sense what I have here so far?  Moving away from a "single button down" --> "held" --> "release" model, instead to "button(s) down" --> "button(s) maybe released or dragged, no buttons, etc."  I would like to carry this simpler model down to the terminal/grid layer, replacing the six {left/right/middle}_button_{signal/release_signal} functions with three functions {press/release/motion}.  On the plugins layer all the existing Actions could stay as-is so as not to break existing code.

Sounds perfect. From my side, as long as the plugin API (and especially the protobuf layer) doesn't change we're not breaking anything, so should be good. It sounds like this structure is much more in-tune with the way these events are encoded as seen in your changes, so let's go with it.

>     * What Action's would you like to see for Motion events?  (My vote would be a Action::MouseMotion(point, buttons).)

Sounds good to me! Note that when you get to serializing these actions to protobuffs, you can just ignore them as unsupported actions here: https://github.com/zellij-org/zellij/blob/main/zellij-utils/src/plugin_api/action.rs#L1295


I'm excited about this feature! Already thinking of how to use it in the UI.

AutumnMeowMeow:
@imsnif OK most of the core new function is here.  :)  Some things are broken though, and I could use your pointers.  (Also converted this to draft in the futile hope it would not spin up the whole CI stuff on every commit...sigh.)

* cargo build/run/test work fine. However cargo xtask build/run/test do not.  They cannot see the input::mouse crate, no idea why.  Any ideas on how to fix?
* I temporarily commented out the other left/right/middle-click functions, which now breaks text selection of course.  But along the way I also removed the MouseHold stuff, and see a few decently large functions that got axed in zellij-server-src/tab/mod.rs.  Do we actually need those?  Or would those functions still work via only LeftClick / LeftRelease / etc.?
* Somewhere along the way the mouse position is getting a couple off-by-ones, i.e. not 0-based.  I suspect that is an artifact of the find_active_tab stuff I cargo culted from the left_click code, e.g. failing to properly handle the conversion from absolute coordinates returned by termwiz to the relative-pane coordinates needed before sending to the terminal.

Thank you for your help! :-)

zellij-server/src/panes/grid.rs
```diff
@@ -1921,6 +1922,90 @@ impl Grid {
...
+                    event.position.line() - 1,
+                ));
+                Some(String::from_utf8_lossy(&msg).into())
+            },
+            MouseMode::Sgr => Some(format!(
+                "\u{1b}[<{:?};{:?};{:?}{}",
+                self.mouse_buttons_value_sgr(event),
+                // AZL: Why is event.position not staying 0-based on
+                // both axes?
+                event.position.column(),
```
AutumnMeowMeow, line 1999: I see that I missed the relative coordinate conversion in handle_mouse_event().  I'll fix that.

zellij-server/src/tab/mod.rs
```diff
@@ -3280,172 +3297,6 @@ impl Tab {
...
-                        client_id,
-                    )
-                    .with_context(err_context)?;
-                    return Ok(true); // we need to re-render in this case so the selection disappears
-                }
-            }
-        }
-        Ok(false) // we shouldn't even get here, but might as well not needlessly render if we do
-    }
-
```
AutumnMeowMeow, line 3448: The three functions above are decently large.  Is there logic anywhere else to handle changing the selection buffer size as the mouse is dragged around?  If not, then we will need to put that in handle_mouse_event().

zellij-utils/src/input/mouse.rs
```diff
@@ -2,102 +2,146 @@ use serde::{Deserialize, Serialize};
...
+    // Keyboard modifier flags can be encoded with events too.  They
+    // are not often passed on the wire (instead used for
+    // selection/copy-paste and changing terminal properties
+    // on-the-fly at the user-facing terminal), but alt-mouseclick
+    // usually passes through and is testable on vttest.  termwiz
+    // already exposes them too.
+    pub shift: bool,
+    pub alt: bool,
+    pub ctrl: bool,
+
```
AutumnMeowMeow, line 29: Also FYI that Swing and WIN32 both expose keyboard modifiers with mouse events.  Soooo when zellij is running natively on Windows we'll have the infrastructure to do things like Ctrl-Click / Alt-Drag / etc. :-)

imsnif: Honestly, even `Alt` + `mouse_click` would be amazing. I'm thinking of using it for multiple selection of panes.

AutumnMeowMeow:
Refactoring termwiz_mouse_convert and from_termwiz got the builds all working, thank you @imsnif ! :-)

A few more glitchy things fixed, and now viola, actual proof of concept:

https://github.com/user-attachments/assets/18292703-fd86-4581-b0b6-2b819e1964a3

imsnif:
Hey @AutumnMeowMeow - I hope you are well and don't mind the mention. I want to get this merged and released in the next version, as it will complement some of the planned new features very nicely. Is this something you'd like to keep working on? (no pressure - there is no release date for now)

darrenburns:
Very cool stuff here! Looking forward to this 🙂

imsnif:
After receiving permission from @AutumnMeowMeow - I'm continuing the work on this. Just rebased from `main`.

imsnif:
I spent some time with this and at this point almost everything is working:

1. Copying text with the mouse is working
2. Dragging floating panes with the mouse is working
3. Scrolling focused/unfocused panes with the mouse is working
4. Pinning panes with the mouse is working
5. Focusing panes with left/right click is working
6. When Zellij does not intercept the event for one of the above, we serialize it to the pane properly (thanks to @AutumnMeowMeow 's fantastic work)

Left to do functionality-wise (mostly minor quirks):
* do not clear the "copied to clipboard" message when the mouse just moves without a button press
* text selection doesn't release properly when release event happens outside of pane (probably need to turn self.selecting into an Option<PaneId> to solve this)
* if we focus a pane with a right click or middle click, we shouldn't also send the pane the right click
* marking text under and left to a pinned pane doesn't work

Otherwise I'd like to get the tests/formatting working, write some more tests for this, and then we can merge so we can start adding really cool UI/UX stuff with these crazy good bits of functionality.

imsnif:
Right, so - worked on this some more. Everything is mostly fixed and working!

More things to do:
1. Fix one e2e test
2. Adjust the default plugins for the change in API (I decided to issue a breaking behavior change here rather than trying to adapt to the old events) so that stuff like clicking on tab names to focus them will work again
3. Conditionally send a notification to plugins about the state change when clicking a mouse action changed something (we used to do this on every click but this would be quite spammy now!)
4. Write some integration tests for AnyEvent

All in all - not a lot more to do. I'm hoping to have this merged soon.

imsnif:
Everything is done and working great! :tada: 

Once the CI passes I'm going to merge this. If @AutumnMeowMeow has the time and feels like testing this, it would be most welcome (even after the merge - three's still some time before the release).

Thanks again for all your hard work!

jondo2010:
Thank you, been waiting for this since discovering Zellij a couple years ago already :smile:

PR #3558, opened 2024-08-19
feat(ux): reload config at runtime

imsnif:
This is the beginning of a series of changes that will allow fully reloading and reapplying the Zellij configuration file (by default `~/.config/zellij/config.kdl`) at runtime.

## What this does?
This PR adds a filesystem watcher to the config file on `zellij-client`. When the file is created or modified, we reload its contents and send them to `zellij-server` who reapplies them to the current session if they changed.

In this PR, aside from this infrastructure, we add support for reapplying the `keybinds` block and the `default_mode` option setting from the configuration file. Nothing else works yet!

## What's next?
Adding support for all the rest of the config file (themes, configuration options, plugin aliases, etc.) - upcoming in next PRs!

### DIFF
--- a/zellij-client/src/lib.rs
+++ b/zellij-client/src/lib.rs
@@ -12,11 +12,18 @@ use log::info;
 use std::env::current_exe;
 use std::io::{self, Write};
 use std::path::Path;
-use std::path::PathBuf;
 use std::process::Command;
 use std::sync::{Arc, Mutex};
 use std::thread;
 use zellij_utils::errors::FatalError;
+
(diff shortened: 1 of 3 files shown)


jaeheonji:
It's REALLY AWESOME! Thank you for doing this 😍 (And sorry for not being able to contribute these days)

imsnif:
> It's REALLY AWESOME! Thank you for doing this 😍 (And sorry for not being able to contribute these days)

Always happy to see you around @jaeheonji ! You will be welcome back when you have the time :)

PR #3561, opened 2024-08-21
feat(plugins): API to temporarily bind keys to send a message to a specific plugin id

imsnif:
This adds the `MessagePluginId` bindable keybind action. Similar to `MessagePlugin`, this will create a pipe that will send a message to a specific plugin - only here it will be done by this plugin's runtime ID.

The purpose is to use this with the `reconfigure` plugin API to essentially allow plugins to temporarily bind global keys in the session to send messages to themselves.

Example:

```rust
let own_plugin_id = Some(get_plugin_ids().plugin_id);
let save_configuration_to_file = false;
reconfigure(format!(r#"
    keybinds {{
        shared {{
            bind "F1" {{
                MessagePluginId {own_plugin_id}  {{
                    name "F1"
                }}
            }}
            bind "F2" {{
                MessagePluginId {own_plugin_id} {{
                    name "F2"
                }}
            }}
            bind "F3" {{
                MessagePluginId {own_plugin_id} {{
                    name "F3"
                }}
            }}
        }}
    }}
"#), save_configuration_to_file);
// ...
impl State {
    fn pipe(&mut self, pipe_message: PipeMessage) -> bool {
        let mut should_render = false;
        if pipe_message.name == "F1" {
            eprintln!("F1 key pressed globally");
        }
        if pipe_message.name == "F2" {
            eprintln!("F2 key pressed globally");
        }
        if pipe_message.name == "F3" {
            eprintln!("F3 key pressed globally");
        }
        should_render
    }
}
```

Note: this sort of keybind action should only be used in this case. It is not serialized with the user's config and should not be trusted to persist across sessions. If placed statically in a configuration it might have unintended consequences (seeing as plugin ids are assigned at runtime).

### DIFF
--- a/zellij-server/src/plugins/mod.rs
+++ b/zellij-server/src/plugins/mod.rs
@@ -134,6 +134,7 @@ pub enum PluginInstruction {
         cwd: Option<PathBuf>,
         skip_cache: bool,
         cli_client_id: ClientId,
+        plugin_and_client_id: Option<(u32, ClientId)>,
     },
     CachePluginEvents {
         plugin_id: PluginId,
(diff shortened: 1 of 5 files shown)


Lockszmith-GH:
> Note: this sort of keybind action should only be used in this case. It is not serialized with the user's config and should not be trusted to persist across sessions. If placed statically in a configuration it might have unintended consequences (seeing as plugin ids are assigned at runtime).

Just asking for clarification about this:
Does this means that if a plugin wants to allow it's users to have configuratble keybindings - it should manage it's own configuration and assign keybindings through this new mechanism?

imsnif:
> Just asking for clarification about this: Does this means that if a plugin wants to allow it's users to have configuratble keybindings - it should manage it's own configuration and assign keybindings through this new mechanism?

I think it depends on what the plugin wants to do. This mechanism is best used if the plugin wants to assign/change keys at runtime. Otherwise the plugin can use the "classic" [plugin configuration](https://zellij.dev/documentation/plugin-api-configuration) (either through layouts, cli or pipes).

PR #3564, opened 2024-08-23
feat(ux): reload config options at runtime

imsnif:
This concludes the work began in https://github.com/zellij-org/zellij/pull/3558 and continued in https://github.com/zellij-org/zellij/pull/3559, to now allow reloading everything defined in the config at runtime with a few notable exceptions (see below).

## What now also works (can be reloaded at runtime from the config or changed from a plugin)
* Reloading `simplified_ui` (i.e. changing arrow font support)
* Reloading `default_shell`
* Reloading `pane_frames`
* Reloading `copy_command` (i.e. clipboard replacement for terminals not supporting OSC52)
* Reloading `copy_to_clipboard` (i.e. the clipboard destination)
* Reloading `copy_on_select`
* Reloading `auto_layout`
* Reloading `rounded_corners`
* Reloading `hide_session_name`
* Reloading `scrollback_editor`

## What still requires a restart
Most of these were not worked on due to some internal limitations that would have required considerable development to get around. I deemed these to be high-effort/low-reward and will look into implementing hot-reloading for them as well if there's a need I missed while making this judgement call.

* The `plugins` (plugin aliases) - this is because it has not yet been decided whether reloading these should reload the existing plugins loaded with aliases or not. More thought needs to be devoted to the best behavior in this context.
* `on_force_close` - due to internal limitations
* `session_serialization`, `serialize_pane_viewport`, `scrollback_lines_to_serialize` - due to internal limitations
* `default_layout` - this requires some reworking of the layout implementation inside Zellij
* `mouse_mode` - currently being worked on in a different context
* `scroll_buffer_size` - due to internal limitations
* `mirror_session` - does not make sense as this only applies to new sessions
* `layout_dir`, `theme_dir` - due to internal limitations
* `styled_underlines` - does not make a lot of sense to change without finer control of scrollback
* `disable_session_metadata` - due to internal limitations
* `support_kitty_keyboard_protocol` - due to internal limitations

### DIFF
--- a/zellij-server/src/lib.rs
+++ b/zellij-server/src/lib.rs
@@ -263,6 +263,13 @@ impl SessionMetaData {
     }
     pub fn propagate_configuration_changes(&mut self, config_changes: Vec<(ClientId, Config)>) {
         for (client_id, new_config) in config_changes {
+            self.default_shell = new_config.options.default_shell.as_ref().map(|shell| {
+                TerminalAction::RunCommand(RunCommand {
+                    command: shell.clone(),
+                    cwd: new_config.options.default_cwd.clone(),
+                    ..Default::default()
+                })
+            });
             self.senders
                 .send_to_screen(ScreenInstruction::Reconfigure {
                     client_id,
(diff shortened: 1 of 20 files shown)


chardskarth:
Hi! Looking forward for the auto reload of configuration on runtime. Which version of zellij is this included in?

aaraney:
@chardskarth, it looks like it was released with [0.41.0](https://github.com/zellij-org/zellij/blob/main/CHANGELOG.md#0410---2024-11-04).

PR #3570, opened 2024-08-25
fix(terminal): repeat retry screen instruction sending during stdin cache replay (cont)

manueldeprada:
We come from #3506 

The sleeping time has been set to 5ms, as suggested. As noted in https://github.com/zellij-org/zellij/pull/3506#issuecomment-2299829281, entering the while loop does not significantly impact performance.

### DIFF
--- a/zellij-server/src/route.rs
+++ b/zellij-server/src/route.rs
@@ -10,6 +10,8 @@ use crate::{
     screen::ScreenInstruction,
     ServerInstruction, SessionMetaData, SessionState,
 };
+use std::thread;
+use std::time::Duration;
 use uuid::Uuid;
 use zellij_utils::{
     channels::SenderWithContext,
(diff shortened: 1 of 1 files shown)


manueldeprada:
bumping this!!

imsnif:
Not forgotten :) Thanks for the reminder. Will get to it before the release, it's just one item out of about 50.

imsnif:
Hey, this looks good to me. Want to run `cargo fmt` so that CI passes and then I'll merge?

manueldeprada:
> Hey, this looks good to me. Want to run `cargo fmt` so that CI passes and then I'll merge?

done!!

imsnif:
Running `cargo fmt` from the project root should take care of this. It might be an issue of a rust version - in that case `rustup update` and then `cargo fmt` might do the trick.

manueldeprada:
> Running `cargo fmt` from the project root should take care of this. It might be an issue of a rust version - in that case `rustup update` and then `cargo fmt` might do the trick.

done! hope it passes now

imsnif:
Thank you!

PR #3579, opened 2024-08-31
refactor(themes): adjust `catppuccin-latte` & add comments

sgoudham:
Hey :wave:, this PR introduces some changes as part of merging https://github.com/catppuccin/zellij/pull/11 downstream.

We noticed that the `fg` and `bg` keys for `catppuccin-latte` were the same which seemed incorrect, however, we couldn't find any noticeable changes on the UI when changing these keys. I'd appreciate any insight into why the theme seemed to function well even with the foreground and background keys set to the same colour.

I've also added comments to map `white` and `fg` to the Catppuccin palette colour of `text`. 

Hope these changes are fine and let me know if there's anything else I need to edit, thanks!

### DIFF
--- a/zellij-utils/assets/themes/catppuccin.kdl
+++ b/zellij-utils/assets/themes/catppuccin.kdl
@@ -4,7 +4,7 @@
 themes {
   catppuccin-latte {
     bg "#acb0be" // Surface2
-    fg "#acb0be" // Surface2
+    fg "#4c4f69" // Text
     red "#d20f39"
     green "#40a02b"
     blue "#1e66f5"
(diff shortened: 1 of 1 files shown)


imsnif:
Hey @sgoudham and @jaeheonji - I didn't notice this PR, and when working on some fixes for the new UI I also noticed this issue. I changed fg to `#5c5f77` (Subtext 1) because it looked best to me. In our context it's mostly about adjusting the color of unselected ribbons. Here's an example of how it looks in the new UI:

img-2024-10-10-162606
img-2024-10-10-162546

What do you think?

Otherwise (maybe related, maybe not) I'm a little concerned about the contrast of the magenta and red variants (as can be seen in these screenshots with `Ctrl g` and `<n> New Pane` (the n part) respectively). Do you have any thoughts?

jaeheonji:
Hey! Sorry for late reply.

When I first added the catppuccin theme to zellij, I added it slightly differently from the catppuccin color rules. The reason is that because of some UI errors and unused colors that we had previously, I added the most appropriate color expression to zellij at that time.

In my opinion, I think it is fine to modify the color palette suitable for zellij as long as the look and feel of the theme matches. So, @imsnif I think that one looks good.

jaeheonji:
After the theme settings were changed, I believe the colors will be displayed correctly.

[catppuccin-latte.kdl](zellij-utils/assets/themes/catppuccin-latte.kdl)

Therefore, I will close this PR.

PR #3658, opened 2024-10-10
fix(ui): set background color for UI components according to theme

imsnif:
This fixes a minor issue in the UI components where `Text` and its various consumers would not set the theme background as their background - causing the UI to look a bit disjointed at times.

### DIFF
--- a/zellij-server/src/ui/components/nested_list.rs
+++ b/zellij-server/src/ui/components/nested_list.rs
@@ -27,22 +27,32 @@ pub fn nested_list(
             break;
         }
         let mut reset_styles_for_item = RESET_STYLES;
-        if line_item.text.selected {
-            reset_styles_for_item.background = None;
-        };
+        reset_styles_for_item.background = None;
+        reset_styles_for_item.foreground = None;
         let padding = line_item.indentation_level * 2 + 1;
         let bulletin = if line_item.indentation_level % 2 == 0 {
             "> "
         } else {
             "- "
         };
-        let text_style = reset_styles_for_item.bold(Some(AnsiCode::On));
+        let text_style = if line_item.text.selected {
+            reset_styles_for_item
+                .bold(Some(AnsiCode::On))
+                .foreground(Some(style.colors.white.into()))
+                .background(Some(style.colors.bg.into()))
+        } else {
+            reset_styles_for_item
+                .bold(Some(AnsiCode::On))
+                .foreground(Some(style.colors.white.into()))
+                .background(Some(style.colors.black.into()))
+        };
         let (mut text, text_width) = stringify_text(
             &line_item.text,
             Some(padding + bulletin.len()),
             &coordinates,
             style,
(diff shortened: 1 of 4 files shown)


dj95:
Is there any way to set the background color explicitly in the UI components? After testing the commit with a translucent terminal, it looks a bit weird as text from UI components suddenly has a background.

Screenshot 2024-10-12 at 22 38 58

imsnif:
I think this should work correctly when the terminal background is set to
the same color as the theme background, then also the translucency should
be okay

dj95:
Unfortunately I cannot get it to work. It always draws the background. Especially on normal text (`print_text_with_coordinates`). I could change the color by changing the `black` theme property, but not the `bg`. And I cannot get it translucent.

Do you have any hints at this?

imsnif:
My apologies, yeah - I meant `bg`. We're overhauling the theme definition so right now it's a bit wonky. Your terminal creates transparency by changing a specific color to transparent (the programs running inside the terminal itself are not aware of transparency). So if your terminal emulator's background color is the same as the theme black color, it should work.

imsnif:
This is what happens with the new status-bar, which also uses the text component (for example for `Ctrl g` and `Alt +`). I just checked and this works with transparency as expected (my terminal background is black and I'm working with the default theme).

dj95:
Screenshot 2024-10-15 at 17 50 01

Thanks for your quick reply! Both are set to `#000000`, but it unfortunately doesn't work. Is there any way I can further debug?

imsnif:
I'd try to look into how the terminal or external application (eg. for me it's picom) sets the translucency and go from there. Maybe my assumption is wrong.

dj95:
It's a terminal issue. In Rio everything works as expected. Sorry to bother you.

But it might be relevant for the next release, so I could add it to the documentation, when I find out what's going on there.

imsnif:
Might be different ways of setting transparency... maybe the affected terminal only sets transparency to text without background. If this is the case, we need to be able to handle that as well. If you could look further into it, it'll be appreciated.

dj95:
Some terminals seem to handle it different, e.g. Wezterm. It sets the background color, if it's a truecolor, even if it's the same one. If I understood correctly from the Wezterm comment, it would set the transparency though, if the background color is set to '`default`'.

Ref: https://github.com/wez/wezterm/issues/1425#issuecomment-999638132

Here are some terminals I've tested. All set to `#000000` as background color and `#000000` as black color. Just the opacity value is not the same:

| Name | Screenshot | Support |
|-------|-----------|----------|
| Wezterm | Screenshot 2024-10-15 at 18 24 18 | :x: |
| iTerm2 |  Screenshot 2024-10-15 at 18 20 23 | :x: |
| kitty | Screenshot 2024-10-15 at 18 21 26 | ✅ |
| Alacritty | Screenshot 2024-10-15 at 18 23 34 | :x: |

Issue #3479, opened 2024-07-06
bug: parallel downloads of the same plugin via web locations break the plugin


dj95:
When a plugin is loaded via http multiple times in parallel, the plugin breaks for zellij. On the first glance, it does not look like a common issue, but it is commonly reproduced with status bar plugins. When a user defines a layout with multiple tabs, on a clean cache zellij will start the download multiple times and corrupts the file.

**Basic information**

`zellij --version`: v0.40.1

`stty size`: 20 140

`uname -av` or `ver`(Windows): Linux 57ffe693fca0 6.5.0-35-generic #35-Ubuntu SMP PREEMPT_DYNAMIC Sat Apr 27 00:18:38 UTC 2024 aarch64 GNU/Linux

Excerpt from the zellij log:

```
ERROR  |zellij_server::plugins::w| 2024-07-06 17:33:31.943 [async-std/runti] [zellij-server/src/plugins/wasm_bridge.rs:1224]: Io(Custom { kind: NotFound, error: VerboseError { source: Os { code: 2, kind: NotFound, message: "No such file or directory" }, message: "could not rename `/root/.cache/zellij/11443609859275959894305566542538288992.part` to `/root/.cache/zellij/11443609859275959894305566542538288992`" } })
ERROR  |zellij_server::plugins::w| 2024-07-06 17:33:31.947 [async-std/runti] [zellij-server/src/plugins/wasm_bridge.rs:1224]: failed to start plugin 1 for client 1

Caused by:
    0: No such file or directory (os error 2): '/root/.local/share/zellij/plugins.wasm'
    1: No such file or directory (os error 2): ''
    2: failed to load plugin from disk
```

## Minimal reproduction

Loading the following layout will show the error line at the bottom. Removing all tabs except one, clearing the cache and loading the layout will work.

```javascript
layout {
  default_tab_template {
    children
    pane size=1 borderless=true {
      // plugin location="file:./zjstatus.wasm.2" {
      plugin location="https://github.com/dj95/zjstatus/releases/latest/download/zjstatus.wasm" {
        format_left   "{mode} #[fg=#89b4fa,bold]{session}"
        format_center "{tabs}"
        format_right  "{command_git_branch} {datetime}{notifications}"
        format_space  ""
      }
    }
  }

  tab {
    pane split_direction="horizontal" {
      pane size="70%"
      pane
    }
  }
  tab {
    pane split_direction="horizontal" {
      pane size="70%"
      pane
    }
  }
  tab {
[...]

Issue #3036, opened 2023-12-28
Zellij crashing when making web request in plugin


lannonbr:
## Issue description

When I run a plugin that tries to make a HTTP request using the `web_request` command, it crashes zellij and the following error is outputted in the logs:

```
WARN   |zellij_utils::ipc        | 2023-12-28 11:29:29.029 [router    ] [zellij-utils/src/ipc.rs:223]: Error in IpcReceiver.recv(): InvalidMarkerRead(Error { kind: UnexpectedEof, message: "failed to fill whole buffer" }) 
ERROR  |zellij_client            | 2023-12-28 11:29:29.030 [router    ] [zellij-client/src/lib.rs:356]: Received empty message from server 
```

## Minimal reproduction

https://github.com/lannonbr/zellij-http-plugin-example

I'm running it by compiling the plugin with `cargo build` and then running it in an active zellij session with the `zellij plugin` command.

```
zellij plugin -- file:./target/wasm32-wasi/debug/zellij-http-plugin-example.wasm
```

## Other relevant information

Zellij verison: `0.39.2`

Rustc --version --verbose output:

```
rustc 1.74.0 (79e9716c9 2023-11-13)
binary: rustc
commit-hash: 79e9716c980570bfd1f666e3b16ac583f0168962
commit-date: 2023-11-13
host: aarch64-apple-darwin
release: 1.74.0
LLVM version: 17.0.4
```

imsnif:
Hey, thanks for the heads up! For me this works

![img-2023-12-28-174331](https://github.com/zellij-org/zellij/assets/795598/611356c1-9c2d-4c0c-992d-95317357c132)

What I did:
1. Started a new Zellij session
2. Cloned your repository: https://github.com/lannonbr/zellij-http-plugin-example
3. Did `cargo build`
4. Did `zellij plugin -- file:target/wasm32-wasi/debug/zellij-http-plugin-example.wasm`
5. Focused on the plugin and pressed the `Down` arrow
6. Got the above in the logs

The error logs are unfortunately not very helpful, they're just the client disconnecting. The only thing I can think of that would cause this is running this plugin on an older version of Zellij...

lannonbr:
So I tried the following:
- recompiling the plugin with the lastest version of rust (1.75)
- cleared my zellij cache folder
- made sure that both the zellij instance and `zellij-tile` were both on 0.39.2
- Tried running your weather-pal plugin as I was using such as a template on how to write a.

[...]

PR #3668, opened 2024-10-12
fix(plugins): force use curl system lib on macOS

dj95:
There were in the past some problems with downloading plugins via HTTPS on macOS (https://github.com/zellij-org/zellij/issues/3036#issuecomment-2024214955, https://github.com/zellij-org/zellij/issues/3479#issuecomment-2407490553).

The cause is, that curl is not properly recognised when building zellij, especially after the latest changes to the isahc crate in favour of surf. This results in zellij not being linked to `libcurl`, which utilises some other dependencies in the background to implement the functionality, that results in crashes for certain scenarios. 

This pull request fixes the issue, as it forces `curl-sys` to use system libs of macOS instead of choosing it on its own. On my macOS with nix as build environment and in a VM with curl installed via brew, the issue is fixed and zellij is not crashing anymore.

In case there are some errors on linux, even though it was tested, we could introduce a target selector (`[target.'cfg(target_os = "macos")'.dependencies]`).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5128,6 +5128,7 @@ dependencies = [
  "colorsys",
  "common-path",
  "crossbeam",
+ "curl-sys",
  "directories",
  "expect-test",
  "futures",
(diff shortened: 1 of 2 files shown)


imsnif:
Thank you!

Issue #3720, opened 2024-11-04
zellij-utils: isahc dependency forces vendored libcurl dependency by default


tranzystorekk:
Found in zellij `0.41`

`isahc`'s default features include `static-curl`: https://github.com/sagebind/isahc/blob/5c533f1ef4d6bdf1fd291b5103c22110f41d0bf0/Cargo.toml#L22

This in turn makes `curl-sys` hardcode the static vendored `libcurl` build, so distro maintainers will be forced to patch this away if they want to use system-provided `libcurl`. Unfortunately, `curl-sys` crate doesn't provide any convenient env var override, like e.g. `git2` with its `LIBGIT2_NO_VENDOR`.

My proposal to improve this in a backwards-compatible way is to add some feature switch to zellij that controls this behavior (sth like `system-curl` would be best, but cargo wants features to be strictly additive).

imsnif:
Hey @tranzystorekk - thanks for pointing this out and for your patience with my reply! Would you like to whip-up a PR that would add an opt-in compile-time feature for this change? I think it will help other distro maintainers too.

PR #3745, opened 2024-11-08
chore: add vendored-openssl feature

theoparis:
This allows using the system openssl installation. If you want me to I can make it vendored by default, with `--no-default-features` to disable it.


tranzystorekk:
Related: #3720

If `isahc` is patched to use system libcurl, openssl linking becomes its responsibility instead.

theoparis:
Should this be closed now that your PR is merged?

tranzystorekk:
> Should this be closed now that your PR is merged?

Probably, unless you have other reasoning for this PR

Issue #3190, opened 2024-03-09
support focusing pane by name


lucaferranti:
Zellij is already great, thanks for the great work!

One feature I find myself missing from time to time is being able to select what pane to focus on based on name. Currently I can do

```
zellij action focus-next-pane
zellij action focus-previous-pane
```

**it would be great to support**

```
zellij action focus-pane-name <name>
```

to switch focus on a pane with a given name, regardless of its position. Motivation for this is more robust communication between panes. I currently have a macro in my editor to send code from the editor to the terminal, but this is a bit fragile as it relies on a very specific layout with panes on a very specific position. Being able to select panes by name would be more robust.

PR #3747, opened 2024-11-08
Add focus-pane-with-id CLI action

topisani:
Allows scripts to better control the pane focus. My use case is to avoid this horrible hack used in the kakoune zellij integration:
https://github.com/mawww/kakoune/blob/8c49c8ee404fecc110338c08eeed515b63caa93c/rc/windowing/zellij.kak#L46-L59

Instead, a pane can now be focused from its `$ZELLIJ_PANE_ID` environment variable.
Possibly a fix for https://github.com/zellij-org/zellij/issues/3190

### DIFF
--- a/zellij-server/src/screen.rs
+++ b/zellij-server/src/screen.rs
@@ -4320,7 +4320,14 @@ pub(crate) fn screen_thread_main(
                 screen.log_and_report_session_state()?;
             },
             ScreenInstruction::FocusPaneWithId(pane_id, should_float_if_hidden, client_id) => {
+                let client_id = if screen.get_active_tab(client_id).is_ok() {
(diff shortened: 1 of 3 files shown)


imsnif:
Hey, I appreciate you putting the time into this but I'm afraid in Zellij-world we encourage users to perform such actions with plugins.

As you have discovered, the CLI is not aware of multiple clients and can never be (because multiple clients being focused on the terminal is a Zellij concept). Meaning that when there is more than one client connected to the session, the results of this action will be unexpected to all users involved.

Plugins are client aware. It's possible to write a plugin that does this and then load it into Zellij, have it communicate with kakoune and generally do the right thing.

Thank you for your understanding.

Issue #3332, opened 2024-05-06
Can't create pannel upward


JonatasAmaral:
## Issue description
On pointing a direction for creating a new page, "Up" and "Left" act just like "Down" and "Right" respectively.

Given the bellow example, i was expecting, the on creating new panel on upward direction, to "panel 3" to end up above "panel 2", it got bellow though.

![zellij-new-up](https://github.com/zellij-org/zellij/assets/8677724/fa4bdd92-d6ed-48e4-9dae-ecb7298868d4)

The equivalent happens to --direction "Left" too, creating panel to the right instead.

## Minimal reproduction

Open a zellijs session, and run the command to create new panel:
```bash
zellij action new-pane --direction "Up"
```

## Other relevant information
Zellij version: 0.40.1 (latest)

mikegreiling:
I am also experiencing this issue. Running zellij v0.43.1

PR #3748, opened 2024-11-08
Add a --size parameter to `zellij run`

topisani:
This has been an important thing missing for me in automating zellij with [kakoune](https://github.com/mawww/kakoune).

Currently, the `--size` parameter allows either a percentage or an initial row/column count. Importantly, this row/column count does not result in a fixed-size pane, but is just an initial size, that is translated to percentage constraint. This is more useful in my case, but maybe that should be added as a separate SplitSize type instead.

Architecture-wise, this required passing the flag down through a lot of layers, and im unsure if there are better ways to do it.

Additionally, `--direction left` and `--direction up` have been fixed, which should close https://github.com/zellij-org/zellij/issues/3332

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -37,6 +37,7 @@ fn main() {
             y,
             width,
             height,
+            size,
         })) = opts.command
         {
             let cwd = cwd.or_else(|| std::env::current_dir().ok());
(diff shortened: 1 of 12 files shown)


topisani:
@imsnif In the context of [your comment](https://github.com/zellij-org/zellij/pull/3747#issuecomment-2464654133), i assume the same applies here. The second commit (direction bug fix) still seems relevant though to me though, right?

PR #3766, opened 2024-11-12
chore: add vendored_curl feature

### LINKED ISSUE #3720: zellij-utils: isahc dependency forces vendored libcurl dependency by default

tranzystorekk:
Closes #3720
Supersedes #3745

The additive and by-default nature of the `vendored-curl` feature is the cleanest way I could come up with to retain the old vendored behavior and also achieve the "off switch" that I talk about in the issue above.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -80,8 +80,10 @@ pkg-fmt = "tgz"
 
 [features]
 # See remarks in zellij_utils/Cargo.toml
-default = ["zellij-utils/plugins_from_target"]
+default = ["plugins_from_target", "vendored_curl"]
+plugins_from_target = ["zellij-utils/plugins_from_target"]
 disable_automatic_asset_installation = ["zellij-utils/disable_automatic_asset_installation"]
+vendored_curl = ["zellij-utils/vendored_curl"]
 unstable = ["zellij-client/unstable", "zellij-utils/unstable"]
 singlepass = ["zellij-server/singlepass"]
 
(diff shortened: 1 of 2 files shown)


tranzystorekk:
Note: due to https://github.com/alexcrichton/curl-rust/issues/321 even with `vendored_curl` disabled, the `curl` crate will silently fall back to vendored if no system curk has been found, so this is a best-effort solution

dj95:
Seems to work well on macOS. It properly links to libcurl of the system (in the nix-store in my case). Downloading plugins from within a layout also still works.

Thanks for asking me, @imsnif !

Screenshot 2024-11-15 at 14 27 47

imsnif:
Great, thanks @dj95 !

And thanks for putting this together @tranzystorekk

Issue #3702, opened 2024-10-25
please drop dependency on common-path


WeepingClown13:
A bit of context first; I am packaging zellij for Debian, for a while now. There are tons of dependencies that had to be and has to be packaged first, as per Debian policy. Currently I have finished packaging all dependencies/dev-dependencies of zellij-utils, except [common-path](https://crates.io/crates/common-path).

`common-path` is currently version 1.0.0, and the last update to it was 6 years ago. That much is quite common, but this one also points to an nonexistent repository as source. Examining the [source code](https://docs.rs/crate/common-path/1.0.0/source/src/lib.rs), there is only so much - non complex - logic existing there. with less than 50 lines of actual library code excluding tests. At that point, having it as a separate package becomes only a lot of maintenance burden. Examining the common-path usage in zellij, it only seems to be used at a single place in zellij-utils.

In conclusion, it'd make life a lot of easier for downstream packaging if this dependency can be avoided. And that leaves some questions.

1. Can the usage of common-path itself easily be removed? Would that affect anything severely?
2. Are there better or well maintained alternatives that could be used, if the utility it provides is a must have?
3. Even better, is it possible to zellij itself to inline its own logic for this so that the utility is retrieved and it's one less dependency?

Please share your thoughts!

imsnif:
Hey, thank you for your efforts! I would be very happy to see Zellij packaged for Debian. common-path is only used in one place in the code base: https://github.com/zellij-org/zellij/blob/main/zellij-server/src/session_layout_metadata.rs#L252 

We use it there to find the common path of terminal working directories when serializing a layout for the session resurrection. This code isn't overly complex but can potentially be error prone. I'm happy to accept a PR that does this manually instead using the STD lib's primitives. I guess it shouldn't be overly complex. Wdyt?

imsnif:
My apologies - I was a bit unclear. I would be happy to use the STD lib's primitives. :)

Right now I'm mostly concerned with getting the next release out the door, but once that's done - if you haven't gotten this done yet I'll try to whip something up. This honestly sounds like a 5-10 lines of code function to me, but I'm always open to having missed something.

joshuamegnauth54:
You may consider vendoring this dependency for the next release too. Based on the code, the two functions seem easy to replicate, only use the standard library, and likely won't ever need maintenance. Plus, the original author included unit tests too. :smile_cat: 

The main logic is:
[...]

PR #3780, opened 2024-11-15
Vendoring common_path

s1syph0s:
As mentioned in #3702, this PR should help packaging zellij by avoiding dependency on `common-path`, whose repository is not available anymore.

One thing I haven't done is mentioning the original author. What's the best way to mention them in the source code? Should I just mention it through a comment or should I add it to rustdoc?


imsnif:
We would need to include the original license verbatim in a comment: https://docs.rs/crate/common-path/1.0.0/source/LICENSE-MIT

s1syph0s:
Thank you for your feedback @imsnif! I agree with you, that `rand` is not a good dependency for the tests. I will fix that soon

imsnif:
Great work @s1syph0s! Thanks for putting this together.

PR #3795, opened 2024-11-21
fix(tab): focus largest pane instead of topmost

### LINKED ISSUE #3006: Stack pane position not persisting when switching tabs
https://github.com/zellij-org/zellij/assets/28514482/e5390ac0-fc55-41e7-bd8b-fc5ce00f9bf0

## Issue description

Here is my layout config

```
layout {
	default_tab_template {
		pane size=1 borderless=true {
			plugin location="zellij:compact-bar"
		}
		children
	}

	swap_tiled_layout name="stack" {
		tab {
			pane stacked=true {
				pane
			}
		}
	}

	swap_tiled_layout name="tile" {
		tab {
		  pane
		}
	}
}
```

The issue I'm noticing is the following:

Suppose I'm focusing on pane 2 of tab 2 in a stacked view, I toggle to tab 1 and then toggle back to tab 2, the focus is on pane 1 and not on the original pane of focus (pre-tab-toggle). The expected behaviour would be that the focus remains on the originally focused pane in the stacked view after toggling between tabs.

## Other relevant information

Attached a recording.

Also I have noted that this does not occur for the `GoToTab` action, or if I explicitly click on a tab, or if I go to next/previous tab in tab mode. It seems like it exclusively occurs for the [MoveFocusOrTab](https://zellij.dev/documentation/keybindings-possible-actions#movefocusortab) action.

Pyxels:
When multiple Panes have the same x coord in a `MoveTabOrFocus` event that switches tabs, instead of picking the pane at the top to focus, choose the one biggest one. This ensures that in stacked layouts, the focused pane is opened again when moving, instead of always focusing the top pane.

## Current Behaviour
When switching tabs in `MoveTabOrFocus`, pick the one at the border, if multiple, pick the topmost.

## Fixed Behaviour
When switching tabs in `MoveTabOrFocus`, pick the one at the border, if multiple, pick the *largest*, then the topmost.

This ensures that when stacked panes are used, the previously focused is once again focused instead of always using the first.

## Considerations
This should not influence existing workflows using normal tiled panes all to much, since the previous implementation was quite arbitrary anyway, but is a big fix for stacked layouts, since this "keeps the layout" instead of resetting.

closes #3006

### DIFF
--- a/zellij-server/src/panes/tiled_panes/tiled_pane_grid.rs
+++ b/zellij-server/src/panes/tiled_panes/tiled_pane_grid.rs
@@ -1039,14 +1039,14 @@ impl<'a> TiledPaneGrid<'a> {
                 Direction::Left => {
                     let x_comparison = a.x().cmp(&b.x());
                     match x_comparison {
-                        Ordering::Equal => b.y().cmp(&a.y()),
+                        Ordering::Equal => a.rows().cmp(&b.rows()).then(b.y().cmp(&a.y())),
                         _ => x_comparison,
                     }
                 },
                 Direction::Right => {
                     let x_comparison = b.x().cmp(&a.x());
                     match x_comparison {
-                        Ordering::Equal => b.y().cmp(&a.y()),
+                        Ordering::Equal => a.rows().cmp(&b.rows()).then(b.y().cmp(&a.y())),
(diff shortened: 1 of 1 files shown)


Pyxels:
closed in #4037

PR #3799, opened 2024-11-22
fix(cli): fix the detach message

### LINKED ISSUE #3798: Zellij prints "Bye from Zellij!" instead of "Session detached" when I'm detaching the session
**Basic information**

`zellij --version`:  zellij 0.41.2

`stty size`: 22 126

`uname -av`: Darwin macaw.local 23.6.0 Darwin Kernel Version 23.6.0: Thu Sep 12 23:36:12 PDT 2024; root:xnu-10063.141.1.701.1~1/RELEASE_ARM64_T6020 arm64 arm Darwin

I launch `zellij setup --clean` and press `^O d` to detach the session. I get the "Bye from Zellij!" message.

According to https://github.com/zellij-org/zellij/commit/b4cf0e7a818ac85c14e598ac302d2f4a11f6cd30#diff-06572a96a58dc510037d5efa622f9bec8519bc1beab13c9f251e97e657a9d4edR61 and https://github.com/zellij-org/zellij/blob/09689eae8b96ddb95713e6612ec17007ced91306/zellij-utils/src/ipc.rs#L131, it's supposed to print "Session detached" in this case, so that I can tell that zellij hasn't quit completely.

(I was initially going to file an FR, but then found that the feature I was going to ask for is supposed to be implemented, it's just not working)

`zellij attach` still works, so I know that it did detach.

louiswpf:
Keys are interpreted on the server after commit 5a437b79aac5 ("refactor(server): interpret keys on server so they can be rebound").

Since the server exits with "ExitReason::Normal" instead of "ExitReason::NormalDetached" when detaching a session, the detach message becomes "Bye from Zellij!" instead of "Session detached".

Correct the exit reason to fix this.

Fixes: #3798

### DIFF
--- a/zellij-server/src/lib.rs
+++ b/zellij-server/src/lib.rs
@@ -1219,7 +1219,7 @@ pub fn start_server(mut os_input: Box<dyn ServerOsApi>, socket_path: PathBuf) {
                     let _ = os_input.send_to_client(
                         *client_id,
                         ServerToClientMsg::Exit {
-                            exit_reason: ExitReason::Normal,
+                            exit_reason: ExitReason::NormalDetached,
                         },
                     );
                     remove_client!(*client_id, os_input, session_state);


Arcovion:
Bumping this PR as I feel like it's been overlooked - hopefully it'll be picked up for the next release 🤞

PR #3814, opened 2024-11-27
zellij fourify

EyalRoginski:
Add the "Fourify" option, that splits the current pane into 4 panes!
image


Zykino:
Isn’t it "just" a 4 pane layout that you can include in your config and, for example, load as `zellij action new-tab --layout fourify`?
Or if you want 4 panes arranged like this you may defines a swap-layout that arrange them like that when there are 4.

Issue #2974, opened 2023-12-01
hide_floating_panes doesn't seem to work


realsnick:
Hi, here is the layout i'm loading, btop always shows even though `hide_floating_panes=true` 

PS. is there a way to setup keybinding to show/hide the pane?
```
    tab name="Dev" focus=true hide_floating_panes=true {
        pane split_direction="vertical" {
            pane name="editor" borderless=true {
                edit "."
            }
        }
        pane split_direction="vertical" size=8 {
            pane name="cargo run" borderless=true {
                command "cargo" 
                args "run"
            } 
            pane size="20%" borderless=true
        }
        floating_panes {
            pane {
                x "10%"
                y "10%"
                width "80%"
                height "80%"
                command "btop"
            }
        }
    }
```

imsnif:
Works for me...

What I'm doing: I placed a `layout` node around it and loaded it with `zellij -l /path/to/layout.kdl`. The layout is loaded and btop is only triggered when I trigger the floating panes with `Ctrl p` + `w`. Also, this is the only way to trigger floating panes unless they're a plugin (and then you can focus the pane specifically with the `LaunchOrFocus` binding).

imsnif:
Alright. Could you please find a minimal reproduction?

realsnick:
this is minimum without default_tab_template... this one fail with error

```layout {
    tab focus=true hide_floating_panes=true {
        pane split_direction="vertical" {
            pane
        }
        pane split_direction="vertical" size=8 {
            pane
            pane
        }
        floating_panes {
            pane {
                x "10%"
                y "10%"
                width "80%"
                height "80%"
            }
        }
    }
}

``` 

this is minimum with default_tab_template

```
layout {
    default_tab_template {
        pane size=1 borderless=true {
            plugin location="zellij:tab-bar"
        }
        children
        pane size=2 borderless=true {
            plugin location="zellij:status-bar"
        }
    }

    tab focus=true hide_floating_panes=true {
        pane split_direction="vertical" {
            pane
        }
        pane split_direction="vertical" size=8 {
            pane
            pane
        }
        floating_panes {
            pane {
                x "10%"
                y "10%"
                width "80%"
                height "80%"
            }
        }
    }
}

```

realsnick:
the most important for me is to be able to do keybinding to toggle a pane. not sure if its a feature to be added or already exists

realsnick:
looks like the code samples doesn't work
[...]

PR #3846, opened 2024-12-06
Set hide_floating_panes in parse_tab_node_with_template

zhyuri:
Fixes https://github.com/zellij-org/zellij/issues/2974

When the layout uses `default_tab_template`, the `hide_floating_panes` option on the `tab` was ignored. With this change the user can hide floating panes while using `default_tab_template`.


theogf:
That fixes a quite annoying issue, can we get that reviewed/merged?

tylerhellner:
It's been over a year and a half. Can you please review it @imsnif? It's a very annoying issue.

jon301:
Looking forward to this fix as well 🙏

PR #3853, opened 2024-12-08
fix clippy warnings

sharma-shray:


mo8it:
This is probably outdated and can be closed. Are you open to having a PR to fix all Clippy warnings?

Issue #2794, opened 2023-09-18
Flickering in Nushell when used with Zellij and WezTerm


yoav-lavi:
When e.g. displaying completions over a certain length, the output form Nushell will flicker under Zellij when used in WezTerm.

This does not happen when using Nushell directly under WezTerm (see below)

With Zellij:

https://github.com/zellij-org/zellij/assets/14347895/7d6890b1-82fa-4764-acfe-6939e8a9145e

Without Zellij:

https://github.com/zellij-org/zellij/assets/14347895/ba8df253-a58e-4bbc-a867-b5d8f3c03158

I'm using Zellij WezTerm and Nushell under Nix (darwin).

Possibly Related:

https://github.com/nushell/nushell/issues/9279
https://github.com/nushell/nushell/issues/6214

Log:

```log
WARN   |zellij_utils::input::layo| 2023-09-18 14:24:35.448 [main      ] [zellij-utils/src/input/layout.rs:962]: Failed to read swap layout file: /Users/yoavlavi/.config/zellij/layouts/layout.swap.kdl. Error: Os { code: 2, kind: NotFound, message: "No such file or directory" } 
INFO   |zellij_client            | 2023-09-18 14:24:35.451 [main      ] [zellij-client/src/lib.rs:152]: Starting Zellij client! 
INFO   |zellij_server            | 2023-09-18 14:24:35.458 [main      ] [zellij-server/src/lib.rs:245]: Starting Zellij server! 
INFO   |zellij_server            | 2023-09-18 14:24:35.515 [main      ] [zellij-server/src/lib.rs:913]: Compiling plugins using Cranelift 
INFO   |zellij_server::plugins   | 2023-09-18 14:24:35.515 [wasm      ] [zellij-server/src/plugins/mod.rs:137]: Wasm main thread starts 
INFO   |zellij_server::plugins::p| 2023-09-18 14:24:35.529 [async-std/runti] [zellij-server/src/plugins/plugin_loader.rs:487]: Loaded plugin 'compact-bar' from cache folder at '/Users/yoavlavi/Library/Caches/org.Zellij-Contributors.Zellij' in 5.527375ms 
INFO   |zellij_server::plugins::p| 2023-09-18 14:24:35.529 [async-std/runti] [zellij-server/src/plugins/plugin_loader.rs:487]: Loaded plugin 'compact-bar' from cache folder at '/Users/yoavlavi/Library/Caches/org.Zellij-Contributors.Zellij' in 6.388667ms 
[...]

PR #3884, opened 2024-12-19
Fix Synchronized Output (mode 2026) reporting

darrenburns:
Hi, thank you for Zellij!

I help maintain [Textual](https://github.com/textualize/textual/) and some dependent TUI apps like [Posting](https://github.com/darrenburns/posting) which make use of mode 2026.

I started using Zellij yesterday (loving it!) and noticed that the mode 2026 query response was not being parsed as expected and so the output could not be synchronized.

Looking into the code of Zellij, I think the response is missing a `?` character. If I'm understanding correctly, response should be of the format `CSI ? 2026 ; N $ y` where N can be any value in the range 0-4 (inclusive).

References:

- https://gist.github.com/christianparpart/d8a62cc1ab659194337d73e399004036
- https://vt100.net/docs/vt510-rm/DECRPM.html

## Before

Notice the tearing effect as the cursor moves through the list. It's particularly noticeable as the cursor moves over `PATCH`.

https://github.com/user-attachments/assets/332aaae5-2844-4ac5-a5f6-894df26871f4

## After

On a build containing this fix I cannot reproduce the tearing effect:

https://github.com/user-attachments/assets/d228626d-94c8-40a6-b3ed-becda0662285

Possibly fixes https://github.com/zellij-org/zellij/issues/2794

### DIFF
--- a/zellij-server/src/panes/grid.rs
+++ b/zellij-server/src/panes/grid.rs
@@ -2745,7 +2745,7 @@ impl Perform for Grid {
                 for param in params_iter.map(|param| param[0]) {
                     match param {
                         2026 => {
-                            let response = "\u{1b}[2026;2$y";
+                            let response = "\u{1b}[?2026;2$y";
                             self.pending_messages_to_pty
                                 .push(response.as_bytes().to_vec());
                         },


imsnif:
Great find, thanks @darrenburns !! Reading the spec I think we should also conditionally return `1` instead of `2` when we're mid-frame. But might be better to only solve this problem if we find that it exists.

Just a general note though: I find in this case that it's best to assume the terminal supports these notations and use them to mark rendering frames anyway without querying. This is what Zellij does on the rendering side. This being a private notation, terminals not familiar with it just drop it and continue on. It is good to return a proper response to the query though.

PR #3902, opened 2024-12-30
Upgrade wasmtime to v24.0.2 to remediate GHSA-c2f5-jxjv-2hh8

mamccorm:
Upgrades wasmtime to v24.0.2, to remediate:  GHSA-c2f5-jxjv-2hh8 (CVE-2024-51745).

Referenced previous upgrade contribution for guidance: https://github.com/zellij-org/zellij/pull/3685

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -13,11 +13,11 @@ dependencies = [
 
 [[package]]
 name = "addr2line"
-version = "0.21.0"
+version = "0.22.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "8a30b2e23b9e17a9f90641c7ab1549cd9b44f296d3ccbf309d2863cfe398a0cb"
+checksum = "6e4503c46a5c0c7844e948c9a4d6acd9f50cccb4de1c48eb9e291ea17470c678"
 dependencies = [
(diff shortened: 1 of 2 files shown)


imsnif:
Hi - thanks for this, but Zellij doesn't support windows. I'd rather not upgrade through 2 breaking changes if it can be avoided.

That being said, @WeepingClown13 - does this help you in your packaging efforts somehow? If it's enough of a help I might reconsider.

mamccorm:
Thanks @imsnif. Vuln scanners match GHSA-c2f5-jxjv-2hh8 to zellij (via wasmtime dep), but if zellij doesn't support windows or leverage the affected code in wasmtime, then this is probably more of a false positive

WeepingClown13:
@imsnif I am still waiting for wasmtime to be available in Debian for some [complicates reasons](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1087400), and thus my zellij packaging is stuck. The maintainer or rust-wasmtime source package is Debian also might be thinking about bumping the wasmtime version in Debian. Do you think you could wait until I can communicate regarding whether v24 is a reasonable common ground?

imsnif:
Sure thing @WeepingClown13 !

Issue #3891, opened 2024-12-22
KDL 2.0.0 support


zkat:
[KDL 2.0.0 has been released!](https://github.com/kdl-org/kdl/releases/tag/2.0.0)

Along with this new spec release, I've released [`kdl-rs@6.0.0`](https://github.com/kdl-org/kdl-rs/releases/tag/v6.0.0), which supports both KDL 1.0.0 and KDL 2.0.0, and is able to translate document text between both versions.

This means that zellij would not need to change much to add 2.0.0 support, except update some `KdlValue`-related API calls/`match`es, since that enum has changed.

If the `v1-fallback` feature is enabled, you would even be able to smoothly support both v2 and v1, safely, since both versions are unambiguously parseable. You can even convert v1 docs to v2 docs automatically, and vice-versa!

```rust
let mut doc = kdl::parse(input)?;
doc.ensure_v1();

let v1 = doc.to_string();
```

`kdl-rs`'s 2.0 parser is able to collect many syntax errors in a single pass, and so can give richer error messages than either the 1.0 parser (for now).

Always happy to answer questions! I hope this is useful to y'all!

imsnif:
Very cool! The meticulous work on backwards compatibility is very much appreciated. I'm looking forward to seeing this implemented.

Incidentally, some really nice related work from our community: https://github.com/dj95/kdl-fmt

zkat:
Unless someone claims this soon, I’ll probably give this an initial pass. It shouldn’t be much effort.

One thing that’s going to be interesting is what the errors should be: most people out there will likely be on v1 for a while, until docs, defaults, etc get updated. But the kdl-rs fallback mechanism only prints v2 errors if both formats fail (which would be the case in a truly malformed v1 file, for example).

I might want to poke at kdl-rs and make the fallback a bit more heuristic: that is, if both fail, do a quick scan of the document for “hints”, like unprefixed keywords, triple quotes, unquoted strings, or even the optional `/- kdl-version 1/2` hint.

I might do that first, actually tbh.

ck3mp3r:
Here's another quirk for you folks, and this might just me missing something; when using kdl-rs to generate values that only have single words or phrases joined by legal, non white-space characters, no quotes are added. Zellij however rejects unquoted values, even if they are legal according to spec.
I'm currently adding functionality to a utility I am writing to generate Zellij layouts programatically, hence why I stumbled on this. Would be good to know if I am missing something and this is just a red herring...

[...]

PR #3908, opened 2025-01-02
feat(config): add KDL v2 support

zkat:
Fixes: https://github.com/zellij-org/zellij/issues/3891

Starting this off as a draft, so we can discuss the changes and see what needs adjusting. I had to update dependencies a bit in order to resolve some cargo resolution conflicts, and with that came a bump in MSRV, as well as the removal of the deprecated `backtrace` feature for `insta`.

This uses kdl-rs' built-in "v1-fallback" system: It will first try to parse a KDL document as if it were v2. If that fails, it will try again with the legacy v1 parser (which is identical to the one Zellij is currently using, so v1 docs will parse the same). If _both_ fail, kdl-rs will try to apply some heuristics to figure out which parser's errors to display, and if the heuristics fail, default to v2 errors. Note: It is *completely safe* for well-formed v1 or v2 documents to be parsed with either parser. If a document just so happens to parse successfully under both parsers, the returned data will be _identical_.

Since KDL v2 also supports multiple diagnostics, I took the liberty of restructuring the KdlError type such that it would support that, too, and integrated it with the new error types from kdl-rs. This change also means that Zellij will be able to return One Big Err with a bunch of diagnostics in it, instead of erroring and exiting on the first failure (DeserializationErrors will already do this--but you will now be able to _append_ errors either to that, or to the post-deserialization semantic checks).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4,15 +4,6 @@ version = 3
 
 [[package]]
 name = "addr2line"
-version = "0.17.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "b9ecd88a8c8378ca913a680cd98f0f13ac67383d35993f86c90a70e3f137816b"
-dependencies = [
- "gimli 0.26.1",
-]
-
-[[package]]
-name = "addr2line"
 version = "0.21.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "8a30b2e23b9e17a9f90641c7ab1549cd9b44f296d3ccbf309d2863cfe398a0cb"
@@ -21,10 +12,19 @@ dependencies = [
 ]
 
 [[package]]
-name = "adler"
-version = "1.0.2"
+name = "addr2line"
+version = "0.24.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "f26201604c87b1e01bd3d98f8d5d9a8fcbb815e8cedb41ffccbeb4bf593a35fe"
+checksum = "dfbe277e56a376000877090da837660b4427aad530e3028d44e0bffe4f89a1c1"
+dependencies = [
+ "gimli 0.31.1",
+]
+
+[[package]]
+name = "adler2"
+version = "2.0.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "512761e0bb2578dd7380c6baaa0f4ce03e84f95e960231d1dec8bf4d7d6e2627"
 
 [[package]]
 name = "ahash"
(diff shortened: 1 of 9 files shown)


zkat:
I’ll… figure out what’s going on with the tests soon 😅

might be a fallback issue.

imsnif:
Hey! So - I'm starting to take a look at this.

Right now on this branch, Zellij doesn't start. We get a `Deserialization error: Failed to parse KDL document`. My guess is because we're failing to deserialize the default layout (I think this is also why most of the tests are failing).

I think an easier example to iterate over would be this test: https://github.com/zellij-org/zellij/blob/main/zellij-server/src/tab/unit/tab_integration_tests.rs#L6567 - in which we're failing to serialize one of these: https://github.com/zellij-org/zellij/blob/main/zellij-server/src/tab/unit/tab_integration_tests.rs#L6574-L6588 and https://github.com/zellij-org/zellij/blob/main/zellij-server/src/tab/unit/tab_integration_tests.rs#L6592-L6608 

I'm not too sure why every now and then we get a failure for the 1st and at other times we get a failure for the second. I can try to take a deeper look if something doesn't jump out.

I'd recommend skipping our build system and running it directly inside the `zellij-server` folder with `cargo test -- base_layout_is` for quick iteration.

theAkito:
Any chance at least partial v2 support may be added? It is particularly mocking to see tons of syntax errors, when editing those KDL files with an up-to-date language server and then getting errors like this from Zellij.

```
zellij

   ╭─[/home/user/.config/zellij/config.kdl:1:1]
 1 │ keybinds clear-defaults=#true {
   ·                         ▲
   ·                         ╰── Please refer to https://github.com/kdl-org/kdl/blob/main/SPEC.md#value for valid KDL value syntaxes.
 2 │     locked {
   ╰────
  help: For more information, please see our configuration guide: https://zellij.dev/documentation/configuration.html
```

While the referenced documentation specifically says, that [plain text booleans are not allowed](https://github.com/kdl-org/kdl/blob/main/draft-marchan-kdl2.md#identifier-string), plus [other notes regarding booleans' special treatment](https://github.com/kdl-org/kdl/blob/main/draft-marchan-kdl2.md#boolean).

Would be a huge & immediate DX improvement, if those booleans would be fixed.

imsnif:
While I would definitely want to see KDL v2.0 support in Zellij, the migration is not trivial (nothing to do with KDL and everything to do with infrastructure changes in a large code base being time consuming and difficult). Add to that the admin work that would have to go with it (communicating this to users, migration paths, updating documentation/screencasts/tutorials, etc.) and this becomes a large project with limited impact on the quality of life of users.

Zellij is underfunded and our development resources are stretched thin. I prioritize what I think is right.

 I absolutely love KDL and am dying to see V2.0 in Zellij, but I just can't afford to spend the time on it presently.

zkat:
I've been kinda crispy and busy with a new job and other life stuff, but if there's anything from the list of things blocking this that you think me or the community can pick up to lighten your load, @imsnif, just lmk. I imagine at least some of those things you would want to end up doing, yourself.

imsnif:
For sure. I think if you or someone from the community would like to make a PR that can parse v2 files as easily as it can v1, I can do the rest. Not for this upcoming version in *crosses fingers* a few weeks time, but I can make the time for next one.

zkat:
isn't that basically what this PR is?

imsnif:
It is, yes! But I think there's still a bit of work to be done here.

zkat:
absolutely! Just making sure we're on the right track. And if anyone else wants to help me with this PR, I could use the help, too :)

Issue #3055, opened 2024-01-09
tab completion for delete-session


benzwt:
It would be nice to be able to delete-session powered with tab-completion. Just like removing stopped containers with docker rm <tab>.

For example,
> zellij ls
```
inventive-elephant [Created 7h 10m 22s ago]
```
> zellij d inven + \<tab>\
will give
```
sellij d inventive-elephant
```

cristiand391:
I think if you use fish you get sesion completinon (haven't tried it):
https://github.com/zellij-org/zellij/blob/main/zellij-utils/assets/completions/comp.fish

for zsh I wrote  a custom completion script using fzf's completion API:
[Screencast from 01-13-2024 01:03:32 AM.webm](https://github.com/zellij-org/zellij/assets/6853656/1bbbc007-a1dc-4e0c-bb8d-fc62fb766b1e)

NorfeldtKnowit:
Since there are currently 134 open PRs and I'm not using fish but zsh I have made this temp fix

https://gist.github.com/NorfeldtKnowit/7556325d969446b27ea048ab18d49e25

PR #3915, opened 2025-01-05
 Add fish shell completions for the `delete-session` subcommand

s3igo:
This PR partially addresses issue #3055 by adding fish shell completion for the `delete-session` subcommand.
It enables session name suggestions when typing `zellij d ` followed by `<tab>`.

The newly added `complete` commands are in the same order as the SUBCOMMANDS section in the `zellij --help` output.


lgarron:
I would be glad to see this land!

(I was confused about why it wasn't working, which led me to this PR.)

PR #3942, opened 2025-01-16
fix(terminal): Support kitty keyboard protocol setting with "="

### LINKED ISSUE #3852: `Ctrl-Backspace` is mapped to `Ctrl-H` even when `support_kitty_keyboard_protocol true`

I test this by using `fish_key_reader`, which comes with `fish`:

1. Open `zellij`
2. Run `fish_key_reader`
3. Type `Ctrl-Backspace`

Expected: The terminal sees `Ctrl-Backspace`

Observed: The terminal sees `Ctrl-H`.

This is on macOS 15.1.1 iTerm 3.5.10 with `CSI u` enabled. `Ctrl-Backspace` is sent to the shell as `Ctrl-Backspace` consistently outside `zellij`, and the issue only occurs when I'm in `zellij`. So I'm fairly certain it's an issue with `zellij` and not my other configuration.

I tried enabling `support_kitty_keyboard_protocol true` in `config.kdl`, then `killall zellij` to make sure `zellij` is running nowhere on the system. This still doesn't fix the issue.

faho:
From
https://sw.kovidgoyal.net/kitty/keyboard-protocol/#progressive-enhancement:

> The escape code for requesting enhancements is:

> CSI = flags ; mode u

That means it can be *set* with CSI =, not just pushed/popped onto the stack with CSI < and CSI >.

This is important because otherwise zellij would interpret this as restore_cursor_position, which would move the cursor.

This would be hit by the next fish-shell release, which sends CSI = 5 u and CSI = 0 u to enable/disable the kitty keyboard enhancements unconditionally,
so if you used any program that saved the cursor position it would constantly move your cursor back on every prompt.

Fixes #3852

### DIFF
--- a/zellij-server/src/panes/grid.rs
+++ b/zellij-server/src/panes/grid.rs
@@ -2992,6 +2992,17 @@ impl Perform for Grid {
                 "\u{1b}[?0u"
             };
             self.pending_messages_to_pty.push(reply.as_bytes().to_vec());
+        } else if c == 'u' && intermediates == &[b'='] {
+            // kitty keyboard protocol without the stack, just setting.
+            // 0 disables, everything else enables.
+            let count = next_param_or(0);
+            if !self.explicitly_disable_kitty_keyboard_protocol {
+                if count > 0 {
+                    self.supports_kitty_keyboard_protocol = true;
+                } else {
+                    self.supports_kitty_keyboard_protocol = false;
+                }
+            }
         } else if c == 'u' {
             self.restore_cursor_position();
         } else if c == '@' {


zellij-server/src/panes/grid.rs
```diff
@@ -2992,6 +2992,17 @@ impl Perform for Grid {
...
+            // 0 disables, everything else enables.
+            let count = next_param_or(0);
+            if !self.explicitly_disable_kitty_keyboard_protocol {
+                if count > 0 {
+                    self.supports_kitty_keyboard_protocol = true;
+                } else {
+                    self.supports_kitty_keyboard_protocol = false;
+                }
+            }
         } else if c == 'u' {
```
faho, line 3006: An alternative is to only do restore_cursor_position if the intermediates are *empty*.

That won't fix the kitty keyboard protocol with fish, of course, but it will fix the cursor movement after you've run something like `mc` or `gh pr create` which saves the cursor position.

r-vdp (approved):
I ran into the issue where the cursor position is wrong after running `gh pr create` with fish 4 in zellij. I tested this patch and that resolves it.

imsnif:
Good catch! I totally overlooked this when implementing the protocol. Thank you.

Also, on a personal note - thank you for fish. It's my favorite shell <3

Issue #3719, opened 2024-11-04
ERROR IN PLUGIN - check logs for more info - v0.41.0


pdanpdan:
# 2. Issues with the Zellij UI / behavior / crash

## Issue description
After update from prev version I get `ERROR IN PLUGIN - check logs for more info` when zellij starts, with default config.

```
ERROR IN PLUGIN - check logs for more info
┌ status-bar ───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┐
│Loading status-bar...                                                                                                                                                      │
│Attempting to load from memory... NOT FOUND                                                                                                                                │
│Attempting to load from cache... NOT FOUND                                                                                                                                 │
│Compiling WASM                                                                                                                                                             │
│ERROR:  failed to start plugin 1 for client 1                                                                                                                              │
│                                                                                                                                                                           │
│Caused by:                                                                                                                                                                 │
│    0: failed to recover cache dir                                                                                                                                         │
│    1: WebAssembly translation error                                                                                                                                       │
│    2: Invalid input WebAssembly code at offset 712955: reference-types not enabled: zero byte expected                                                                    │
│ERROR IN PLUGIN - check logs for more info
```

## Minimal reproduction
- remove all configs for zellij
- start zellij (without any parameter)

## Other relevant information
- using https://github.com/zellij-org/zellij/releases/download/v0.41.0/zellij-x86_64-unknown-linux-musl.tar.gz
- zellij is in `/usr/local/bin`
- PopOS running on Intel CPU
- terminal is wezterm, but the same happens in other terminals
```
Distributor ID: Pop
Description:    Pop!_OS 22.04 LTS
Release:        22.04
Codename:       jammy
```

How can I find the logs with the error?

jscarrott:
Exact same issue on Arch updated via binstall.

giyany:
Hi @pdanpdan, thanks for the swift report. Naturally, this is important.
For general information, zellij logs should write to /$temp_dir/zellij-<UID>/zellij-log/zellij.log. in most systems $temp_dir, is /tmp with some expections (/var/folders/dr/ in MAC).

imsnif:
Looking into this - meanwhile, does this happen with a cargo install on the same machines? Does clearing the cache folder (you can find its location with `zellij setup --check`) help? (beware, the latter will remove all resurrectable sessions from the relevant version)

[...]

PR #3955, opened 2025-01-25
Update to Wasmtime 29.0.1

bjorn3:


bjorn3:
This was previously blocked on an MSRV bump, which https://github.com/zellij-org/zellij/pull/3945 has done now.

bjorn3:
Are the end to end test failures spurious? A quick look suggested that there are more PR's which failed in the exact same way.

imsnif:
Hey @har7an - could the e2e failures in the CI be related to the recent toolchain upgrade? I saw you said it made some changes in the build system.

imsnif:
Seems like main passes fine with the MSRV upgrade, so I'm not sure it's related. I'll try to take a closer look in the next few days.

har7an:
> Hey @har7an - could the e2e failures in the CI be related to the recent toolchain upgrade? I saw you said it made some changes in the build system.

They could, at least I can't prove they aren't.

Looking at your (@bjorn3) CI run though I suspect it hit some timeouts. The e2e-tests usually take something between 15-20 minutes (so they did on my PR as well), yours took 67 minutes.

imsnif:
Looking through the logs of the e2e run for this branch, it seems we're crashing with `Received empty message from server` - which is probably a catch all for "The server crashed and we swallowed the error somewhere". Again, happy to take a closer look in a few days if no-one finds the exact problem until then.

imsnif:
Hey, so I'm looking into this and it seems that with this branch the generic musl build crashes on startup with this error:

```
ERROR  |zellij_utils::errors::not| 2025-01-28 14:11:57.143 [main      ] [zellij-utils/src/errors.rs:692]: Panic occured:
             thread: main
             location: At zellij-server/src/lib.rs:1494:70
             message: called `Result::unwrap()` on an `Err` value: cannot create an engine with GC support when none of the collectors are available; enable one of the following features: `gc-drc`, `gc-null`
```

Any hits, @bjorn3 ?

bjorn3:
https://github.com/zellij-org/zellij/commit/95ff0853c6a852ced785a8a0ea9ae71cbb43b29f enabled the gc feature of Wasmtime which I think with newer Wasmtime version requires either the gc-drc or gc-null features to also be enabled to select which garbage collector to use. The gc-null feature doesn't garbage collect at all and as such isn't suitable for Zellij which uses long lived wasm instances. So enabling the gc-drc feature hopefully fixes this.

imsnif:
Alright, iirc we enabled the gc feature to fix https://github.com/zellij-org/zellij/issues/3719 - so I think we still need it.
Any negative trade-offs to adding the `gc-drc` feature here? More dependencies, unsupported platforms, or...?

bjorn3:
There shouldn't be. In fact I think the code that gc-drc enables was always enabled when the gc feature is enabled previously.

imsnif:
The e2e tests work by building a musl binary, sticking it in a docker container and sending commands through SSH to it. The build itself worked properly but crashed on first run when we tried to compile the plugins with the specified error.

I didn't see this when I checked out this branch and did a `cargo x run --singlepass`, but I guess this was because we used a different compiler (I think it's mapped to winch for us, right?)

I just checked now and before your fix this also crashes for `cargo x run` without `--singlepass`, so I think we're good.

har7an:
So to wrap up: e2e tests didn't exactly fail as such, just the binary we built crashed right on start. Is that correct? Or do I still need to investigate stuff w.r.t. the toolchain update?

imsnif:
> So to wrap up: e2e tests didn't exactly fail as such, just the binary we built crashed right on start. Is that correct? Or do I still need to investigate stuff w.r.t. the toolchain update?

That's correct, but if you could still help me out with https://github.com/zellij-org/zellij/pull/3957 I'd really appreciate it!

PR #3957, opened 2025-01-27
feat(ui): stacked resize

imsnif:
https://github.com/user-attachments/assets/53ce1597-2c13-4e85-bd50-aa7d427b93cd

This changes the default Zellij resize algorithm to a new one called "stacked resize". It is demonstrated in the short video above.

In a gist:
1. When resizing a pane with Alt +/- (non-directional resize), Zellij will try to first resize it by 30% (instead of the current 5%) and if unable to do so, will try to create a stack out of it with its neighbors. This happens in each direction, first preferring the direction which contains neighbors geometrically contained within the pane's relevant boundaries (horizontal/vertical depending on the direction).
2. When already inside a stack and doing a negative non-directional resize (`Alt -`) Zellij will break out a pane from this stack.
3. When adding new panes with `Alt n`, if the swap layouts are dirty (the swap layout indication, by default on the upper right, is not lighted), Zellij will attempt to split the currently focused pane. If it is a stack, Zellij will add a pane to this stack.
4. When splitting a stack right or down (`Ctrl p` + `r` or `Ctrl p` + `d`), Zellij will now split the whole stack in this direction as if it were a normal pane.

This new resize algorithm can be opted-out of by specifying `stacked_resize false` in the config (can also be updated at runtime).


har7an:
For some reason I don't understand:

1. how the cache key `v0-rust-format-Linux-846f4f1d-a50227c7` came to be, because the commit hash in question doesn't contain any of that
2. Why the *build* jobs produce incredible amounts of debug output where the *formatting* job doesn't

This looks like a very interesting problem indeed. I'll have another look tomorrow.

imsnif:
Thank you @har7an - I really appreciate it! Let me know if you want me to jump in so we can put our heads together.

har7an:
Hey @imsnif,

I think I sort of solved the puzzle. Here's the failure from CI:

image

and I get the exact same error when trying to build locally. But you said:

> The tests are failing due to a compilation error, however the actual code in this branch is different than what the error shows.

I had a look at `zellij-utils/src/pane_size.rs`, line 43 (as mentioned in the error) and it seems you renamed `is_stacked` (what I had used on main) to `stacked`.

So the diff to fix your error should be:

```diff
diff --git (-)/zellij-utils/src/pane_size.rs (+)/zellij-utils/src/pane_size.rs
index c86eaba2..d337e9ac 100644
--- (-)/zellij-utils/src/pane_size.rs
+++ (+)/zellij-utils/src/pane_size.rs
@@ -40,7 +40,7 @@ impl std::hash::Hash for PaneGeom {
         self.y.hash(state);
         self.rows.hash(state);
         self.cols.hash(state);
-        self.is_stacked.hash(state);
+        self.stacked.hash(state);
     }
 }
```

and I think you can undo that "do not cache CI stuff" commit again. :)

har7an:
`cargo x make` finished successfully on my PC, so I think that fix really does the trick.

imsnif:
I swear to you, I looked at that file 7 times both on Github and on my PC and it wasn't there. Either git voodoo or me just working too many hours. :)

Thanks for the second pair of eyes.

Issue #742, opened 2021-09-25
Feature: apply layout at runtime


imsnif:
We should be able to allow defining several possible layouts in the layout yaml file: https://zellij.dev/documentation/layouts.html

We should then have some keyboard shortcut that would switch between these layouts, provided that the same amount of panes are open on screen (it would rearrange the open panes to fit the new layout alternate).

spacemaison:
I'll work on this over the weekend if no one else has taken this up yet.  It looks like I need to:

1) Modify the layout file to accept multiple "template" layouts
2) Add `Action::NextLayout` / `Action::PreviousLayout` variants and bind keys to them
3) Add `ScreenInstruction::NextLayout` / `ScreenInstruction::PreviousLayout` variants to be called from the router.
4) Add an `update_layout` method to both the screen and tab to handle resizing and re-rendering panes.
5) (Possibly) Add a layouts mode in the status bar to surface the keybindings to the user. I'm not sure if this is necessary.

Does that all sounds about right?

imsnif:
Sounds good to me!

raphCode:
I think it could also be useful to "add" a new layout to current session, opening more tabs as required by the new layout.

This would fit my workflow because I only keep a single terminal window with zellij open at all times, for weeks:
When I want to work on e.g. zellij, I could just open a prepared layout that adds a new Tab with panes running vim, cargo watch, and some shell for git commands while leaving all previously opened Tabs/Panes intact.

mikegreiling:
Would love to be able to do this, specifically so I can switch back and forth to the `compact-bar` at will.

The interface is excellent and very friendly for new users, but once you learn the commands it feels a bit needlessly busy.

I'm at a sort of middle ground, where I can get by on memory most of the time, but still need to reference things once in a while. It would be great if I could start off with the compact bar, but be able to switch to a layout containing the full default status bar on-the-fly using a keyboard shortcut in the instances where I need it.

imsnif:
Check out [zellij-forgot](https://github.com/karimould/zellij-forgot).

mikegreiling:
> Check out [zellij-forgot](https://github.com/karimould/zellij-forgot).

Oh, that is perfect. Thank you 🙏

patrick-motard:
How does this feature differ from swap layouts?

pkleineb:
[...]

PR #3984, opened 2025-02-12
Feature: Apply layout at runtime; Implementing #742

pkleineb:
As said in the title this PR aims to implement a system for applying layouts at runtime based on one layout file.

What I did:
- added capabilities in the layout kdl parser for parsing multiple `layout` blocks (normal syntax applies)
- added a new `LayoutConfig` struct that stores the currently selected layout and all layouts that got parsed. It also implements utility functions for switching to layouts and iterating over all of them
- Since we are changing the main Layout struct from `Layout` to `LayoutConfig` I moved functionality like `list_available_layouts()` to `LayoutConfig` and marked the old functions as deprecated in rust itself
- added `Action::PreviousLayout` and `Action::NextLayout` for Keybindings
- added `ScreenInstruction::PreviousLayout` and `ScreenInstruction::NextLayout` telling the screen to switch Layouts
- added `screen.update_layout` switching to the screens currently active layout using multiple new `ScreenInstruction`'s

I gotta be honest this is my first time working with this big of a codebase and therefore I might not have used best practice or overlooked something during changing from `Layout` to `LayoutConfig` that breaks some Plugins, but since I don't have many plugins I wouldn't know.

I also implemented some unit and e2e tests for the functionality.

Let me know if there is something missing and I will try to implement that as soon as possible. Also if I should squash commits let me know!


Zykino:
Hello,

Thanks for this big looking contribution (I’m not a maintainer, I’m "just" helping peoples that have questions on Discord).

Not sure how to tell you but… do you know about the swap layouts feature?
* Announcement post: https://zellij.dev/news/stacked-panes-swap-layouts/
* Documentation: https://zellij.dev/documentation/swap-layouts.html

The description I see looks awfully lot like this feature and I fear you work on re-implementing that already exist :sweat_smile: 

PS: I see in the issue you tagged in the title that you also see the similarity. So maybe a comparative description could help?
Note that `zellij action new-tab -l <layout-name>` create a new tab using the named layout, which as I understand it is what the comment you mentioned would need?

pkleineb:
@Zykino Yeah I was also quite confused about how this exactly differs from Swaplayouts. I also dont have much experience with them.

But isnt it that you cannot create new Tabs in Swaplayouts? With this feature it would allow you to create layouts that differ in tab amount and therefore be unique to Swaplayouts.

Also I am not quite sure on how to understand your command down below. Since the command would be called from outside of zellij in the terminal right? This is rather a zellij local keybindable action.

Let me know if I confused something.

PR #3988, opened 2025-02-13
feat(ux): tips on startup

imsnif:
This adds 12 unique Zellij usage tips that will appear when a session start, and can be browsed or dismissed.

It's possible to disable them through the config by adding `show_startup_tips false` to the Zellij configuration, or through the tips themselves by pressing `Ctrl c` - which will add this line to the config for you.

Some samples:

img-2025-02-13-163519
img-2025-02-13-163500
img-2025-02-13-163413


imsnif:
Thanks, I hope to add the ability to customize these in the future!

PR #4015, opened 2025-02-22
Ctrl + p, Ctrl + n to lnavigate lists in sesssion manager up and down, and shift + tab to navigate session manager tabs backwards

nicole-jf:
Allows navigation of lists in session manager wiht Ctrl+p and Ctrl+n (these keyshortcuts are used by default by vim, nvim, helix, zsh and fish, for similar purposes, such as navigating lists of suggested completions, these are established keyshortcuts) which can be more ergonimical than using the up and down arrow keys, and allows using Shift+Tab to navigate the tab bar backwards.

### DIFF
--- a/default-plugins/session-manager/src/main.rs
+++ b/default-plugins/session-manager/src/main.rs
@@ -197,11 +197,19 @@ impl State {
     fn handle_new_session_key(&mut self, key: KeyWithModifier) -> bool {
         let mut should_render = false;
         match key.bare_key {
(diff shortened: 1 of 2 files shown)


jonas-srlabs:
Could we please merge this? It has been open for 1.5 years and would be a great quality-of-life improvement. Ctrl+n and Ctrl+p are muscle memory for many people. However, this should probably be extended to the plugin manager, layout manager, and similar plugins.

PR #4026, opened 2025-02-26
Fix compiler warnings

har7an:
After the recent rust toolchain upgrade, we started getting a bunch of warnings during compilations. It turns out they were mostly harmless, but still not pretty to look at.

This PR removes the current compiler warnings and makes `cargo x build` run warning-free. It does remove a bit of code to get there though. As of right now I've still got an enum in the Client implementation to weed through, where apparently we're not using internal fields in certain variants. I'm currently investigating whether these enum variants are still needed in the first place.

@imsnif What's your stance on deleting "unused" code? In particular the functions `handle_right_mouse_release` and `handle_middle_mouse_release` are only used in tests, for example. We're only handling mouse hold/release for left clicks explicitly. Would you prefer I remove that code entirely, or mark it as `#[cfg(test)]` so it sticks around in case we need it in the future?


imsnif:
Things are still not settled in that area... could we remove the warning?

imsnif:
Actually - I think these unused warnings were introduced by my recent changes with `AnyEvent` and are not something that comes from new rules in the new rust version, right? If so you can just leave them. I'll deal with them when I finalize stuff before the release.

har7an:
*Sorry for the noise, but changes in `main` seemed to add a few things I hadn't caught before.*

> Actually - I think these unused warnings were introduced by my recent changes with AnyEvent and are not something that comes from new rules in the new rust version, right?

I suspect so, but I can't prove it without digging deeper.

In any case, they're all gone now. I saw in your commit message for `AnyEvent` that you're aware that mouse button releasing is currently unfinished? Assuming that is still the case, I marked the `handle_*_mouse_release` functions as `#[cfg(test)]`, so the tests still pass and you don't have to restore the code from the git history.

There's still the mystery around the CliPipe enum variant thingies I'm not quite sure of. It appears to do nothing much on the client side, but removing the code (and all of its appendages) entirely broke a lot of other stuff. So I followed the compilers recommendation and stubbed the `String`s with `()`. Afaict, these enum fields weren't ever read (not even in tests?), and CI at least says there's no harm done here. I'll keep the PR around until tomorrow and see if I can figure something out about that enum, but other than that this is done IMO.

PR #4033, opened 2025-03-01
fix(about): Correct typo.

jlewallen:
Corrects apparent typo in about plugin - "Chagne" vs "Change"

My first contribution here, as meager as it is. Happy for any feedback.


har7an:
Hey, thanks for your contribution and sorry for having you wait so long!

Unfortunately it seems like this change has already been implemented in 4fd0bac67, so I'm closing this PR. If you think that isn't the case, feel free to reopen the PR.

PR #4040, opened 2025-03-04
fix(utils): use correct `MIN_TERMINAL_HEIGHT`

TheLostLambda:
Pulls out `MIN_TERMINAL_*` constants from `zellij-server` into `zellij-utils` so that they can be used everywhere, without relying on comments to keep magic numbers in sync!


TheLostLambda:
I never saw any of the effects of this bug, but I'm sure it'll hurt someone someday...

TheLostLambda:
Yikes, I'll need to figure out how in the word this is breaking a test later...

PR #4051, opened 2025-03-09
fix(stacked): stacked panes can't resize status-bar

TheLostLambda:
Fixed a number of other typos and issues I came across, but the last commit is the one that actually solves the resize issue!

There is another experiment on `tll-pane-resizer-with-stackinfo` for the future when we have <N panes above / below>

I might need @imsnif's help to write a test if one is wanted!

### DIFF
--- a/default-plugins/about/src/tips.rs
+++ b/default-plugins/about/src/tips.rs
@@ -323,7 +323,7 @@ impl Page {
                             }
                     )),
                     ActiveComponent::new(TextOrCustomRender::Text(
-                            Text::new("Press TAB to go to Chagne Mode Behavior")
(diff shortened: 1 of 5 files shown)


TheLostLambda:
I think those e2e tests are unrelated to this change? They seem to be related to the status-bar layout from a skim, but let me know if that's something I need to fix!

imsnif:
Thank you very much for this! I fixed the race in the e2e tests. Otherwise I don't think we need an explicit test for this, as this general area is unfortunately not very well tested. The comment is great so that we'll know what to expect when refactoring this. If you'd like to keep working on this area, I'd be very happy to provide whatever help I can.

PR #4057, opened 2025-03-11
session-manager: ctrl-k keybinding to kill selected session

q60:
i love zellij, but it doesn't love my rather obscure keyboard. on some small keyboards like mine delete key is absent. i can add it to my keyboard using QMK on a different layer, but i don't have any space left for an additional key

to me it seems a great idea to replace `Delete` with more keyboard-friendly `Ctrl-k`, as it is available on every keyboard and in my opinion better *feels* with other `Ctrl` prefixed bindings of session manager

extremely unrelated to this but i've figured there's no way to use system clipboard by selecting text in shell panes/tabs on Kitty terminal without disabling mouse mode first. i am able to copy to system clipboard from helix, but not using `Ctrl-Shift-C` from my shell panes. don't know if it's a but or not. configuring `copy_command`/`copy_clipboard` doesn't help too

meow

### DIFF
--- a/default-plugins/session-manager/src/main.rs
+++ b/default-plugins/session-manager/src/main.rs
@@ -343,7 +343,7 @@ impl State {
                     self.renaming_session_name = Some(String::new());
                     should_render = true;
                 },
-                BareKey::Delete if key.has_no_modifiers() => {
+                BareKey::Char('k') if key.has_modifiers(&[KeyModifier::Ctrl]) => {
                     if let Some(selected_session_name) = self.sessions.get_selected_session_name() {
                         kill_sessions(&[selected_session_name]);
                         self.reset_selected_index();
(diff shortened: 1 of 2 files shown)


Zykino:
`Ctrl k` is so obscure for most peoples. I’m not a maintainer but knowing the philosophy (loosely transcribed from memory):
> Default plugins are here for a default experience that is ok for most peoples. If peoples want configurable bindings (or different ones) they should fork the plugin and customize it to their needs. The aim is for the community to take the burden of maintaining the customized experience while the core devs can focus on the zellij-core (and its plugin’s API).

So I suspect that this PR will not be merged and you are encouraged to fork and customize any plugin as you want.

q60:
i don't see this keybinding as obscure albeit i don't really have hopes for this getting merged, but i do hope for a some sort of discussion. this `Delete` without modifiers looks a bit strange and awkward amongst other bindings with modifiers, in my opinion, especially when i try to imagine using it with a generic keyboard, as this delete key is pretty far away. i had read the contribution guidelines before creating this PR so i don't expect a swift response from maintainers so i guess i'll just wait for now  
  
forking is not a problem for me generally, but only if it's something elaborate i add to forks and no one wants to merge it. in this case it's just a replacement of a single binding, so i suppose don't have any motivation to maintain a fork just for that (that i need to keep upstream and probably resolve conflicts on a regular basis). that would be not a fun job........ idk  
  
i prefer stable version of zellij, so i don't really want to write a flake for it to use my fork, that should provide release versions

PR #4077, opened 2025-03-21
(fix) show more realistic serialization_interval

joshgoebel:
This value is in seconds, not milliseconds.


joshgoebel:
There are other places that list `1`, etc... happy to change more if I was given some direction.  This default of 10000 caused me to lose a lot of time trying to figure out why none of my sessions were serializing. Changing it to `10` worked much better.

joshgoebel:
> (also, the comment right above the line happens to say that they are in seconds).

**Yes, but the example of 10000 seems to contradict this.**  10,000 seconds is nonsensical as a value for this.

If a realistic value harmful in the fixtures?

---

**Thanks for the doc update though!**

PR #4094, opened 2025-03-23
feat(zellij-utils): Add source `client_id` to pipe messages

EmrysMyrddin:
# Proposal
## Description
Adds a the `client_id` of the client being the source of the pipe message.

## Problem

I want to create a plugin that allows to write chars to a specific pane based on it's name.
For this, I want my plugin to react to a pipe message (which can be sent from the CLI or a KeyBind), then tries to find a pane with a matching title in the focused tab.

But, since a pipe message is always sent to all clients, even when it is triggered by a KeyBind, I can't know which tab is focused by the client how sent the KeyBind.

# Solution

After a lot of trial and error, I think that adding the source `client_id` to the `PipeMessage` structure is the minimal way to fix my problem.

With the source `client_id`, my plugin can know from which tab the pipe message was sent by keeping track of which tab each client is currently focusing. 

If it's coming from the CLI, my plugin can look for a `pane_id` argument in the pipe message. The CLI caller can send the current pane id from `ZELLIJ_PANE_ID` env variable.

It's not perfect (the cli source is a bit cumbersome), but it's minimal, and allows for way better behavior in multiplayer mode.

### DIFF
--- a/zellij-server/src/plugins/mod.rs
+++ b/zellij-server/src/plugins/mod.rs
@@ -145,6 +145,7 @@ pub enum PluginInstruction {
     MessageFromPlugin {
         source_plugin_id: u32,
         message: MessageToPlugin,
+        source_client_id: ClientId,
     },
     UnblockCliPipes(Vec<PluginRenderAsset>),
     Reconfigure {
@@ -672,6 +673,7 @@ pub(crate) fn plugin_thread_main(
                         // send to specific plugin(s)
                         pipe_to_specific_plugins(
                             PipeSource::Cli(pipe_id.clone()),
+                            cli_client_id,
                             &plugin_url,
                             &configuration,
                             &cwd,
@@ -690,9 +692,11 @@ pub(crate) fn plugin_thread_main(
                         );
                     },
                     None => {
+                        log::warn!("PIPE MESSAGE TO ALL PLUGINS");
                         // no specific destination, send to all plugins
                         pipe_to_all_plugins(
                             PipeSource::Cli(pipe_id.clone()),
+                            cli_client_id,
                             &name,
                             &payload,
                             &args,
@@ -724,7 +728,14 @@ pub(crate) fn plugin_thread_main(
                     pipe_messages.push((
                         Some(plugin_id),
                         Some(client_id),
-                        PipeMessage::new(PipeSource::Keybind, name, &payload, &args, is_private),
+                        PipeMessage::new(
+                            PipeSource::Keybind,
+                            client_id,
+                            name,
+                            &payload,
+                            &args,
+                            is_private,
+                        ),
                     ));
                 } else {
                     match plugin {
@@ -732,6 +743,7 @@ pub(crate) fn plugin_thread_main(
                             // send to specific plugin(s)
                             pipe_to_specific_plugins(
                                 PipeSource::Keybind,
+                                cli_client_id,
                                 &plugin_url,
                                 &configuration,
                                 &cwd,
@@ -753,6 +765,7 @@ pub(crate) fn plugin_thread_main(
                             // no specific destination, send to all plugins
                             pipe_to_all_plugins(
                                 PipeSource::Keybind,
+                                cli_client_id,
                                 &name,
                                 &payload,
                                 &args,
@@ -768,6 +781,7 @@ pub(crate) fn plugin_thread_main(
                 wasm_bridge.cache_plugin_events(plugin_id);
             },
             PluginInstruction::MessageFromPlugin {
+                source_client_id,
                 source_plugin_id,
                 message,
             } => {
@@ -795,6 +809,7 @@ pub(crate) fn plugin_thread_main(
                         // send to specific plugin(s)
                         pipe_to_specific_plugins(
                             PipeSource::Plugin(source_plugin_id),
+                            source_client_id,
                             &plugin_url,
                             &Some(message.plugin_config),
                             &None,
@@ -819,6 +834,7 @@ pub(crate) fn plugin_thread_main(
                             None,
                             PipeMessage::new(
                                 PipeSource::Plugin(source_plugin_id),
+                                source_client_id,
                                 message.message_name,
                                 &message.message_payload,
                                 &Some(message.message_args),
@@ -834,6 +850,7 @@ pub(crate) fn plugin_thread_main(
                             None,
                             PipeMessage::new(
                                 PipeSource::Plugin(source_plugin_id),
+                                source_client_id,
                                 message.message_name,
                                 &message.message_payload,
                                 &Some(message.message_args),
@@ -845,6 +862,7 @@ pub(crate) fn plugin_thread_main(
                         // send to all plugins
                         pipe_to_all_plugins(
                             PipeSource::Plugin(source_plugin_id),
+                            source_client_id,
                             &message.message_name,
                             &message.message_payload,
                             &Some(message.message_args),
@@ -955,6 +973,7 @@ fn populate_session_layout_metadata(
 
 fn pipe_to_all_plugins(
     pipe_source: PipeSource,
+    from_client_id: u16,
     name: &str,
     payload: &Option<String>,
     args: &Option<BTreeMap<String, String>>,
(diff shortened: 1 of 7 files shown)


imsnif:
Hey, thanks for this and good find. I think it makes sense all in all, but I'd rather not add the cli_client_id, since it's misleading (these clients are usually extremely short lived and don't have many of the other guarantees clients have). We could do this by making the field optional in the protobuf declaration.

Also, if you'd like you can add the `PaneId` (also as an optional field) to the message coming from the CLI to perform the tracking you suggested.

EmrysMyrddin:
@imsnif I've updated the PR to only send the `client_id` when it's not a CLI client.
I've also added the `pane_id` as you suggested, so that Plugin devs don't have to manually track the client focused pane just to know which pane was focused when a KeyBind was pressed.

Since I'm already modifying the protobuf, perhaps we can merge #4095 into this PR ? Since it's already a breaking change, we can take the opportunity to clean up this.

imsnif:
Hey, from a cursory look this seems fine - thanks! I appreciate the thought with the protobuffs in xtask, but I'd rather not conflate things. Can we remove it for now?

As for the `PaneId` - if we just add stuff to protobuf files, it's not a breaking change. It's only a breaking change if we change stuff. As I mentioned in the previous PR, changing stuff around might or might not be a breaking change (might also be a change in some implementations and not in others) - but I'd really rather not spend time on this to find out. My resources are limited and I try to spend very deliberate time on changes/additions that will benefit users/plugin developers. Thanks for understanding!

EmrysMyrddin:
Ok I see :-) Actually It seems I can use the `PaneId` defined in `event.proto`, so it's not mandatory to move things.

For the xtask thing, sure I will remove it. Do you want me to open a new PR for it ?

After testing, it seems the PaneId is not set for `KeybindPipe` action. Do you know why ? Should I change this too, or you prefere the `pane_id` to be set only for the `CliPipe` action ?

EmrysMyrddin:
@imsnif May I ask you if you have time to take a new look at this PR ?

imsnif:
My apologies, I am juggling a lot of things and am merely Human. Best bet is to try to make this change as small and concise as possible so that I can review and merge it in one go. I'll try to take a look in the next few weeks. Thanks for understanding.

EmrysMyrddin:
No problem, I fully understand that :-) Maintaining alone this kind of project is not an easy task!

I've cleaned up the PR. The changes should be as minimal as I can think about.

EmrysMyrddin:
@imsnif Apologies for the ping here, I've rebased my PR and made sure it includes only the most minimal changes possible.

It's just passing adding the `client_id` where as parameter so that it can be passed around :-) 
Do you think you could have a chance to review this ?

zellij-server/src/plugins/mod.rs
```diff
@@ -719,9 +724,12 @@ pub(crate) fn plugin_thread_main(
                         );
                     },
                     None => {
+                        log::warn!("PIPE MESSAGE TO ALL PLUGINS");
```
imsnif: Let's please remove this

zellij-server/src/plugins/mod.rs
```diff
@@ -754,14 +763,24 @@ pub(crate) fn plugin_thread_main(
...
+                            is_private,
+                        ),
                     ));
                 } else {
                     match plugin {
                         Some(plugin_url) => {
                             // send to specific plugin(s)
                             pipe_to_specific_plugins(
                                 PipeSource::Keybind,
+                                Some(cli_client_id),
```
imsnif: I think this should be a `None`. Even though in this case cli_client_id is will always be None (since this is coming from a keybinding), we shouldn't re-use it in case this behavior somehow changes in the future. Then, if we want to add the cli_client_id we'll do it explicitly.

zellij-server/src/plugins/mod.rs
```diff
@@ -785,6 +804,8 @@ pub(crate) fn plugin_thread_main(
                             // no specific destination, send to all plugins
                             pipe_to_all_plugins(
                                 PipeSource::Keybind,
+                                Some(cli_client_id),
```
imsnif: Same here

zellij-server/src/plugins/unit/plugin_tests.rs
```diff
@@ -6191,6 +6191,7 @@ pub fn unblock_input_plugin_command() {
     std::thread::sleep(std::time::Duration::from_millis(500));
 
     let _ = plugin_thread_sender.send(PluginInstruction::CliPipe {
+        pane_id: None,
```
imsnif: Let's maybe make one of these a `Some` for the sake of completeness in the tests?

zellij-utils/src/plugin_api/pipe_message.proto
```diff
@@ -10,6 +12,8 @@ message PipeMessage {
     optional string payload = 5;
     repeated Arg args = 6;
     bool is_private = 7;
+    optional uint32 clientId = 8;
+    optional api.event.PaneId paneId = 9;
```
imsnif: These should be snake case

imsnif (changes requested):
Thank you for your patience! I'll get the CI to run the tests and I also left some comments.

zellij-server/src/plugins/unit/plugin_tests.rs
EmrysMyrddin: So just one with a fake `PaneId::Terminal` ?
Or perhaps tow, one `PaneId::Terminal` and one `PaneId::Plugin` ?

EmrysMyrddin:
Thanks for the review! 
I've addressed all the comments and fixed the failing CI 🙂

EmrysMyrddin:
By the way I don't know if you want me to clean up the commit history or if you usually squash on merge?

imsnif:
Hey @EmrysMyrddin - thank you for your endless patience!

Is this still relevant for you? If so I have some dedicated time to get community PRs merged in the next couple of weeks. There have also been some changes in main that should make this change much easier. In particular we're not tracking client_ids for cli_pipes natively, eg. here: https://github.com/zellij-org/zellij/blob/main/zellij-server/src/route.rs#L1744

LMK if you want to keep working on getting this merged.

EmrysMyrddin:
Hi!

>thank you for your endless patience!
Ahah no problem 🙂 You are already giving a lot to the community by making zellij!

Yes, I would still be happy to work on this :-) I still have the project to use this for Helix interoparibility with other tools (waiting for them to have a proper plugin system, which will probably not happen any time soon)

> In particular we're not tracking client_ids for cli_pipes natively, eg. here: [main/zellij-server/src/route.rs#L1744](https://github.com/zellij-org/zellij/blob/main/zellij-server/src/route.rs?rgh-link-date=2026-01-14T07%3A48%3A12Z#L1744)

I'm not sure to understand what you mean?

imsnif:
Ah, sorry - typo. I meant "now tracking..." As in the cli_client_id send to the plugin thread should always be populated for pipes. I remember there were some difficulties with that in this PR.

If I remember wrong though, all good. Would you like to rebase to main and maybe help me out with a short summary of where things stand?

EmrysMyrddin:
Hi @imsnif! Sorry, I didn't have much time to work on this lately.

I've made a rebase, it should now be mergeable :-)

As for the `cli_client_id`, you made the following comment: https://github.com/zellij-org/zellij/pull/4094#discussion_r2375838735

From what I understand of the code, as of today, I should revert the change that removed the `Some(cli_client_id)` as the `source_client_id` in the case of a `KeyBind` event. Because now, this `cli_client_id` should pretty much always makes sense ?

imsnif:
Hey! So, in short what needs to happen:

1. When piping from the CLI, we need to include the PaneId and not the ClientId (the latter will always be wrong as it's the CLI's client id)
2. When piping from a keybinding, we need to include the ClientId and not the PaneId (as here the latter is currently always None - since we might change it in the future for one reason or another, I'd rather we explicitly pass `None` rather than passing the `PaneId` from `route` which is currently always None but might change in the future)

Makes sense?

EmrysMyrddin:
Ok, I think I see the idea!

Not sure to fully get why you wouldn't want to forward the `paneId`. I mean yes, today it is always `None` in the case of a keybinding. But if in the future we make the change so that the `PaneId` is actually tracked and set when a keybind occurs, it will make our code works without needing any modifications.

You fear that the `paneId` could be misleading in the future and could in this scenario represent something else than the paneId from where the keybind occured ?

imsnif:
Fair question. From my point of view: I am unable to tell the future or what we'll be using this field for if we decide to use it. I'd rather the person implementing that change will make an explicit decision than have an implicit one unknowingly made for them.

PR #4095, opened 2025-03-23
fix(zellij-utils): Fix duplication of `PaneId` in protobuf

EmrysMyrddin:
# Problem

The `PaneId` message in the proto file has been duplicated to avoid import cycles.

While this works, it's not ideal to have duplicated code like that.

And you actually have to make a new copy of it each time you want to use a `PaneId` in a proto file.

# Solution

Since the `PaneId` and `PaneType` only have native types, they don't have to import any other proto file, the import cycle can actually be resolved by moving those message definition in there own file.

This is simpler, and also allows to use the `PaneId` type anywhere it is needed without duplicating it again.

### DIFF
--- a/zellij-utils/assets/prost/api.event.rs
+++ b/zellij-utils/assets/prost/api.event.rs
@@ -102,7 +102,7 @@ pub struct ClientInfo {
     #[prost(uint32, tag = "1")]
     pub client_id: u32,
     #[prost(message, optional, tag = "2")]
-    pub pane_id: ::core::option::Option<PaneId>,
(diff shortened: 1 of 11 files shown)


imsnif:
Thanks for this, but I think any change to protobuf files might be considered breaking at least in certain implementations. While I'm not 100% sure of this, I don't feel this warrants the risk.

EmrysMyrddin:
Oh, you think the name of the message changes something ?
Do you want me to go verify this ? Or you don't want to touch this anyway ?

Issue #3337, opened 2024-05-08
Restore selected tab, pane, and cursor position when attaching to a Zellij session


rockboynton:
I get disconnected from my Zellij session often (ssh connection drops). Zellij does a great job at restoring basically all of my session tabs, layout, even content. However, thing it is missing is to restore the last selected tab, pane, and cursor position. 

Currently each time I reattach it takes me to the first tab I have open, and I have to navigate back to where I was working which is an inconvenience. I think this would be a very nice feature suitable for core. 

If you can point me in the right direction for where state is serialized and stored/restored, I would even be happy to take this on myself!

Apologies if this request already exists; I searched but didn't find a similar issue.

PR #4115, opened 2025-04-01
feat(zellij-server): restore tab position when re-attaching

### LINKED ISSUE #3416: Zellij attach doesn't remember last opened pane/tab
If I launch a Zellij session, and then disconnected, when I reconnect to that Zellij session afterwards (eg.. `zellija a), it doesn't remember my last opened pane/tab  etc.

For example, if I have say 12 tabs open, and I'm opened on tab 7 - if I get disconnected, or intentionally disconnect (e.g. with `zellij d`) when I re-open it just opens up tab 1 by default.

(I know Zellij isn't screen/tmux etc, but from memory, most of those will remember the last opened position etc. I also find it more intuitive/helpful, for it to just resume your session as it was.)

XOR-op:
Fix #3416 #3337 .

This PR focuses on improving user experience: when users detach all clients, the next time users attach to the session, they expect to see the last tab they were at from the last connection, instead of always being the first tab. The changes in this PR allow a session to remember the last tab position on exit globally, and restore to it. When there is more than zero active client, no behavior will be changed.

### DIFF
--- a/zellij-server/src/screen.rs
+++ b/zellij-server/src/screen.rs
@@ -672,6 +672,7 @@ pub(crate) struct Screen {
     connected_clients: Rc<RefCell<HashSet<ClientId>>>,
     /// The indices of this [`Screen`]'s active [`Tab`]s.
     active_tab_indices: BTreeMap<ClientId, usize>,
+    global_last_active_tab_index: usize,
     tab_history: BTreeMap<ClientId, Vec<usize>>,
     mode_info: BTreeMap<ClientId, ModeInfo>,
     default_mode_info: ModeInfo, // TODO: restructure ModeInfo to prevent this duplication
@@ -740,6 +741,7 @@ impl Screen {
             style: client_attributes.style,
             connected_clients: Rc::new(RefCell::new(HashSet::new())),
             active_tab_indices: BTreeMap::new(),
+            global_last_active_tab_index: 0,
             tabs: BTreeMap::new(),
             overlay: OverlayWindow::default(),
             terminal_emulator_colors: Rc::new(RefCell::new(Palette::default())),
@@ -1489,6 +1491,8 @@ impl Screen {
             self.active_tab_indices.iter().next()
         {
             *first_active_tab_index
+        } else if self.tabs.contains_key(&self.global_last_active_tab_index) {
+            self.global_last_active_tab_index
         } else if self.tabs.contains_key(&0) {
             0
         } else if let Some(tab_index) = self.tabs.keys().next() {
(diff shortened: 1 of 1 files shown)


XOR-op:
Hi @imsnif,

I understand review of PRs will be delayed, but this PR has been waiting 7 months without feedback and currently in the second page of PR list, lacking visibility. It's a 5-line fix for an issue affecting many users. It should not create additional burdens for you; however, this is important for enhancing the user experience for many users (see related issues).

Would appreciate any feedback when you have a moment.

imsnif:
Hi @XOR-op - thanks for the reminder. I can understand the frustration of having this going without feedback for a long while. This is not how I would like things to be handled.

Having said that, there are quite a few similar seemingly minor enhancements/fixes in a similar state. Many of them are similarly beneficial to user experience and seem similarly innocuous. All of them, including this one, are not trivial five-minute fixes/reviews for various reasons (an example in this case is that the aforementioned client_ids used as keys don't always refer to actual users and might cause havoc when combined with certain plugins/cli commands in certain situations).

I am aware of this issue and this fix - I promise I have not forgotten about it, and I promise to get to it. I cannot however promise an ETA. I'm doing my best maintaining, developing and directing this project and appreciate both your patience and trust in my prioritization of maintainer attention.

XOR-op:
@imsnif Thanks for all the efforts you put for this project! I will wait until our maintainers have enough time. One request I have is could you approve CI workflow for this PR? Given non ETA is promised, I may need to catch up with the upstream. Allowing me to fix issues introduced by updates can reduce the burdens of reviewers when they have time.

imsnif:
Hey @XOR-op - I'm taking a look at this, and maybe I misunderstand something, but this doesn't seem to be working.

What I'm doing:
1. Start a new session, start a new tab and move to it
2. Detach with `Ctrl o` + `d`
3. Reattach with "zellij attach <session-name>" -> still focused on tab 1.

Any clues?

XOR-op:
Hi @imsnif - That sounds strange. Today I rebased my commit to the latest main branch, and tested it again on both macOS and Linux. On both platform it works as intended.

What I'm doing:
0. Rebase my commit to the current main branch. No conflict there.
1. Build from source, and then kill the existing zellij daemon process given the handling logic is on the server side.
2. Run `zellij attach -c test`, a new session is created with tab 1
3. Create a new tab, now there are tab 1 and tab 2, and zellij is showing tab 2 now.
4. Detach with `Ctrl o`+`d`
5. Reattach again with `zellij attach -c test`. Now it is focused on tab 2.

Is the above procedure what you went through? I guess the most possible reason is you may connect the session to an older server. A less possible reason is that the implementation is flaky, while I didn't observe a flaky behavior in my testing. Let me know if I can do anything that helps you solve the issue.

imsnif:
My apologies! I was testing the wrong branch. This works great.

Thank you!

PR #4118, opened 2025-04-03
fix(deps): fix --locked build on rust 1.86

tranzystorekk:
Noticed when rebuilding 0.42.1 for rust 1.86.

The cause is https://github.com/bytecodealliance/wasmtime/issues/10184, and the fix is updating the `wit-parser` dependency to 0.221.3

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4084,7 +4084,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "c17a3bd88f2155da63a1f2fcb8a56377a24f0b6dfed12733bb5f544e86f690c5"
 dependencies = [
  "leb128",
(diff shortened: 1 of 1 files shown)


imsnif:
Thanks for this. I haven't tested this yet, but even if we merge this I wouldn't want to release a version with this yet. We already have far-reaching changes in main and back-porting this change to a patch version will be a bit of work. Can you patch things from your side until the next release?

tranzystorekk:
Yeah, patching on the void side is no problem, i made this pr in case mainly for anyone that tests the --locked build on main branch, and to make the problem/solution known :)

imsnif:
I just merge this and will test-drive the unreleased changes (mostly because we already had stuff in `main`) for a week or so before releasing a patch. Thanks again.

PR #4194, opened 2025-05-15
fix(theme): fix tokyo-night-storm selection bg

imsuck:
Taken from https://github.com/mbadolato/iTerm2-Color-Schemes/blob/f3d53cc/alacritty/tokyonight-storm.toml.


imsuck:
I just realized it doesn't quite match up with tokyo-night-dark selection (which is more grey than blue). Is there any problem with this?

Issue #3992, opened 2025-02-15
Runtime Configuration not picked up when config.kdl is a symlink


steveclarke:
## Issue description

I use [stow](https://www.gnu.org/software/stow/) for managing my dotfiles. Stow symlinks my configuration files into `./config/zellij/config.kdl`. I wanted to try out the new dynamically reloading configuration changes but it wouldn't work for me.

### Works
Create a `./config/zellij.kdl` file directly in the config folder. Open the config file and change the theme. You will see the theme change immediately.

### Doesn't Work
Create `/tmp/zellij.kdl` and symlink it to `./config/zellij.kdl`. Open the config file and change the theme. You will not see the change reflected until you quit and re-start Zellij.

marvin-yorke:
I can confirm, same configuration (symlinks managed by `stow`), same issue

mrmarkwell:
I have the same issue (using stow for dotfiles) and this is still happening on 0.42.2.

airtonix:
same don't want to copy my config to the place with the stuff to do the things

sestrella:
I have the same issue for the `~/config/zellij/config.kdl` file generated by Home Manager, which is a symlink to the Nix store.

PR #4205, opened 2025-05-20
fix: support symlinks on runtime config reload

simifalaye:
Fixes: #3992 

Currently, the config auto-reload feature can't detect symlinks. Users may use symlinks with applications such as gnu-stow so config file auto-reload should detect these changes by default. I have changed the code to resolve the symlink if it is one and then watch that file instead.

### DIFF
--- a/zellij-client/src/lib.rs
+++ b/zellij-client/src/lib.rs
@@ -10,6 +10,7 @@ mod stdin_handler;
 
 use log::info;
 use std::env::current_exe;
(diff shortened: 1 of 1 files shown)


Lockszmith-GH:
I guess you never read #3794

simifalaye:
@Lockszmith-GH No, I had not seen that. Thanks for letting me know, I'll keep that in mind for any future contributions to this project.

PR #4208, opened 2025-05-24
fix(themes): improve solarized-dark contrast and color consistency

carlocaione:
The solarized-dark theme has several issues:

- The selection for tables and lists is not visible
- It is using base values that should not be present in the dark version of the theme
- The normal relationship for background and body text is base03:base0 and base02:base1 for highlighted content
- Contrast is totally wrong for some elements

This is making the appearance of the solarized-dark theme very inconsistent and wrong.

To fix:

- Standardize text color to base1 (147 161 161) for better readability
- Fix ribbon_unselected background from base3 to base1 for dark theme consistency
- Improve selection backgrounds using base02 for better contrast

Before:

Screenshot 2025-05-24 at 09 53 52

After:

Screenshot 2025-05-24 at 09 54 25

### DIFF
--- a/zellij-utils/assets/themes/solarized-dark.kdl
+++ b/zellij-utils/assets/themes/solarized-dark.kdl
@@ -1,7 +1,7 @@
 themes {
     solarized-dark {
         text_unselected {
-            base 238 232 213
+            base 147 161 161
             background 7 54 66
             emphasis_0 203 75 22
             emphasis_1 42 161 152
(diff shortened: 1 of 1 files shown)


carlocaione:
Is there any interest in this patch?

Zykino:
I’m sure there is interrest in this PR but as the [CONTRIBUTING](https://github.com/zellij-org/zellij/blob/main/CONTRIBUTING.md) state:

> If you're still eager to contribute minor fixes, please note that we might take a long while to get to them.

I suspect this sort of PR that only modify theme related aspects of the code will only be read and merged when the maintainer decide to tidy up for a release. (Those are usually not frequent, I would say about between 2 and 4 time a year? No guarantee that old cadence will be maintained.)

PR #4222, opened 2025-06-03
fix(ui): properly redraw boundaries when colors change

### LINKED ISSUE #3373: Can't style unfocused panes with Zellij

CleanShot 2024-05-24 at 12 22 46

Unfocussed pane borders are always white with no control.

**Basic information**
`zellij --version`: zellij 0.40.1
`stty size`: n/a

## Further information
```
themes {
  custom {
          fg "#000000"
          bg "#000000"
          black "#000000"
          red "#000000"
          green "#000000"
          yellow "#000000"
          blue "#000000"
          magenta "#000000"
          cyan "#000000"
          white "#000000"
          orange "#000000"
  }
}
```

## Issue description
I tried this in 3 terminals. In alacritty terminal, I noticed that the white color was controlled by the terminal theme.

## Minimal reproduction
Set all the theme colors to black. Observe white border.

## Other relevant information
Really annoying, because the only thing I wanted to change was make the unfocussed pane less prominent 🙃

nh:
Fixes an issue where pane border colors wouldn't update correctly when switching focus between panes. This enables the frame_unselected theme property to work properly and also fixes mouse hover effects on inactive panes.

The problem was in the boundary rendering logic that would skip redrawing boundaries if a boundary already existed at that position, even if the color had changed. Now boundaries are redrawn whenever any property (including color) changes.

Fixes #3373

### DIFF
--- a/zellij-server/src/ui/boundaries.rs
+++ b/zellij-server/src/ui/boundaries.rs
@@ -584,20 +584,37 @@ impl Boundaries {
     ) -> Result<Vec<CharacterChunk>> {
         let mut character_chunks = vec![];
         for (coordinates, boundary_character) in &self.boundary_characters {
-            let already_on_screen = existing_boundaries_on_screen
-                .and_then(|e| e.boundary_characters.get(coordinates))
-                .map(|e| e == boundary_character)
-                .unwrap_or(false);
-            if already_on_screen {
-                continue;
+            let should_render = if let Some(existing) = existing_boundaries_on_screen {
+                existing
+                    .boundary_characters
+                    .get(coordinates)
+                    .map(|existing_boundary| existing_boundary != boundary_character)
+                    .unwrap_or(true)
+            } else {
+                true
+            };
+
+            if should_render {
+                character_chunks.push(CharacterChunk::new(
+                    vec![boundary_character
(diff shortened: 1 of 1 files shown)


imsnif:
Thanks for putting this together. Can you please offer a step by step reproducer to the issue you're trying to fix?

nh:
Sure I can do that tomorrow. Essentially it's exactly as described in issue #3373, unfocused pane borders are always white (or inherited from the terminal foreground color, which is usually white).

I thought this might be a known issue, as none of the built in themes specify frame_unselected. When it is specified, the unfocused panes will use the color from the theme - but the colors do not update correctly as you move between panes. This patch fixes the rendering behavior.

imsnif:
I haven't looked too deeply into this issue yet (it was on my TODO list for next version), but any fix for it will have to also include the coloring of the pane frames (which are the default). From a brief look at this PR, I think it doesn't address it. Would you like to see this through?

nh:
Ok it turns out I'm an idiot and it's already working properly in v0.43, my 'fix' was doing nothing. Sorry! I've updated the built in themes with "frame_unselected" values, I'll submit that as a separate PR.

PR #4223, opened 2025-06-04
fix(themes): add missing frame_unselected values to all themes

nh:
Zellij v0.43 now works correctly with "frame_unselected" theme colors. Updated the built in themes to set these colors - derived from legacy theme "bg" value, and if not available, created by darkening the "frame_selected" base color for dark themes, and lightening it for light themes.

### DIFF
--- a/zellij-utils/assets/themes/ansi.kdl
+++ b/zellij-utils/assets/themes/ansi.kdl
@@ -77,6 +77,14 @@ themes {
             emphasis_2 2
             emphasis_3 5
         }
+        frame_unselected {
+            base 8 8 8
+            background 0
+            emphasis_0 0
+            emphasis_1 0
+            emphasis_2 0
+            emphasis_3 0
+        }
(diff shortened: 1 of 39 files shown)


zellij-utils/assets/themes/ansi.kdl
```diff
@@ -77,6 +77,14 @@ themes {
             emphasis_2 2
             emphasis_3 5
         }
+        frame_unselected {
+            base 8 8 8
```
arkriny, line 81: Shouldn't this be "7"? Since it's an ANSI theme

imsnif:
Beautiful, thank you! You saved me a lot of work. I do think there needs to be a bit more contrast between the selected and unselected frame colors. Ideally they should not be in the same color. In addition, as @skewb1k mentioned, in the `ansi` theme I think the color should be just `7`. Would you like to make these changes @nh ?

PR #4225, opened 2025-06-06
feat(ui): add optional tooltip with key hints to the compact bar

imsnif:
https://github.com/user-attachments/assets/27b9a5ad-1188-4d32-9efd-0b59de018612

This adds a configuration to the `compact-bar` that will allow us to have a tooltip displayed with the available keys for our current mode (since the compact bar does not show this).

The tooltip will be displayed:
1. When we switch to a mode that is not our base mode (normal or locked, depending on our configuration)
2. When we toggle it with the configured key (eg. `F1`). In this case, it will also appear in modes where it does not normally appear (eg. base mode, rename tab, rename pane, etc.) This key can also hide the bar when it is shown.

Example of how to enable this in the Zellij configuration:

```kdl
plugins {
    // ...
    compact-bar location="zellij:compact-bar" {
      tooltip "F1"
    }
    // ...
}
```


elythh:
awesome feature, thank you !

c02y:
Nice feature, but

I got issue:
1. Should have mentioned to use `default_layout "compact"`
2. `F1` opens a new pane which cannot be closed 
3. `F1` again, it opens another smaller floating pane which cannot be closed 

https://github.com/user-attachments/assets/df3da670-436b-40dd-924f-f809e8e70e51

PR #4150, opened 2025-04-19
reencode termwiz keycodes as current_buffer isnt reliable

kewaopcode:
It's not reliable because current_buffer can sometimes contain multiple instructions at once, which are mistakenly treated as a single one. For example, if it holds [KeyCode, MouseMove], the parser might interpret it as one large KeyCode and print it directly to the terminal, leading to issues like this: https://github.com/zellij-org/zellij/issues/932

PR #4266, opened 2025-07-04
fix(input): properly handle cases where STDIN is fragmented

imsnif:
This fixes cases where due to a fragmented STDIN (eg. ssh connections) we would occasionally not handle events properly and instead print them to screen (eg. mouse movements, focus changes, etc.)

This fixes it by buffering STDIN if we detect partial ANSI instructions. While not a perfect solution (eg. in non-kitty keyboard mode `Alt [` is ambiguous - it could signify a fragmented sequence or just the keybinding - we treat it as the key), anything we miss should be extremely rare. In any case, we timeout our buffering after 50ms.


imsnif:
I'm closing this in favor of work I'm doing here, which I think is more robust and makes a lot more sense than this guesswork: https://github.com/zellij-org/zellij/pull/4150#issuecomment-3045559491

Issue #4023, opened 2025-02-25
Cwd does not update as expected


cipherquax:
**Basic information**

`zellij --version`: 0.41.2

`stty size`: 29 104

`uname -av` or `ver`(Windows): Linux fedora 6.12.15-200.fc41.x86_64

## Issue description

After using Resurrection mechanism (but also after load the layout from the file), the current directory does not seem to be saved correctly.

Here is a step-by-step reproduction:

1. Create a new session and enter directory `x`.
2. Use the shortcut `Ctrl+Q` to quit the session.
3. Reattach to the session you just left, and you will still be in directory `x`.
4. Now change the directory to `y`.
5. Use `Ctrl+Q` again and reattach to the session.
6. Upon reattachment, you will still be in directory `x`, not `y`.

## Minimal reproduction

- Create a layout file (e.g., layout.zellij) with the following content:

```
layout {
    tab name="Tab 1" {
        pane size=1 borderless=true {
            plugin location="zellij:tab-bar"
        }
        pane cwd="/x"
    }
}
```
- Launch Zellij using the custom layout:

> zellij --layout layout.zellij

- Within the session:
  - Verify that you are in directory /x.
  - Change the directory to /y.
  - Optionally, run a command such as pwd to confirm you are in /y.

- Execute command:

> zellij action dump-layout | grep cwd

- Observation: The output still shows cwd="/x", even though the working directory was changed to /y.

## Other relevant information

- If you simplify the layout to only:

```
layout {
    tab name="Tab 1" {
        pane cwd="/x"
    }
}
```
the issue does not occur. This indicates that the inclusion of an extra pane (such as the plugin pane) may be related to the bug.

Guria:
I have cwd mostly ignored in most cases when loan kdl layout file.
```
zellij --version
zellij 0.43.1
```

Guria:
works well only if i have `cwd "/absolute/path"` on layout defined

azinsharaf:
i have the same issue. could you please share your layout? how did you define `cwd` there?

Guria:
Mine is shared in ticket description.
Also most likely it shares the root cause with https://github.com/zellij-org/zellij/issues/4689#issuecomment-4153862810

Issue #3374, opened 2024-05-24
tabs / panges lose their working directory on restore after a reboot


n8henrie:
## Issue description

I am new to zellij. I have about 15 named tabs in a named session (`main`). Each tab has a working directory for a project (from which the tab name is derived). I have been able to disconnect and reconnect to this session and am glad to see this works well!

I just had to reboot my machine, and upon resurrecting this session, I found that about half of the tabs had retained their name, but their working directory had been changed to a different directory (`~/git/espanso` in this case, I doubt the specific directory is relevant except that it was not `$HOME` or `/` or something empty). Oddly, the other half of the tabs had retained their working directory.

## Minimal reproduction

Somewhat difficult since it requires a full machine reboot, and I just got done changing all of the tabs back to the proper directory. I will report back if this recurs though.

## Other relevant information

```console
$ zellij --version
zellij 0.40.1
$ sw_vers
ProductName:            macOS
ProductVersion:         14.5
BuildVersion:           23F79
```

Is there a user-facing way for me to inspect the cached files -- I assume in `~/Library/Caches/org.Zellij-Contributors.Zellij` -- to see if there is something that seems off? All I see is a bunch of UUIDs, so I'm not entirely sure where to look.

Dialga:
Following this, my cwd in open tabs also aren't being restored.

n8henrie:
Continues to be an issue.

While the cache file is itself being updated, It doesn't seem that the CWD is being updated in the cache file.

As one can see, I'm currently in session `main`, in a tab called `icwasm`, which has only one pane.

<img width="407" alt="Screenshot 2024-07-15 at 06 01 38" src="https://github.com/user-attachments/assets/9baad930-608f-4af8-aab4-7882967575ed">

From here, one can see that the session layout file *is* being updated every second, but the reported cwd for this tab / pane is wrong:

```console
$ date; stat ~/Library/Caches/org.Zellij-Contributors.Zellij/0.40.1/session_info/"${ZELLIJ_SESSION_NAME}"/session-layout.kdl
Mon Jul 15 06:03:20 MDT 2024
  File: /Users/n8henrie/Library/Caches/org.Zellij-Contributors.Zellij/0.40.1/session_info/main/session-layout.kdl
  Size: 13363           Blocks: 32         IO Block: 4096   regular file
Device: 1,22    Inode: 688103593   Links: 1
[...]

Issue #3041, opened 2023-12-29
Resurrection seems to stop changing CWD after reattaching


Parry-97:
## Issue description

As the title implies, the serialized CWD after resurrection does not change.
before reattaching the directory changes ok and after reattaching it stops.
  
## Minimal reproduction

1. I start off inside a test folder,
2. In here I recreated the default config with zellij setup --dump-config >> basic.kdl
3. I edited it to include serialization_interval 1 (no other changes)
4. I created a session named example with zellij --config basic.kdl a example
5. I move to a different directory called first
6. I quit from zellij
7. I reattach to the example session hence resurrecting it
8. I can see that the working directory is first
9. I move to another directory called second
10. I quit from zellij
11. I reattach to the example session
12. The working directory is still first
13. I move to another directory called third
14. I quit from zellij
15. I reattach to the example session
16. The working directory is still first

## Other relevant information

Various `zellij action dump-layout` for troubleshooting:

1. The result of `dump-layout` when starting Zellij:
```
layout {
    cwd "/home/pops/test"
    tab name="Tab #1" focus=true hide_floating_panes=true {
        pane size=1 borderless=true {
            plugin location="zellij:tab-bar"
        }
        pane command="zellij" focus=true {
            args "action" "dump-layout"
            start_suspended true
        }
        pane size=2 borderless=true {
            plugin location="zellij:status-bar"
        }
    }
    new_tab_template {
        pane size=1 borderless=true {
            plugin location="zellij:tab-bar"
        }
        pane
        pane size=2 borderless=true {
            plugin location="zellij:status-bar"
        }
    }
    swap_tiled_layout name="vertical" {
        tab max_panes=5 {
            pane size=1 borderless=true {
                plugin location="zellij:tab-bar"
            }
            pane {
                pane split_direction="vertical" {
                    pane
                    pane {
                        children
                    }
                }
            }
            pane size=2 borderless=true {
                plugin location="zellij:status-bar"
            }
        }
        tab max_panes=8 {
            pane size=1 borderless=true {
                plugin location="zellij:tab-bar"
            }
            pane {
                pane split_direction="vertical" {
                    pane {
                        children
                    }
                    pane {
                        pane
                        pane
                        pane
                        pane
                    }
                }
            }
            pane size=2 borderless=true {
                plugin location="zellij:status-bar"
            }
        }
        tab max_panes=12 {
            pane size=1 borderless=true {
                plugin location="zellij:tab-bar"
            }
            pane {
[...]

Issue #2978, opened 2023-12-03
CWD never changes when syncing the session file


Nukiloco:
Thank you for taking the time to file this issue! Please follow the instructions and fill in the missing parts below the instructions, if it is meaningful. Try to be brief and concise.

**Basic information**

`zellij --version`:
zellij 0.39.0
`stty size`:
28 72

`uname -av` or `ver`(Windows):

List of programs you interact with as, `PROGRAM --version`: output cropped meaningful, for example:
`nvim --version`: NVIM v0.5.0-dev+1299-g1c2e504d5  (used the appimage release)
`alacritty --version`: alacritty 0.7.2 (5ac8060b)

**Further information**
Reproduction steps, noticeable behavior, related issues, etc

Nukiloco:
So it looks like I can change the cd on a new session, however if I quit from it, and then attach back to it, changing the directory will no longer save. Note that these new sessions are not using symlinks.

Nukiloco:
I changed the title of this to more reflect of whats currently happening. I guess I will make another bug report about the tab information being randomly wiped for no reason?

towry:
Happens on zellij 0.39.2

plotnw:
I can reproduce this on version 0.40.0 as well.

jelenv:
It's happening to me on 0.40.1

febrianamir:
Looks like I have to quit the session by detaching from it (`Ctrl+o d`) to save the current working directory and log history.

PR #4287, opened 2025-07-12
fix: populate layout metadata with the right cmd and cwd

vincentrou:
Add cmd from sysinfo in terminal_ids_to_commands if no cmd are return by ps. So then, cmd and cwd are udpated in "session-layout.kdl" even if no cmd is runing in pane.

Fix related issue : #2978, #3041, #3374, #4023


imsnif:
Great fix, thank you!

PR #4288, opened 2025-07-14
Add gruber-darker theme

haxfn:
Based on gruber-darker theme colors from github.com/rexim/gruber-darker-theme/


imsnif:
Thank you!

PR #4301, opened 2025-07-18
feat: close unfocused panes action

### LINKED ISSUE #2756: How to close other panes or tabs??
many window manager support closeOtherPane or closeOhterTab action to close unActive panes. 

i cant find how to do this with zellij. 

with zellij action , i cant do this , 

with zellij plugin, i think it can do this,  but i dont known how to provide a Action like Plugin_CloseOhter, i cant find this in plugin doc

will zellij implement this future?  

By the way , How to move a pane to new tab or to exist tab, i also cant do it now.

chasinglogic:
This adds a new action to close the unfocused panes similar tmux's kill-pane -a option.

Fixes #2756 at least for panes.


chasinglogic:
Converting to draft while I add tests for this, but would appreciate feedback on if this feature is acceptable or not.

abloch:
please also consider looking at #4594

PR #4334, opened 2025-08-03
fix(terminal): handle omitted flags in kitty keyboard protocol push sequence

### LINKED ISSUE #4333: Kitty keyboard protocol: `CSI > u` with omitted flags fails to set flags to 0
NOTE: Description written with LLM assistance.  Bug found by human frustration :)

## Issue description

Zellij fails to implement the push operation for the Kitty keyboard protocol's progressive enhancement stack, specifically the `CSI > u` sequence.  When flags are omitted in the push sequence, Zellij does not default to 0 as required by the protocol.  Moreover, omitted flags still cause the progressive enhancements to be enabled.
Pop operations (`CSI < u`) function correctly, so a pop is required to recover from a push with no flags.

Effects for the user is that after existing some applications, the terminals does not respond to keys like up/down, return, etc, and prints kitty keycodes instead.  I ran into this with https://github.com/sst/opencode/, which led me to find that any application using https://github.com/charmbracelet/bubbletea/ v2 is affected, but since Zellij isnt handling this part of the spec, there are likely more applications affected.

## Minimal reproduction

1. Start Zellij with Kitty keyboard protocol support enabled in a host terminal that supports it.
2. In a pane, send the `CSI > u` sequence to the terminal, e.g. via `printf "\e[>u"` from a shell.
3. Query the current flags with `CSI ? u`, e.g. `printf "\e[?u"` from a shell, and observe the printed value.

### Expected

Flags are set to 0.

### Actual

Flags are set to 1.

## Other relevant information

The behaviour of push with omitted flags is equivalent to push with flags.  Protocol details from [Kitty documentation](https://sw.kovidgoyal.net/kitty/keyboard-protocol/#progressive-enhancement) state that push should handle omitted flags as 0.

2asoft:
The `CSI > u` sequence (push) should default omitted flags to `0` per the [Kitty keyboard protocol specification](https://sw.kovidgoyal.net/kitty/keyboard-protocol/#progressive-enhancement). Previously, omitted flags incorrectly defaulted to 1, causing progressive enhancements to be enabled when they should remain disabled.

This fix ensures push with no flags (`printf "\e[>u"`) correctly sets flags to `0`, matching the protocol specification and making push behavior consistent with the existing `CSI = u` implementation introduced in #3942.

Fixes #4333.

### DIFF
--- a/zellij-server/src/panes/grid.rs
+++ b/zellij-server/src/panes/grid.rs
@@ -3142,8 +3142,14 @@ impl Perform for Grid {
         } else if c == 'u' && intermediates == &[b'>'] {
             // Zellij only supports the first "progressive enhancement" layer of the kitty keyboard
             // protocol
+            // 0 disables, everything else enables.
+            let count = next_param_or(0);
             if !self.explicitly_disable_kitty_keyboard_protocol {
-                self.supports_kitty_keyboard_protocol = true;
+                if count > 0 {
+                    self.supports_kitty_keyboard_protocol = true;
+                } else {
+                    self.supports_kitty_keyboard_protocol = false;
+                }
             }
         } else if c == 'u' && intermediates == &[b'<'] {
             // Zellij only supports the first "progressive enhancement" layer of the kitty keyboard


2asoft:
I think some basic tests for this functionality would be useful, but I am not sure of a good way to write them.  Open to suggestions there.

aymanbagabas:
Hi there! I've encountered this working on Bubble Tea v2 and had to make it use the `CSI =` notation to work around this and make Kitty keyboard work properly in Bubble Tea applications within Zellij.

CC/ @imsnif

imsnif:
Good catch, thank you!

PR #4345, opened 2025-08-06
fix(web): set secure cookie flag only for https

imsnif:
Previously, we would set the `secure` cookie flag (meaning the cookie should only be sent on https connections) to our session_token cookie whether we were on http or https. Most browsers have a "localhost exception" for this, meaning they send secure cookies on non-https connections so long as they are to localhost. Safari does not have this exception, so users were unable to log-in before they set an https certificate.

This fixes the issue by conditionally setting the `secure` flag depending on whether the connection is https or not. Since the web client does not perform any http (or indeed, any remote) requests other than to the server, this is not considered risky.


imsnif:
@tlinford - mind giving this a test on safari to make sure we fixed the issue you found?

tlinford:
Seems fine now.

PR #4351, opened 2025-08-07
Update dependencies to support more architectures

OriginCode:
This includes two dependency version changes:

- wasmtime{,-wasi} => 30.0.2
- nix => 0.29, indirectly depended by termwiz => 0.23

It now supports more architectures including loongarch64, ppc64el, and mips64

Superseding https://github.com/zellij-org/zellij/pull/4137

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2514,7 +2514,7 @@ version = "1.1.8"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "c0aeb26bf5e836cc1c341c8106051b573f1766dfa05aa87f0b98be5e51b02303"
 dependencies = [
- "nix 0.29.0",
+ "nix",
  "winapi",
 ]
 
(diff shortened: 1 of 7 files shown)


MingcongBai:
@imsnif Can you kindly take a look at this? @OriginCode has attempted at this for two other times at this point:

- https://github.com/zellij-org/zellij/pull/4080
- https://github.com/zellij-org/zellij/pull/4137

But all were ignored... We have confirmed that this fix would make Zellij work on LoongArch for quite a while now - if there are other assistance that you would need, please do let us know.

Thanks!

imsnif:
I'm sorry, but this is absolutely not a small change. It upgrades our runtime several breaking changes ahead and requires a lot of manual testing for behavior changes. We want to upgrade wasmtime to the newest LTS (iirc it should be released this week). I think @tlinford wanted to take a look at it.

OriginCode:
> I'm sorry, but this is absolutely not a small change. It upgrades our runtime several breaking changes ahead and requires a lot of manual testing for behavior changes. We want to upgrade wasmtime to the newest LTS (iirc it should be released this week). I think @tlinford wanted to take a look at it.

Nice news on wasmtime. What about nix?

imsnif:
> Nice news on wasmtime. What about nix?

Sorry, I didn't mean to ghost you - but just looking into this would involve some work and manual testing (this is a library for system interaction so can't be easily tested). Once wasmtime is out of the way we can give this a closer look.

PR #4373, opened 2025-08-19
fix(web-client): cmd handling, always use relative urls

tlinford:


tlinford:
closing as relative url approach is not sufficient yet

PR #4376, opened 2025-08-19
fix(web): avoid sending single meta keypresses, cleanup

### LINKED ISSUE #4359: CMD key on MacOS prints character while in Web Client
## Issue description

While using the new Web Client feature on MacOS, pressing the CMD key is printing a character to the terminal. For example:

- Pressing `CMD+T` to open a new tab prints `m`. 
- Pressing `CMD+SHIFT+]` to go to the next browser tab prints `M`

Image

## Minimal reproduction

1. Configure Zellij Web Client to use HTTPS using a cert valid for: localhost, 127.0.0.1, ::1, and local IP address
2. Configure Zellij Web Client to use port `11000` and bind to `0.0.0.0`
3. Start it: `zellij web`
4. Connect to the web server using a browser and provide your token
   - I've tried: Safari, Chrome Version 138.0.7204.184 (Official Build) (arm64) and Version 139.0.7258.67 (Official Build) (arm64), Firefox 141.0.3 (aarch64)
5. Press `CMD+T` to open a new tab.

## Other relevant information

- Zellij v0.43.0
- Tried on MacOS 15.6 and 26.0 Beta
- My shell is ZSH 
- Font is "Fira Code"

tlinford:
- extend the logic already used to skip single modifier key presses to meta
- refactor: extract the code into a function, make it simpler and use event.key instead of deprecated event.keyCode
- cleanup leftover axum cors config from feature development

Fix #4359

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 
 ## [Unreleased]
+* fix: meta key handling in web client (https://github.com/zellij-org/zellij/pull/4376)
 
 ## [0.43.1] - 2025-08-08
 * fix: pane rename backspace regression (https://github.com/zellij-org/zellij/pull/4346)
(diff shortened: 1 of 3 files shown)


thiblahute:
That one was annoying, thanks for fixing :-)

mokeyish:
When can we release the version? This almost makes the web unusable.

Issue #1429, opened 2022-05-22
Thread 'stdin_handler' panicked along suspending system into sleep mode


matu3ba:
Unfortunately the issue is not easily reproducable and I dont have a backtrace/the panic handler did not tell what code broke.
Further, I broke the logs in `/tmp/zellij-1000/zellij-log/` by starting a fresh instance, but I expect those would not tell much either.

The following output information were printed:
```txt
  × Thread 'stdin_handler' panicked.
  ├─▶ At /home/misterspoon/.cargo/registry/src/github.com-1ecc6299db9ec823/zellij-client-0.29.1/src/os_input_output.rs:256:14
  ╰─▶ could not poll stdin for readiness: Os { code: 4, kind: Interrupted, message: "Interrupted system call" }
  help: If you are seeing this message, it means that something went wrong.
        Please report this error to the github issue.
        (https://github.com/zellij-org/zellij/issues)

        Also, if you want to see the backtrace, you can set the `RUST_BACKTRACE` environment variable to `1`.
```

**Basic information**

`zellij --version`: zellij 0.29.1
`tput lines`: 30 (not relevant)
`tput cols`: 126 (not relevant) 
`uname -av` or `ver`(Windows): Linux pc 5.17.9-arch1-1

List of programs you interact with as, `PROGRAM --version`: output cropped meaningful, for example:
`nvim --version`: NVIM v0.5.0-dev+1299-g1c2e504d5  (used the appimage release)
`alacritty --version`: alacritty 0.7.2 (5ac8060b)

**Further information**
To reproduce, I suspect that suspending system into sleep mode and waking up should reproduce the issue eventually. 
Signaling code is very hard, so I would not be surprised that some dependency does not handle the signal interrupt mask or handlers correctly.

If you give me some pointers what you use to poll stdin I'll fix the signaling code myself. Ideally a portable solution should be used.
My suspection is that signal interrupts are not handled at all such that the bubbled up failure to poll crashes the application.

raphCode:
Does this crash your session / the zellij server or are you able to reattach to your session afterwards?

matu3ba:
Thanks for the pointers!

1. Is `EINTR` already handled in Rust by retrying or are C semantics used?
2. `SIGSTOP and SIGCONT signals` are not handled, but I need to check first what the signaling semantics of mio poll are.
3. `.expect("could not poll stdin for readiness");` reads like it doesnt handle this edge case (but I must check first)

raphCode:
[...]

PR #4397, opened 2025-09-05
fix: Thread 'stdin_handler' panicked along suspending system into sleep mode (#1429)

fabianbuettner:
The panic could be observed almost each time after waking it up from suspend. According to a post in users.rust-lang.org[1] some system calls in POSIX can be interrupted and the application needs to handle this, usually by ignorin the error and re-issuing the system call.

[1]: https://users.rust-lang.org/t/how-to-fix-interrupted-system-call/44605/2


fabianbuettner:
Sorry, I referenced the wrong issue

PR #4449, opened 2025-09-29
Migrate from wasmtime to wasmi

imsnif:
UPDATE: After rigorous testing, improvements and optimizations, this seems to improve things for us significantly and is no longer an experiment.

### What changed?

1. The main difference is that now rather than compiling .wasm files, we interpret them. This is a much faster process (compile times would number in seconds, now they number in single-digit milliseconds).

2. Point 1 means that we no longer need to keep a hard-drive cache of compiled assets. The main benefit being ease of plugin development (no need to worry about cache-busting) and reduced code complexity.

3. Instead of using an async runtime, we now use a threadpool with a "pinned thread" model. Plugins are pinned to a specific thread in the pool, making sure to deallocate their memory on the same thread that allocated it to reduce memory fragmentation (this was actually an unrelated problem in the way we've been doing things up until now that caused some minuscule memory leakage, it's now plugged).

4. Our memory footprint has decreased (mostly due to 3), as well as our binary size.

5. Plugin performance *has* taken a minor hit for compute-heavy operations. This is of course not Humanly noticeable for UI elements, but it could be noticeable for scanning large amounts of text synchronously (eg. file contents fuzzy matchers). This can be mitigated by adding [these optimizations](https://github.com/wasmi-labs/wasmi/blob/main/docs/usage.md#webassembly-optimizations) to the plugin compilation (similar optimizations have been added to the Zellij compilation itself). If plugin authors are affected by this, I would be happy to work around these problems by adding dedicated API methods that would perform this sort of work on the Zellij side.

~~This is (for the time being) an experiment to migrate our wasm/wasi runtime from wasmtime to wasmi. The main benefit is being able to use a wasm interpreter rather than a compiler. Doing this, we will be able remove our plugin cache entirely - relying instead on just loading the `.wasm` files on demand whenever a plugin is loaded. This should make plugin development *significantly* easier, as well as reduce a great amount of complexity from our own code. It will also reduce our resource utilization and binary size.~~

~~From tests I have run locally this has no performance impact whatsoever for us, and if anything actually makes things a little bit faster on first run (no compilation step needed).~~

~~Before I move forward with this, I'd like to check with a few stakeholders:~~

~~@dj95 - if you have the time, would you be willing to give this branch a go and test zjstatus? Namely some of the more advanced features such as pipes, complex and time sensitive filesystem access, etc.~~

~~@WeepingClown13 - as far as I understand it, the main hurdle in packaging Zellij for Debian was troubles packaging wasmtime. Since this removes the wasmtime crate entirely and adds wasmi instead (which should be a far simpler beast) - will this help your efforts? Or will it hurt them because now you'll have to package something entirely new?~~

~~@tlinford - could you give this a go on a mac to see if it works fine? (just a smoke test: start the app, open some tabs, attach from another terminal session).~~

~~Thanks all!~~

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 * feat: allow attaching to remote Zellij sessions over https (eg. `zellij attach https://example.com/my-cool-session`) (https://github.com/zellij-org/zellij/pull/4460)
 * build: Update Rust toolchain to 1.90.0 (https://github.com/zellij-org/zellij/pull/4457)
 * feat: allow plugins to read pane scrollback (https://github.com/zellij-org/zellij/pull/4465)
+* infra: migrate wasm runtime from wasmtime to wasmi (https://github.com/zellij-org/zellij/pull/4449)
 
 ## [0.43.1] - 2025-08-08
 * fix: pane rename backspace regression (https://github.com/zellij-org/zellij/pull/4346)


tlinford:
Smoke test on mac went well.

Regarding performance, did my first test with `cargo x run`, and in that case I had the feeling that things were slower than I'm used to. With a release binary did not notice anything off.

imsnif:
> Regarding performance, did my first test with `cargo x run`, and in that case I had the feeling that things were slower than I'm used to. With a release binary did not notice anything off.

Sower, how so? What was slow?

tlinford:
> Sower, how so? What was slow?

Noticed a bit of lag in the tab bar and status bar when changing tabs for example. I'll get you a video as soon as I manage.

tlinford:
https://github.com/user-attachments/assets/afa07639-9a34-4a9f-97c6-257235377a68


https://github.com/user-attachments/assets/388c613a-5a11-4681-8ef6-f87c1f3ff879

Did a `cargo x run` on both `main` and `wasmi-migration`.

WeepingClown13:
wasmi is not packaged in debian, but I gave the dependency tree a look and it doesn't seem to have a lot missing, mostly just wasi crates itself. So packaging that _might_ not be all that hard, although I don't know about the complexicity of wasi codebase itself. But one thing I see is wiggle in the wasmi dependency tree and that to my understanding depends on a bunch of wasmtime stuff. So untimately the chances are that it will be still stuck. But I'd guess if wiggle is the only thing that'll now need wasmtime, the overall situation might be at least somewhat better. So feel free to go ahead with whatever direction you want, with the note that in the long run it'll be easier for the packaging efforts when the dependency tree doesn't keep changing all the time.

imsnif:
@WeepingClown13 - thank you for your patience with our changing dependency tree! I can definitely understand how it might make things difficult, long and frustrating.

One of the main reasons for this experiment is to make it easier to package Zellij for Debian - so we have a mutual goal and I'm willing to go the distance in order to make this possible, easy and quick. Would you be willing to look a little deeper into this to see how much of a blocker the wiggle dependency (and other similar ones) might be? If there is still trouble there, I'm happy to explore other options. Let me know!

WeepingClown13:
@imsnif it seems wasmtime for wiggle is an optional dependency as well as a dev dependency. While wiggle does add it as a default feature, wasmi is using wiggle with default_features=false and thus it should not be a problem to just forget about the wasmtime dependency. wiggle usage in wasmi seems to be very minimal. I will have to patch out a bunch of things, but it should be possible to package that as well. From what I can see now anyway.

imsnif:
> @imsnif it seems wasmtime for wiggle is an optional dependency as well as a dev dependency. While wiggle does add it as a default feature, wasmi is using wiggle with default_features=false and thus it should not be a problem to just forget about the wasmtime dependency. wiggle usage in wasmi seems to be very minimal. I will have to patch out a bunch of things, but it should be possible to package that as well. From what I can see now anyway.

Gotcha. Thanks for taking a look @WeepingClown13 ! So to make sure I understand you correctly: if we move forward with this branch it should get you unstuck with packaging Zellij for debian, right?

I'm still looking into other options (seeing as there are some issues with wasmi), but I want to make sure whichever option we choose will have the benefit of unstucking your packaging efforts. So: if we switch instead of wasmi to wasmedge, that will be even better for you, right? Since wasmedge is afaict already packaged for debian? (not saying we'll go this path, there are other considerations, just wanting to see what our options are).

WeepingClown13:
Right, wasmi switch should make it possible to move forward with the packaging again, as per my current understanding at least. wasmedge, or any wasm runtime/engine whatnot packaged in debian, should indeed make that easier as well.

Thanks for the packaging side considerations, that helps a lot. Whichever one you go with, I'd just hope that it's a long term option.

imsnif:
> Thanks for the packaging side considerations, that helps a lot. Whichever one you go with, I'd just hope that it's a long term option.

Awesome - I'll definitely keep that in mind. I'm not happy with engine hopping either!

When you have a chance, could you let me know exactly which patching you need to do to get wasmi to work without wiggle? I'll replicate that environment and give it a test locally to make sure we're not losing anything important.

Otherwise for you and all others in this thread: I'm going to keep testing out wasmi. Right now it has a memory leak I haven't yet managed to pin-point (and then there's the slowness issue reported on macos above in non --release builds). If we solve both of these issues and find no other outstanding ones, we'll move forward with this. Otherwise I'll start by doing another experiment with wasmedge and see where that leaves us.

WeepingClown13:
I will not be patching wasmi to work without wiggle, rather I will patch out the wasmtime dependency and features from wiggle when packaging. They are not used by wasmi anyway, so it should be inconsequential, and from the wiggle side also it should be similar as the dependency on wasmtime is optional.

dj95:
Finally managed to get the new version up and running. There's at least one breaking change within this version, but I'm not sure if it's due to this MR or another change differing from the latest release. I cannot specify layout and config file via CLI parameters with a relative path. Zellij won't find it anymore and then load the default layout.

e.g.

```bash
zellij -s test -n ./tests/zjstatus/layout.kdl # doesn't work anymore
zellij -s test -n $PWD/tests/zjstatus/layout.kdl # does work
```

As the status bar is one new instance per new tab, the startup time has a really large impact on how seamless the experience is. When testing, the missing performance on plugin start is significant! I've compiled this version with `cargo x build -r`. Not sure if it's completely a release build because there are a ton of wasi log lines in zellijs log.

**With wasmi**

https://github.com/user-attachments/assets/e25c50d2-c768-4435-a4eb-991b305658ec

**With wasmtime**

https://github.com/user-attachments/assets/b7e575d7-d392-46e3-901f-e04addd4483c

I'll perform further testing in the next days. Functionality seems to work though (I've tested pipes manually and complex fs access is working within the command widgets), just the performance seems degraded - especially on startup.

imsnif:
Thanks for testing @dj95 ! The layout relative path is a main thing, I've heard reports elsewhere but thanks for mentioning this specific case. I'll add it to the list and take a better look before releasing.

It's also good input regarding the loading speed. I'm guessing you're also on a mac? I seem to remember we also had similar issues specifically with mac when loading plugins in the early days (with wasmer) - which is what got us to create the plugin cache in the first place.

I'll try to troubleshoot this a bit (hopefully @tlinford would be willing to run a custom build and show me some logs). Worst case, we'll bring back the memory cache, then only the first load will be slow (in release from @dj95 's screencast it seems to be on the order of magnitude of 1s right? I'm not missing something? this seems acceptable for first plugin instance load IMO).

imsnif:
@dj95 - I'm trying to troubleshoot the performance issue, would you be willing to help try out some things?

Namely, could you try:
1. Compiling this branch in release mode (`cargo x build --release`) but changing the following section in the Zellij Cargo.toml:
```toml
[profile.release]
lto = true
strip = false
debug = true
```
To:
```toml
[profile.release]
lto = true
codegen-units = 1
```
(warning, this will slow down compilation significantly)

2. If the above doesn't help (or doesn't help enough), could you try in addition to optimize the zjstatus build and then load it? This will involve:
* Changing the release profile (similar to step 1) but in zjstatus:
```toml
[profile.release]
lto = "fat"
codegen-units = 1
panic = "abort"
```
* Running `binaryen`'s `wasm-opt` to optimize the created wasm code: https://github.com/WebAssembly/binaryen?tab=readme-ov-file#wasm-opt

dj95:
Yes, it's within the margin of 1 second on initial start. But as every new tab has an initial start without the caching, it feels really sluggish for status bars. For most other plugins it should be sufficient IMO, as they do not spawn as often as a status bar will on new tabs.

---

Sure! I've tested the steps above, except `wasm-opt` and it works insanely fast right now, when using the optimised zjstatus build.


https://github.com/user-attachments/assets/64e25a8c-bde7-49a3-a7c9-0cd89746f7b7

imsnif:
Very cool @dj95 ! Could you test both in isolation to find out which one gives us the boost, the compilation profile on Zellij or zjstatus? Or must it be both?

dj95:
Best performance is with both optimisations. I would say the profile in zjstatus is about 75% of the boost. But it's just a feeling, no real numbers :D

WeepingClown13:
At this point it is confirmed that this is the route, right? I want to look into the packaging when I get some free time.

imsnif:
> At this point it is confirmed that this is the route, right? I want to look into the packaging when I get some free time.

Yes! Just from about an hour ago :) I was planning to reach out to you after the merge (currently fixing the e2e tests)

PR #4484, opened 2025-10-23
fix(zellij-server): unblock building on 32bit targets

tranzystorekk:
After the migration to `wasmi` (https://github.com/zellij-org/zellij/pull/4449), we should be able to build for 32bit targets (wasm runtimes have notoriously had issues with such targets, see https://docs.wasmtime.dev/stability-tiers.html#unsupported-features-and-platforms).

While testing with builds on Void Linux for i686, I noticed the static asserts in our code that hardcode support for 64bit platforms only. These were introduced as an optimization (https://github.com/zellij-org/zellij/pull/3043), but I'd argue that this is a bit overly pedantic, especially when it effectively limits possible target platforms for no reason.

Hence, my proposal to drop them altogether as a simple solution. I'm aware that we could instead rewrite those to take platform-specific pointer size into account, but especially in the case of `TerminalCharacter` it becomes cumbersome to cleanly express what the expected size should be.

When testing I at least made sure that the new reported size only goes **down** for the affected target.

cc @aidanhs for comments and suggestions

### DIFF
--- a/zellij-server/src/panes/terminal_character.rs
+++ b/zellij-server/src/panes/terminal_character.rs
@@ -161,7 +161,7 @@ pub enum RcCharacterStyles {
     Rc(Rc<CharacterStyles>),
 }
 
-// #[cfg(target_arch = "x86_64")]
+#[cfg(target_arch = "x86_64")]
 const _: [(); 8] = [(); std::mem::size_of::<RcCharacterStyles>()];
 
 impl From<CharacterStyles> for RcCharacterStyles {
@@ -931,7 +931,7 @@ pub struct TerminalCharacter {
 
 // This size has significant memory and CPU implications for long lines,
 // be careful about allowing it to grow
-// #[cfg(target_arch = "x86_64")]
+#[cfg(target_arch = "x86_64")]
 const _: [(); 16] = [(); std::mem::size_of::<TerminalCharacter>()];
 
 impl TerminalCharacter {


imsnif:
@aidanhs - if you're around, available, willing and have the time to help us out by giving some input on this - it would be much appreciated!

evan0greenup:
@tranzystorekk , what is the current progress for this PR?

tranzystorekk:
From my POV the PR is good to merge, but as mentioned earlier, we were hoping for the original code author to have a say.

aidanhs:
Oh sorry, I thought I'd replied to this. Perhaps you could just config those two items to exist on x86_64 platforms, to prevent accidental regressions? Think it's `#[cfg(target_arch = "x86_64")]`

aidanhs:
But I'm happy to do that as a followup if you just want to get this merged

tranzystorekk:
Just spitballing here, since I'm not sure if it's even reasonably widely used, but maybe https://doc.rust-lang.org/reference/conditional-compilation.html#target_pointer_width would be more suitable / acceptably generic?

aidanhs:
Oh sure, that works too, hadn't come across that item before. Don't really mind what it is, as long as it gets caught somehow if CI tries to build after someone regresses it - either of these will work.

imsnif:
Thanks @aidanhs ! @tranzystorekk - I'm fine with either option. @tranzystorekk - I like the suggestion of adding a CI entry for this, wdyt?

tranzystorekk:
I'm not sure what kind of additional CI entry you have in mind? Assuming we only want to test this on 64bit targets, we already run a build step on standard ubuntu-latest runner, which should fail on a regression.

imsnif:
For sure - I was thinking about adding a 32bit one, since we now want to support them too. Even just to run the test suite to make sure it compiles.

tranzystorekk:
Hm, github CI doesn't offer native 32bit runners, so we'd have to set up some kind of cross-build, I frankly don't remember the last time I had to do that...

tranzystorekk:
My general idea for this PR was to unblock building on 32 and potentially follow up from there later, officially supporting a set of targets is a huge leap

imsnif:
> My general idea for this PR was to unblock building on 32 and potentially follow up from there later, officially supporting a set of targets is a huge leap

Alright, I'm game for that. Just that we won't be able to guarantee we'll keep this support in the future without some sort of automated test. Want to adjust the conditional compilation so we can get this merged?

tranzystorekk:
Made the changes and verified that both 64 and 32 build successfully.

aidanhs:
Lgtm

imsnif:
Hey @tranzystorekk - could you try changing the compilation target to using a target architecture instead? While I can't see how this can be related, the same (somewhat related) test on mac is consistently failing here.

tranzystorekk:
Yeah, doesn't seem related to this change, still weird nonetheless...

imsnif:
For those keeping track at home: with the changes commented out this passed 2/3 times. Otherwise it passed 1/8 times. I think the numbers are too small to assume causality, but still quite odd. I'm going to merge this as it did pass the one time and deal with the flaky test on main.

Thanks @tranzystorekk and @aidanhs !!

PR #4542, opened 2025-12-09
feat: layout-independent keybindings for non-English keyboards

### LINKED ISSUE #1355: keyboard layout independent keybindings
The keybindings now only work on the English layout. Sway has good solution.
https://wiki.archlinux.org/title/Sway#Keybindings_and_keyboard_layouts

nikita-voronoy:
Fixes #1355

Keybindings now work on non-English keyboard layouts (Ukrainian, Russian, Greek, etc.).

**Problem:** When using a non-English keyboard layout, keybindings don't work:
- `Alt+а` is sent instead of `Alt+a`
- `Ctrl+П` is sent instead of `Ctrl+g`

**Solution:** Leverage Kitty keyboard protocol's extended features:

1. **Unicode support for Alt+keys**: Changed `u8` to `u32` parsing to handle Unicode codepoints > 255

2. **Base layout key for Ctrl+keys**: Enabled "Report alternate keys" flag (`>5u` instead of `>1u`) and parse the base-layout-key field: `CSI unicode:shifted:base-layout ; modifiers u`

**Test:** Switch to non-English layout, test `Alt+n`, `Ctrl+g` - should work.

### DIFF
--- a/zellij-client/src/keyboard_parser.rs
+++ b/zellij-client/src/keyboard_parser.rs
@@ -6,6 +6,8 @@ enum KittyKeysParsingState {
     Ground,
     ReceivedEscapeCharacter,
     ParsingNumber,
+    ParsingShiftedKey,
+    ParsingBaseLayoutKey,
     ParsingModifiers,
     DoneParsingWithU,
     DoneParsingWithTilde,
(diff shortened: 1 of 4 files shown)


SacrilegeWasTaken:
We're really need it. Can help with the implementation or anything.

Issue #3951, opened 2025-01-22
Zellij prevents neovim from detecting OSC52 clipboard support


kika:
Zellij prevents neovim 0.10.* from detecting OSC52 clipboard support (which zellij supports). 

## Minimal reproduction
kitty (0.34.0) terminal on MacOS runs an ssh session (using `kitten ssh` for proper terminfo) to an ubuntu machine:
1. Without zellij on a remote machine - no problem, neovim (`nvim --clean`) detects OSC52 support in kitty and installs fallback OSC52 clipboard handler
2. In the zellij pane neovim doesn't install the OSC52 fallback clipboard handler.

I've sprinkled `$VIMRUNTIME/plugins/osc52.lua` with a bunch of `vim.print()`s and discovered that this code works differently in these two situations:
```lua
require('vim.termcap').query('Ms', function(cap, found, seq)
  if not found then
    vim.print('No term cap, return')
    return
  end
--- ....
end)
```

Without zellij the `Ms` capability is found, with it - not. The `infocmp` result in both sessions is exactly the same and kitty is recognized as a terminal emulator.

loralb:
Having a similar problem, where the `paste` keybind of my terminal emulator (tried with both ghostty and konsole) does not work when inside zellij. This happens while working on my local machine and also when working on a remote machine, connected through ssh.

`zellij --version: zellij 0.41.2`

Basically I can't paste text from my clipboard when inside a zellij session, no matter what I am running (text editor, shell, whatever)

Enzime:
Feel free to try out my PR #4545 that should fix this issue with Neovim running the latest nightlies :)

ifnk:
why not merge this ?

khaneliman:
> why not merge this ?

This is an issue, not a pr. https://github.com/zellij-org/zellij/pull/4545 was merged. This should probably be closed if @Enzime resolved this with the merge.

imsnif:
I implemented opt-in support for reading the paste buffer (assuming the host terminal itself supports it) under a new `dangerously_enable_paste_buffer_read true` in the config (defaulting to false): https://github.com/zellij-org/zellij/pull/5472

At the time of writing still unreleased. Follow the project on github if you'd like to be notified when this is released.

Issue #2647, opened 2023-07-25
forward clipboard read osc sequence


theimpostor:
It appears that reading the clipboard via osc is not supported yet. Can it be added? Many terminals have added support for this, either by default (e.g. alacritty) or gated behind a config option (iterm2, kitty). Passing through the clipboard read command and allowing the terminal to handle/reject seems like it would be useful.

whmountains:
#2637 is related to this.  Having the ability to paste into a Zellij session would be very helpful for those of us who "literally live in the remote session."

Lockszmith-GH:
[osc](https://github.com/theimpostor/osc) is a clipboard tool for the modern age, since [August](https://github.com/theimpostor/osc/releases/tag/v0.3.9) is explicitly excluding Zellij due to incompatibility - links to this issue on the README.md page.

This is a shame, as this is the only non-server based clipboard stdin/stdout app I've found.

If no one can spare the time to take a look at it, I would appricate some guidance at to how this would be tackled.

Some general pointers, and I'd be willing to attempt a PR for this.

imsnif:
Zellij intentionally does not support the `paste` part of OSC52, just the copy part. 

Why? Think about everything you're copying to your clipboard (intentionally or unintentionally through just marking it in some DEs). Now imagine *anything* you run in your terminal (intentionally or unintentionally) would be able to read this clipboard without you even noticing it happened.

Would you be comfortable with that? Neither would I. I'm so uncomfortable with it that I don't even want software that I maintain to have this ability.

I realize this can be very convenient and empathize with users who want this, but this is IMO too reckless to have in any sort of terminal.

I might consider adding it behind a user visible prompt (eg. `program <X> wants to read this text <TEXT> from your clipboard, allow? (y/n)`)

jshu-git:
Chiming in here since I've been looking forward to this too for a while. To preface, I have been daily driving with zellij at work since its release, and it has replaced (almost all of) tmux for and I don't see myself ever going back. (I also use neovim with clipboard='unnamedplus').

[...]

PR #4545, opened 2025-12-11
feat(zellij-server): report OSC 52 clipboard support over DA1

### LINKED ISSUE #2637: Zellij Clipboard Functionality: Limitations for Integrating Neovim Registers
I am a user of **Neovim** and heavily use its registers. It is convenient to press `y` to copy some text from **Neovim** running on the remote host in Zellij to my local laptop's system clipboard.

When I used **tmux**, it had integrated `tmux set-buffer` and `tmux save-buffer` functions that allowed me to pipe their contents into tmux's clipboard. Then, **tmux** copies these contents to my local clipboard using **osc52**. **Neovim** integrated with tmux's clipboard by default, and you can find more information about it by running `:help clipboard` in nvim.

However, after switching to **Zellij**, the advanced terminal multiplexer, I discovered that it doesn't have tools like `zellij action copy` and `zellij action paste`, which means that **Neovim** cannot dump its registers' content to the clipboard.

I am wondering if this is a limitation of Zellij's clipboard functionality or if I have some misunderstanding about how it works.

Enzime:
As Zellij always handles OSC 52 from inner programs and then forwards it appropriately (`copy_command`, `copy_clipboard`), I think it makes sense to unconditionally report OSC 52 support.

This makes clipboard support work in `neovim` running inside of `zellij` without any configuration changes.

Fixes https://github.com/zellij-org/zellij/issues/2637 and https://github.com/zellij-org/zellij/issues/3951

See also https://github.com/zellij-org/zellij/issues/2647


Enzime:
Marking this as a draft as I'm testing this PR over SSH to ensure that this PR actually works

Enzime:
After a bunch of debugging, I've confirmed that this PR does fix Neovim's ability to copy to the system clipboard through OSC 52 when run over SSH inside Zellij

However, it depends on https://github.com/neovim/neovim/pull/34860 which is only available in nightly builds currently

When I check the primary device attributes (DA1) in Ghostty outside of Zellij I get:

```
$ echo -en '\e[c'; sleep 0.2
^[[?62;22;52c
```

With Zellij on `main`:

```
$ echo -en '\e[c'; sleep 0.2
^[[?62;4c%
```

With this PR:

```
$ echo -en '\e[c'; sleep 0.2
^[[?62;22;52c
```

adamcstephens:
I can confirm this works with neovim nightly.

Enzime:
@imsnif any chance you can take a look at this? neovim 0.12 is now released including https://github.com/neovim/neovim/pull/34860

imsnif:
> @imsnif any chance you can take a look at this? neovim 0.12 is now released including [neovim/neovim#34860](https://github.com/neovim/neovim/pull/34860)

I was honestly just this second reviewing this to add to the upcoming patch :) Good timing.

Thanks for this!

airtonix:
I think this needs to be revisited.

1. start a Zellij Session on my laptop, within that session start neovim.
2. name that session "Dotfiles"
3. from my desktop, ssh to my laptop
4. zellij attach Dotfiles
5. There's no SSH_TTY or other usual envvars to indicate that this is a ssh connection that resumed a session.

PR #4554, opened 2025-12-17
Additional Themes

jeffwindsor:
Included Themes:

- [Duskfox](https://github.com/EdenEast/nightfox.nvim) from nightfox
- [Evergreen](https://github.com/everviolet/nvim) fall and spring
- [Tender](https://github.com/jacoborus/tender.vim)
- [Japanese Indigo](https://github.com/vimpunk/vim-japanese-indigo)
- [Melange](https://github.com/savq/melange-nvim)


imsnif:
Hey - thanks for this, and sorry for the delay. I think some (all?) of the themes here make the controls in eg. the configuration plugin (but also other places) a bit hard to distinguish:

theme-issue

Want to take a closer look, maybe using https://github.com/imsnif/theme-tester ?

jeffwindsor:
@imsnif I updated the themes to make them all more readable.  The theme-tester was a great help.  Let me know if these are better

jeffwindsor:
also looking into an external theme manager `tinty` from https://github.com/tinted-theming .  It looks promising for generating base-16 and base-24 themes.  If it works out I will try to add zellij to their supported list of command line apps.  Right now I have it working via a hook and a generated "theme.kdl" that is always referenced.

PR #4560, opened 2025-12-19
fix(cli): resolve --layout paths relative to CWD instead of layout_dir

### LINKED ISSUE #4557: Regression when starting zellij with a layout from a local file
## Issue description

Opening zellij with a layout from a local file works fine with the latest published version (v.0.43.1), but on `main` the layout is ignored.

## Minimal reproduction

- Go to any directory
- Write down a small `kdl` file with a layout
- Run `zellij --layout ./your_layout.kdl`

Actual behavior: The layout is ignored.

Expected behavior: The layout is used.

## Other relevant information

The regression is introduced by PR #4383, which moves the path resolution to the server. It makes uses of the code path introduced in PR #4278, whose behavior is inconsistent with the one documented in PR #1426, which states:

> A layout is considered local if:
> layout.extension().is_some() || layout.components().count() > 1
> 
> This check produces the following results:
> "foo": Abs
> "foo.yml": Local
> "./foo": Local
> "foo/bar": Local
> "foo/bar.yml": Local

I can attempt a fix.

yacinehmito:
Fixes #4557

## Summary

Fixes a regression introduced in PR #4383 where CLI `--layout` paths were incorrectly resolved relative to `layout_dir` (typically `~/.config/zellij/layouts/`) instead of the current working directory.

This broke commands like `zellij --layout ./my-layout.kdl` which previously worked per PR #1426's semantics.

## Changes

- Add `LayoutInfo::from_cli_arg()` for CLI `--layout` paths (resolved from CWD)
- Keep `LayoutInfo::from_config()` for config's `default_layout` setting (resolved from `layout_dir`)
- Update `zellij-client` to use `from_cli_arg()` for CLI layout arguments
- Simplify `Layout::from_layout_info()` since path resolution now happens at the caller level

## Behavior After Fix

| Source | Path | Resolution |
|--------|------|------------|
| CLI | `zellij --layout my-layout.kdl` | From CWD |
| CLI | `zellij --layout ./foo.kdl` | From CWD |
| CLI | `zellij --layout default` | Built-in layout |
| Config | `default_layout "my-layout.kdl"` | From `layout_dir` |

## Testing

- Added unit tests for `LayoutInfo::from_cli_arg()` and `LayoutInfo::from_config()`
- Added integration tests for `Layout::from_layout_info()` with relative and absolute paths
- All 213 tests pass

### DIFF
--- a/zellij-utils/src/input/layout.rs
+++ b/zellij-utils/src/input/layout.rs
@@ -1291,19 +1291,14 @@ impl Layout {
         });
         (available_layouts, layouts_with_errors)
     }
-    pub fn from_layout_info(
-        layout_dir: &Option<PathBuf>,
-        layout_info: LayoutInfo,
-    ) -> Result<Layout, ConfigError> {
+    pub fn from_layout_info(layout_info: LayoutInfo) -> Result<Layout, ConfigError> {
         let mut should_start_layout_commands_suspended = false;
         let (path_to_raw_layout, raw_layout, raw_swap_layouts) = match layout_info {
-            LayoutInfo::File(layout_name_without_extension, _layout_metadata) => {
-                let layout_dir = layout_dir.clone().or_else(|| default_layout_dir());
+LayoutInfo::File(layout_path_str, _layout_metadata) => {
+                // Path resolution (relative to CWD or layout_dir) is done by the caller
+                let layout_path = PathBuf::from(&layout_path_str);
                 let (path_to_layout, stringified_layout, swap_layouts) =
-                    Self::stringified_from_dir(
-                        &PathBuf::from(layout_name_without_extension),
-                        layout_dir.as_ref(),
-                    )?;
+                    Self::stringified_from_path(&layout_path)?;
                 (Some(path_to_layout), stringified_layout, swap_layouts)
             },
             LayoutInfo::BuiltIn(layout_name) => {
@@ -1334,19 +1329,16 @@ impl Layout {
         layout
     }
     pub fn from_layout_info_with_config(
-        layout_dir: &Option<PathBuf>,
         layout_info: &LayoutInfo,
         config: Option<Config>,
     ) -> Result<(Layout, Config), ConfigError> {
         let mut should_start_layout_commands_suspended = false;
         let (path_to_raw_layout, raw_layout, raw_swap_layouts) = match layout_info {
-            LayoutInfo::File(layout_name_without_extension, _layout_metadata) => {
-                let layout_dir = layout_dir.clone().or_else(|| default_layout_dir());
+LayoutInfo::File(layout_path_str, _layout_metadata) => {
+                // Path resolution (relative to CWD or layout_dir) is done by the caller
+                let layout_path = PathBuf::from(&layout_path_str);
                 let (path_to_layout, stringified_layout, swap_layouts) =
-                    Self::stringified_from_dir(
-                        &PathBuf::from(layout_name_without_extension),
-                        layout_dir.as_ref(),
-                    )?;
+                    Self::stringified_from_path(&layout_path)?;
                 (Some(path_to_layout), stringified_layout, swap_layouts)
             },
             LayoutInfo::BuiltIn(layout_name) => {
--- a/zellij-utils/src/input/unit/layout_test.rs
+++ b/zellij-utils/src/input/unit/layout_test.rs
@@ -2361,3 +2361,71 @@ fn layout_node_with_name_and_only_floating_panes() {
     assert_eq!(layout.tabs.len(), 1);
     assert_eq!(layout.tabs[0].0, Some("floating-only".to_string()));
 }
+
+#[test]
+fn from_layout_info_resolves_relative_path_from_cwd() {
+    use crate::data::LayoutInfo;
+    use std::env;
+    use std::fs;
+
+    let temp_dir = env::temp_dir().join("zellij_test_layout_info_relative");
+    if temp_dir.exists() {
+        fs::remove_dir_all(&temp_dir).unwrap();
+    }
+    fs::create_dir_all(&temp_dir).unwrap();
+
+    let layout_content = r#"
+layout {
+    pane
+    pane
+}
+"#;
+    let layout_file = temp_dir.join("test-layout.kdl");
+    fs::write(&layout_file, layout_content).unwrap();
+
+    env::set_current_dir(&temp_dir).unwrap();
+
+    let layout_info = LayoutInfo::File("./test-layout.kdl".to_string());
+
+    let layout = Layout::from_layout_info(layout_info).expect("Failed to load layout");
+
+    fs::remove_dir_all(&temp_dir).unwrap();
+
+    if let Some((tiled_panes, _)) = &layout.template {
+        assert_eq!(tiled_panes.children.len(), 2);
+    } else {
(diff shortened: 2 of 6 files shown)


zellij-utils/src/input/unit/layout_test.rs
```diff
@@ -2242,3 +2242,79 @@ fn env_var_missing() {
...
 }
+
+#[test]
+fn from_layout_info_resolves_relative_path_from_cwd() {
+    use crate::data::LayoutInfo;
+    use std::env;
+    use std::fs;
+
+    let temp_dir = env::temp_dir().join("zellij_test_layout_info_relative");
+    let _ = fs::remove_dir_all(&temp_dir);
```
har7an: This should `unwrap()` since failure to remove the directory may mean we keep state from a previous test.

zellij-utils/src/input/unit/layout_test.rs
```diff
@@ -2242,3 +2242,79 @@ fn env_var_missing() {
...
+
+    let original_cwd = env::current_dir().unwrap();
+    env::set_current_dir(&temp_dir).unwrap();
+
+    let layout_info = LayoutInfo::File("./test-layout.kdl".to_string());
+    let layout_dir = Some(std::path::PathBuf::from("/some/other/layout/dir"));
+
+    let result = Layout::from_layout_info(&layout_dir, layout_info);
+
+    env::set_current_dir(&original_cwd).unwrap();
```
har7an: Iiuc there is no need here to (re)store the original CWD, since it should not affect subsequent tests. Or is this in fact required?

zellij-utils/src/input/unit/layout_test.rs
```diff
@@ -2242,3 +2242,79 @@ fn env_var_missing() {
...
+    let _ = fs::remove_dir_all(&temp_dir);
+
+    assert!(
+        result.is_ok(),
+        "Relative path should resolve from CWD: {:?}",
+        result.err()
+    );
```
har7an: This is needlessly verbose and should be equivalent to a `result.expect(...)` I think. Also the error message is misleading since `from_layout_info` may fail for other (unrelated) reasons.

zellij-utils/src/input/unit/layout_test.rs
```diff
@@ -2242,3 +2242,79 @@ fn env_var_missing() {
...
+
+    let result = Layout::from_layout_info(&layout_dir, layout_info);
+
+    let _ = fs::remove_dir_all(&temp_dir);
+
+    assert!(
+        result.is_ok(),
+        "Absolute path should work: {:?}",
+        result.err()
+    );
```
har7an: Could you rewrite this like above, including a check at the end to see if the expected number of child panes matches please?

zellij-utils/src/input/layout.rs
```diff
@@ -1245,18 +1245,16 @@ impl Layout {
         available_layouts
     }
     pub fn from_layout_info(
-        layout_dir: &Option<PathBuf>,
+        _layout_dir: &Option<PathBuf>,
```
har7an: Why did you keep the second argument if it's not required any longer?

har7an (changes requested):
Thank you for the contribution!

I did a few manual tests and while your PR definitely restores some of the now broken functionality, it's not 100% equivalent to behavior in release 0.43.1.

Most notably, when I add a config file `layouts/test.kdl` and set the `default_layout` in `config.kdl` to the bare string `"test"`, 0.43.1 correctly resolved that to `layouts/test.kdl` whereas this version here currently doesn't (according to my testing). However I do admit that I can't currently tell where this originates and I strongly suspect this has another underlying issue deeper down in the code, i.e. it's unrelated to your fix.

Since another release is probably due soon I'll start looking into the bug I mentioned now, but I'd still like to take your patch since I think it's good! I just left a few remarks and I think GitHub would prefer if you rebased your code. :)

zellij-utils/src/input/layout.rs
yacinehmito: It triggers a lot of changes. I am happy to make it though, I just didn't want to go through the effort if the fix turned out not to be the proper one.

zellij-utils/src/input/unit/layout_test.rs
yacinehmito: I don't actually know. I assumed that `env::set_current_dir` effects the process, and thus needs restoration, but I didn't try without. I am not sure of what Rust does or does not do regarding test isolation.

I removed it and it seemed to work though.

zellij-utils/src/input/layout.rs
yacinehmito: Removing the parameter causes a chain reaction that leads to removing `layout_dir` from `PluginEnv`. I separated this to its own commit at the tip of the branch. Let me know whether it belongs here. If it does not, I'll drop it, but there will be a warning for the unused field.

yacinehmito:
Thank you very much @har7an for the review.

I rebased the branch on `main` and all commits have passing tests.

The branch has now 2 commits for easier review:
- The original changes + the fixes following your review
- The removal of `layout_dir` from `PluginEnv`, to fix a warning that flags the field as unused. See [this comment](https://github.com/zellij-org/zellij/pull/4560#discussion_r2679633917).

I can squash or reorganize them if requested.

(Disclaimer: This work was heavily AI-assisted but I reviewed every change)

> Since another release is probably due soon

[I submitted a PR to propose a feature](https://github.com/zellij-org/zellij/pull/4598). The present fix was initially a side quest from the work on that new feature. I'd love for it to land in the upcoming release so that I can retire my fork.

zellij-utils/src/input/unit/layout_test.rs
```diff
@@ -2361,3 +2361,71 @@ fn layout_node_with_name_and_only_floating_panes() {
...
+    let layout_content = r#"
+layout {
+    pane
+    pane
+}
+"#;
+    let layout_file = temp_dir.join("test-layout.kdl");
+    fs::write(&layout_file, layout_content).unwrap();
+
+    env::set_current_dir(&temp_dir).unwrap();
```
har7an, line 2386: Sorry for the confusion, but I just had trouble with this myself in another patch I'm writing. It appears that changing the `cwd` does indeed affect other tests, at least that's what I observed. So it appears we really have to restore the `cwd` as you did before.

yacinehmito: I will restore it when I do the rebase.

yacinehmito:
I really don't mind waiting. Thank you for taking a good look and for the pleasant interaction.

har7an:
@yacinehmito The changes are in, you can do the rebase now.

yacinehmito:
@har7an I rebased. The changes I had made in the separate commit were not required anymore.

har7an:
It seems a few tests are broken now, could you fix those please? Also I think you forgot to [restore the CWD][1] in one of your integration tests.

[1]: https://github.com/zellij-org/zellij/pull/4560#discussion_r2682977759

har7an:
@yacinehmito The issue you described should be fixed via #4640. Could you give it a try and report your findings please? If the issue persists, feel free to reopen this PR!

Sorry for "denying" you this contribution, but we've got a list of things to fix before the next release and this was part of it. If your issue is solved now, I'll happily take a look at your other PR though once that is rebased!

Issue #4165, opened 2025-04-28
Logs spammed with "creation time is not available on this platform currently"


valankar:
**Basic information**

`zellij --version`:
zellij 0.42.2

I'm using the binary from:

https://github.com/zellij-org/zellij/releases/download/v0.42.2/zellij-x86_64-unknown-linux-musl.tar.gz

`uname -av` or `ver`(Windows):
❯ uname -av
Linux cachyos 6.14.3-1-cachyos #1 SMP PREEMPT_DYNAMIC Mon, 21 Apr 2025 08:29:00 +0000 x86_64 GNU/Linu

Starting zellij in a clean setup (i.e. no `.config/zellij` directory) is resulting in log spam of:

```
ERROR  |zellij_server::background| 2025-04-28 06:21:15.881 [async-std/runti] [zellij-server/src/background_jobs.rs:536]: Failed to read created stamp of resurrection file: Error { kind: Unsupported, message: "creation time is not available on this platform currently" }
ERROR  |zellij_server::background| 2025-04-28 06:21:15.881 [async-std/runti] [zellij-server/src/background_jobs.rs:536]: Failed to read created stamp of resurrection file: Error { kind: Unsupported, message: "creation time is not available on this platform currently" }
ERROR  |zellij_server::background| 2025-04-28 06:21:15.881 [async-std/runti] [zellij-server/src/background_jobs.rs:536]: Failed to read created stamp of resurrection file: Error { kind: Unsupported, message: "creation time is not available on this platform currently" }
ERROR  |zellij_server::background| 2025-04-28 06:21:15.881 [async-std/runti] [zellij-server/src/background_jobs.rs:536]: Failed to read created stamp of resurrection file: Error { kind: Unsupported, message: "creation time is not available on this platform currently" }
ERROR  |zellij_server::background| 2025-04-28 06:21:15.881 [async-std/runti] [zellij-server/src/background_jobs.rs:536]: Failed to read created stamp of resurrection file: Error { kind: Unsupported, message: "creation time is not available on this platform currently" }
ERROR  |zellij_server::background| 2025-04-28 06:21:15.881 [async-std/runti] [zellij-server/src/background_jobs.rs:536]: Failed to read created stamp of resurrection file: Error { kind: Unsupported, message: "creation time is not available on this platform currently" }
[...]

PR #4579, opened 2026-01-03
build: Update Rust toolchain to 1.92.0

har7an:
since the previous version 1.90 is EOL. From reading the changelogs I didn't spot any changes that may be problematic for us, but I'll stick around in case things start acting weird.

On a related note: The cargo book now has a [chapter about build performance optimizations][1] that might be worthwhile to study. From a quick glance we already reap most of the benefits mentioned there, but it won't hurt to take a closer look. This is something for after the next release though as I don't want to debug the impact on end user builds...

[1]: https://doc.rust-lang.org/stable/cargo/guide/build-performance.html

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 * feat: command sequences, conditionally blocking CLI commands, new plugin APIs (https://github.com/zellij-org/zellij/pull/4546)
 * feat: read-only web authentication-tokens (https://github.com/zellij-org/zellij/pull/4551)
 * feat: allow overriding layouts at runtime (https://github.com/zellij-org/zellij/pull/4566)
+* build: Update Rust toolchain to 1.92.0 (https://github.com/zellij-org/zellij/pull/4579)
 
 ## [0.43.1] - 2025-08-08
 * fix: pane rename backspace regression (https://github.com/zellij-org/zellij/pull/4346)


Zykino:
This should make some peoples really happy see #4165.

Also I don’t see an MSRV policy for zellij. Neither in the readme nor on contributing. Is the policy "keeping a working version and only updating once the version becomes EOL"?

I’ve seen multiples that may suit more or less zellij’s development as I understand it:
* keep on oldest possible at all time
* update at the same time as Debian
* update only at editions (after the dust settle a bit maybe)
* update as needed.

I understand this is not directly a comment to your PR but it looks like you kind of know when updating the toolchain is important, so stating this somewhere would be interesting (so I can reference there when peoples ask for an update). Even if the policy is kind of nebulous like "as needed".

har7an:
> This should make some peoples really happy see https://github.com/zellij-org/zellij/issues/4165.

I see, that's an interesting effect. But the issue should be gone with any Rust toolchain >= 1.87, which is already the case (it's stuck with 1.90 atm).

> Also I don’t see an MSRV policy for zellij. Neither in the readme nor on contributing. Is the policy "keeping a working version and only updating once the version becomes EOL"?

That's a difficult question. We don't really have a well-defined policy and there has been a bit of internal discussion about handling this in the past as well.

Our main goal is to write a Terminal Workspace (or Terminal Multiplexer, if you prefer that). The toolchain is mostly a tool to achieve that goal. So in principle our preferred approach is "use any Rust version that works and keep using that". However, we allow (and advertise) installing zellij using `cargo install ...`. The issue is that (as far as I'm aware), we have absolutely no means to influence what Rust version is used when you `cargo install` things locally. Which means it will use whatever toolchain you have installed.

In my experience, this tends to be a recent toolchain rather than something old. Also in the past (at least as far as I know) we've had more issues with people using "more recent" toolchains than "older toolchains". Hence, I've been leaning more towards keeping the toolchain updated in the past months.

But I think your comment is a good opportunity to write down a couple of thoughts about the whole MSRV thing and maybe revise a few past decisions. So I hope you don't mind if I perform a little brain dump here.

W.r.t. your proposals:

> - keep on oldest possible at all time

This is bound to cause trouble with users that use more modern toolchains and `cargo install` the package.

> - update at the same time as Debian

I don't like pulling in that kind of dependency, because it raises so many follow-up questions: Why put that kind of focus on Debian and not `$MY_FAVORITE_DISTRO`? What motivates Debian to updating the toolchain version and how/why should that influence us? Also I have no idea of whether/how Debian handles bugs or backports in their Rust toolchains, i.e. is the vanilla Rust toolchain for version x.y **really the same** as what Debian ships as x.y? My experience tells me it probably isn't in some subtle ways. (side-note: I don't think any of the maintainers use Debian themselves)

> - update only at editions (after the dust settle a bit maybe)

Editions are few and far between and as #4165 demonstrates, things break from time to time. Although I admit things break at regular intervals even if we keep the toolchain updated, but differently. Note that we're still using the 2021 edition. I attempted switching to 2024 out of curiosity, but the amount of breakages (caused by downstream dependencies etc.) was so absurd I don't think this will be a thing in the near future.

> - update as needed.

Sure, but what constitutes a "need"? If we wait for bugs like #4165 to show up before updating, that means users are stuck with a (potentially) broken version of zellij until we push another update. If in the meantime a large PR lands on main, this puts us in a difficult position: We can either update the toolchain on `main` and make a new release including a feature that may have undergone minimal testing. Or we pick up a branching model where we go back to the last release version, branch-off of that only to update the toolchain, make a release and then merge that back into `main`. This is possible for sure, it's just more work on the maintainers and, personally, I think it's hard to keep track of such branches on the side while developing features.

> I understand this is not directly a comment to your PR but it looks like you kind of know when updating the toolchain is important, so stating this somewhere would be interesting (so I can reference there when peoples ask for an update). Even if the policy is kind of nebulous like "as needed".

It's not so much a matter of "knowing", really. It's more a matter of getting around to do it properly. I decided some time in September or October (I think) to track the issue of toolchain updates more seriously. Now we're here. :-)


## A different approach to Rust version handling

So I think we can agree that, for some reason or another, the current situation isn't optimal. Also, I want to make this very clear, I don't have a lot of evidence to go on. The current "MSRV" policy is mostly caused by the fact that we allow (and always have allowed) people to install zellij with `cargo install ...`. How many people actually do that I cannot tell (*side-note: Downloads on crates-io as far as I know are also caused by e.g. third-parties crawling the website for research, security or perhaps even ML purposes nowadays*). Much less I can tell what Rust toolchain versions they use.

I've often thought in the past that some of our "problems" or creative workarounds would magically go away if only we didn't have to support the `cargo install ...` method of installation. That includes:

- Toolchain management: If `cargo install ...` wasn't a thing any more, we could gain full control of the toolchain version. Source installation (from the repo) would still be possible, but it would (according to our "design") require a very specific Rust toolchain (see `rust-toolchain.toml`). In a more extreme scenario, we could demand that zellij be compiled with one specific Rust toolchain version (nothing newer or older allowed) using something like `compile_error!` and maybe a feature like `allow-incompatible-rust-versions` to bypass that.
- (Binary) assets: Right now we have to pre-pack all binary assets (*plugins*, but also configs, themes, ...) into the bundle we upload to `crates.io`. Hence, `zellij-utils` has become a sort of "carrier" for things that aren't source code, simply because we have no other way to ensure that the pre-built plugins are available in the final application. `crates.io` reports that `zellij-utils` v0.43.1 has a size of 5.21 MiB. The actual `zellij` crate weighs in at only 46.8 KiB.
- *I'm sure I forgot a third one...*

On a related note I've had a look at a few other Rust binaries I'm using. Their MSRV policies seem to be:

- `git-cliff`: stable
- `ripgrep`: Has an MSRV, but development targets stable
- `fd-find`: Has an MSRV, but development targets 1.90
- `bat`: 1.87 with a note anyone can bump it as required
- `topgrade`: 1.84.1

I can't make out an established versioning scheme here, and I'm not aware of one from past experience. So it seems making up our own is fine.

Now what to do? I think we could:

1. Stick with the status quo, perform regular toolchain updates and see where it leads us.
2. Use an old toolchain (or stop updating ours now) but update CI so it keeps running (at regular intervals) a build with `stable` and notifies us if things break there
3. Use an old toolchain and forbid installation using `cargo install ...`, then update whenever we have a compelling reason (CVE, bugs, ...)
4. Use evidence for updating all our components: Check out something like `ossf` and their corresponding GHA job to get feedback about toolchains and dependencies that have known vulnerabilities, then decide on a case-by-case basis which affect us and act accordingly.

Personally I lean towards option 1, which is why I'm currently pursuing it. It ensures that development targets a recent toolchain, which is likely what people use when they run `cargo install ...` locally. It also ensures (hopefully) that we catch bugs from recent toolchain versions during development ourselves, rather than in "production". And it keeps the installation method of `cargo install ...` available.

Keeping the project goal in mind (see above, "Writing a Terminal Workspace"), 4 should probably be the way to go. But that involves actual work with GHA and sitting down to analyze the impact of vulnerabilities on our project. This needs time and, likely, a lot of refactoring. And I haven't done security/vulnerability assessment before, so that's a lot of uncharted territory to cover for me.

Personally I refrain from solutions involving CI since, as much as I try, I can't motivate myself to properly pick up GHA (I develop my own projects on GitLab). This has become even worse since GitHub docs, apparently, defaults to forwarding search queries to Copilot if no result is available. That's not a good reason of course, but that's how the situation is for me. :/


Getting back to your comment: Much of this probably doesn't answer your question, and I agree it might be a good idea to write down our MSRV approach somewhere. I'll draft a paragraph and put it up here for discussion. In case I don't get around to it, maybe we should keep 1.90 for the next release and then decide on something sensible (i.e. something with an actual *reason* behind it) after that.

har7an:
A little later than expected, but I have drafted a paragraph for the `CONTRIBUTING.md`. It needs a little more work, I'll open the PR tomorrow. :-)

Thanks for bringing it to my attention @Zykino !

Zykino:
Thanks a lot to you for this lengthy response! Even if I what I think can easily be passed to new users would be:

> Zellij now try to stay up to date with the MSRV, but it may lag a bit. Especially a lot of attention is made on manual testing so closer to the following release the rust version will be freeze ahead of time. For more info go to the `CONTRIBUTING.md` and for ever more details see [this message](https://github.com/zellij-org/zellij/pull/4579#issuecomment-3707789907)

Not sure it will happen often that issues are because of rust version.

PR #4581, opened 2026-01-03
feat: propagate terminal bell from inactive tabs and show visual indicator

### LINKED ISSUE #3245: Tab-bar inactive tab does not pass bell
I have the same problem as described in : Implement passing the bell signal to the terminal #981 
Issue seems to be solved the bell is coming from a pane from active tab. But does work if bell is coming from an inactive (unfocused) tab.

mikekelly:
## Summary

Fixes #3245

- Propagates terminal bell signals (BEL character) from panes in **inactive tabs** to the terminal emulator
- Adds a visual indicator (`" *"`) to tab names in the tab-bar/compact-bar when a bell occurs in a background tab
- Clears the indicator when the user switches to that tab

## Background

Previously, bell signals from inactive tabs were silently dropped, making it impossible to receive notifications from background processes (e.g., long-running builds, AI coding agents, email clients).

## Implementation

- **Server:** Drain pending bells from ALL tabs in `render_to_clients()`, not just the active tab
- **Tab state:** Added `has_pending_bell_notification` field to track bell state per tab
- **Plugin API:** Added `has_bell` field to `TabInfo` struct and protobuf schema
- **Plugins:** Updated `tab-bar` and `compact-bar` to display `" *"` suffix on tabs with pending bells

## Test plan

- [x] Added unit test `bell_in_inactive_tab_sets_has_bell_flag` 
- [x] All existing tests pass (`cargo xtask test`)
- [x] Manual testing: bell sound plays AND visual indicator appears when bell triggered in inactive tab
- [x] Visual indicator clears when switching to the affected tab

### DIFF
--- a/zellij-server/src/background_jobs.rs
+++ b/zellij-server/src/background_jobs.rs
@@ -64,6 +64,7 @@ pub enum BackgroundJob {
         BTreeMap<String, String>, // context
     ),
     HighlightPanesWithMessage(Vec<PaneId>, String),
+    HighlightPanesForBell(Vec<PaneId>),
     RenderToClients,
     QueryZellijWebServerStatus,
     Exit,
@@ -88,6 +89,9 @@ impl From<&BackgroundJob> for BackgroundJobContext {
             BackgroundJob::RenderToClients => BackgroundJobContext::ReportPluginList,
             BackgroundJob::HighlightPanesWithMessage(..) => {
                 BackgroundJobContext::HighlightPanesWithMessage
+            },
+            BackgroundJob::HighlightPanesForBell(..) => {
+                BackgroundJobContext::HighlightPanesWithMessage // reuse same context
             },
             BackgroundJob::QueryZellijWebServerStatus => {
                 BackgroundJobContext::QueryZellijWebServerStatus
@@ -487,6 +491,18 @@ pub(crate) fn background_jobs_main(
                         task::sleep(std::time::Duration::from_millis(FLASH_DURATION_MS)).await;
                         let _ = senders.send_to_screen(
                             ScreenInstruction::ClearPaneFrameColorOverride(pane_ids),
+                        );
+                    }
+                });
+            },
+            BackgroundJob::HighlightPanesForBell(pane_ids) => {
+                // Unlike HighlightPanesWithMessage, this doesn't auto-clear.
+                // The highlight persists until the pane is focused.
+                task::spawn({
+                    let senders = bus.senders.clone();
+                    async move {
+                        let _ = senders.send_to_screen(
+                            ScreenInstruction::AddHighlightPaneFrameColorOverride(pane_ids, None),
                         );
                     }
                 });
--- a/zellij-server/src/screen.rs
+++ b/zellij-server/src/screen.rs
@@ -1604,15 +1604,16 @@ impl Screen {
 
             // Drain pending bells from ALL tabs (including inactive ones) and send to all clients.
             // This ensures bells from background tabs are propagated to the terminal.
-            // Also add visual indicators to unfocused panes with bells.
+            // Also collect panes that need bell highlighting (unfocused panes with bells).
             let mut has_bell = false;
+            let mut panes_to_highlight: Vec<PaneId> = Vec::new();
             let all_client_ids: Vec<ClientId> =
                 self.connected_clients.borrow().keys().copied().collect();
             for (tab_index, tab) in self.tabs.iter_mut() {
                 let panes_with_bells = tab.drain_pending_bells();
                 if !panes_with_bells.is_empty() {
                     has_bell = true;
-                    // Add bell highlight to unfocused panes in this tab
+                    // Collect unfocused panes that need bell highlighting
                     // A pane is considered unfocused if no client has it as their active pane
                     for pane_id in panes_with_bells {
                         let is_focused_by_any_client = all_client_ids.iter().any(|client_id| {
@@ -1624,10 +1625,16 @@ impl Screen {
                                 && tab.get_active_pane_id(*client_id) == Some(pane_id)
                         });
                         if !is_focused_by_any_client {
-                            tab.add_highlight_pane_frame_color_override(pane_id, None, None);
+                            panes_to_highlight.push(pane_id);
                         }
                     }
                 }
+            }
+            // Send bell highlight via background job to avoid render races
+            if !panes_to_highlight.is_empty() {
+                let _ = self.bus.senders.send_to_background_jobs(
+                    BackgroundJob::HighlightPanesForBell(panes_to_highlight),
+                );
             }
             if has_bell {
                 // Send the bell character (BEL, \x07) to all connected clients


Zykino:
You looks to have made a lot of work/change.

Can you help me understand how this is different from this implementation done from a plugin author with current’s zellij: https://github.com/cristiand391/zj-status-bar?tab=readme-ov-file#tab-alerts

(If you did not knew, we have a list of community made [awesome plugins](https://github.com/zellij-org/awesome-zellij), we try to only provide basic functionality with the default plugins and let the community create more elaborate ones as they see fit.)

mikekelly:
@Zykino it's much different - this doesn't require any special usage like prefixing commands with `zw` - it visually represents standard alerts from terminal sessions in unfocused tabs, which seems in line with basic functionality (and usability)? Even if I wanted to, I couldn't achieve this simply with a plugin because it requires extending `plugin_api/event.proto` to expose the necessary information.

PabloLION:
Need this feature :) Nice work.

PabloLION:
coming from #4595, I'm thinking whether it's better to include "pane-level granularity" here? or maybe in another PR (possibly based on this one) and implement the "pane-level granularity". 
For pane bells, apart from adding visual indicator, changing the border style (color) of the pane is also an option.

mikekelly:
> coming from #4595, I'm thinking whether it's better to include "pane-level granularity" here? or maybe in another PR (possibly based on this one) and implement the "pane-level granularity". For pane bells, apart from adding visual indicator, changing the border style (color) of the pane is also an option.

awesome sounds good - if you want to add some more on top definitely create a PR and add some commits!

PabloLION:
> add some more on top definitely create a PR and add some commits

I'm waiting for this to merge happily.

default-plugins/tab-bar/src/tab.rs
```diff
@@ -102,6 +102,10 @@ pub fn tab_style(
     } else if tab.is_sync_panes_active {
         tabname.push_str(" (SYNC)");
     }
+    // Show bell indicator for inactive tabs with pending bell notifications
+    if tab.has_bell && !tab.active {
+        tabname.push_str(" *");
```
imsnif: Let's change this to `tabname.push_str(" [!]");`

zellij-server/src/unit/screen_tests.rs
```diff
@@ -4200,3 +4200,96 @@ pub fn send_cli_change_floating_pane_coordinates_action() {
...
+    for (i, tabs) in tab_updates.iter().enumerate() {
+        eprintln!(
+            "Update {}: {:?}",
+            i,
+            tabs.iter()
+                .map(|t| (t.position, t.name.clone(), t.active, t.has_bell))
+                .collect::<Vec<_>>()
+        );
+    }
+    eprintln!("=== END TAB UPDATES ===");
```
imsnif: Let's please remove the debugging block

mikekelly:
Thanks for the review! I've addressed the feedback:

**Changes made:**

1. ✅ Changed bell indicator from `" *"` to `" [!]"` (in both tab-bar and compact-bar)

2. ✅ Removed debugging block from the test

3. ✅ Added pane-level bell notification - unfocused panes that receive a bell now get a frame highlight, which clears when the pane is focused (following the `HighlightAndUnhighlightPanes` pattern as suggested)

4. ✅ Rebased from main

5. ✅ Made the bell indicator configurable via plugin configuration. Users can customize it in their layout:
   ```kdl
   plugin location="tab-bar" {
       bell_indicator " *"
   }
   ```
   Or disable it entirely with `bell_indicator ""`

The default indicator remains `" [!]"`.

imsnif:
>     3. ✅ Added pane-level bell notification - unfocused panes that receive a bell now get a frame highlight, which clears when the pane is focused (following the `HighlightAndUnhighlightPanes` pattern as suggested)

Have you tried this? Doesn't seem to work for me.

schrieveslaach:
Sorry for starting a side discussion. I was looking if zellij could support notification ANSI escape sequences (like [in Wezterm](https://wezterm.org/config/lua/config/notification_handling.html)):

```bash
$ printf "\e]777;notify;%s;%s\e\\" "title" "body"
$ printf "\e]9;%s\e\\" "hello there"
```

I already looked at the code and I was wondering if the [`ring_bell: bool` in grid.rs](https://github.com/zellij-org/zellij/blob/cdbdf2b6679c580d4b0c8332694cd8ee9117d49f/zellij-server/src/panes/grid.rs#L350) should be then refactored into an enum:

```rust
#[derive(Clone)]
pub struct Grid {
    // …
    // instead of ring_bell
    pub pending_notifications: Vec<AnsiNotification>,
    // …
}

enum AnsiNotification {
    RingBell,
    Notification {
        title: Option<String>,
        body: String
    }
}
```

Let me know if you are open for a PR that moves into that direction once this PR has been merged.

imsnif:
> Sorry for starting a side discussion. I was looking if zellij could support notification ANSI escape sequences (like [in Wezterm](https://wezterm.org/config/lua/config/notification_handling.html)):
> 
> ```shell
> $ printf "\e]777;notify;%s;%s\e\\" "title" "body"
> $ printf "\e]9;%s\e\\" "hello there"
> ```
> 
> I already looked at the code and I was wondering if the [`ring_bell: bool` in grid.rs](https://github.com/zellij-org/zellij/blob/cdbdf2b6679c580d4b0c8332694cd8ee9117d49f/zellij-server/src/panes/grid.rs#L350) should be then refactored into an enum:
> 
> ```rust
> #[derive(Clone)]
> pub struct Grid {
>     // …
>     // instead of ring_bell
>     pub pending_notifications: Vec<AnsiNotification>,
>     // …
> }
> 
> enum AnsiNotification {
>     RingBell,
>     Notification {
>         title: Option<String>,
>         body: String
>     }
> }
> ```
> 
> Let me know if you are open for a PR that moves into that direction once this PR has been merged.

Hey, thanks for the input. Would be happy to consider this (even though I imagine there will be considerably more work involved in getting this to work), but let's not conflate things. Let's get this merged.

mikekelly:
> > ```
> > 3. ✅ Added pane-level bell notification - unfocused panes that receive a bell now get a frame highlight, which clears when the pane is focused (following the `HighlightAndUnhighlightPanes` pattern as suggested)
> > ```
> 
> Have you tried this? Doesn't seem to work for me.

yeah I smoke tested it and it seems to work fine for me? 
image

imsnif:
> > > ```
> > > 3. ✅ Added pane-level bell notification - unfocused panes that receive a bell now get a frame highlight, which clears when the pane is focused (following the `HighlightAndUnhighlightPanes` pattern as suggested)
> > > ```
> > 
> > 
> > Have you tried this? Doesn't seem to work for me.
> 
> yeah I smoke tested it and it seems to work fine for me? image

Doesn't for me :) Could be a matter of render races. Can you please use the background jobs method like I linked to? That should sort it.

mikekelly:
@imsnif apologies missed that suggestion - addressed in [0cce720](https://github.com/zellij-org/zellij/pull/4581/commits/0cce7200e2d561968b5a2bd3327c84ae0c4aef4c)

nerdrew:
Question about this feature: does the bell notification show up in `zellij ls` output?

imsnif:
@mikekelly - the pane indication is still not working for me:


https://github.com/user-attachments/assets/e65a255c-6d47-46d7-b98e-c1c2b1936a83

mikekelly:
@imsnif is it ok if you explicitly describe what behaviour your'e expecting. It's designed to highlight a pane on the other tab if it's not in focus - in your demo when you switch it briefly highlights the pane but it was left as the pre-selected pane on the tab so it quickly shifts back to the focused highlighting. This was an intentional design decision, are you saying that behaviour is wrong?

imsnif:
In short: we should get a visual indication even if we have just one tab and are focused on another pane.

mikekelly:
Ok, and in the case of multiple tabs - if notified pane in a background tab was left as the selected pane, when the tab is opened should that pre-selected pane show with bel notification highlight or focused highlight? At the moment it briefly shows as bel notified highlighting and then switches to focused highlighting.

imsnif:
Good point. I think the best case would be the same behavior as the Tab. We add `[!]` to its title which gets cleared as soon as it's focused. I do also want to have the highlight effect though if we're already focused on this tab.

mikekelly:
Ah ok so not just colour highlighting around the pane - each _pane_ title should also have the notification label appended to it?

Raises another question - should the tab retain its notification until all of the notified panes have been cleared with a focus? Or should the tab lose its notification label as soon as its been opened regardless of whether the panes haven't had their notification status cleared?

imsnif:
The color highlight is not really noticeable if you're not focused on the tab. Even if it just appears as you focus and disappears 400ms later, the user is usually too busy parsing the tab's contents to figure out something is flashing off. So it should only happen if a user is focused on the same tab.

The `[!]` indications on the pane are additions. I think it's safe to remove the indication on the tab as soon as we're focused on it, regardless of if we've focused on the pane(s) with the bell or not.

imsnif:
Still interested in this feature?

GeertJohan:
Was looking exactly for this, very happy to see there's being worked on it, thanks! :pray: 
I think a different color tab and pane when it was alerted will be more distinct than the `[!]` text. I have a lot of tabs with different names and [!] will probably not be notable enough unless I read the tabs one by one. A distinct color will stand out. It doesn't have to be flashing.

imsnif:
I implemented this in https://github.com/zellij-org/zellij/pull/4781

PR #4583, opened 2026-01-04
fix(security): validate session name in delete-session to prevent directory deletion

0xthierry:
## Summary

When `delete-session` receives a path instead of a session name (e.g., I lost my entire `~/Work/Sideprojects` folder due to a shell alias conflict), `PathBuf::join()` replaces the cache directory base path entirely, causing `remove_dir_all()` to delete the user's directory.

  ### The bug

  Expected: ~/.cache/zellij/.../session_info/my-session
  Actual:   /home/user/some/path  (base path replaced!)

  1. `session_info_folder_for_session()` calls `ZELLIJ_SESSION_INFO_CACHE_DIR.join(session_name)`
  2. Rust's `PathBuf::join()` with an absolute path **replaces the entire base path**
  3. `std::fs::remove_dir_all()` then deletes the absolute path directly

  ### The fix

  Add `validate_session_name()` to reject session names containing `/`, `.`, `..`, or empty strings - the same validation that already exists for session creation in `assert_session_ne()`.

  ## Test plan

  - [x] `cargo test --lib -p zellij-utils` - 208 tests pass
  - [x] `cargo fmt` - formatted
  - [ ] `zellij delete-session "/tmp/test"` - rejects with "Session name cannot contain '/'"
  - [ ] `zellij delete-session ".."` - rejects with "Invalid session name"
  - [ ] `zellij delete-session ""` - rejects with "Session name cannot be empty"

### DIFF
--- a/zellij-utils/src/sessions.rs
+++ b/zellij-utils/src/sessions.rs
@@ -430,7 +430,9 @@ pub fn assert_dead_session(name: &str, force: bool) {
 
 pub fn validate_session_name(name: &str) -> Result<(), String> {
     if name.trim().is_empty() {
-        return Err("Session name cannot be empty. Please provide a specific session name.".to_string());
+        return Err(
+            "Session name cannot be empty. Please provide a specific session name.".to_string(),
+        );
     }
     if name == "." || name == ".." {
         return Err(format!("Invalid session name: \"{}\".", name));


imsnif:
Thanks! Could you make the function return a `Result<(), String>` and then do an `if let Err(e) = ...`. I usually don't nitpick to this level, but in this case I think returning errors as an option would be a bit confusing.

0xthierry:
Thank you for the review @imsnif

imsnif:
Hey @0xthierry - just a quick note, could you make the error a `String` rather than a `&str` and then return more descriptive errors such as we had before? (eg. `eprintln!("Invalid session name: \"{}\".", name);`)

Thanks!

imsnif:
Thank you!

PR #4585, opened 2026-01-06
docs(CONTRIBUTING): Describe rust toolchain versioning policy

har7an:
and mention the lack of a MSRV. This should allow people to discover how toolchain updates are handled, even though there currently isn't much of an explanation.

For a more detailed explanation on the background of this decision, see [this comment][1]

[1]: https://github.com/zellij-org/zellij/pull/4579#issuecomment-3707789907


har7an:
@Zykino What do you think about this proposal? Does that make the situation clearer?

Zykino:
Thanks, it is a good message that I can point to peoples if the need to "be on the latest rust version" is needed again. I will try to remember to ping you in the issue if it ever happens again :)

PR #4593, opened 2026-01-09
fix: make `TerminalAction::OpenFile` also inherit client/pane cwd

### LINKED ISSUE #4592: Open scrollback editor with the panel's current working directoy
Sometimes I use "Edit scrollback in default editor" with Neovim to save the logs of the last command I ran. However, when Neovim is spawned, its working directory is not set to the `cwd` of the current pane, but instead to the working directory of the Zellij process (which is `~` in my case). As a result, I need to navigate back to my project directory before saving the scrollback.

I would like the editor to open in the same `cwd` as the current pane.

In #4593, I implement this behavior, but the way I implemented it may affect more than just scrollback editing. So, I am opening this issue for discussion if necessary.

Rodrigodd:
Fix #4592, but I am not sure if this is the proper fix. The previous code appeared to explicitly exclude TerminalAction::OpenFile in the fill_cwd function, and I do not know what other features this change might affect besides the scrollback editor.

---

If a `TerminalAction` does not currently have a specified working directory, it will be set to the cwd of the client/pane that triggered the action. Previously this was only done for `TerminalAction::RunCommand`, with this change it is now also done for `TerminalAction::OpenFile`.

This fix the issue where the "edit scrollback" action would not inherit the working directory from the pane, but from the zellij process.

### DIFF
--- a/zellij-server/src/pty.rs
+++ b/zellij-server/src/pty.rs
@@ -945,47 +945,51 @@ impl Pty {
         }
     }
     fn fill_cwd(&self, terminal_action: &mut TerminalAction, client_id: ClientId) {
-        if let TerminalAction::RunCommand(run_command) = terminal_action {
-            if run_command.cwd.is_none() {
-                run_command.cwd = self
-                    .active_panes
-                    .get(&client_id)
-                    .and_then(|pane| match pane {
-                        PaneId::Plugin(plugin_id) => self.plugin_cwds.get(plugin_id).cloned(),
-                        PaneId::Terminal(id) => {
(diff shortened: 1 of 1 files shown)


imsnif:
Hey @Rodrigodd - I did a brief merge from main but I think there are some minor fixes beyond the syntax error I introduced (sorry!) - would you like to patch these up and I'll merge?

Rodrigodd:
@imsnif I rebased the changes. Thanks for taking a look!

imsnif:
Works great, thank you!

PR #4594, opened 2026-01-09
feat: add remote pane/tab renaming and dump-layout --with-ids

### LINKED ISSUE #4591: zellij action to rename other panes and tabs
zellij's has the ability to programmatically rename panes and tabs with `zellij action rename-pane/tab`. this ability, however, is limited to the currently active pane/tab only.

when executing long-running scripts on many tabs and panes, each one requiring my attention at another moment, it could be really beneficial to make the script run something like `zellij action rename-pane 👇 $ZELLIJ_PANE_ID` to catch my attention to the correct pane.

additionally - having the ability to do the same for renaming tabs could be even more beneficial but is slightly harder as the env vars currently only reveal the pane_id, not the tab id.

my suggestion is to:
- add an optional pane_id parameter to the `rename-pane` subcommand to rename a remote pane
- add an optional tab_index parameter to the `rename-tab` subcommand to rename a remote tab
- add a `--with-ids` optional flag to the `dump-layout` subcommand to allow yielding the tab's index from the pane's id (after parsing the kdl tree)

demo:
[asciicast](https://asciinema.org/a/0P6L0gud6M3lpLRn)

thank you for considering my suggestion and thank you even more for a terrific utility. I truly enjoy it a lot

abloch:
## Summary

This PR implements the three enhancements requested in #4591 for better scripting support:

- **`zellij action rename-pane <name> [pane_id]`** - optionally specify a pane ID to rename a non-active pane remotely
- **`zellij action rename-tab <name> [tab_index]`** - optionally specify a tab index to rename a non-active tab remotely
- **`zellij action dump-layout --with-ids`** - include pane IDs, tab indices, and `is_plugin` flags in the layout output for scripting

These features enable workflows where multiple long-running processes need to be managed and renamed without switching focus to each pane or tab individually.

## Test plan

- [x] Existing tests pass (`cargo xtask test`)
- [x] New unit test for `--with-ids` serialization added in `session_serialization.rs`
- [x] Code formatted with `cargo xtask format`
- [x] Clippy passes (`cargo xtask clippy`)

Closes #4591

demo:
[asciicast](https://asciinema.org/a/0P6L0gud6M3lpLRn)


abloch:
issue closed - PR no longer needed

PR #4596, opened 2026-01-09
Fix bash auto-start for Cosmic DE

axelbdt:
When I installed zellij on Cosmic DE (via NixOS if it matters), and added the auto-launch feature for bash, Cosmic DE tried to run the script and launch Zellij at startup, preventing the DE to start.

Adding a check for interactivity session made every thing run smoothly.

Here's a PR if interested.


erikced:
While helpful to solve, the issue here is not arguably not in zellij, `.bashrc` should only be sourced for interactive shells. Debian, Ubuntu, Arch etc add guards for this at the top of the default `.bashrc`.

axelbdt:
That's fair, thanks for the feedback.

Issue #4458, opened 2025-10-04
PR idea feedback request - extending zellij action dump-screen


emchristiansen:
Would you be interested in a PR that adds options to `zellij action dump-screen`?

1. **`--pane-id <ID>` flag** to specify which pane to capture (defaults to current pane)
   *Reason:* Focusing a pane before capture changes the user's view, which is disruptive for monitoring/automation use cases

2. **`--raw-terminal` or `--ignore-overlays` flag** to capture the base terminal content, excluding Zellij UI overlays (search mode, copy mode, etc.)
   *Reason:* I'm building an app that periodically captures TUI state, and user interactions like entering search mode shouldn't affect the captured content

Use case: Non-intrusive monitoring of TUI applications running in Zellij panes

Would this align with Zellij's goals? Happy to discuss implementation details if this sounds useful.

PR #4598, opened 2026-01-11
feat(cli): add --pane-id option to dump-screen action

yacinehmito:
Adds the ability to dump a specific terminal pane's content by passing `--pane-id <ID>` to the `dump-screen` CLI action. When omitted, the existing behavior of dumping the focused pane is preserved.

This change is relevant to better support pane content inspection by AI coding agents.

Indeed, to get the content of a specific pane with the current API, one needs to cycle through them, focus the relevant one, dump the content, then go back to the original one. This logic is shown in [zellij-pane-tracker](https://github.com/theslyprofessor/zellij-pane-tracker), whose design complexity is due to the limitation that this PR addresses.

Related to #4458.

### DIFF
--- a/docs/MANPAGE.md
+++ b/docs/MANPAGE.md
@@ -155,7 +155,9 @@ ACTIONS
   Right, Up, Down).
 * __Clear__ - clears current screen.
 * __DumpScreen: <File\> [--pane-id <ID\>]__ - dumps the pane content to the specified file.
-  If --pane-id is provided, dumps the specified terminal pane; otherwise dumps the focused pane.
+  If --pane-id is provided, dumps the specified pane; otherwise dumps the focused pane.
+  <ID\> can be a bare integer (eg. 1), a terminal pane id (eg. terminal_1) or a plugin pane id (eg. plugin_1).
+  A bare integer is equivalent to a terminal pane id with the same number.
 * __DumpLayout: <File\>__ - dumps the screen in the specified or default file.
 * __EditScrollback__ - replaces the current pane with the scrollback buffer.
 * __ScrollUp__ - scrolls up 1 line in the focused pane.


abloch:
duplicates #4594

yacinehmito:
> duplicates #4594

@abloch Why so? The changes seem complementary. Is there a way to specifically dump a pane with #4594?

abloch:
@yacinehmito 

you're right. I mistakenly got confused in this pr and read `dump-layout` instead of `dump-screen`

sorry for the confusion

basnijholt:
I am really missing this functionality!

I have some commands that automatically manage git worktrees and opens agentic coding agents in e.g., Zellij tabs (https://github.com/basnijholt/agent-cli).

The experience in Zellij (which I prefer over tmux) is suboptimal and race-y due to the requirement of first going to the tab to read its content.

I am also working on a multi-agent orchestrator where each pane/tab is its own agent. This feature will make it significantly easier to support Zellij instead of tmux only.

imsnif:
Hey @yacinehmito - thanks for this. I would definitely like to see this feature, but I'm a bit confused by this PR. Seems like we're adding and removing a whole lot of stuff that doesn't seem related to the change? (eg. removing `layout_dir` from the plugin state, adding LayoutInfo transformations and tests, etc.) could you help me understand the intention here?

yacinehmito:
Hello @imsnif.
This PR was based another (#4560) that was since closed because of conflicting changes.
I need to rework this one. Will do now. Thanks for the patience.

yacinehmito:
This is fixed. Automated tests are passing. I am still to do a full build and e2e manual test. I will let you know once done.

zellij-server/src/screen.rs
```diff
@@ -4610,19 +4610,34 @@ pub(crate) fn screen_thread_main(
...
-                        client_id,
-                        full
-                    ),
-                    ?
-                );
+                match pane_id {
+                    Some(pane_id) => {
+                        // Dump specific pane - search across all tabs
+                        for tab in screen.get_tabs_mut().values_mut() {
+                            if tab.get_pane_with_id(pane_id).is_some() {
```
imsnif: Let's use `has_pane_with_id` rather than `get_pane_with_id`

zellij-server/src/screen.rs
```diff
@@ -4610,19 +4610,34 @@ pub(crate) fn screen_thread_main(
...
+                                tab.dump_terminal_screen(Some(file.clone()), pane_id, full)?;
+                                break;
+                            }
+                        }
+                    },
+                    None => {
+                        // Existing behavior: dump focused pane
```
imsnif: I know LLMs do this, but let's please remove the extraneous comment

zellij-utils/src/cli.rs
```diff
@@ -683,13 +683,17 @@ pub enum CliAction {
     MovePaneBackwards,
     /// Clear all buffers for a focused pane
     Clear,
-    /// Dump the focused pane to a file
+    /// Dump a pane to a file. If no pane ID is specified, dumps the focused pane.
```
imsnif: Let's change this to `Dumps the viewport and optionally scrollback of a pane to STDOUT` (might as well update this to correct wording while we're here)

zellij-utils/src/input/actions.rs
```diff
@@ -657,9 +658,10 @@ impl Action {
             CliAction::MovePaneBackwards => Ok(vec![Action::MovePaneBackwards]),
             CliAction::MoveTab { direction } => Ok(vec![Action::MoveTab { direction }]),
             CliAction::Clear => Ok(vec![Action::ClearScreen]),
-            CliAction::DumpScreen { path, full } => Ok(vec![Action::DumpScreen {
+            CliAction::DumpScreen { path, full, pane_id } => Ok(vec![Action::DumpScreen {
                 file_path: path.as_os_str().to_string_lossy().into(),
                 include_scrollback: full,
+                pane_id: pane_id.map(PaneId::Terminal),
```
imsnif: This needs to behave a bit differently. We need to parse this string so that it accepts `1` (terminal with pane id 1), `terminnal_1` (same) and `plugin_1` (plugin with id 1). There are other examples of this elsewhere in the file. The `--help` of the CLI command also needs to reflect this.

imsnif:
Hey @yacinehmito - any updates on this? I don't want to rush you but I'm starting to wrap things up for the upcoming release and would like to have this included.

yacinehmito:
Sorry I missed the notification for the review. I am on it right now.

yacinehmito:
@imsnif I think I addressed all comments. Thank you for the review!

imsnif:
Hey @yacinehmito - thanks for the quick fixes! I tested this locally and works great. Only thing I'd like to change (sorry, did not notice this before) is to make `path` optional. If it is not provided, we should dump to STDOUT. I think the coding agent crowd will appreciate this behavior.

Want to do it or shall I?

imsnif:
He @yacinehmito - I implemented dumping to STDOUT locally but this PR has `maintainer_can_modify false`. Mind changing it to true so I can push my changes?

yacinehmito:
@imsnif I cannot change it as this PR was created from a repository in an organization, not a user, [as per the documentation](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork). I personally just learned about this limitation, and it sucks... Never forking in an organization anymore.

I'll implement it.

imsnif:
No worries, but no need to implement it. I already have it working and finished. I'll just do it in another PR. Will merge this one  a bit later today.

imsnif:
Thank you @yacinehmito !! I think you're going to make lots of people happy with this.

PR #4617, opened 2026-01-20
Add optional support for raw scrollback with ansi

### LINKED ISSUE #3937: EditScrollback and DumpScreen `--raw` or `--ansi` please?
Currently, there is no way to recover the ansi codes used for the current buffer/pane or window.

Use case:
You would like to share your terminal snippet that does not have markdown highlighting support on Discord. Discord now supports the `ansi` code block.

1. Enter `ScrollBack --raw` with your `$EDITOR`
2. Copy section with ansi codes
3. Put into `ansi` markdown code block. The ansi codes will correctly highlight the buffer!

I am open to workarounds or other programs. I couldn't find any :(. 

It would also be useful because you can use this to convert asni to svg, useful for ∞-res code snapshots.

lemunozm:
fixes https://github.com/zellij-org/zellij/issues/3937#issuecomment-3768482297

The current approach follows the minimum required changes to make it work. Next follow-ups or options could be:

- Give also support for `edit_scrollback_for_pane_with_id` (not sure if practically need it)
- Instead of duplicate codepaths, refactorize `edit_scrollback` and `dumb_*` methods to receive a bool `isRaw` or similar. But this could break the API and consequently some plugins.

To make this work, you need to modify in the `config.kdl` with:

```diff
-        bind "e" { EditScrollback; SwitchToMode "locked"; }
+        bind "e" { EditScrollbackRaw; SwitchToMode "locked"; }
```

A showcase of how this work (along with the `neovim` plugin `baleia`):

https://github.com/user-attachments/assets/72927098-4237-4d11-9b7f-39d8c072dbe7


lemunozm:
I'm fixing some minor visualization detail. I'll ping once I think it's done

lemunozm:
Ok, now after some minor fixes, when I open a scrollback in neovim I can see exactly the same characters & colors everywhere, even in complex scenarios. i.e: I can open a scrollback in a panel with neovim into neovim, and I see exactly the same colors.

zellij-server/src/panes/grid.rs
```diff
@@ -288,6 +288,41 @@ macro_rules! dump_screen {
...
+        for line in &$lines {
+            if line.is_canonical && !is_first {
+                buf.push_str("\u{1b}[m\n");
+                last_styles = None;
+            }
+
+            let last_non_space = line
+                .columns
+                .iter()
+                .rposition(|tc| tc.character != ' ')
```
imsnif: So, here we want the last space character that doesn't have styles (otherwise we will truncate space characters with backgrounds - a real world example is `delta`'s git diff). I recently fixed this bug in our own rendering engine if you'd like to see how it's done: https://github.com/zellij-org/zellij/pull/4680

zellij-server/src/panes/grid.rs
```diff
@@ -288,6 +288,41 @@ macro_rules! dump_screen {
...
+macro_rules! dump_screen_with_ansi {
+    ($lines:expr) => {{
+        use std::fmt::Write;
+        let mut is_first = true;
+        let mut buf = String::new();
+        let mut last_styles: Option<RcCharacterStyles> = None;
+
+        for line in &$lines {
+            if line.is_canonical && !is_first {
+                buf.push_str("\u{1b}[m\n");
```
imsnif: I think we should be able to get away with not resetting the styles here.

imsnif (commented):
Hey @lemunozm - thanks for this (and for your patience waiting for my review)!

I left some comments. Otherwise please rebase from main, fix the compilation errors and I think we should be good to go.

zellij-server/src/panes/grid.rs
lemunozm: I think you're right, the next character should add the correct style value, in all scenarios. Removed.

lemunozm:
Should be ready now!

imsnif:
> Should be ready now!

Did I miss something, or are we still missing the fix for my comment about the last-non-styled-space instead of the last space?

lemunozm:
Oops, here it is: [131c8e0](https://github.com/zellij-org/zellij/pull/4617/commits/131c8e00dac432284b929723d3b85274f2f10e98)

zellij-server/src/panes/grid.rs
```diff
@@ -304,7 +304,10 @@ macro_rules! dump_screen_with_ansi {
             let last_non_space = line
                 .columns
                 .iter()
-                .rposition(|tc| tc.character != ' ')
+                .rposition(|tc| {
+                    tc.character != ' '
+                        || !matches!(tc.styles.background, Some(AnsiCode::Reset) | None)
```
imsnif: Maybe I'm missing something, but shouldn't this be `&&`?

lemunozm: I think it's correct. That condition is equivalent to:
- `tc.character != ' '` 
or
- `tc.character == ' ' && !matches!(tc.styles.background, Some(AnsiCode::Reset) | None)` which means, a space with style is found

imsnif: Fair... can we either simplify the conditional or break it down to variables with meaningful names so that my poor brain doesn't have to parse it when troubleshooting? :)

lemunozm: Makes sense! It's not the most understandable code ever. I think now much better

imsnif:
Thanks for the refactor @lemunozm ! Could you rebase from main? That should fix the CI failures.

lemunozm:
Done!

imsnif:
Nice. Thank you!

dieggsy:
how does this actually work?

If I do e.g. `eza -l --color=always | less`, I see a different set of sequences than `eza -l --color` and then inspect the scrollback

CKolkey:
For anyone who finds this via a google search: The api was modified before release. 

Use:
```kdl
EditScrollback { ansi true; };
```

not 
```kdl
EditScrollbackRaw;
```

PR #4623, opened 2026-01-21
Fix partial sequences parsing

### LINKED ISSUE #4553: Scrolling fast causes input spam; Unhandled osc
## Issue description

I'm not entirely sure how to describe this, so I'll do my best. I am only testing this on macOS, haven't tried from Windows or a linux machine. When I scroll quickly/fast, I get this character spam input. It's most prevalent with a freespin mouse wheel, but can also get it to occur by just scrolling quickly with a standard mouse wheel. It will look something like:

```bash
<65;45;17M65;45;17M65;45;17M<65;45;17M65;45;17M<65;45;17M65;45;17M65;45;17M<65;45;17M65;45;17M65;45;17M<65;45;17M
```

The longer the spin, the more the output. Freespinning for more than a second causes it to seemingly infinitely spam the input, and ultimately have to force close the tab/pane.

I've tested with deleting my zellij configuration entirely, and using the one I have and I get the same behavior with just whatever default configurations `zellij`. Oddly, I do not see this behavior using `iTerm2`. But I do get this behavior with `Ghostty`, as well as the default macOS `Terminal` app. I do not get this behavior in any terminal with `tmux`, nor outside of a `zellij` session. I use `zsh` with `powerlevel10k`, but also tested with `bash` and also got the same behavior.

### Zellij Logs

```markdown
INFO   |zellij_client            | 2025-12-16 13:31:58.867 [main      ] [zellij-client/src/lib.rs:232]: Starting Zellij client! 
INFO   |zellij_server            | 2025-12-16 13:31:58.878 [main      ] [zellij-server/src/lib.rs:581]: Starting Zellij server! 
INFO   |zellij_server            | 2025-12-16 13:31:58.883 [main      ] [zellij-server/src/lib.rs:1761]: Compiling plugins using Cranelift 
WARN   |zellij_server::route     | 2025-12-16 13:31:58.883 [server_router] [zellij-server/src/route.rs:1068]: Server not ready, trying to place instruction in retry queue... 
WARN   |zellij_server::route     | 2025-12-16 13:31:58.883 [server_router] [zellij-server/src/route.rs:1213]: Server ready, retrying sending instruction. 
INFO   |zellij_server::plugins   | 2025-12-16 13:31:58.883 [wasm      ] [zellij-server/src/plugins/mod.rs:243]: Wasm main thread starts 
INFO   |zellij_server::plugins::p| 2025-12-16 13:31:58.889 [async-std/runti] [zellij-server/src/plugins/plugin_loader.rs:534]: Loaded plugin '/Users/username/Library/Caches/org.Zellij-Contributors.Zellij/11443609859275959894305566542538288992' from cache folder at '/Users/username/Library/Caches/org.Zellij-Contributors.Zellij' in 2.01925ms 
INFO   |wasmtime_wasi::preview1  | 2025-12-16 13:31:58.891 [async-std/runti] [/Users/brew/Library/Caches/Homebrew/cargo_cache/registry/src/index.crates.io-1949cf8c6b5b557f/wasmtime-wasi-29.0.1/src/preview1.rs:1217]: environ_sizes_get; 
WARN   |zellij_server::panes::gri| 2025-12-16 13:31:58.965 [screen    ] [zellij-server/src/panes/grid.rs:2734]: Unhandled osc: [[49, 51, 51], [65]] 
WARN   |zellij_server::panes::gri| 2025-12-16 13:31:58.965 [screen    ] [zellij-server/src/panes/grid.rs:2734]: Unhandled osc: [[49, 51, 51], [66]] 
ERROR  |zellij_utils::ipc        | 2025-12-16 13:31:59.124 [unnamed   ] [zellij-utils/src/ipc.rs:201]: Failed to flush ipc sender: Broken pipe (os error 32) 
WARN   |zellij_server::panes::gri| 2025-12-16 13:31:59.437 [screen    ] [zellij-server/src/panes/grid.rs:2734]: Unhandled osc: [[49, 51, 51, 55], [82, 101, 109, 111, 116, 101, 72, 111, 115, 116, 61, 109, 105, 108, 49, 105, 64, 109, 105, 108, 45, 109, 97, 99, 109, 105, 110, 105, 45, 109, 52, 112]] 
WARN   |zellij_server::panes::gri| 2025-12-16 13:31:59.437 [screen    ] [zellij-server/src/panes/grid.rs:2734]: Unhandled osc: [[49, 51, 51, 55], [67, 117, 114, 114, 101, 110, 116, 68, 105, 114, 61, 47, 85, 115, 101, 114, 115, 47, 109, 105, 108, 49, 105]] 
WARN   |zellij_server::panes::gri| 2025-12-16 13:31:59.437 [screen    ] [zellij-server/src/panes/grid.rs:2734]: Unhandled osc: [[49, 51, 51, 55], [83, 104, 101, 108, 108, 73, 110, 116, 101, 103, 114, 97, 116, 105, 111, 110, 86, 101, 114, 115, 105, 111, 110, 61, 49, 52], [115, 104, 101, 108, 108, 61, 122, 115, 104]] 
WARN   |zellij_server::panes::gri| 2025-12-16 13:31:59.437 [screen    ] [zellij-server/src/panes/grid.rs:2734]: Unhandled osc: [[49, 51, 51], [68], [48]] 
WARN   |zellij_server::panes::gri| 2025-12-16 13:31:59.437 [screen    ] [zellij-server/src/panes/grid.rs:2734]: Unhandled osc: [[49, 51, 51, 55], [82, 101, 109, 111, 116, 101, 72, 111, 115, 116, 61, 109, 105, 108, 49, 105, 64, 109, 105, 108, 45, 109, 97, 99, 109, 105, 110, 105, 45, 109, 52, 112]] 
WARN   |zellij_server::panes::gri| 2025-12-16 13:31:59.437 [screen    ] [zellij-server/src/panes/grid.rs:2734]: Unhandled osc: [[49, 51, 51, 55], [67, 117, 114, 114, 101, 110, 116, 68, 105, 114, 61, 47, 85, 115, 101, 114, 115, 47, 109, 105, 108, 49, 105]] 
WARN   |zellij_server::panes::gri| 2025-12-16 13:31:59.438 [screen    ] [zellij-server/src/panes/grid.rs:2734]: Unhandled osc: [[49, 51, 51], [65]] 
WARN   |zellij_server::panes::gri| 2025-12-16 13:31:59.438 [screen    ] [zellij-server/src/panes/grid.rs:2734]: Unhandled osc: [[49, 51, 51], [66]] 
WARN   |zellij_server::panes::gri| 2025-12-16 13:34:02.575 [screen    ] [zellij-server/src/panes/grid.rs:2734]: Unhandled osc: [[49, 51, 51], [65]] 
WARN   |zellij_server::panes::gri| 2025-12-16 13:34:02.575 [screen    ] [zellij-server/src/panes/grid.rs:2734]: Unhandled osc: [[49, 51, 51], [66]] 
WARN   |zellij_server::panes::gri| 2025-12-16 13:34:02.587 [screen    ] [zellij-server/src/panes/grid.rs:2734]: Unhandled osc: [[49, 51, 51], [67], []] 
WARN   |zellij_server::panes::gri| 2025-12-16 13:34:02.587 [screen    ] [zellij-server/src/panes/grid.rs:2734]: Unhandled osc: [[49, 51, 51], [68], [49, 51, 48]] 
WARN   |zellij_server::panes::gri| 2025-12-16 13:34:02.590 [screen    ] [zellij-server/src/panes/grid.rs:2734]: Unhandled osc: [[49, 51, 51, 55], [82, 101, 109, 111, 116, 101, 72, 111, 115, 116, 61, 109, 105, 108, 49, 105, 64, 109, 105, 108, 45, 109, 97, 99, 109, 105, 110, 105, 45, 109, 52, 112]] 
WARN   |zellij_server::panes::gri| 2025-12-16 13:34:02.590 [screen    ] [zellij-server/src/panes/grid.rs:2734]: Unhandled osc: [[49, 51, 51, 55], [67, 117, 114, 114, 101, 110, 116, 68, 105, 114, 61, 47, 85, 115, 101, 114, 115, 47, 109, 105, 108, 49, 105]] 
WARN   |zellij_server::panes::gri| 2025-12-16 13:34:02.593 [screen    ] [zellij-server/src/panes/grid.rs:2734]: Unhandled osc: [[49, 51, 51], [65]] 
WARN   |zellij_server::panes::gri| 2025-12-16 13:34:02.593 [screen    ] [zellij-server/src/panes/grid.rs:2734]: Unhandled osc: [[49, 51, 51], [66]] 
WARN   |zellij_server::panes::gri| 2025-12-16 13:34:03.968 [screen    ] [zellij-server/src/panes/grid.rs:2734]: Unhandled osc: [[49, 51, 51], [65]] 
WARN   |zellij_server::panes::gri| 2025-12-16 13:34:03.968 [screen    ] [zellij-server/src/panes/grid.rs:2734]: Unhandled osc: [[49, 51, 51], [66]] 
WARN   |zellij_server::panes::gri| 2025-12-16 13:34:03.968 [screen    ] [zellij-server/src/panes/grid.rs:2734]: Unhandled osc: [[49, 51, 51], [67], []] 
INFO   |zellij_server::plugins::w| 2025-12-16 13:34:04.002 [wasm      ] [zellij-server/src/plugins/wasm_bridge.rs:321]: Bye from plugin 0 
INFO   |zellij_client            | 2025-12-16 13:34:04.002 [main      ] [zellij-client/src/lib.rs:669]: Bye from Zellij! 
```

## Minimal reproduction

1. Open Ghostty or Terminal on macOS
2. Scroll down as fast as you can

Image

### MacOS

macOS Tahoe 26.2 (has been occurring for many OS versions)

### Zellij

zellij 0.43.1

### Ghostty

Ghostty 1.2.3

Version
  - version: 1.2.3
  - channel: stable
Build Config
  - Zig version   : 0.14.1
  - build mode    : builtin.OptimizeMode.ReleaseFast
  - app runtime   : apprt.Runtime.none
  - font engine   : font.main.Backend.coretext
  - renderer      : renderer.generic.Renderer(renderer.Metal)
  - libxev        : kqueue

jgiannuzzi:
Fixes #4553 (and probably others).

~_Currently set as draft because it relies on an unreleased version of `termwiz`_~

This PR adds non-blocking stdin debounce for fragmented input handling. We always parse with `maybe_more = true` and send events immediately. Finalization is deferred until 50ms of inactivity, avoiding latency while correctly handling escape sequences split across multiple reads (e.g., over SSH).


imsnif:
Thanks for the fix and your continuing efforts! I understand the approach with the poll and can see how it might be the path of least resistance. My issue here though is that this is a legacy piece of code that we're trying to move past (afaik the only place that uses it is the mouse-hold polling, and now that we support AnyEvent we can probably get rid of that too).

The direction I'd like to go is having an async `select!` that listens to several channels (two in this case: proper STDIN and a debounced finalization event). Something like: https://github.com/zellij-org/zellij/blob/main/zellij-client/src/lib.rs#L313

Do you think it can work?

jgiannuzzi:
Ah I think I finally understand what you meant! I will try to solve it this way in the coming days. Should I also try to get rid of the mouse-hold polling, or is it ok to keep it around for now?

imsnif:
Nice! My apologies for not being clearer earlier.

Thank you for your offer to also take care of the mouse polling! I'd be happy to take you up on it, but since that one will require lots of manual testing, let's separate it to a separate future PR so as not to make this very vital fix dependent on it.

imsnif:
Also, forgot to mention: in my mind this should still not be a very big fix. I think most code can be copied as is, with these changes essentially acting as a wrapper. If I'm wrong and you find yourself going down a rabbit-hole of complexity, I urge you to stop and consult me so as not to do extra work.

jgiannuzzi:
to run the stdin loop asynchronously, should I create a tokio runtime in a thread, or should I spawn a new task on the global runtime?

imsnif:
I think we should be able to get away with using the global runtime. If we receive messages on something other than STDIN or the debounce though, we should probably use a dedicated thread (I don't remember atm but pretty sure we don't).

jgiannuzzi:
I thought I had this working, but it turns out that the session switching is now broken. I'll try something else!

imsnif:
Heads-up - I removed the mouse polling during some related work. You might want to rebase from main.

imsnif:
He @jgiannuzzi - I don't mean to stress but I'm starting to close in on a release in the not-too-far future (I hope). What's the status here?

jgiannuzzi:
Hey @imsnif, I haven't had a chance to continue on this but I will prioritize it as I'd love to get it into the next release! We do depend on a release of `termwiz` with my fix from https://github.com/wezterm/wezterm/pull/7504 so we may want to ask Wez to cut a new one.

imsnif:
I don't think we're going to get a new release with this, so let's vendor the dep with the fix for now. Otherwise, what's still left to be done here? I'd like to see this fix in the upcoming version release, so if you don't have the time I can pick this up.

jgiannuzzi:
I have got this mostly working in a local branch. The fragmented sequences are handled correctly, the session switch works fine, but I have one remaining issue with losing the first byte after a session switch. This is because we use `tokio::io::stdin` which loses bytes on cancellation. Tomorrow I will try to tackle this with using a single thread that reads stdin and sends the bytes into a channel - the parsing loop can then read from the channel asynchronously in the `select!`. Once that works I'll force push and let you know. If you're happy with the approach I'll vendor the latest unreleased version of `termwiz`. Any guidance on how you'd like to do the censoring is appreciated!

imsnif:
> I have got this mostly working in a local branch. The fragmented sequences are handled correctly, the session switch works fine, but I have one remaining issue with losing the first byte after a session switch. This is because we use `tokio::io::stdin` which loses bytes on cancellation. Tomorrow I will try to tackle this with using a single thread that reads stdin and sends the bytes into a channel - the parsing loop can then read from the channel asynchronously in the `select!`. 

It could be that I'm conflating things, but I think we already deal with this problem here: https://github.com/zellij-org/zellij/blob/main/zellij-client/src/os_input_output.rs#L163

Are we by any chance skipping this? I just don't want us to add more moving parts if we don't have to.

jgiannuzzi:
> It could be that I'm conflating things, but I think we already deal with this problem here: https://github.com/zellij-org/zellij/blob/main/zellij-client/src/os_input_output.rs#L163
> 
> Are we by any chance skipping this? I just don't want us to add more moving parts if we don't have to.

What we do in there ends up in a synchronous system call and thus cannot be cancelled, resulting in bad interactions with `select!`. I tried rewriting this function as `async` by using `tokio::io::stdin` but the problem remains because tokio uses a blocking thread to read from stdin with the same synchronous system call that cannot be cancelled.

In other words, in order to use `tokio::select!` for the debounce mechanism (instead of the simple timer approach I originally used) we need to make a few more things properly async and cancellation-safe.

I'll force-push what I currently have so you can take a look.

jgiannuzzi:
The interesting things happen in commits [`a60b509`](https://github.com/zellij-org/zellij/pull/4623/commits/a60b50913d4be221b6099d3ff44bdeee12bcc095?w=1) (use async stdin) and https://github.com/zellij-org/zellij/pull/4623/commits/d2029c7de60742c1da96c43db423277b230b1639 (debounce mechanism).

jgiannuzzi:
Alright, I have changed `AsyncStdinReader` to be cancellation-safe using a "pump" thread that does blocking reads on stdin and sends the bytes into a channel (https://github.com/zellij-org/zellij/pull/4623/commits/c60a581c6c81b3e5768d57df0d7e562034ed6acd). This solves the lost bytes issue I was facing. Apart from the vendoring of `termwiz`, I think this PR is ready to be reviewed. I would really recommend going commit by commit as I tried to group the changes logically to make the review easier. I'm looking forward to your input @imsnif!

imsnif:
Hey @jgiannuzzi - this is very cool, great work! I especially liked the trick with cancelling the token on the reconnect loop iterations.

However, I'm a bit concerned making these sorts of sweeping changes close to a release. I *do* however want to get this fix in. Do you think we can do it with minimal changes by adding a synchronous `select` to the STDIN loop with a subchannel that will debounce on input?

jgiannuzzi:
I fully understand your concern so I'll give the synchronous `select` a try. Once the release is done, I guess you'd be open to using this async approach, right?

I'd still like your guidance on vendoring `termwiz` btw!

imsnif:
> I fully understand your concern so I'll give the synchronous `select` a try. Once the release is done, I guess you'd be open to using this async approach in, right?

For sure. Maybe even as part of a bit of a refactor, because as I'm sure you noticed that area is a bit more of a mess than I'd like it to be.

> I'd still like your guidance on vendoring `termwiz` btw!

Ah, sorry! Please place it in a new vendored/termwiz inside the repo root and expose it as is in the relevant commit (make sure the license is there too).

jgiannuzzi:
hey @imsnif, I have a few questions:
- If I naively vendor `termwiz` using `cargo vendor` and copying it in `vendored/termwiz`, a lot of dependencies get added to our `Cargo.lock` via its optional features and dev dependencies. None of them are actually used to build Zellij but it's a bit "polluting". If I modify `termwiz`'s `Cargo.toml` by removing optional and dev dependencies, I can keep the existing `Cargo.lock` almost identical (apart to the change to `termwiz` itself (source and checksum). I could also remove the benchmarks and examples to further reduce "stuff we don't need". Do you have any preference around this?
- There have been a few unreleased changes since `termwiz` 0.23.3 that have nothing to do with my fix and introduce even more vendoring complexity (like new unreleased `wezterm-*` crates), so I would vendor 0.23.3 in a commit, and then apply just my fix to `termwiz` in a separate commit. Does that sound good?
- With regards to the synchronous approach, I want to clarify that the `select` you are talking about is _not_ a Unix select which wouldn't be portable, but rather a `crossbeam` one over receiving from a channel that is populated by a stdin "pump" thread. If that is the case, I would actually simply use `recv_timeout` on that channel as it would be the only one we would reading from. Does that sound right to you or did you have something else in mind?

jgiannuzzi:
I have a synchronous version with `recv`/`recv_timeout` on a stdin "pump" thread [here](https://github.com/jgiannuzzi/zellij/commit/ea4aea96581aa6ae22593e0b218412a561f3b648?w=1). This works exactly as expected in my testing!

imsnif:
Hey @jgiannuzzi - regarding vendoring: let's remove the optional+dev dependencies (we do this anyway as is) as well as the examples/benchmarks. Just be sure to keep the license in. I'd rather do the minimum update possible (i.e. until your fix) so as to bring in just what we need and no more (as much as possible). I'm intending to roll our own STDIN parsing library anyway (this and other issues taught us we should really own this part of our solution), so this will be temporary in any case.

About recv_timeout: I could be missing something, but if we do that won't the thread have a busy-ready of `timeout` regardless of whether input is received or not? We just want a debounce.

jgiannuzzi:
> regarding vendoring: let's remove the optional+dev dependencies (we do this anyway as is) as well as the examples/benchmarks. Just be sure to keep the license in. I'd rather do the minimum update possible (i.e. until your fix) so as to bring in just what we need and no more (as much as possible). I'm intending to roll our own STDIN parsing library anyway (this and other issues taught us we should really own this part of our solution), so this will be temporary in any case.

Understood, will keep it minimal!

> About recv_timeout: I could be missing something, but if we do that won't the thread have a busy-ready of `timeout` regardless of whether input is received or not? We just want a debounce.

No basically `recv` and `recv_timeout` block the thread, there is no busy wait. Also note that `recv` is essentially calling the equivalent of `recv_timeout` internally but with the timeout set to `None`. My proposed solution keeps the existing logic and adds an extra thread that is responsible for looping (and blocking) over stdin reads, passing the result over a channel, and then the original thread that does the parsing reads from that channel, but with the ability to timeout when finalization is needed.
I actually think it's even more elegant than my async solution with `tokio::select!` and I think that after the release we should do something similar (but with async tasks and proper cancellation).

I will update this branch with proper vendoring and the synchronous solution from my other branch. Shout if you think this is not the way to go!

imsnif:
> No basically `recv` and `recv_timeout` block the thread, there is no busy wait. Also note that `recv` is essentially calling the equivalent of `recv_timeout` internally but with the timeout set to `None`. My proposed solution keeps the existing logic and adds an extra thread that is responsible for looping (and blocking) over stdin reads, passing the result over a channel, and then the original thread that does the parsing reads from that channel, but with the ability to timeout when finalization is needed. I actually think it's even more elegant than my async solution with `tokio::select!` and I think that after the release we should do something similar (but with async tasks and proper cancellation).

Ah yeah - now I see the `needs_finalization` and that makes sense. I thought you meant for the recv_timeout to be unconditional.

jgiannuzzi:
@imsnif this is now ready for review!

One thing about the vendoring that I'm not entirely sure of: what is the consequence with regards to publishing crates? Will users still be able to build Zellij with `cargo install --locked zellij`? Will your release workflow try to upload our version of `termwiz` (which would fail)?

imsnif:
Hey @jgiannuzzi - you are right, vendoring it like this will present some issues (namely that crates.io won't allow us to publish it, since we don't own it and it also won't allow us to publish a path dependency). I think the best way forward would be to turn the crate into a module and have it live in zellij-utils (`zellij-utils/src/vendored/termwiz` - be sure to include the license file there). We'll also need to move its dependencies to the zellij-utils Cargo.toml. A brief check tells me the only conflicting dep is `nix`, but it's not in any path we use so shouldn't be an issue. If the compiler complains about it, we can stub it out in the code. Are you comfortable making this change? I'm happy to lend a hand if not.

Otherwise, could you please rebase from main (there are some minor logical conflicts due to internal API changes) and run cargo fmt?

jgiannuzzi:
Yes I was thinking along the same lines about making it a module instead. I'm comfortable with doing that, hopefully in the next few hours.

Noted about the rebase!

jgiannuzzi:
Just to give an update: doing the vendoring properly is taking a bit longer than I anticipated but I am getting close. It's quite likely a tomorrow thing though.

jgiannuzzi:
@imsnif rebasing and vendoring `termwiz` as a module done (and this time I ran `cargo fmt` on the imported `termwiz` source)

imsnif:
Hey @jgiannuzzi - thanks for this! I trimmed the hedges, because 100K+ lines and 40+ dependencies are not good for my anxiety.

Otherwise, this LGTM. Cool solution with the pump and timeout. Will merge once CI passes.

jgiannuzzi:
@imsnif thank you so much for the trimming! I wanted to keep it strictly like what we had before to avoid introducing any extra changes close to the release but had this in mind too. I'm really impressed you got it done so quick and this actually means a whole 41 less dependencies than before the vendoring 🚀

I created a follow-up PR to remove the unecessary `winapi` features that got added to `zellij-utils/Cargo.toml`: #4816. I'll work on rebasing my async version in the coming days so you can look at it after the release is done.

PR #4633, opened 2026-01-23
feat(web): add WebSocket heartbeat to prevent false disconnections

thiblahute:
Idle WebSocket connections are often silently closed by network intermediaries (proxies, load balancers, firewalls) after their timeout expires. This causes false "connection lost" messages in the web client even though the underlying session is still alive.

Add application-level heartbeat messages on the control WebSocket:
- Server sends Heartbeat every 30 seconds (before typical 60s proxy timeout)
- Client responds with HeartbeatResponse
- Server closes connection if no response within 45 seconds

Application-level heartbeats are used instead of WebSocket protocol ping/pong because some proxies intercept protocol-level pings without forwarding them, whereas application messages verify the entire path.

### DIFF
--- a/zellij-client/src/web_client/websocket_handlers.rs
+++ b/zellij-client/src/web_client/websocket_handlers.rs
@@ -18,12 +18,17 @@ use axum::{
 };
 use futures::StreamExt;
 use std::sync::{atomic::AtomicBool, Arc};
+use tokio::sync::Mutex;
+use tokio::time::Instant;
 use tokio_util::sync::CancellationToken;
 use zellij_utils::{
     input::mouse::MouseEvent,
     ipc::{ClientToServerMsg, PixelDimensions, ResizeCause},
     pane_size::SizeInPixels,
 };
+
+const PING_INTERVAL_SECS: u64 = 30;
+const PONG_TIMEOUT_SECS: u64 = 45;
 
 pub async fn ws_handler_control(
     ws: WebSocketUpgrade,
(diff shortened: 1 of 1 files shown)


imsnif:
@tlinford - could you take a look? I think you mentioned something about this.

the-color-cyan:
Was just looking into implementing WebSocket ping/pong due to disconnections I was getting routing through a Cloudflare tunnel (non-enterprise plans have an enforced 100s WebSocket idle-timeout). Would it be valuable to have RFC-compliant ping/pongs as a default with app level as a fallback or a config/feature flag? There seems to be a few cases where ping/pongs would succeed in keeping the connection alive at the transport layer when app-level heartbeats would not. Namely:

- JavaScript event loop blocking
- Server-side processing locks
- Browser tab freezes
- Initial keep-alive

Given that there's effectively no overhead for ping/pongs - I thought they might still be valuable to have. That being said, I'm sure the overhead for heartbeats is very minimal. Any thoughts on this?

thiblahute:
Thanks @the-color-cyan, agreed, and I've reworked the PR accordingly: the heartbeat is now
RFC 6455 protocol-level Ping/Pong (server pings the control socket every 30s, closes after
45s without a Pong). It is all simpler.

I'd skip the app-level fallback/flag for now: it would only add detection of a wedged page,
which isn't the failure mode here (and a busy tab is exactly when we don't want to
disconnect). Easy to add later if a concrete need shows up.

Also rebased onto current main.

thiblahute:
@imsnif Any way to get that one reviewed please?

PR #4637, opened 2026-01-24
chore: Fix the issue template to match existing labels

Zykino:
Hope this works


Zykino:
@imsnif I think you were interested in this.

With this change (if it work) every issue will have a tag according to the template use for the issue creation. Are you OK with this or maybe you prefer for some issue not to have tags by default?

imsnif:
Hey, I don't have strong opinions about this. If you think it's worth it, go for it.

Issue #4635, opened 2026-01-24
KeybindPipe Action Times Out and Freezes Zellij when Triggered via Keybinding


GenceErdagi:
## Issue description

I've been working on a plugin, [zjide-manager](https://github.com/GenceErdagi/zjide-manager), to enhance pane layout management. The goal is to give users finer control over swap layout switches, enabling an IDE-like environment where features like sidebars or terminals can be toggled via stateful feature flags. To support this, I've been experimenting in my fork on the [go-to-swap-layout](https://github.com/GenceErdagi/zellij/tree/go-to-swap-layout) branch, where I introduced a custom layout switch method based on swap names. This approach avoids the flicker and unpredictability of cycling through swap lists when multiple layouts are defined.

However, when I tried to keybind the pipe message to the layout manager (see branch [keybind-pipe-timeout-fix](https://github.com/GenceErdagi/zellij/tree/keybind-pipe-timeout-fix)), I encountered a critical issue. While the CLI pipe works instantaneously as expected, using the `MessagePlugin` method consistently times out and locks up Zellij, producing error logs. I've implemented a fix for this timeout issue in the linked branch, but I wanted to open this report to document the problem and ensure my solution aligns with the project guidelines.

### Expected Behavior
- **CLI:** Running `zellij action keybind-pipe ...` returns immediately after dispatching the pipe.
- **Keybinding:** Invoking the same pipe via `keybinds` in the config should behave identically—no delay and no timeout.

### Actual Behavior
- The **CLI path** works as expected with no delay.
- The **Keybinding path** hangs for approximately 1 second and logs `Action KeybindPipe did not complete within 1s timeout`. This happens even though the plugin actually receives the pipe and the pane/layout change occurs after the freeze.

### Root Cause Analysis
- `Action::KeybindPipe` (the keybinding path) creates a completion channel but never passes a `NotificationEnd` signal to the plugin instruction. Consequently, the router waits on that channel until the 1-second timeout elapses.
- `PluginInstruction::KeybindPipe` lacks any completion notification mechanism, so nothing can signal the router to unblock.
- In contrast, `Action::CliPipe` (the CLI path) explicitly drops its completion sender, which explains why the CLI call returns immediately. The keybinding path is missing this signaling, causing the discrepancy.

**Code Reference (Before Fix):**

```rust
// zellij-server/src/route.rs
senders
    .send_to_plugin(PluginInstruction::KeybindPipe {
        ...
        // ❌ no NotificationEnd passed here
[...]

PR #4638, opened 2026-01-24
fix: pass NotificationEnd to KeybindPipe to prevent 1s timeout

GenceErdagi:
As mentioned at the #4635  messaging an plugin using keyboard shortcuts are causing freezes and timeouts due to NotificationEnd is not passed. the fix that i found is passing down the completion signal to the `PluginInstruction::KeybindPipe` in files `zellij-server/src/plugins/mod.rs` and `zellij-server/src/route.rs`

@imsnif if any more suggestions about the change i would love to take a look at for improvements.

### DIFF
--- a/zellij-server/src/plugins/mod.rs
+++ b/zellij-server/src/plugins/mod.rs
@@ -964,7 +964,6 @@ pub(crate) fn plugin_thread_main(
                 plugin_and_client_id,
                 notification_end,
             } => {
-                let _ = notification_end; // signal completion
                 let should_float = floating.unwrap_or(true);
                 let mut pipe_messages = vec![];
                 let floating_pane_coordinates = None; // TODO: do we want to allow this?
--- a/zellij-server/src/plugins/wasm_bridge.rs
+++ b/zellij-server/src/plugins/wasm_bridge.rs
@@ -1089,7 +1089,7 @@ impl WasmBridge {
         &mut self,
         messages: Vec<(Option<PluginId>, Option<ClientId>, PipeMessage)>,
         shutdown_sender: Sender<()>,
-        notification_end: Option<NotificationEnd>,
+        mut notification_end: Option<NotificationEnd>,
     ) -> Result<()> {
         let plugins_to_update: Vec<(
             PluginId,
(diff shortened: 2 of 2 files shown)


imsnif:
Hey @GenceErdagi - I have given this some thought and I think your solution would work better after all. Mainly because allowing plugins to arbitrarily block keybinding input implicitly is probably more chaotic than helpful.

That being said, the notification_end needs to be threaded a little farther. Namely we need to send it all the way into the plugin executor and drop it here: https://github.com/zellij-org/zellij/blob/main/zellij-server/src/plugins/wasm_bridge.rs#L1159-L1176

This way we can be certain we'll only release the keybinding sequence to process the next message once the plugin has processed the message.

Would you like to implement this?

GenceErdagi:
Thanks for your time and interest. As soon as i will be free will take a look at it.

GenceErdagi:
Okey i have drilled down the notification_end to the plugin_execution, but i needed to supress the supress the CliPipe and MessageFromPlugin with None since changing them requires additonal modifications to other places like tests ( also they were working just fine )

imsnif:
Thanks! It's fine not to change the other paths, but we need to pass the same notification end and not clone it (the return channel is not Clone and so will signal as soon as the original is dropped). I think these are shadowed here somewhere which might be the reason they need to be cloned. If you remove the shadowing (the `let _ = ...`) the compiler might forgive you :)

GenceErdagi:
Sorry for delay but i needed to get my understanding deeper in rust and what is going on here ( neither english nor rust are my primary languages :) ) so i am sorry for misunderstanding the issues. So before implementing and pushing i must ask, after deleting the line in [mod.rs](https://github.com/zellij-org/zellij/pull/4638/changes/085b506344e93dcea90d8dd707582d78acac7b0e#diff-0b345f3f38523fd8a80b8f68723ebf85970c593779112c45e80021cb17bca181R967) to get rid of said shadowing and removing the `.copy()` calls, i have used the `let notification_end = notification_end.take();` instead of copy. So not to contaminate the commit history further i wanted to be sure that this implementation would right. 

Thank you for your patiance again.

imsnif:
No worries! I think you don't even need to do that - pass the variable as is and it should do the right thing, not let bindings or anything.

GenceErdagi:
Okey i have fixed the remaining shadownings and `.copy()` instances in fix but i could not find a way to get rid of [`let notification_end = notification_end.take()`](https://github.com/zellij-org/zellij/pull/4638/changes/b3619fcbbaba2bdf636320c6a56a25e51c5e216d#diff-03bbc0b5848e63bf937e99e4c1840af9853a24bf65dbb1f7248b5aadbb4694deR1134) to be able to compile. I dont know if this is close to what you are expecting.

imsnif:
Hey @GenceErdagi - thanks for the changes. The notification_end needs to be dropped right after this line: https://github.com/zellij-org/zellij/pull/4638/changes#diff-03bbc0b5848e63bf937e99e4c1840af9853a24bf65dbb1f7248b5aadbb4694deR1169 (as in - it must be inside the closure), otherwise we'll still have races. `take` is alright (because it does not clone), but I think we might not need it if we put it in that line.

GenceErdagi:
Line you have pointed still inside the match exprassion, I dont know how we can drop there. Am i missing something ?

imsnif:
You are totally right, this was my mistake. I got confused by the indentation :sweat_smile: 

Thanks for this!

Issue #4509, opened 2025-11-14
Unable to use C-\ as keybind


betaprior:
# Issues with the Zellij UI / behavior / crash

## Issue description

Background: I'm migrating from a tmux, where `C-\` was the leader key to avoid conflicts with emacs/vim bindings. In my `config.kdl` I have the following: 

```
    locked {
        bind "Ctrl \\" { SwitchToMode "normal"; }
    }
// ....
support_kitty_keyboard_protocol false
```

I find that I need to use `support_kitty_keyboard_protocol false` in Zellij in order to use default Emacs keybindings such as `M-:` or `M->`. 

## Minimal reproduction

With the above settings in any terminal (tried WezTerm, Alacritty, Ghostty), the `C-\` binding is failing to properly switch modes as expected.

Furthermore, in WezTerm the `C-\` binding is failing to work in Zellij _regardless_ of the value of  `support_kitty_keyboard_protocol`. This binding works just fine with tmux.

## Other relevant information

Using Zellij 0.43.1 on Fedora Linux.

vthunder:
@betaprior I ran into the same issue and fixed the kitty support so M-<, M->, etc will work. See my PR/fork above.

vthunder:
Also, 🙌 for C-\ as the escape sequence, I have used that for decades in screen! It is the superior choice for emacs users imo.

betaprior:
@vthunder Thanks for looking into this! I see that the PR above was not accepted by the maintainer. I'm curious, were you able to figure out why `C-\` binding is broken with kitty protocol disabled?  I would strongly prefer to keep it disabled, since I see no benefit in it and as you've discovered it breaks emacs keybinds.

vthunder:
No — I am happy with my patched version so far, so I'm running it with kitty protocol enabled. Haven't used it extensively, but so far so good.

m-morino:
@betaprior I dug into this — the root cause is in termwiz's `build_basic_key_map()`. It maps `0x01`–`0x1A` (Ctrl+A through Ctrl+Z) but just... doesn't cover `0x1C`–`0x1F`. So Ctrl+\ sends byte `0x1C`, termwiz doesn't know what to do with it, and you get `KeyCode::Char('\x1c')` with no modifiers. The keybind never fires.

Same kind of thing was already worked around for Ctrl+H and Ctrl+J in `cast_termwiz_key()` (see #3307). Here's a fix that extends that to the four missing bytes:

```diff
[...]

Issue #3789, opened 2024-11-18
Shift modifier not working/recognised in conjunction with other modifier keys when inside Zellij session


blixenkrone:
**Basic information**

`zellij --version`:
0.41.1

`stty size`:
78 150

## Issue description
Whenever I am inside a zellij in any terminal (tried 3 different ones), the shift modifier key is not picked up as a modifier key, when used in combination with others.
This is mostly an issue when using terminal editors like vim, since I have a keybinding with `shift+alt+s` to do :w or write_file.

As the most minimal example, a use case flow would be:
- I can write uppercase letters ie. ABCD with no issue
- I can not write my files with the key combination of shift+alt+s. Only alt+s combination is picked up.

If I am _not_ inside Zellij, I can do above use flow without issues. The bug seem to only exists when I am within a zellij session.

## Minimal reproduction
Terminal: Ghostty
Editor: Helix
1 Enter a zellij session
2 Open a file with Helix (or vim)
3 Hit shift+alt+s
4 Only alt+s gets picked up.

## Extra relevant information
1 No /tmp/zellij-log file was produced when running with `$ zellij --debug` for some reason.
2 I don't think it makes sense to show this in a video recording, but I am happy to record one if needed.

Thanks for a great tool!

blixenkrone:
@imsnif Just FYI this thread was created from a Reddit thread with another user with same issue.
I'm trying to setup a bug bounty for implementing the solution, but can't find anywhere to donate to whomever fixes this (if not you).
Thanks!

imsnif:
Hey @blixenkrone - this issue is on my to-fix (or rather, to investigate and then fix if it's a Zellij issue). My GitHub Sponsors is linked at the top of this repo and here for convenience: https://github.com/sponsors/imsnif - do with this info as you wish.

My personal opinion on bounty systems as a full-time OSS dev: please avoid bounty systems and middle-person OSS sponsoring platforms in general. Give directly to maintainers, everything else is noise at best or actively takes money out of our pockets at worst. Not going to expand on this further, just something to think about.

Thanks for your report and for caring about OSS sustainability.

Iorvethe:
[...]

PR #4642, opened 2026-01-26
feat(kitty): support REPORT_ALTERNATE_KEYS for shifted key combinations

vthunder:
## Summary

- Adds support for Kitty keyboard protocol flag 4 (REPORT_ALTERNATE_KEYS) in addition to flag 1 (DISAMBIGUATE_ESCAPE_CODES)
- Fixes Shift+Alt key combinations not being correctly forwarded to non-Kitty applications (like Emacs)
- When terminal sends `CSI base:shifted;modifiers u`, we now parse and use the shifted key

## Problem

When pressing M-< (Alt+Shift+comma) in Emacs inside Zellij, Emacs received M-, instead. The shift modifier was being lost because Zellij only supported DISAMBIGUATE_ESCAPE_CODES (flag 1), which sends the base key with modifier flags. Non-Kitty applications expect the actual shifted character.

Related: #4509 (user there is also running into this issue)

## Solution

Enable REPORT_ALTERNATE_KEYS (flag 4) which causes the terminal to send both base and shifted keys in format `ESC[44:60;4u` (base=comma, shifted=<). When serializing to non-Kitty apps, we now use the shifted key directly.

## Test plan

- [x] Unit tests for keyboard parser with shifted key format
- [x] Roundtrip tests for protobuf serialization with shifted_key
- [x] Manual testing: M-< in Emacs now correctly triggers `beginning-of-buffer`

### DIFF
--- a/zellij-client/src/keyboard_parser.rs
+++ b/zellij-client/src/keyboard_parser.rs
@@ -6,6 +6,8 @@ enum KittyKeysParsingState {
     Ground,
     ReceivedEscapeCharacter,
     ParsingNumber,
+    ParsingShiftedKey,   // After first colon - parsing shifted key
+    SkippingAlternateKey, // After second colon - skip alternate key until ; or terminator
     ParsingModifiers,
     DoneParsingWithU,
     DoneParsingWithTilde,
@@ -15,6 +17,7 @@ enum KittyKeysParsingState {
 pub struct KittyKeyboardParser {
     state: KittyKeysParsingState,
     number_bytes: Vec<u8>,
+    shifted_key_bytes: Vec<u8>,  // For REPORT_ALTERNATE_KEYS: the shifted form
     modifier_bytes: Vec<u8>,
 }
 
@@ -23,6 +26,7 @@ impl KittyKeyboardParser {
         KittyKeyboardParser {
             state: KittyKeysParsingState::Ground,
             number_bytes: vec![],
+            shifted_key_bytes: vec![],
             modifier_bytes: vec![],
         }
     }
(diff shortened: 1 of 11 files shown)


imsnif:
This is very likely https://github.com/zellij-org/zellij/issues/3789#issuecomment-2567760536

(tl;dr - emacs or whichever relevant plugin does the STDIN parsing in this case, needs to make sure which progressive enhancement of the KKP the terminal it's running supports, rather than assuming the binary of "supports KKP/doesn't support KKP")

vthunder:
I see. I didn't read the whole thread in detail, but I think I got the gist of it:
- zellij only implements level 1 of the kitty protocol (what this PR is changing)
- other apps assume that if level 1 is supported, other levels will be as well, leading to mismatched expectations
- other terminals actually support the whole kitty protocol, or not at all, hence no mismatched expectations

I should note that it's not just emacs, I see the same issue in other apps as well.

Would you accept a change to add support for more of the kitty protocol, or do you actively want to support only level 1? I agree apps should be handling the negotiation better, but supporting more of the protocol would "just fix" the UX for these apps, no? But if you don't want it, we can just close out this PR of course.

Thanks for your attention (and for zellij in the first place).

imsnif:
>     * zellij only implements level 1 of the kitty protocol (what this PR is changing)
> 
>     * other apps assume that if level 1 is supported, other levels will be as well, leading to mismatched expectations
> 
>     * other terminals actually support the whole kitty protocol, or not at all, hence no mismatched expectations

Yep, you got it.

> I should note that it's not just emacs, I see the same issue in other apps as well.
> 
> Would you accept a change to add support for more of the kitty protocol, or do you actively want to support only level 1? I agree apps should be handling the negotiation better, but supporting more of the protocol would "just fix" the UX for these apps, no? But if you don't want it, we can just close out this PR of course.

Thanks for the offer. It might be something to look at in the future, though I'm a bit ambivalent. I have some concerns about this protocol - this inherent design issue only affirming my biases, and I suspect that the more layers we peel off as we implement, the more such issues will arise.

Even putting all this aside, this PR only deals with the negotiation between Zellij and its controlling terminal, as well as the reporting of the protocol level to the underlying panes. It also seems to skip a few levels of progressive enhancements and considering aforementioned design issue, it's anyone's guess how many issues this will create in various implementations. I'd estimate this is about 20% of the work needed.

As a personal note - if I may - while I understand in some corners of the tech industry this is common place these days, I personally tend to avoid PRs that are heavily written by LLMs. What tends to happen is that these PRs are reviewed, then the review is copy/pasted to the LLM, the LLM addresses the review and the cycle continues. It feels like I might as well just use an LLM directly. Just my personal opinion and preference of course. I understand where you're coming from and am not in any way saying you did something wrong. Just that if I were in your place, I'd appreciate this being pointed out to me.

vthunder:
Thanks for your response, and I totally understand. I'll close out the PR to minimize clutter.

Re: your note—to be honest, I wouldn't have made this change in the first place had it not been for the LLM, and so for my own personal use it is the difference between being able to use zellij versus not at all. Truly a game-changer. But I 100% get where you're coming from as well.

Thanks again!

PR #4654, opened 2026-01-29
Allow saving current session for resurrection on demand

imsnif:
Rather than waiting for the `serialization_interval` (by default, saving session's for resurrection every 60 seconds), this allows users to explicitly save their session for resurrection.

As a CLI command:
```
zellij action save-session
```

As a plugin API:
```
save_session();
```

From the `session-manager` interface:

image


mloskot:
Is there any plan to issue a new release that will include this feature?

PR #4666, opened 2026-02-01
docs: Remove the "testing plugins" section from CONTRIBUTING.md

emesterhazy:
This is no longer relevant after the switch from wasmtime to wasi in
https://github.com/zellij-org/zellij/pull/4449.


emesterhazy:
Ok, it turns out this flag is obsolete after https://github.com/zellij-org/zellij/pull/4449. I've updated the commit to delete this entire section now that it's irrelevant.

PR #4669, opened 2026-02-01
feat(session-manager): Make shift + tab move to previous tab

### LINKED ISSUE #4416: Improve session manager behavoiur
Make session manager close on escape in "New Session" menu instead of clear form.
Make shift +  tab to go towards menu list

emesterhazy:
This holding shift to change the direction of an action is a pretty well established UX pattern. There's no reason not to implement it in the session-manager, and users have noticed that it's missing.

- Fixes https://github.com/zellij-org/zellij/issues/4416

### DIFF
--- a/default-plugins/session-manager/src/main.rs
+++ b/default-plugins/session-manager/src/main.rs
@@ -260,6 +260,10 @@ impl State {
                 self.toggle_active_screen();
                 should_render = true;
             },
+            BareKey::Tab if key.has_modifiers(&[KeyModifier::Shift]) => {
+                self.toggle_active_screen_backwards();
+                should_render = true;
+            },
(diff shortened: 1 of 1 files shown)


basnijholt:
Shift tab is used by e.g., Claude Code to move permission modes.

Personally, I use shift arrow keys:
https://github.com/basnijholt/dotfiles/blob/5a73b9c9bf01e999429b6492839d76bcc33ff01f/configs/zellij/config.kdl#L133-L136

emesterhazy:
Personally I don't think a conflict with Claude code should prevent this change from being made. There are a lot of other shortcuts that conflict with terminal applications, and the solution for those is to lock Zellij before using them.

This change is also only for the session manager, so I suspect that it's ok to capture shift+tab when the session manager is open instead of forwarding the keys to the active terminal window.

PR #4670, opened 2026-02-01
Expose tab name to panes via environment variable (e.g., ZELLIJ_TAB_N…

dovansy1998:
## Problem
Panes have no way to know which tab they belong to or what that tab is called. That limits:

- Showing the tab name in shell prompts 
- Scripts that depend on the current tab
- Logging or tagging output by tab

ZELLIJ_SESSION_NAME and ZELLIJ_PANE_ID already exist; there was no equivalent for the tab.
## Solution
Introduce a ZELLIJ_TAB_NAME environment variable that is set for every terminal pane and reflects the name of the tab it belongs to.
When set: On spawn (new pane, layout, re-run). Value is the current name of the tab containing the pane.
When it changes: Tab name is updated in pty state when the user switches tabs or renames a tab. New panes (and re-run) get the updated name; already-running processes keep their original env (env cannot be changed for existing processes).

### DIFF
--- a/zellij-server/src/os_input_output.rs
+++ b/zellij-server/src/os_input_output.rs
@@ -157,6 +157,7 @@ fn handle_openpty(
     cmd: RunCommand,
     quit_cb: Box<dyn Fn(PaneId, Option<i32>, RunCommand) + Send>, // u32 is the exit status
     terminal_id: u32,
+    tab_name: Option<&str>,
 ) -> Result<(RawFd, RawFd)> {
     let err_context = |cmd: &RunCommand| {
         format!(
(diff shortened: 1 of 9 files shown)


Zykino:
What does happens if the pane is moved to a new tab? (`Ctrl t, b` by default) I think we can also move to an existing tab.
You cannot change environment variable of already running processes. At least not as far as I know.

imsnif:
Yeah - as @Zykino says - I'd rather have a query interface here. Something like:

```bash
$ zellij action query-current-tab-name
my-tab-name
$
```

Should probably also be combined with `query-current-pane-name` while we're there.

dfee:
I think if a pane has an ID (immutable) and a tab has an ID (immutable), and there's a query interface that relates pane IDs to tab IDs (mutable), and then another query (or the same one) that relates a tab ID to a tab name (mutable), then all problems are solved cleanly.

PR #4672, opened 2026-02-03
Purge `async_std` from the codebase

har7an:
Remove `async_std` from the zellij codebase in favor of `tokio`, the de-facto standard async runtime.

## User-facing changes

*None*.

Unless, of course, I missed something. But I think that's unlikely... :crossed_fingers:

## Why?

There are multiple reasons:

1. Having multiple async runtimes adds a lot of bloat to the code
    - Many more dependencies to worry about
    - Two async state machines at runtime
    - Bigger binary size
    - Potential for incompatibilities in code (mostly w.r.t. trait impls)
2. [`async_std` is unmaintained since August 2025][1]

There is nothing in `async_std` that `tokio` can't deliver, while `tokio` has many things to offer that `async_std` never had. So, we're moving all our async code to `tokio`, which will enable us to hook into other parts of the tokio ecosystem more seamlessly (think about `tracing` to replace `log`, for example).

## Potential issues

I haven't had a need to juggle runtime instances ad-hoc before, so I'm
wondering whether the code could be simplified somehow. I'm fairly confident
that the code is basically sane at least.

I also replaced `async_channel` while doing the refactoring because it seemed
like an easy enough target to get rid of. The channels previously supplied by
that crate are now native `tokio` channels. Since I didn't find anything about
the capacity of `async_channel` channel types, I chose to go with unbounded
channels. This may cause issues with lacking backpressure under very high load,
but I'm willing to try this out and see what happens. Note that this only
applies to WASM workers, i.e. the plugin system.

## Next steps

All of these are optional, of course, but I think we should:

1. Check for cases where async and sync code are nested beyond depth 1, i.e.
   sync code spawns an async task which then spawns a sync thread somewhere.
   I'm not sure this is actually a thing in our codebase, but I saw at least
   one instance of nesting that did this.
2. Go through all async code and make sure it doesn't interact with sync
   primitives, such as `std::sync::*` (like a `std::sync::Mutex`). I've fixed
   at least one bit of code where this was the case. This has the potential to
   stall threads/tasks for extended amounts of time, defying the idea of async
   code (since the task now blocks and it doesn't actually *know* there is a
   yield point, preventing other tasks from running).

[1]: https://rustsec.org/advisories/RUSTSEC-2025-0052.html

### DIFF
--- a/zellij-client/src/lib.rs
+++ b/zellij-client/src/lib.rs
@@ -37,13 +37,34 @@ use crate::web_client::control_message::{
 static ASYNC_RUNTIME: OnceLock<tokio::runtime::Runtime> = OnceLock::new();
 use std::sync::OnceLock;
 
-pub(crate) fn async_runtime() -> tokio::runtime::Handle {
+/// Spawn an async runtime for this client instance.
+///
+/// The number of workers can be configured to any nonzero value. Passing zero or `None` will spawn
+/// one worker per physical CPU on the current machine.
+pub(crate) fn async_runtime(maybe_number_of_workers: Option<usize>) -> tokio::runtime::Handle {
     match tokio::runtime::Handle::try_current() {
         Ok(handle) => handle.clone(),
         _ => {
+            let number_of_workers = match maybe_number_of_workers {
+                Some(value) if value > 0 => {
+                    log::debug!(
+                        "Creating client async runtime with {} tasks based on user request",
+                        value
+                    );
+                    value
+                },
+                _ => {
+                    let cpus = num_cpus::get_physical();
+                    log::debug!(
+                        "Creating client async runtime with {} tasks based on CPU count",
+                        cpus
+                    );
+                    cpus
+                },
+            };
             let runtime = ASYNC_RUNTIME.get_or_init(|| {
                 tokio::runtime::Builder::new_multi_thread()
-                    .worker_threads(4)
+                    .worker_threads(number_of_workers)
                     .thread_name("zellij client async-runtime")
                     .enable_all()
                     .build()
@@ -476,10 +497,11 @@ pub fn start_remote_client(
     token: Option<String>,
     remember: bool,
     forget: bool,
+    async_worker_tasks: Option<usize>,
 ) -> Result<Option<ConnectToSession>, RemoteClientError> {
     info!("Starting Zellij client!");
 
-    let runtime = crate::async_runtime();
+    let runtime = crate::async_runtime(async_worker_tasks);
 
     let connections = remote_attach::attach_to_remote_session(
         runtime.clone(),
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5160,6 +5160,7 @@ dependencies = [
  "mio 0.8.11",
  "nix 0.23.1",
  "notify",
+ "num_cpus",
  "regex",
  "serde",
  "serde_json",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -88,6 +88,7 @@ miette = { version = "5.7.0", default-features = false, features = ["fancy"] }
 nix = { version = "0.23.1", default-features = false }
 notify-debouncer-full = { version = "0.3.1", default-features = false }
 notify = { version = "6.1.1", default-features = false, features = ["macos_kqueue"] }
+num_cpus = { version = "1.16", default-features = false }
 prost = { version = "0.11.9", default-features = false, features = ["std", "prost-derive"] }
 regex = { version = "1.5.5", default-features = false, features = ["perf", "std"] }
 serde = { version = "1.0", default-features = false, features = ["derive", "std"] }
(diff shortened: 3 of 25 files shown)


har7an:
## TODOs

- [x] Make sure a runtime instance is spawned early on for client and server respectively
- [x] Ensure client and server each have their own runtime instance and there is no unintended "sharing"
- [ ] Get rid of `crate::async_runtime_blurb().spawn(...)` code by ensuring a runtime exists and can be reused with `tokio::spawn`

Especially that last one will be a huge boost in ergonomics, I hope. At least if it works the way I assume it does.

har7an:
"Reusing" a spawned runtime with bare `tokio::spawn` is only possible from within an async task, and we can't provide that because we have too much sync code. Since the code seems to work: LGTM!

har7an:
@imsnif IIRC you wanted to take a look at this before I merge "[...] to make sure no functionality/performance was broken". Let me know what you think.

imsnif:
Yes, thanks for the heads up! I'll give this a shake in the next few days.

zellij-server/src/pty.rs
```diff
@@ -1704,17 +1702,15 @@ impl Pty {
...
-                        self.bus
-                            .os_input
-                            .as_mut()
-                            .with_context(err_context)
-                            .fatal()
-                            .kill(Pid::from_raw(child_fd))
-                            .with_context(err_context)
-                            .fatal();
-                    });
+                    let err_context = || format!("failed to kill child processes for pane {id}");
```
imsnif, line 1762: Just out of curiosity, why did we remove the async task here?

zellij-server/src/pty.rs
```diff
@@ -1704,17 +1702,15 @@ impl Pty {
...
+                    self.bus
+                        .os_input
+                        .as_mut()
+                        .with_context(err_context)
+                        .fatal()
+                        .kill(Pid::from_raw(child_fd))
+                        .with_context(err_context)
+                        .fatal();
```
imsnif: Also, I know this was here before - but let's make this non_fatal. I don't want to crash the whole app because of an OS error here.

zellij-client/src/lib.rs
```diff
@@ -36,6 +34,26 @@ use crate::web_client::control_message::{
...
+static ASYNC_RUNTIME: OnceLock<tokio::runtime::Runtime> = OnceLock::new();
+use std::sync::OnceLock;
+
+pub(crate) fn async_runtime() -> tokio::runtime::Handle {
+    match tokio::runtime::Handle::try_current() {
+        Ok(handle) => handle.clone(),
+        _ => {
+            let runtime = ASYNC_RUNTIME.get_or_init(|| {
+                tokio::runtime::Builder::new_multi_thread()
+                    .worker_threads(4)
```
imsnif: Can we make this configurable? Also, why did we choose 4? (just curious)

zellij-server/src/pty.rs
har7an: Because, as I understand it, the code created the async task to run a sync function call, blocking until the function call returned. I.e. rather than executing it directly in the current thread, it asks `tokio` to park this operation on one of their blocking workers, waiting for that to return.

This seemed like an unnecessary "context switch" to me. Or did I misunderstand what the code did before?

zellij-client/src/lib.rs
har7an: > Can we make this configurable?

I assume you mean as in "make this a constant with a pretty name rather than a magic number"? In that case: Definitely.

> Also, why did we choose 4?

[I could ask you the same question][1]. ;)

I guess it's fair to assume people have at least that many distinct threads available in PCs nowadays? We could also stick with the default [which takes the number of cores, apparently][2]. I don't have strong feelings here.

[1]: https://github.com/zellij-org/zellij/blame/13a82eea6d74d8168e538b8755131ed793a7f054/zellij-server/src/global_async_runtime.rs#L11
[2]: https://docs.rs/tokio/1.49.0/tokio/runtime/struct.Builder.html#method.worker_threads

har7an (commented):
Great, thanks for the review! See answers above (or wherever GH puts them).

zellij-client/src/lib.rs
imsnif: LOL, fair enough :)

By making this configurable though I mean adding this to the config (will require a restart of course and not be updated on config changes). Namely I want to give users the ability to tinker with this without recompiling - eg. for few or many resources.

har7an: I see. I'm wondering whether that's a good idea, though, since right now the client impl mixes a lot of sync and async code. This option may not meet users expectations w.r.t. the final load the application creates on their machines. Like, as far as I know we have separate threads for handling "stdin", "input" (server messages I think), "signals", and message routing at least, none of those is async and we have no means to limit their resources meaningfully.

I'm still gonna add it, just wanted to tell you my concerns with this setting.

har7an: Side-note: Right now, (apparently) the only code in the zellij-client that makes use of the async runtime is the web client implementation. So the config option only applies to web clients. I did my best to document this, too.

imsnif:
Merging this as @har7an is AFK, gave his go-ahead and we might depend on this elsewhere. Great work!

Issue #4535, opened 2025-12-04
zellij action / zellij run silently fail when given --session argument


jeff-hykin:
**Basic information**

`zellij --version`: 0.43.1

`stty size`: 49 154

`uname -av` or `ver`(Windows): Darwin Jeffs-MBP-3 21.6.0 Darwin Kernel Version 21.6.0: Mon Aug 22 20:19:52 PDT 2022; root:xnu-8020.140.49~2/RELEASE_ARM64_T6000 arm64 arm64 MacBookPro18,3 Darwin

## Further information

Quick reproduction:

```sh
# expected/normal part:
zellij attach -c s1
# manually type: zellij action new-pane
# manually type: zellij run -- git diff
# (both work as expected)
# detach from the session

zellij --session s1 action new-pane
zellij attach -c s1
# expected: a new pane, actual: no new pane
# detach again

zellij --session s1 run -- git diff
zellij attach -c s1
# expected: a new pane with git diff inside, actual: no new pane

zellij --session s1 action write-chars 'echo hi\n'
zellij attach -c s1
# expected: the main pain with "echo hi" run, actual: no new chars
```

The session-targeted commands (e.g. `zellij --session s1 run -- git diff`) take an unusually long time to run (around 3 full seconds), which makes me wonder if something is failing or timing out.

jeff-hykin:
After a bit more testing, the commands work if the session is attached *somewhere*, but not if it is detached.

If that is intended, then I'd expect them to error when the session is detached everywhere (e.g. the action won't actually happen)

imsnif:
Fixed in https://github.com/zellij-org/zellij/pull/4626 - will be released in the next version.

Thanks!

PR #4676, opened 2026-02-03
fix(screen): tab actions work on sessions with no attached clients

alasano:
## Summary

Extends the fix from #4626 (which addressed #4535) to all other tab-related CLI actions that had the same silent failure pattern when targeting a session with no attached clients.

### The Problem

After #4626, `zellij run --session` and `zellij action new-pane` correctly work on detached sessions. However, other tab actions still silently returned success without doing anything:

- `zellij --session X action close-tab`
- `zellij --session X action rename-tab "name"`
- `zellij --session X action go-to-next-tab`
- etc.

### The Fix

Added a helper method `get_active_tab_index_with_fallback()` that mirrors the fallback logic in `add_client()`:
1. Specified client's active tab
2. First connected client's active tab
3. `global_last_active_tab_index` if that tab exists
4. First tab by index

Applied this fallback to 9 methods:
- `switch_tab_next` / `switch_tab_prev`
- `close_tab`
- `update_active_tab_name` / `undo_active_rename_tab`
- `move_active_tab_to_left` / `move_active_tab_to_right`
- `move_focus_left_or_previous_tab` / `move_focus_right_or_next_tab`

Also modified `switch_tabs()` to accept `Option<ClientId>` and update `global_last_active_tab_index` when no client is connected, so tab move operations work correctly in detached sessions.

### Additional Cleanup

Refactored all affected methods to use the existing `client_id()` helper for consistent client ID resolution, reducing code duplication.

## Test Plan

- [x] All 90 existing screen tests pass
- [ ] Manual testing with detached session:
  ```bash
  zellij --session test
  # Create multiple tabs, detach
  
  zellij --session test action go-to-next-tab
  zellij --session test action rename-tab "NewName"
  zellij --session test action close-tab
  
  # Reattach and verify actions took effect
  zellij attach test
  ```

### DIFF
--- a/zellij-server/src/screen.rs
+++ b/zellij-server/src/screen.rs
@@ -1310,6 +1310,10 @@ impl Screen {
 
                     self.log_and_report_session_state()
                         .with_context(err_context)?;
+                    // Update global_last_active_tab_index to track the current tab
+                    // This ensures subsequent actions target the correct tab even after
+                    // ephemeral clients (like CLI action clients) disconnect
+                    self.global_last_active_tab_index = new_tab_index;
                     return self.render(None).with_context(err_context);
                 },
                 Err(err) => Err::<(), _>(err).with_context(err_context).non_fatal(),
--- a/zellij-server/src/screen.rs
+++ b/zellij-server/src/screen.rs
@@ -154,13 +154,15 @@ macro_rules! active_tab_and_connected_client_id_with_first_tab_fallback {
                         Err(err) => Err::<(), _>(err).non_fatal(),
                     }
                 } else {
-                    match $screen.get_indexed_tab_mut(0) {
-                        Some(first_tab) => {
-                            $closure(first_tab, None);
-                        },
-                        None => {
-                            log::error!("Not tabs found!");
-                        },
+                    // No clients connected - use fallback to find the appropriate tab
+                    if let Some(fallback_tab_index) = $screen.get_active_tab_index_with_fallback($client_id) {
+                        if let Some(fallback_tab) = $screen.get_indexed_tab_mut(fallback_tab_index) {
+                            $closure(fallback_tab, None);
+                        } else {
+                            log::error!("Failed to get tab at fallback index: {}", fallback_tab_index);
+                        }
+                    } else {
+                        log::error!("No tabs found!");
                     }
                 };
             },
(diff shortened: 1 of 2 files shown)


alasano:
Pushed an additional commit — I missed `switch_active_tab`, which is the core method that `go_to_tab`, `go_to_tab_name`, and `toggle_tab` all delegate to. It had the same silent failure pattern when no clients are connected.

The dispatch handlers for `GoToTab` and `GoToTabName` also needed fixing — they were skipping the operation entirely when `active_tab_indices` was empty, never reaching the method at all.

Same fix pattern as the rest: uses `client_id()` helper and falls back to updating `global_last_active_tab_index`.

alasano:
Found another related bug — `action new-pane --direction <left|right|up|down>` was also silently failing on detached sessions.

PR #4626 claimed to fix this but missed a code path. The issue is in `zellij-server/src/tab/mod.rs` line 1430. The directional split path (`vertical_split` / `horizontal_split`) requires a non-optional `ClientId` to determine which pane to split. When PR #4626's macro correctly fell back with `client_id: None`, the entire split block was skipped and it returned `Ok(())` without creating the pane.

Fixed by falling back to directionless tiled pane placement when `client_id` is `None`, so the pane still gets created even if we can't do a directional split.

alasano:
Found another issue — `action new-tab` didn't properly update focus in detached sessions. When you created a new tab and then created a new pane, the pane would be added to the original tab instead of the new one.

The problem was that `switch_active_tab` only updated `global_last_active_tab_index` when `effective_client_id` was `None`. But `apply_layout` creates ephemeral client entries for CLI actions, causing `switch_active_tab` to take the client path which didn't update the global index.

Fixed by updating `global_last_active_tab_index` on every successful tab switch, regardless of whether a client exists. This ensures subsequent actions always target the correct tab, even after ephemeral CLI clients disconnect.

alasano:
Update: The previous fix for `new-tab` focus wasn't complete. The real issue was the `active_tab_and_connected_client_id_with_first_tab_fallback!` macro hardcoding `get_indexed_tab_mut(0)` as its fallback, causing `new-pane` to always target Tab #1 instead of the currently active tab.

Updated the macro to use `get_active_tab_index_with_fallback()` which properly consults `global_last_active_tab_index`. Now `new-pane` correctly targets whichever tab was most recently switched to in a detached session.

imsnif:
Hey @alasano - thanks for this!

Great work on threading here, but I'd prefer not to add this sort of fallback logic to actions such as changing pane focus or switching tabs. Creating new panes/tabs now returns their index (for tabs) or pane_id (for panes). I think a better CLI path would be to use those IDs as arguments to the relevant commands (eg. `zellij action rename-tab --index <index>`).

It's now also possible to use `zellij [--session <name>] action list-tabs` or `zellij [--session <name>] action list-panes` to get a (json formatted) list of tabs/panes and their IDs so you don't have to rely on keeping this info at all.

Wdyt?

alasano:
> Hey @alasano - thanks for this!
> 
> Great work on threading here, but I'd prefer not to add this sort of fallback logic to actions such as changing pane focus or switching tabs. Creating new panes/tabs now returns their index (for tabs) or pane_id (for panes). I think a better CLI path would be to use those IDs as arguments to the relevant commands (eg. `zellij action rename-tab --index <index>`).
> 
> It's now also possible to use `zellij [--session <name>] action list-tabs` or `zellij [--session <name>] action list-panes` to get a (json formatted) list of tabs/panes and their IDs so you don't have to rely on keeping this info at all.
> 
> Wdyt?

Hey @imsnif thanks for the reply! Sorry for the delay. All good, the latest versions of zellij fix most of my issues :) . I have some remaining bug fixes that are minor, I'll submit those separately based on latest head. 

I'll close the PR here for now

PR #4684, opened 2026-02-06
fix: compose NFD Hangul Jamo to NFC syllables in terminal grid

### LINKED ISSUE #3148: The Korean alphabet appears broken.

Folder names and file names written in Korean are broken.
The folder name is displayed normally in iterm 2, but it is not displayed properly in zellij. I am attaching an image related to this.

스크린샷 2024-02-18 오후 9 51 22

Please attach the files that were created in `/tmp/zellij-1000/zellij-log/` to the extent you are comfortable with.

The contents of the '$TMPDIR/zellij-501/ ' path have been compressed and attached.
[zellij-501.zip](https://github.com/zellij-org/zellij/files/14322589/zellij-501.zip)

**Basic information**

`zellij --version`: zellij 0.39.2

`stty size`: 26 103

`uname -av` or `ver`(Windows):
Darwin HyunGwangui-MacBookPro.local 23.3.0 Darwin Kernel Version 23.3.0: Wed Dec 20 21:30:44 PST 2023; root:xnu-10002.81.5~7/RELEASE_ARM64_T6000 arm64

## Further information
Reproduction steps, noticeable behavior, related issues, etc

arkjun:
## Summary

On macOS, filenames containing Korean characters are stored in NFD (Normalization Form Decomposed), where each Hangul syllable is split into separate Jamo characters (leading consonant + vowel + optional trailing consonant). The terminal grid's `add_character()` drops zero-width characters, and since `unicode-width` reports Jungseong (vowels U+1160–U+11A7) and Jongseong (trailing consonants U+11A8–U+11FF) as width 0, these combining Jamo were silently discarded. This caused Korean folder names to display as broken sequences of leading consonants only.

**Example:** NFD `폴더` (폴더) → displayed as `ᄑᄃ` instead of `폴더`

## Root Cause

1. macOS HFS+ stores filenames in NFD, decomposing Hangul syllables into Jamo
2. `unicode-width` returns width 0 for Jungseong and Jongseong Jamo
3. `Grid::add_character()` has an early return: `if character_width == 0 { return; }` which drops these combining characters

## Fix

Added Unicode composition in `Grid::print()` using `unicode_normalization::char::compose()`. When a new character can compose with the preceding character (e.g., leading consonant + vowel → syllable, or partial syllable + trailing consonant → full syllable), the composed result replaces the preceding character in the viewport via a new `replace_preceding_character_in_viewport()` helper.

This approach:
- Handles the full NFD → NFC pipeline incrementally (char by char as VTE delivers them)
- Works for all Unicode composition, not just Korean
- Has no performance impact on non-composable character sequences (single `compose()` check returns `None`)

## Changes

- `Cargo.toml` / `zellij-server/Cargo.toml` — added `unicode-normalization` dependency
- `zellij-server/src/panes/grid.rs` — composition logic in `print()` + `replace_preceding_character_in_viewport()` helper
- `zellij-server/src/panes/unit/grid_tests.rs` — 5 new tests covering NFD Hangul composition

## Test Plan

- [x] All 737 existing tests pass (`cargo test -p zellij-server`: 737 passed, 0 failed)
- [x] 5 new tests added:
  - `nfd_hangul_jamo_composed_to_nfc_syllable` — basic NFD→NFC for "폴더명"
  - `nfd_hangul_with_ansi_color_composed_correctly` — composition preserves ANSI styles
  - `nfd_hangul_mixed_with_ascii` — mixed Korean/ASCII text
  - `nfd_hangul_two_jamo_syllable` — two-Jamo syllables (no trailing consonant)
  - `nfc_hangul_not_affected_by_composition` — pre-composed NFC input unchanged
- [x] Code formatted with `cargo xtask format`

Fixes #3148

### DIFF
--- a/zellij-server/src/panes/grid.rs
+++ b/zellij-server/src/panes/grid.rs
@@ -13,6 +13,7 @@ use std::{
     str,
 };
 
+use unicode_normalization::char::compose;
 use vte;
 use zellij_utils::{
     consts::{DEFAULT_SCROLL_BUFFER_SIZE, SCROLL_BUFFER_SIZE},
@@ -1775,6 +1776,19 @@ impl Grid {
     fn set_preceding_character(&mut self, terminal_character: TerminalCharacter) {
         self.preceding_char = Some(terminal_character);
     }
+    fn replace_preceding_character_in_viewport(&mut self, new_char: &TerminalCharacter) {
+        if let Some(preceding) = &self.preceding_char {
+            let preceding_width = preceding.width();
+            let preceding_x = self.cursor.x.saturating_sub(preceding_width);
+            if let Some(row) = self.viewport.get_mut(self.cursor.y) {
+                let abs_idx = row.absolute_character_index(preceding_x);
+                if let Some(cell) = row.columns.get_mut(abs_idx) {
+                    *cell = new_char.clone();
+                }
+            }
+            self.output_buffer.update_line(self.cursor.y);
+        }
+    }
     pub fn start_selection(&mut self, start: &Position) {
         let old_selection = self.selection;
         self.click.record_click(*start);
@@ -2500,6 +2514,17 @@ impl Grid {
 impl Perform for Grid {
     fn print(&mut self, c: char) {
         let c = self.cursor.charsets[self.active_charset].map(c);
+
+        // Try to compose with preceding character for proper Unicode normalization
+        // (e.g. NFD Hangul Jamo → NFC syllable: ᄑ + ᅩ + ᆯ → 폴)
+        if let Some(ref preceding) = self.preceding_char {
+            if let Some(composed) = compose(preceding.character, c) {
+                let composed_tc = TerminalCharacter::new_styled(composed, preceding.styles.clone());
+                self.replace_preceding_character_in_viewport(&composed_tc);
+                self.preceding_char = Some(composed_tc);
+                return;
+            }
+        }
 
         let terminal_character =
             TerminalCharacter::new_styled(c, self.cursor.pending_styles.clone());
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4199,6 +4199,21 @@ dependencies = [
 ]
 
 [[package]]
+name = "tinyvec"
+version = "1.10.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "bfa5fdc3bce6191a1dbc8c02d5c8bffcf557bafa17c124c5264a458f1b0613fa"
+dependencies = [
+ "tinyvec_macros",
+]
+
+[[package]]
+name = "tinyvec_macros"
+version = "0.1.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "1f3ccbac311fea05f86f61904b462b55fb3df8837a366dfc601a0161d0532f20"
+
+[[package]]
 name = "to_method"
 version = "1.1.0"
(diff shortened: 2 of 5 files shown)


imsnif:
@tlinford - I know you've looked a bit into this in the past: what do you think? Does this cover everything it needs to?

zellij-server/src/panes/grid.rs
```diff
@@ -1775,6 +1776,19 @@ impl Grid {
     fn set_preceding_character(&mut self, terminal_character: TerminalCharacter) {
         self.preceding_char = Some(terminal_character);
     }
+    fn replace_preceding_character_in_viewport(&mut self, new_char: &TerminalCharacter) {
+        if let Some(preceding) = &self.preceding_char {
+            let preceding_width = preceding.width();
+            let preceding_x = self.cursor.x.saturating_sub(preceding_width);
+            if let Some(row) = self.viewport.get_mut(self.cursor.y) {
+                let abs_idx = row.absolute_character_index(preceding_x);
+                if let Some(cell) = row.columns.get_mut(abs_idx) {
```
daveth3t3chg33k, line 1785: Extrapolate the function to modules for better readability

tlinford:
> @tlinford - I know you've looked a bit into this in the past: what do you think? Does this cover everything it needs to?

I'm not sure honestly, I have to study a bit!
The main doubt is whether fixing this issue with this approach could have unwanted side effects/wrong behaviors, and at that point it would just be better fix the root cause by removing the whole drop of zero width characters and handling the grapheme clusters in a sensible way.

imsnif:
> I'm not sure honestly, I have to study a bit! The main doubt is whether fixing this issue with this approach could have unwanted side effects/wrong behaviors, and at that point it would just be better fix the root cause by removing the whole drop of zero width characters and handling the grapheme clusters in a sensible way.

What are we missing from this solution for it to be considered handling properly?

tlinford:
> > I'm not sure honestly, I have to study a bit! The main doubt is whether fixing this issue with this approach could have unwanted side effects/wrong behaviors, and at that point it would just be better fix the root cause by removing the whole drop of zero width characters and handling the grapheme clusters in a sensible way.
> 
> What are we missing from this solution for it to be considered handling properly?

I personally don't know enough about unicode normalization and related topics to know if these changes are correct basically 😅. On the other hand if we went with the general solution this problem plus others would be fixed.

imsnif:
Hey @arkjun - thanks for this. I looked into this a bit deeper and found that unfortunately this is a bit of a naive implementation (no fault of your own, I assume you pointed the LLM at it and asked it to fix your specific issue). Namely, the "preceding_character" can't be assumed to always be the character before the cursor (eg. when taking cursor movements into account).

In this case, we need a proper solution for not only this issue, but also for fixing other multi-part unicode composed sequences. We will need some sort of solution that changes `TerminalCharacter` to be able to store either multiple characters (if there's some sort of upper limit to unicode segmentation - my feeling is that there isn't) or to point to some sort of internal index that tracks these.

Are you interested in pursuing a complete implementation? (fair warning: it will likely be about 10x the effort)

arkjun:
@imsnif Thanks for the thorough review and the kind words! That makes total   sense - I can see how cursor movements would break the assumption.
To be honest, a full TerminalCharacter refactor is beyond what I can commit to right now. I'll close this PR for now - feel free to use any of the test cases or approach as reference if someone picks this up later. Happy to help review when a proper implementation comes along!

PR #4685, opened 2026-02-06
refactor: make tab_ids stable

imsnif:
Previously there was a bit of a mess regarding tracking a tab's position and its index. This canonicalizes the following:

1. `tab_position` / `tab_index` -> refer to the tab's ordering (they change if the user moves tabs to change their location)
2. `tab_id` -> new field, a unique unchanging id that refers to a specific tab regardless of its position

The reason we keep both the position and index terminology is that they're already part of the API.

This also adds the `tab_id` field to `TabInfo` so that plugin authors can identify and track tabs reliably.


iwinux:
Is it feasible to release a new version?

PR #4692, opened 2026-02-10
feat(plugin): add open_external API to open URLs in default app

BlackHole1:
Add a new OpenExternal plugin command that allows plugins to open URLs or file paths in the user's default external application. On macOS this uses `open`, on Linux it uses `xdg-open` (or `xdg-email` for mailto: URLs). The command requires RunCommands permission and operates as a fire-and-forget operation with no result returned to the plugin.

### DIFF
--- a/zellij-server/src/plugins/zellij_exports.rs
+++ b/zellij-server/src/plugins/zellij_exports.rs
@@ -246,6 +246,7 @@ fn host_run_plugin_command(mut caller: Caller<'_, PluginEnv>) {
                     PluginCommand::WebRequest(url, verb, headers, body, context) => {
                         web_request(env, url, verb, headers, body, context)
                     },
+                    PluginCommand::OpenExternal { url } => open_external(env, url),
                     PluginCommand::PostMessageTo(plugin_message) => {
                         post_message_to(env, plugin_message)?
                     },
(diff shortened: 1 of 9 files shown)


BlackHole1:
Hi @imsnif, I've rebased the latest main branch code and fixed the format issues.

zellij-server/src/plugins/zellij_exports.rs
```diff
@@ -2456,6 +2457,36 @@ fn web_request(
         ));
 }
 
+fn open_external(_env: &PluginEnv, url: String) {
+    if url.is_empty() {
+        log::error!("URL cannot be empty");
+        return;
+    }
+    std::thread::spawn(move || {
```
imsnif: Since this is fire-and-forget, could we move this to `BackgroundJobs`, just for order's sake?

zellij-utils/src/plugin_api/plugin_command.proto
```diff
@@ -185,17 +185,18 @@ enum CommandName {
...
+   RenameTabWithId = 195;
+   BreakPanesToTabWithId = 196;
+   GetSessionEnvironmentVariables = 197;
+   OpenCommandPaneInNewTab = 198;
+   OpenPluginPaneInNewTab = 199;
+   OpenEditorPaneInNewTab = 200;
+   OpenCommandPaneInPlaceOfPaneId = 201;
+   OpenTerminalPaneInPlaceOfPaneId = 202;
+   OpenEditPaneInPlaceOfPaneId = 203;
+   OpenExternal = 204;
```
imsnif: I know we don't format these files, but can we remove the extra indentation? :)

imsnif (changes requested):
Thanks for this! Left some minor comments.

imsnif:
Also - just noticed: you need to commit the prost assets to (what was created when running with cargo x run or cargo x build)

zellij-server/src/plugins/zellij_exports.rs
BlackHole1: Actually, in the initial implementation, I had put it in BackgroundJobs. However, because we didn't need the result, I moved it out to reduce an unnecessary call.

However, I can move it back.

BlackHole1:
@imsnif All corrected.

imsnif:
Thanks, but... I don't understand the changes to client_server_contract.rs?

PR #4694, opened 2026-02-11
fix(client): enable CSI synchronized output by default to prevent render corruption

### LINKED ISSUE #4693: fix: UI corruption (duplicate pane rendering) over SSH due to disabled synchronized output
## Describe the bug

When using Zellij over SSH (M1 Mac + Ghostty → EC2), pane rendering becomes corrupted over time — panes appear duplicated/overlapping. Detaching and reattaching the session (`zellij attach`) fully restores the display.

## To reproduce

1. SSH into a remote server (e.g. EC2 Amazon Linux)
2. Start Zellij with `TERM=xterm-256color` (the default for most SSH sessions)
3. Work normally with multiple panes for an extended period
4. Observe: pane borders and content gradually become duplicated/overlapping

**Workaround**: `zellij detach` then `zellij attach` restores the display.

## Expected behavior

Pane rendering should remain consistent over time, even over SSH connections with TCP buffering.

## Environment

- Zellij version: 0.41.x (also reproducible on earlier versions)
- OS: EC2 Amazon Linux (server), macOS M1 (client)
- Terminal: Ghostty
- TERM: xterm-256color
- Connection: SSH

## Root cause analysis

The issue is in `zellij-client/src/lib.rs` (lines 282-285 and 951-954):

```rust
let synchronised_output = match os_input.env_variable("TERM").as_deref() {
    Some("alacritty") => Some(SyncOutput::DCS),
    _ => None,  // ← CSI synchronized output disabled for non-Alacritty terminals
};
```

**Synchronized Output is only enabled for Alacritty (via DCS).** For all other terminals (including `xterm-256color`), it is set to `None`.

### Why this causes corruption over SSH:

1. Zellij's server uses differential rendering — only changed lines are sent as VT escape sequences
2. After `Grid::read_changes()` reads changed lines, `output_buffer.clear()` immediately marks them as "sent"
3. Over SSH, large render frames get split across TCP segments and arrive partially at the terminal
4. The terminal receives incomplete frames containing cursor positioning (`\x1b[row;colH`) mid-sequence, causing content to be drawn at incorrect positions
5. Since the server already cleared its output buffer, **those lines won't be re-sent until the next change** to that specific line
6. Over time, the drift between "actual terminal display" and "server's assumed display" accumulates → duplicate pane rendering

### Why detach/reattach fixes it:

`Tab::set_force_render()` + `\x1b[2J` (full screen clear) triggers a complete redraw, resynchronizing the display state.

### Why enabling Synchronized Output fixes it:

CSI Synchronized Output (`\x1b[?2026h` / `\x1b[?2026l`) wraps each render frame in begin/end markers. The terminal buffers the entire frame and renders it atomically, preventing partial frame display even when TCP segments arrive incrementally.

## Proposed fix

Change `_ => None` to `_ => Some(SyncOutput::CSI)` in both locations:

```rust
let synchronised_output = match os_input.env_variable("TERM").as_deref() {
    Some("alacritty") => Some(SyncOutput::DCS),
    _ => Some(SyncOutput::CSI),  // Enable CSI synchronized output by default
};
```

**Safety**: CSI `?2026h`/`?2026l` uses standard private mode sequences. Terminals that don't support it simply ignore these sequences. All major modern terminals support it: Ghostty, kitty, iTerm2, WezTerm, foot, contour, and others.

This is a minimal 2-line change in a single file.

kishisan:
## Summary

- Enable CSI synchronized output (`?2026h`/`?2026l`) by default for all non-Alacritty terminals, preventing UI corruption (duplicate pane rendering) over SSH
- Change `_ => None` to `_ => Some(SyncOutput::CSI)` in two locations in `zellij-client/src/lib.rs`

Fixes #4693

## Problem

When using Zellij over SSH with `TERM=xterm-256color`, pane rendering gradually becomes corrupted — panes appear duplicated/overlapping. This happens because:

1. Zellij uses differential rendering (only changed lines are sent)
2. After reading changes, the output buffer is immediately cleared
3. Over SSH, TCP buffering splits large render frames into segments
4. Partial frames cause the terminal to draw content at wrong positions
5. The cleared buffer means corrupted lines won't be re-sent until next change
6. Drift accumulates over time → duplicate pane display

Synchronized output wraps each frame in begin/end markers (`\x1b[?2026h` ... `\x1b[?2026l`), causing the terminal to buffer and render atomically — preventing partial frame display.

## Why this was disabled

The current code only enables synchronized output for Alacritty (using DCS mode). For all other terminals, it falls through to `None`. The CSI implementation (`SyncOutput::CSI`) already exists in the codebase and was added in #2798, but was never made the default.

## Safety

CSI private mode `?2026` is a well-established standard:
- **Supported by**: Ghostty, kitty, iTerm2, WezTerm, foot, contour, rio, and others
- **Unsupported terminals**: Simply ignore the private mode sequences (no side effects)
- **Alacritty**: Continues to use DCS mode as before (no change to Alacritty behavior)

## Test plan

- [ ] Verify `cargo build` succeeds (CI)
- [ ] Verify `cargo xtask clippy` passes (CI)
- [ ] Test over SSH: confirm pane rendering remains stable over extended use
- [ ] Test with `TERM=alacritty`: confirm DCS mode still works
- [ ] Test with older terminals: confirm no visual side effects from ignored `?2026` sequences

### DIFF
--- a/zellij-client/src/lib.rs
+++ b/zellij-client/src/lib.rs
@@ -281,7 +281,7 @@ pub async fn run_remote_client_terminal_loop(
 
     let synchronised_output = match os_input.env_variable("TERM").as_deref() {
         Some("alacritty") => Some(SyncOutput::DCS),
-        _ => None,
+        _ => Some(SyncOutput::CSI),
     };
 
     let mut async_stdin: Box<dyn AsyncStdin> = os_input.get_async_stdin_reader();
@@ -950,7 +950,7 @@ pub fn start_client(
     let mut pending_instructions = vec![];
     let mut synchronised_output = match os_input.env_variable("TERM").as_deref() {
         Some("alacritty") => Some(SyncOutput::DCS),
-        _ => None,
+        _ => Some(SyncOutput::CSI),
     };
 
     let mut stdout = os_input.get_stdout_writer();


imsnif:
Hi, thanks for this - however

> The current code only enables synchronized output for Alacritty (using DCS mode). For all other terminals, it falls through to `None`. The CSI implementation (`SyncOutput::CSI`) already exists in the codebase and was added in #2798, but was never made the default.

This is not true (I'm guessing LLM context size issues?) see: https://github.com/zellij-org/zellij/blob/e160e9e7ee742f84d6f793ba4741cc422d3fae44/zellij-client/src/stdin_ansi_parser.rs#L270

If you'd like to troubleshoot this further to find out what's causing the issue you're experiencing and come back with more info, that would be great.

PR #4706, opened 2026-02-13
fix: add --timeout flag to zellij web --status

marcusmichaels:
The HTTP client for checking web server status didn't have a timeout configured. This could cause the CLI to hang indefinitely if the server was unreachable.

I've added a 30-second default timeout and lets users override it with `--timeout <seconds>` when running `zellij web --status`. For example:

```
zellij web --status --timeout 60
```

### DIFF
--- a/src/tests/cli.rs
+++ b/src/tests/cli.rs
@@ -1,0 +1,69 @@
+use clap::Parser;
+use zellij_utils::cli::{CliArgs, Command};
+
+#[test]
+fn web_cli_status_alone_works() {
+    let args = CliArgs::try_parse_from(["zellij", "web", "--status"]);
+    assert!(args.is_ok());
+    if let Ok(CliArgs {
+        command: Some(Command::Web(web)),
+        ..
+    }) = args
+    {
+        assert!(web.status);
+        assert!(web.timeout.is_none());
+    } else {
+        panic!("Expected Web command");
+    }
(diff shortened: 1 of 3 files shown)


marcusmichaels:
Hey @imsnif this is just a small fix that handles a timeout on the web server status. Looks like the workflows need approvals to run, could you help with this?

imsnif:
Hey - thanks for this. Looks generally good except:

```
$ zellij web --timeout 1 --status
error: The argument '--status' cannot be used with one or more of the other specified arguments

USAGE:
    zellij web [OPTIONS]

For more information try --help
```

Otherwise, please rebase from main - this should also fix the failing flaky tests.

marcusmichaels:
Thanks @imsnif, I've updated it to work, and also added some cli tests (hope that's okay where I added them) just to make sure everything is still working as expected as I had to remove `exclusive(true)` from the `--status` flag and replace it with `conflicts_with("start")` (which was mainly to avoid having to list and maintain _all_ of the available commands.)

All of the other flags (`--stop`, `--create-token`, etc.) already have `exclusive(true)` or `conflicts_with_all` that includes `status`, so the conflict relationships are preserved.

marcusmichaels:
Tests passing 🥳

imsnif:
Great. Thank you!

PR #4708, opened 2026-02-14
fix: add missing share plugin keybinding to default config

### LINKED ISSUE #4707: fix: Default keybinding for share plugin (Ctrl+o+s) not working
## Description
The documentation at https://zellij.dev/documentation/web-client.html#how-to-start and 
https://zellij.dev/tutorials/web-client/#how-to-set-it-upmentions
using "Ctrl o + s" to launch the share plugin.

## Expected behavior
Pressing Ctrl+o followed by s should launch the share plugin.

## Actual behavior
Nothing happens when pressing Ctrl+o+s. The keybinding is missing from the default config.

## Proposed fix
Add the following binding to the session mode in `zellij-utils/assets/config/default.kdl`:

```kdl
bind "s" {
    LaunchOrFocusPlugin "zellij:share" {
        floating true
        move_to_focused_tab true
    };
    SwitchToMode "Normal"
}
```

hitsan:
## Problem
  The documentation states that pressing `Ctrl+o` followed by `s`
   should launch the share plugin, but this keybinding was
  missing from the default configuration.

  Fixes #4707

  ## Solution
  Add the `bind "s"` keybinding in session mode to the default
  KDL configuration file.

  ## Changes
  - Added share plugin keybinding in
  `zellij-utils/assets/config/default.kdl`
  - Configuration:
    - `floating: true` opens in a floating window
    - `move_to_focused_tab: true`  moves shared session to the focused tab

### DIFF
--- a/zellij-utils/assets/config/default.kdl
+++ b/zellij-utils/assets/config/default.kdl
@@ -115,6 +115,13 @@ keybinds {
         bind "Ctrl o" { SwitchToMode "Normal"; }
         bind "Ctrl s" { SwitchToMode "Scroll"; }
         bind "d" { Detach; }
+        bind "s" {
+            LaunchOrFocusPlugin "zellij:share" {
+                floating true
+                move_to_focused_tab true
+            };
+            SwitchToMode "Normal"
+        }
         bind "w" {
             LaunchOrFocusPlugin "session-manager" {
                 floating true


Zykino:
But now this block of lines you added (L118-L124) is duplicating already existing lines:

https://github.com/zellij-org/zellij/blob/a4e3b62bd8f6da8251c6ab2cba14916d0b5fefae/zellij-utils/assets/config/default.kdl#L153-L159

I think you upgraded zellij from an earlier version and forgot to dump the new config to merge it with your current one.

hitsan:
Thanks for pointing this out!

You're absolutely right — this is already included in the latest default configuration.
It seems I based this on an outdated dump.

I'll close this PR.
Thanks for the clarification!

PR #4711, opened 2026-02-17
perf(plugins): optimize session update handling to reduce unnecessary re-renders

IngmarStein:
This commit optimizes the 'plugin-manager' and 'session-manager' default plugins to avoid re-rendering when receiving 'SessionUpdate' events that do not contain actual state changes.

This fixes an issue where SSH connections would break with a 'broken pipe' error when the client machine sleeps for an extended period (>10 minutes).

The root cause was a constant stream of screen updates sent to all connected clients, even when they were idle. This stream was triggered by the 'ReadAllSessionInfosOnMachine' background job, which runs every second and broadcasts a 'SessionUpdate' event. Plugins subscribing to this event (like the session-manager) would request a render, causing 'render_to_clients' to send VTE bytes to the client.

When a client sleeps, it stops reading from the socket. The server's OS network buffer fills up with these 1Hz updates. Eventually, the connection times out or the buffer overflows, causing the pipe to break. When the client wakes up and tries to send a keepalive or ACK, it receives a RST.

This prevents the network buffer from filling up during sleep while maintaining a responsive experience upon wake-up.

### DIFF
--- a/default-plugins/plugin-manager/src/main.rs
+++ b/default-plugins/plugin-manager/src/main.rs
@@ -3,6 +3,7 @@ use fuzzy_matcher::FuzzyMatcher;
 use uuid::Uuid;
 use zellij_tile::prelude::*;
 
+use std::collections::hash_map::Entry::{Occupied, Vacant};
 use std::collections::{BTreeMap, HashMap};
 
 pub struct SearchResult {
@@ -551,19 +552,32 @@ impl ZellijPlugin for State {
                 should_render = true;
             },
             Event::SessionUpdate(live_sessions, _dead_sessions) => {
+                let mut changed = false;
                 for session in live_sessions {
                     if session.is_current_session {
                         if session.plugins != self.plugins {
                             self.plugins = session.plugins;
                             self.reset_selection();
                             self.update_search_term();
+                            changed = true;
                         }
                         for tab in session.tabs {
-                            self.tab_position_to_tab_name.insert(tab.position, tab.name);
+                            match self.tab_position_to_tab_name.entry(tab.position) {
+                                Occupied(mut entry) => {
+                                    if entry.get() != &tab.name {
+                                        entry.insert(tab.name);
+                                        changed = true;
+                                    }
+                                }
+                                Vacant(entry) => {
+                                    entry.insert(tab.name);
+                                    changed = true;
+                                }
+                            }
                         }
                     }
                 }
-                should_render = true;
+                should_render = changed;
             },
             Event::PaneUpdate(pane_manifest) => {
                 for (tab_position, panes) in pane_manifest.panes {
--- a/default-plugins/session-manager/src/new_session_info.rs
+++ b/default-plugins/session-manager/src/new_session_info.rs
@@ -137,8 +137,9 @@ impl NewSessionInfo {
             },
         }
     }
-    pub fn update_layout_list(&mut self, layout_info: Vec<LayoutInfo>) {
-        self.layout_list.update_layout_list(layout_info);
+    /// Returns a boolean indicating whether the layout list has changed
+    pub fn update_layout_list(&mut self, layout_list: Vec<LayoutInfo>) -> bool {
+        self.layout_list.update_layout_list(layout_list)
     }
     pub fn layout_list(&self, max_rows: usize) -> Vec<(LayoutInfo, bool)> {
         // bool - is_selected
@@ -263,13 +264,18 @@ pub struct LayoutList {
 }
 
 impl LayoutList {
-    pub fn update_layout_list(&mut self, layout_list: Vec<LayoutInfo>) {
+    /// Returns a boolean indicating whether the layout list has changed
+    pub fn update_layout_list(&mut self, layout_list: Vec<LayoutInfo>) -> bool {
+        if self.layout_list == layout_list {
+            return false;
+        }
         let old_layout_length = self.layout_list.len();
         self.layout_list = layout_list;
         if old_layout_length != self.layout_list.len() {
             // honestly, this is just the UX choice that sucks the least...
             self.clear_selection();
         }
+        true
     }
     pub fn selected_layout_info(&self) -> Option<LayoutInfo> {
         if !self.layout_search_term.is_empty() {
--- a/default-plugins/session-manager/src/main.rs
+++ b/default-plugins/session-manager/src/main.rs
@@ -116,16 +116,23 @@ impl ZellijPlugin for State {
                 should_render = true;
             },
             Event::SessionUpdate(session_infos, resurrectable_session_list) => {
+                let mut session_manager_changed = false;
                 for session_info in &session_infos {
                     if session_info.is_current_session {
-                        self.new_session_info
-                            .update_layout_list(session_info.available_layouts.clone());
+                        if self.new_session_info
+                            .update_layout_list(session_info.available_layouts.clone()) {
+                                session_manager_changed = true;
+                            }
                     }
                 }
-                self.resurrectable_sessions
-                    .update(resurrectable_session_list);
-                self.update_session_infos(session_infos);
-                should_render = true;
+                if self.resurrectable_sessions
+                    .update(resurrectable_session_list) {
+                        session_manager_changed = true;
+                    }
+                if self.update_session_infos(session_infos) {
+                    session_manager_changed = true;
+                }
+                should_render = session_manager_changed;
             },
             _ => (),
         };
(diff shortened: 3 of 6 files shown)


imsnif:
Hi, thanks for this. Since this issue is not easily reproducible it's hard for me to ascertain the root cause here. Regardless though, if the issue is on the network/socket side I'd prefer to have a solution on that level rather than working around it on the application layer. I would guess if the above analysis is correct there's likely a common practice to prevent such things?

IngmarStein:
So here's how I reproduced the issue:

1) Connect from a MacBook Pro to a Mac Studio over SSH and start zellij with the default layout.
2) Close the MacBook and wait for ~15 minutes
3) Open the MacBook and find that the session with zellij was dropped (other SSH sessions survive).

I've tried various SSH settings on the client side (`IPQoS`, `TCPKeepAlive`, `ServerAliveInterval`, `ServerAliveCountMax`, …), but eventually I figured that the issue wasn't about keeping an idle connection alive, but rather it's because the server continues to send data, filling the buffer while there is no client awake to receive it. The SSH sessions without zellij survived the client sleep which is why I think it's an issue at the application layer.

I think it may be related to certain layout elements, though. I've removed the status bar plugin from the default layout and that seems to mitigate the issue.

imsnif:
Thanks for the reproduction steps. I do not own nor have access to a mac. From similar attempts in linux I was not able to reproduce it in the past. I believe you there's an issue, but please believe me that reproducing these things is often less trivial than it seems. :)

My intuition tells me there's a network/socket/mac-sleep side solution for this (eg. buffering, retries, keep-alives, reconnects, etc.)

IngmarStein:
Oh, I believe you about the attempts for reproduction! :) This issue has bugged me for a long time and now I finally want to get to the bottom of it.

Maybe you didn't see this issue with Linux clients as macOS has pretty aggressive power saving measures. The default setting for "Wake for network access" is "Only on Power Adapter" which means that Macs, while on battery power, don't wake up to receive incoming network traffic.

The only setting I could imagine would make a difference is to increase the TCP buffer size on the server side, but that would just mean it takes a bit longer to fill it.

Here are the relevant sysctls on the server side as of macOS 26.3:
```
> sysctl net.inet.tcp | grep -E 'rexmt|keep|msl'
net.inet.tcp.keepidle: 7200000
net.inet.tcp.keepintvl: 75000
net.inet.tcp.keepinit: 75000
net.inet.tcp.rexmt_thresh: 3
net.inet.tcp.rexmt_slop: 200
net.inet.tcp.keepcnt: 8
net.inet.tcp.msl: 15000
net.inet.tcp.always_keepalive: 0
```

I do believe that the right solution is to not send traffic when there is no client awake to receive it. A silent SSH shell without activity shouldn't cause more traffic than the keepalives.

imsnif:
Alright - I gave this some thought and I think we have two avenues here:

1. I think the Proper Solution(tm) would be to implement buffering on the `zellij-client` side. This being a different process and on the other side of the IPC channel from the server. This would involve something along the lines of writing to STDOUT in a non-blocking way and buffering messages when we get a `WOULDBLOCK` and friends. These messages would then be replayed to the client when the channel is opened again.

2. In this case, I guess we could also solve the application-level problem by getting plugins not to render (return `false` from the `update` method) when they get a session update event and nothing in the UI would change. I was sure this is already the case, but if not it's a bug and I'd be happy to accept a fix. This however will not solve stuff along the lines of "I'm tailing a log file and putting my laptop to sleep".

Another thing you can try is using `Mosh` instead of ssh (which does these sorts of things internally).

IngmarStein:
I don't think buffering on the client side would work if the client machine is asleep…

I had actually looked at Mosh and Eternal Terminal but Mosh's most recent release was from 2022 and I'd prefer to stick to built-in solutions which don't require additional software to installed on both client and server side.

imsnif:
As far as I understand it, you're SSHing into a remote machine and running Zellij there. So both the client and server process are on the remote machine.

IngmarStein:
True. I think the problem is between the ssh client and the ssh server not the zellij client and server. Because zellij keeps rendering updates in the background, the ssh server tries to send data to the client which is sleeping. Because of this background trickle, the TCP buffer of the server side of the ssh connection is filled. This PR attempts to stop generating new data until a client returns. Does that make sense?

imsnif:
Yes - that's exactly why I think the "proper solution" would be to buffer the messages on the client's STDOUT pipe, where the problem is. :) We would then also be doing this only when the buffer is full (`WOULDBLOCK` and friends, as I said) rather than after an arbitrary amount of time of idleness (eg. 10m)

IngmarStein:
I see. I confused the zellij client and the ssh client, sorry. I re-read your proposal and it makes sense to me now.

> This however will not solve stuff along the lines of "I'm tailing a log file and putting my laptop to sleep".

I think it's fair to solve the persistence of a really idle connection before worrying about this case.

> getting plugins not to render (return false from the update method) when they get a session update event and nothing in the UI would change

I'll have a look at the default plugins again and see if they handle this correctly.

IngmarStein:
I found that the plugin-manager and session-manager unconditionally set `should_render = true`, but I'm not sure if that actually resulted in a render.

imsnif:
Play around with it, let me know if anything solves the problem for you.

IngmarStein:
So I ran this build in an ssh connection, closed the MacBook, and the session was still alive on the next day. I tested again when running on battery and that was also stable. Previously, those sessions would not have survived that long sleep.

default-plugins/plugin-manager/src/main.rs
```diff
@@ -551,19 +551,32 @@ impl ZellijPlugin for State {
...
                         if session.plugins != self.plugins {
                             self.plugins = session.plugins;
                             self.reset_selection();
                             self.update_search_term();
+                            changed = true;
                         }
                         for tab in session.tabs {
-                            self.tab_position_to_tab_name.insert(tab.position, tab.name);
+                            match self.tab_position_to_tab_name.entry(tab.position) {
+                                std::collections::hash_map::Entry::Occupied(mut entry) => {
```
imsnif: Cool trick, I did not know about `Occupied` / `Vacant`. Could we import them at the beginning of the file though rather than use their full path?

default-plugins/session-manager/src/new_session_info.rs
```diff
@@ -263,13 +263,17 @@ pub struct LayoutList {
 }
 
 impl LayoutList {
-    pub fn update_layout_list(&mut self, layout_list: Vec<LayoutInfo>) {
+    pub fn update_layout_list(&mut self, layout_list: Vec<LayoutInfo>) -> bool {
```
imsnif, line 268: Could we please add comments in this and all other similar functions explaining what the `bool` return value means?

zellij-server/src/screen.rs
```diff
@@ -1737,11 +1737,13 @@ impl Screen {
             non_watcher_output_was_dirty = output.is_dirty();
             if non_watcher_output_was_dirty {
                 let serialized_output = output.serialize().context(err_context)?;
-                let _ = self
-                    .bus
-                    .senders
-                    .send_to_server(ServerInstruction::Render(Some(serialized_output)))
-                    .context(err_context);
+                if !serialized_output.is_empty() {
```
imsnif: I'm a little wary of this as it's a double-guard (we have `non_watcher_output_was_dirty` above) and might be confusing to debug in the future. Is this a must? If so, what's being rendered here if we remove the if in the case we're concerned with?

default-plugins/session-manager/src/new_session_info.rs
IngmarStein: Done

default-plugins/plugin-manager/src/main.rs
IngmarStein: Done

IngmarStein:
Dang. Looks like I celebrated too early:

Bildschirmfoto 2026-02-18 um 18 43 20

I think I'll give up at this point. Up to you if you still want this change or not. Let me know :)

imsnif:
Thank you for your efforts! At the moment, close to wrapping up the upcoming version, I only want to merge things that have a meaningful effect on end-users. Fixing this bug would definitely fall into this category, but I think we need more investigation here.

If you'd like to pick this up now or in the future that would be very welcome.

IngmarStein:
Makes sense, we can keep this open for now. I will certainly investigate some more up and down the stack (zellij, ssh, terminal, TCP/IP, OS, …).

PR #4720, opened 2026-02-18
[Windows port PR1] refactor: replace nix::Error with std::io::Error in public APIs

divens:
[PR 1](https://github.com/zellij-org/zellij/issues/316#issuecomment-3920047803) of the Windows port.

- Replace `nix::Error` with `std::io::Error`
- change in `zellij-client/src/os_input_output.rs` l130 is due to formatting


imsnif:
Nice. Next!

PR #4722, opened 2026-02-18
[Windows port PR3] refactor: replace RawFd/AsyncFile, introduce platform-split PTY backend

divens:
[PR3](https://github.com/zellij-org/zellij/issues/316#issuecomment-3920047803) of the Windows port.
- `RawFd` → `i32` in `debug_to_file` (`zellij-utils/src/logging.rs`) (removes the Unix-only `RawFd` type from a cross-platform utility function)
- Platform-split architecture (Unix-specific PTY/signal code extracted from `os_input_output.rs` into `os_input_output_unix.rs`; `os_input_output_windows.rs` added with stubs, gated with `#[cfg(not(windows))]` / `#[cfg(windows)]` in `lib.rs`)
- `UnixPtyBackend` struct (encapsulates `orig_termios` and `terminal_id_to_raw_fd`, composed into `ServerOsInputOutput`)
- Lazy `AsyncFd` initialization in `RawFdAsyncReader` (upstream wraps the fd in `tokio::fs::File` immediately; our version defers `AsyncFd` registration to first read() since `spawn_terminal` runs outside the Tokio runtime, uses epoll via `AsyncFd` instead of `tokio::fs::File`)
- `NullAsyncReader` (type-safe placeholder for held panes, replacing the original `terminal_id` as `i32` in the `RawFd` slot)
- `build_command()` extracted (converts `TerminalAction` → `RunCommand` in shared code rather than inline in the platform-specific `spawn_terminal` free function)
- `nix::unistd::Pid` → `u32` throughout `pty.rs` and `ServerOsApi` trait (`get_cwd(Pid)` → `get_cwd(u32)`, `get_cwds(Vec<Pid>)` → `get_cwds(Vec<u32>)`)
- `RawFd` removed from `ServerOsApi` trait (`read_from_tty_stdout(fd: RawFd, ...)` and `async_file_reader(fd: RawFd)` removed; `spawn_terminal` now returns `Box<dyn AsyncReader>` directly instead of a `RawFd`)
- `terminal_bytes.rs` simplified (no longer takes `RawFd` pid or `Box<dyn ServerOsApi>`; receives `Box<dyn AsyncReader>` directly from caller)
- `pty.rs` reader type cleanup (`id_to_child_pid` changed from `HashMap<u32, RawFd>` to `HashMap<u32, u32>`; all `RawFd`/`AsyncFile` references replaced with `Box<dyn AsyncReader>`)
- 4 test mocks updated (`spawn_terminal` / `re_run_command_in_terminal` return types aligned with new API)
- Integration tests gated (signal tests in `os_input_output_tests.rs` gated with `#[cfg(not(windows))]`)
- tokio net feature added (required for `AsyncFd`)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -99,7 +99,7 @@ strum_macros = { version = "0.27.2", default-features = false }
 suggest = { version = "0.4.0", default-features = false }
 tempfile = { version = "3.2.0", default-features = false }
 termwiz = { version = "0.23.2", default-features = false }
-tokio = { version = "1.40", default-features = false, features = ["fs", "rt-multi-thread", "macros", "time", "process", "signal", "io-util", "io-std"] }
+tokio = { version = "1.40", default-features = false, features = ["fs", "rt-multi-thread", "macros", "time", "process", "signal", "io-util", "io-std", "net"] }
 thiserror = { version = "1.0.40", default-features = false }
 tokio-util = { version = "0.7.15" }
 unicode-width = { version = "0.1.8", default-features = false }
--- a/zellij-server/Cargo.toml
+++ b/zellij-server/Cargo.toml
@@ -18,7 +18,6 @@ byteorder = { version = "1.4.3", default-features = false, features = ["std"] }
 bytes = { version = "1.6.0", default-features = false, features = ["std"] }
 cassowary = { version = "0.3.0", default-features = false }
 chrono = { version = "0.4.19", default-features = false, features = ["std", "clock"] }
-close_fds = { version = "0.3.2", default-features = false }
 crossbeam = { version = "0.8", default-features = false, features = ["crossbeam-channel", "std"] }
 daemonize = { workspace = true }
 highway = { version = "0.6.4", default-features = false, features = ["std"] }
@@ -30,6 +29,7 @@ libc = { workspace = true }
 log = { workspace = true }
 nix = { workspace = true }
 notify-debouncer-full = { workspace = true }
+close_fds = "0.3.2"
 prost = { workspace = true }
 regex = { workspace = true }
 semver = { version = "0.11.0", default-features = false }
--- a/zellij-server/src/lib.rs
+++ b/zellij-server/src/lib.rs
@@ -1,3 +1,10 @@
+#[cfg(not(windows))]
+#[path = "os_input_output_unix.rs"]
+mod os_input_output_unix;
+#[cfg(windows)]
+#[path = "os_input_output_windows.rs"]
+mod os_input_output_windows;
+
 pub mod os_input_output;
 pub mod output;
 pub mod panes;
--- a/zellij-server/src/os_input_output.rs
+++ b/zellij-server/src/os_input_output.rs
@@ -1,21 +1,13 @@
 use crate::{panes::PaneId, ClientId};
 
 use interprocess::local_socket::LocalSocketStream;
-use nix::{
-    pty::{openpty, OpenptyResult, Winsize},
-    sys::{
-        signal::{kill, Signal},
-        termios,
-    },
-    unistd,
-};
-use tokio::{fs::File as AsyncFile, io::AsyncReadExt};
+
+#[cfg(not(windows))]
+use crate::os_input_output_unix::UnixPtyBackend as PtyBackendImpl;
+#[cfg(windows)]
+use crate::os_input_output_windows::WindowsPtyBackend as PtyBackendImpl;
 
 use interprocess;
-use libc;
-use nix;
-use signal_hook;
-use signal_hook::consts::*;
 use sysinfo::{ProcessExt, ProcessRefreshKind, System, SystemExt};
 use tempfile::tempfile;
 use zellij_utils::{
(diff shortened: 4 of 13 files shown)


divens:
@imsnif That's the biggest PR. Definitely take whatever time you need for review.

Also for whatever reason, some jobs failed to launch properly in the Rust CI workflow. Can you relaunch them please?

imsnif:
Cool - I'll go over this in more detail tomorrow. Without looking at the code though, I'm a little hesitant about introducing a new crate. While I do definitely want to break the monolith into more crates, I prefer to be deliberate about it. Introducing a new crate means we're in build-system/ci/different-installation-method/possible-headache-for-packagers territory. Is it absolutely necessary? Can we do with keeping the os_input_output modules and using conditional compilation as needed?

> Also for whatever reason, some jobs failed to launch properly in the Rust CI workflow. Can you relaunch them please?

Yeah, this is a github being github sort of thing. I fixed it in main, if you rebase it should fix it (and re-run).

divens:
Ok, I understand your concerns. What I think would work best then would be to split `zellij-os` crate as modules across `zellij-client` and `zellij-server`. What about structuring the code around an `os` module like so

zellij-server/src/os/mod.rs -> pty.rs, process.rs, shell.rs
zellij-client/src/os/mod.rs -> signals.rs

This way we have :
- no new crate
- portable-pty stays a server-only dependency
- we have a dedicated module for os-related code, with cfg gates in dedicated files

This approach would make the change more manageable.
Would this work for you?

divens:
So I went ahead with the changes I suggested above : no more crate, using a module instead. For now it is only in `zellij-server`, handling PTY operations and initial signal handling refactoring.

Let me know if it looks ok to you.

imsnif:
Hi @divens - thanks for the quick change and great work on this!

Some comments:
1. Most importantly: the `spawn_blocking` in terminal_bytes has to go. This would essentially mean that every terminal pane we open will take up an OS thread all to itself, a thread which will spend most of its time idle. If we open 50-100 panes (which is not an unheard of scenario) this will start being a problem. In the current main we wrap this in an async file abstraction, which drills down to an os-level epoll - meaning essentially the kernel handles this for us. I understand this might be an issue with windows, but I think we can probably find similar solutions there (a few minutes ago we just merged https://github.com/zellij-org/zellij/pull/4672, moving us from async-std to tokio, which I think will make things easier for you both on the unix and windows sides.)
2. A little less critical but also something we need to address: we no longer pass the `original_termios` to the spawned terminals, which is actually a good thing. But the one thing that I don't think portable-pty handles for us is figuring out the user's `VERASE` (see https://www.man7.org/linux/man-pages/man3/termios.3.html). We should probably poll for it before opening a terminal and make sure to pass it along.
3. Naming: I definitely understand where you're coming from with the `os` module and the `pty.rs` file, but I find these a bit confusing. We already have a pty.rs (as well as a pty thread), and while I realize this name is now becoming less relevant I would still like to keep it for the time being. Same for the `os` module versus `os_input_output.rs`. Ideally, I'd like all of the os level abstractions to live in os_input_output.rs - but I realize this might be a tall order. What do you think?
4. Signals: I see these are stubbed out in process.rs - which makes sense as far as my limited understanding of windows goes - but we do need to handle them. I guess this is implemented in the following PRs?
5. In the same venue as point 4 - I'd rather not match `not(unix)` in this case because it's a bit too broad of a catchall. Could we maybe invert it to a windows/not-windows? I'm pretty sure windows is the odd one out here.

I might have more comments once these are addressed. Good stuff so far!

divens:
Hi @imsnif , thanks for the feedback.

I updated the PR.

Removing `spawn_blocking` had me reconsider the approach, and I decided to remove `portable-pty` entirely. On Unix, it abstracted away the file descriptor, making it impossible to have epoll. On Windows, it creates pipes in blocking mode which make it inadequate for overlapped io.

We'll go for a split unix/windows implementation (which is pretty much what's in portable-pty, but we'll have more control).
So on Unix we're back to using openpty and basically the previous code.

Regarding naming and code organization, I removed the os module, the code lives in `os_input_output*`:
- shared core in `os_input_output`
- unix-specific in `os_input_output_unix`
- windows-specific in `os_input_output_windows` (stubs for now, will be implemented in later PRs)
This seems cleaner that cfg gates all over the place in a single file.

For further PRs, it shouldnt affect too much 4/5/6 (signals, IPC, daemonization). It's mostly the Windows implementation of PTY that will differ, we'll use native Windows APIs directly rather than relying on portable-pty.

imsnif:
Right, thanks for the changes @divens - this is much much better!

Only one small technical issue and two styling issues, and then we can merge:
1. The `command_exists` check that we had in the original `handle_openpty` was lost and moved to the `spawn_terminal`. This bypasses at least one path I could find, which is rerunning a command inside a terminal. There can be cases where the command was deleted before the re-run, and then we would crash. Could we return the check to its original position?
2. Styling: there are lots of places where `use` imports are placed in the middle of the file, inside functions, and such. I know this is likely an LLM context issue, but could we please endeavor to place all of them at the beginning of the file? Similarly with fully qualified import paths being used instead of just importing them.
3. A bit of a tangent, but I hope you'll bear with me - could we please remove the LLM's commits and write them under your name? I don't want to give a commercial company free ad-space in our commit log. :)

Great stuff once more, let's keep at it!

divens:
Thanks for your additional feedback and patience @imsnif 

1. Good catch. Fixed now.
2. Agreed, it gets noisy really fast. I think with this PR update, we're striking a good balance between specificity and readability.
3. I understand where you're comming from. Not a problem. I removed the co-authored statements in the commits. 

Let me know if there are more improvements needed.

imsnif:
Amazing. Next!

PR #4737, opened 2026-02-20
feat: add per-pane default foreground/background color support

### LINKED ISSUE #4712: Per-pane default foreground/background color support (OSC 10/11)
Not sure if this is on the roadmap but I've always wanted per-pane background color support. I noticed the OSC 10/11 query handling and OSC 110/111 reset stubs are already in grid.rs, so the foundation is partially there.

I'm working on a PR that adds:
- OSC 10/11 set support (programs inside a pane can set/query default fg/bg)
- OSC 110/111 reset support (fills in existing stubs)
- Layout config: default_fg/default_bg per-pane properties

Colors are applied at render time as a fallback — cells with explicit styles are unaffected.

I considered whether this could be a plugin, but OSC 10/11 handling lives in the VTE dispatch layer and the color fallback is in the output serializer, so it's inherently core terminal emulator behavior (similar to the existing OSC 4 palette support).

Tests pass, still doing manual testing. Would you be interested in a PR?

Here's a live example:

Image

karlbunch:
Add per-pane default foreground and background colors, configurable through four surfaces:

- OSC 10/11 escape sequences (set/query) and OSC 110/111 (reset)
- CLI action: zellij action set-pane-color --pane-id <id> --fg/--bg
- Layout config: default_fg/default_bg properties on pane nodes
- Plugin API: set_pane_color() SDK function and PaneInfo read-side

Colors are applied at render time in `serialize_chunks` as a fallback when cells have no explicit fg/bg, without mutating stored grid data.

Open to ideas on the render approach — applying defaults at serialize time was the least invasive option I could find, but happy to hear alternatives.

Closes #4712

zellij-background-support-v2


karlbunch:
I created a plugin to exercise the code too it's here: [zellij-colorful](https://github.com/karlbunch/zellij-colorful)

zellij-server/src/tab/mod.rs
```diff
@@ -2445,6 +2450,25 @@ impl Tab {
                 .values()
                 .any(|s_p| s_p.1.pid() == PaneId::Plugin(plugin_id))
     }
+    pub fn set_pane_color(
+        &mut self,
+        pane_id: PaneId,
+        fg: Option<String>,
+        bg: Option<String>,
+    ) -> Result<()> {
+        use crate::panes::xparse_color;
```
imsnif: Could we move the parsing to grid? I feel it would make more sense.

zellij-server/src/tab/mod.rs
```diff
@@ -5480,6 +5508,15 @@ pub fn pane_info_for_pane(
     pane_info
 }
 
+fn ansi_code_to_color_string(code: AnsiCode) -> Option<String> {
```
imsnif: Could we also move this to grid? Ideally we'd even do the whole parsing/serializing there - I think it would make more sense. Then it'll also be trivial to add these to plugin panes

zellij-server/src/output/mod.rs
```diff
@@ -124,12 +126,27 @@ fn serialize_chunks_with_newlines(
...
+                    current_character_styles.foreground = Some(fg);
+                }
+            }
+            if current_character_styles.background.is_none()
+                || current_character_styles.background == Some(AnsiCode::Reset)
+            {
+                if let Some(bg) = pane_default_bg {
+                    current_character_styles.background = Some(bg);
+                }
+            }
```
imsnif: Could we move this block to something like `adjust_styles_for_custom_bg_fg` similar to `adjust_styles_for_possible_selection`?

zellij-server/src/output/mod.rs
```diff
@@ -183,12 +202,26 @@ fn serialize_chunks(
...
```
imsnif: Similar to the comment above :)

karlbunch:
Sorry I was traveling I will circle back on this tomorrow morning just got home, but thanks for the review, and for me to "test" I had to kill everything like I had this alias I was using I call zj-wipe:

```
zj-wipe ()
{
    find /run/user/$(id -u)/zellij /tmp/zellij-$(id -u)/contract_version_1 ! -type d 2> /dev/null | xargs -t --no-run-if-empty fuser -k;
    command zellij delete-all-sessions -yf && echo "** cleared sessions **";
    command zellij ls;
    rm -fr ~/.cache/zellij/ /tmp/zellij-$(id -u)/ && echo "** zellij cache and /tmp/zellij-$(id -u) cleared **";
    rm -vfr /run/user/$(id -u)/zellij && echo "** cleared /run/user/$(id -u)/zellij **"
}
```
and of course had to use the latest compiled version and there are proto changes etc.

imsnif:
Hi @karlbunch - I don't want to rush you, but I'm starting to wrap things up for the upcoming release. Would love to have this included. Do you think you'll find the time to do another round or two (hopefully one will be enough) in the next few days?

karlbunch:
Ok, I've made updates locally to address all your comments. I need to head out, might be able to push at lunch break (if I get one). It should be today or this evening, Eastern time.

imsnif:
Hey @karlbunch - thanks for the changes! This looks great now. I pushed some changes, namely:
1. Fixed the stack overflow issue (the CLI you added pushed a certain stack allocated struct over the threshold - just bad luck that it happened here - fixed it by moving stuff around)
2. Added sorting to the protobuf file generation (the sorting was not deterministic so this just happened to generate a large diff in the generated files - again, just bad luck that it happened here)
3. Ran rustfmt
4. Rebased from main

As soon as CI passes I'll merge. Thanks for this great feature!

karlbunch:
Thanks @imsnif sorry about the misses there still learning how to muck around in this project.

Also, would appreciate your feedback on #4750, not sure if that's the best solution, but it works well for the tab-bar plugin I've written for my own use cases.

imsnif:
No worries, it's just bad luck and tedious repo administration :) Speaking of administration: I'll update the snapshots and push once I've built the assets locally and ran the tests to generate them.

As for the other PR, I'll see if I get a chance but I'm currently at a place where I'm trying to reduce changes as much as possible before the next release, so no promises.

karlbunch:
That reminds me I did read the https://github.com/zellij-org/zellij/blob/main/CONTRIBUTING.md and used xtask etc, but I wasn't sure if I should be pushing the snaps etc, it was bit unclear if the artifacts were just bad .gitignore setup etc, I'm used to tests not "building" artifacts so perhaps that was my miss.

imsnif:
All good. Most of the time I cba to run the e2e tests locally and just run+push the snapshots later if the CI fails.
Thanks again!

karlbunch:
I forgot to ask @imsnif do you need/want me to push a PR for the docs site for this change?

imsnif:
Thanks for the offer but no need. I do these in batches before I release the version.

PR #4738, opened 2026-02-20
[Windows port PR4] refactor: split client signal handling into platform files

divens:
[PR4](https://github.com/zellij-org/zellij/issues/316#issuecomment-3920047803) of the Windows port.

Extract signal handling from os_input_output.rs into platform-specific files, mirroring the server-side architecture from PR3:

- os_input_output_unix.rs: AsyncSignalListener (tokio unix signals), BlockingSignalIterator (signal_hook)
- os_input_output_windows.rs: stubs (unimplemented)
- os_input_output.rs: shared SignalEvent enum, AsyncSignals trait, handle_signals() now uses BlockingSignalIterator + SignalEvent

Moved signal-hook to cfg(unix)-only dependency in zellij-client.

### DIFF
--- a/zellij-client/Cargo.toml
+++ b/zellij-client/Cargo.toml
@@ -36,7 +36,6 @@ regex = { workspace = true }
 serde = { workspace = true }
 serde_json = { workspace = true }
 serde_yaml = { version = "0.8", default-features = false }
-signal-hook = { workspace = true }
 termwiz = { workspace = true }
 url = { workspace = true }
 uuid = { workspace = true }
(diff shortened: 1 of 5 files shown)


zellij-client/Cargo.toml
```diff
@@ -49,6 +48,9 @@ tokio-tungstenite = { version = "0.20", features = ["native-tls"] } #TODO: see a
 futures-util = "0.3"
 urlencoding = "2.1"
 
+[target.'cfg(unix)'.dependencies]
```
imsnif: Could we make this `cfg(not(windows))` to match the rest of our related patterns?

imsnif:
Good to merge after addressing the above comment

zellij-client/Cargo.toml
divens: done

imsnif:
Great, next!

PR #4750, opened 2026-02-21
fix: enlarge small tiled panes for plugin permission prompts

### LINKED ISSUE #4749: Plugin permission prompt not visible in small tiled panes (tab-bar, status-bar)
When a third-party plugin runs in a small tiled pane (e.g. a 1-row borderless tab-bar or status-bar), the permission prompt gets crammed into that single row. It's effectively invisible, users may not realize the plugin is waiting for permission approval and can't figure out how to grant it.

Other plugin authors have noted this and suggest workarounds like "click on the bar and press y". But that means users are blindly granting permissions without reading what's being requested. This is a security concern since the prompt lists specific permissions (ReadApplicationState, ChangeApplicationState, etc.) that should be reviewed before approving.

The existing suppressed pane mechanism handles this nicely for headless plugins loaded via `load_plugins`, they get a floating popup. But small tiled panes don't benefit from that path because they're visible, just too small to read.

**Steps to reproduce:**
1. Configure a third-party plugin (not `zellij:` built-in) in a 1-row tab-bar position
2. Clear the permission cache (`rm ~/.cache/zellij/permissions.kdl`)
3. Start zellij — the permission prompt is invisible in the 1-row pane

**Proposed fix:**

When a tiled pane is too small to display the permission prompt, temporarily enlarge it using `set_geom_override` with a frame so the verbose prompt is actually visible and the pane is focusable. After the user responds y/n, `reset_size_and_position_override` restores the original geometry.

I have a working branch I'm doing manual testing with, seems to work well, and it's just a handful of lines in `zellij-server/src/tab/mod.rs`.

karlbunch:
## Summary
- When a plugin in a small tiled pane (e.g. 1-row tab-bar) requests permissions, temporarily enlarge it using `set_geom_override` so the verbose prompt is visible
- Pane gets a frame and becomes selectable/focused for the duration of the prompt
- After y/n response, `reset_size_and_position_override` restores the original geometry
- Fixes #4749

## Test plan
- [ ] Load a third-party plugin in a 1-row borderless tab-bar position with cleared permission cache
- [ ] Verify the pane enlarges and shows the full permission prompt with frame
- [ ] Press y — verify pane shrinks back to 1 row and plugin works normally
- [ ] Press n — verify pane shrinks back and plugin is denied
- [ ] Verify normal-sized panes still show permissions inline as before
- [ ] Verify suppressed/headless plugin permissions still float as before


cloudiful:
I hit the same issue while working on my own tab bar plugin, so I’d love to see this merged soon 🙂

One thing I noticed is that geom_override() is already used in [toggle_pane_fullscreen()](https://github.com/zellij-org/zellij/blob/main/zellij-server/src/panes/tiled_panes/mod.rs#L2534-L2583), so using it as the signal for whether the pane was temporarily enlarged seems too broad. The cleanup also restores hard-coded borderless/selectable values instead of the pane’s original state.

Maybe the safer approach is to track this permission-prompt promotion explicitly and restore from saved state? Something like this:

```rust
struct PermissionPromptRestoreState {
    borderless: bool,
    selectable: bool,
}

permission_prompt_restore_state.insert(
    pid,
    PermissionPromptRestoreState {
        borderless: plugin_pane.borderless(),
        selectable: plugin_pane.selectable(),
    },
);

// ... apply temporary override ...

if let Some(prev) = permission_prompt_restore_state.remove(&pid) {
    active_pane.reset_size_and_position_override();
    active_pane.set_borderless(prev.borderless);
    active_pane.set_selectable(prev.selectable);
}
```

PR #4768, opened 2026-02-25
[Windows port PR8] feature: add Windows support

divens:
[PR8](https://github.com/zellij-org/zellij/issues/316#issuecomment-3920047803) of the Windows port.

This is the final PR of the Windows port effort.

Native Windows runtime for Zellij — ConPTY backend, named pipe IPC, console input handling, signal support, CI, and cross-platform test adaptations.

## Overview

This PR implements all Windows-specific runtime code needed to run Zellij natively on Windows. It builds on PRs 1–7 which refactored the codebase for cross-platform compilation. The key subsystems are:

- **ConPTY pseudo-terminal backend** — full lifecycle management
- **Named pipe IPC** with a dual-pipe protocol (replacing Unix domain sockets)
- **Dual-path console input** — native Console API vs raw VT byte parsing
- **Signal handling** — Ctrl+C/break/close and resize polling
- **Plugin/WASM fixes** — directory handles, path sanitization
- **CI** — Windows build and test jobs

## IPC: Named Pipes with Dual-Pipe Protocol

**Problem:** Unix domain sockets are bidirectional over a single fd. Windows named pipes are created in half-duplex mode. This simplifies implementation of overlapped IO.

**Solution:** Six abstraction functions in `zellij-utils/src/consts.rs` that wrap `interprocess::local_socket`:

| Function | Unix | Windows |
|----------|------|---------|
| `ipc_connect(path)` | `GenericFilePath` connect | `GenericNamespaced` connect |
| `ipc_bind(path)` | `GenericFilePath` listener | `GenericNamespaced` listener + PID marker file |
| `ipc_bind_async(path)` | async variant | async variant + PID marker file |
| `ipc_connect_reply(path)` | — | connects to `{path}-reply` pipe |
| `ipc_bind_reply(path)` | — | binds `{path}-reply` pipe |

The protocol:
```
client  ──main pipe──▶  server   (ClientToServerMsg)
client  ◀─reply pipe──  server   (ServerToClientMsg)
```

**Session discovery:** `ipc_bind()` on Windows writes a marker file at the socket path containing the server PID. Session enumeration uses `readdir` + `is_file()` (same pattern as Unix's `is_socket()`). Stale session cleanup reads the PID from the marker file and checks liveness via `OpenProcess(PROCESS_QUERY_LIMITED_INFORMATION)`.

**Kill/delete session:** Must connect to the reply pipe, send the message on the main pipe, then wait for the Exit response on the reply pipe — otherwise the server may not process the message before the pipe disconnects. Applied in both CLI (`sessions.rs`) and the session-manager plugin (`zellij_exports.rs`).

## ConPTY Backend

`zellij-server/src/os_input_output_windows.rs` replacing stubs with implementation.

**Spawn sequence:**
1. Create overlapped named pipe for output (`FILE_FLAG_OVERLAPPED` read end for IOCP)
2. Create anonymous pipe for input
3. `CreatePseudoConsole` with initial 80x24 size
4. Build `PROC_THREAD_ATTRIBUTE_PSEUDOCONSOLE` attribute list
5. `CreateProcessW` with `EXTENDED_STARTUPINFO_PRESENT | CREATE_UNICODE_ENVIRONMENT`
6. Spawn exit-monitoring thread (`WaitForSingleObject(INFINITE)`)
7. Wrap output handle in async reader (lazy IOCP registration)

**Key decisions:**

- **HPCON attribute passing:** `UpdateProcThreadAttribute` for `PROC_THREAD_ATTRIBUTE_PSEUDOCONSOLE` takes the HPCON *value* as a pointer (`hpcon as *const c_void`), NOT a pointer to the variable. Documented in microsoft/terminal#6705.
- **Lazy async reader:** The output pipe handle is stored raw and only promoted to a tokio `NamedPipeServer` (IOCP registration) on first `read()`, since the tokio reactor isn't available at spawn time.
- **Command resolution:** `CreateProcessW` doesn't reliably resolve bare command names through PATH. `resolve_command()` + `find_executable()` search PATH with PATHEXT extensions (`.exe`, `.cmd`, etc.), replacing the Unix-only `command_exists()`.
- **Signal delivery:** `send_sigint` uses `GenerateConsoleCtrlEvent(CTRL_C_EVENT)` with fallback to `TerminateProcess`.

## Client Input Handling

Two input strategies selected at runtime in `stdin_handler.rs` based on the `TERM` environment variable:

**Path A — Native console (no TERM):** cmd.exe, PowerShell, Windows Terminal. Uses `crossterm::event::read()` which reads `INPUT_RECORD`s via `ReadConsoleInput`. Key events are converted by `cast_crossterm_key()`, mouse events by `from_crossterm_mouse()`.

**Path B — Terminal emulator (TERM set):** Alacritty, etc. Enables `ENABLE_VIRTUAL_TERMINAL_INPUT` on stdin so `ReadFile` returns raw VT bytes. Uses the existing termwiz parser (same as Unix).

**`cast_crossterm_key()`** (260 lines in `zellij-utils/src/input/mod.rs`):
- Strips redundant Shift modifier — Windows reports physical Shift alongside the already-translated character (Shift+1 → `Char('!')` + SHIFT). Unix just reports `!`.
- AltGr handling — AltGr = Ctrl+Alt on Windows. When both are present and the character is printable, both modifiers are stripped. Needed for French AZERTY, German layouts, etc.
- Synthesizes VT raw bytes for special keys (arrows, function keys, Home/End, Ctrl+letter, Alt+letter).

**Console mode race fix:** `enable_vt_input()` sets an explicit console mode (0x0298) rather than read-modify-write, eliminating a TOCTOU race with crossterm's `EnableMouseCapture` which also calls `SetConsoleMode`. Deliberately clears `ENABLE_QUICK_EDIT_MODE` (would intercept mouse events).

**Mouse support in `os_input_output.rs`:**
- VT path: Enables `ENABLE_VIRTUAL_TERMINAL_PROCESSING` on stdout (triggers ConPTY passthrough mode), then writes ANSI mouse-enable sequences.
- Console API path: Uses `crossterm::execute!(EnableMouseCapture)`.

**ANSI terminal queries skipped on Windows:** The startup query mechanism (escape sequences written to stdout, responses read from stdin) hangs on Windows Terminal. Skipped with immediate `DoneParsing`.

## Signal Handling

`zellij-client/src/os_input_output_windows.rs`:

- **`AsyncSignalListener`:** tokio `signal::windows` for ctrl_c/ctrl_break/ctrl_close; polls `crossterm::terminal::size()` every 100ms for resize.
- **`BlockingSignalIterator`:** `SetConsoleCtrlHandler` callback sets an `AtomicBool`; iterator polls the bool + terminal size every 50ms.

## Server Spawn and Stack Size

- **`spawn_server()`:** Uses `CREATE_NO_WINDOW | CREATE_NEW_PROCESS_GROUP` (not `DETACHED_PROCESS` which would leave stdin/stdout as NULL, breaking PTY creation).
- **8MB stack:** Applied to `main()`, pty, screen, and plugin threads. Windows default is 1MB; KDL config parsing uses deep recursion that overflows it.

## Plugin/WASM Fixes

- **Directory handles:** `File::open()` on a directory fails on Windows. Uses `OpenOptionsExt::custom_flags(FILE_FLAG_BACKUP_SEMANTICS)` in `plugin_loader.rs`.
- **Plugin URL paths:** URLs like `zellij:tab-bar` contain colons, illegal in Windows paths. `make_plugin_url_path_safe()` replaces colons with underscores.

## Default Shell

`get_default_shell()` on Windows checks `$SHELL` first (Git Bash, MSYS2), falls back to `$COMSPEC`, then `cmd.exe`.

## Web Server Daemonization

`daemonize_web_server()` spawns a detached process with `CREATE_NO_WINDOW | CREATE_NEW_PROCESS_GROUP`, then polls TCP connect on the server port every 100ms (10s deadline) before exiting the parent.

## Cross-Platform Code Modifications

| File | Change | Reason |
|------|--------|--------|
| `consts.rs` | 6 IPC abstraction functions | Wrap Unix sockets vs Windows named pipes behind a common interface |
| `sessions.rs` | `assert_socket()` + dual-pipe kill/delete | PID-based liveness check; dual-pipe handshake required on Windows |
| `input/mod.rs` | `cast_crossterm_key()`, `from_crossterm_mouse()` | Convert Windows console events to zellij types |
| `os_input_output.rs` (client) | VT/Console mouse branching, `connect_to_server()` dual-pipe | Mouse enable must use ANSI sequences on VT path, Console API on native path |
| `stdin_handler.rs` | Dual input loop, `enable_vt_input()`, skip ANSI queries | Two fundamentally different input sources on Windows |
| `lib.rs` (client) | `InputInstruction::MouseEvent` variant, `spawn_server()` | Crossterm mouse events need separate variant; Windows server spawn |
| `input_handler.rs` | Handle `MouseEvent` instruction | Dispatch crossterm mouse events from native console path |
| `os_input_output.rs` (server) | `resolve_command()`, `new_client_with_reply()`, Windows `run_command_hook()` | PATH/PATHEXT resolution; reply pipe; `cmd /C` for hooks |
| `lib.rs` (server) | Dual-pipe listener, 8MB thread stacks | Reply pipe accept before route thread; stack overflow prevention |
| `pty.rs` | Windows `get_default_shell()` | $SHELL → $COMSPEC → cmd.exe fallback chain |
| `plugin_loader.rs` | `open_dir()` with `FILE_FLAG_BACKUP_SEMANTICS` | Windows requires this flag for directory handles |
| `wasm_bridge.rs` | `make_plugin_url_path_safe()` | Colons in plugin URLs are illegal in Windows paths |
| `zellij_exports.rs` | Dual-pipe `kill_sessions()` | Plugin kill session needs same reply pipe handshake |
| `main.rs` | 8MB stack trampoline | Windows 1MB default stack overflows on KDL parsing |

## Test Adaptations

- **IPC socket tests:** Cross-platform rewrite with `GenericNamespaced` on Windows; `is_socket()` tests gated to `#[cfg(unix)]`
- **Layout tests:** `normalize_layout_debug()` for `\\` → `/` path normalization; `env_var_expansion` gated to `#[cfg(unix)]`
- **Setup tests:** `canonicalize()` to handle Windows `\\?\` long-path prefix
- **Server OS tests:** Cross-platform command helpers (`cmd /K echo`, `findstr`, `timeout`); `spawn_and_read_output` exercises the full ConPTY pipeline
- **Mock trait impls:** Four `FakeInputOutput` impls gain `new_client_with_reply()` stubs

## CI

New jobs in `.github/workflows/rust.yml`:
- **`build-windows`:** `cargo check --no-default-features` on `windows-latest`
- **`test-windows`:** Per-crate `cargo test --no-default-features -p {zellij-utils,zellij-server,zellij-client} --lib`
- **`workflow_dispatch`** trigger for manual runs

### DIFF
--- a/zellij-utils/src/cli.rs
+++ b/zellij-utils/src/cli.rs
@@ -143,7 +143,9 @@ pub struct WebCli {
         display_order = 5
     )]
     pub daemonize: bool,
-    /// Timeout in seconds waiting for the server to start (default: 10)
+    /// Timeout in seconds waiting for the server to start (default: 10).
+    /// Only used on Windows where the daemonized server is polled via TCP.
+    /// On Unix, startup signaling uses pipes and this option is ignored.
     #[clap(long, value_parser, display_order = 6)]
     pub server_startup_timeout: Option<u64>,
     /// Create a login token for the web interface, will only be displayed once and cannot later be


imsnif:
Exciting! Before I go over this, two things:
1. What are the outstanding usability/performance issues right now? I saw there was some back-and-forth in the issue with those testing out this implementation.
2. It would be really great if we could find some way to run the e2e tests on windows. Might not be trivial, but maybe see if there's some sort of quick win here?

src/main.rs
```diff
@@ -15,6 +15,25 @@ use zellij_utils::{
 };
 
 fn main() {
+    // Windows default stack is 1MB vs 8MB on Linux. Zellij's layout/KDL parsing
+    // uses deep recursion that needs more stack space. Spawn the real main on a
+    // thread with an explicit 8MB stack to match Linux behavior.
+    #[cfg(windows)]
+    {
+        const STACK_SIZE: usize = 8 * 1024 * 1024; // 8 MB
```
imsnif: I'm not a fan of this, tbh. Are we 100% certain this is what's happening? What's the path that causes this stack overflow? What's on the stack when this happens?

I could be wrong of course, but this smells a bit like a hallucination to me.

divens:
> 1. What are the outstanding usability/performance issues right now? I saw there was some back-and-forth in the issue with those testing out this implementation.

On my side, there are no outstanding usability or performance issues at the moment. Let's keep in mind this is still a fresh implementation though, so maybe we could label windows support as beta in the upcoming release if we want to be conservative.

I need to spend a bit of time with default plugins and complex layouts. Last time I tried the file picker was not working, most likely due to unix/windows path rules differences. It was a few days back though.

> 2. It would be really great if we could find some way to run the e2e tests on windows. Might not be trivial, but maybe see if there's some sort of quick win here?

I will look into that. We will also need to think about the build/packaging/distribution pipeline for release. In addition to the website and github/release, it would be nice to make it available through winget.

imsnif:
> On my side, there are no outstanding usability or performance issues at the moment. Let's keep in mind this is still a fresh implementation though, so maybe we could label windows support as beta in the upcoming release if we want to be conservative.

We're pre-1.0 anyway, I think there's no need to be more specific than that but I still want to understand the state of things. I remember you mentioned there were performance issues in non-gpu accelerated terminals, is that still the case?

> I will look into that. We will also need to think about the build/packaging/distribution pipeline for release. In addition to the website and github/release, it would be nice to make it available through winget.

Right, forgot to ask about that! Is there an option to create a statically linked executable like we do with linux/mac so that people can just download and run (i.e. the "try without installing" script we have on the website)?

divens:
> We're pre-1.0 anyway, I think there's no need to be more specific than that but I still want to understand the state of things. I remember you mentioned there were performance issues in non-gpu accelerated terminals, is that still the case?

Performance is much better than it was, dragging panes with the mouse now feel responsive both in Windows Terminal and Alacritty. FWIW, I intend to spend time in the coming weeks to do some profiling and have a better understanding of our  performance profile both for the Windows binary and within WSL as a proxy for Linux.

> Right, forgot to ask about that! Is there an option to create a statically linked executable like we do with linux/mac so that people can just download and run (i.e. the "try without installing" script we have on the website)?

I don't see any reason it wouldn't be possible. I'll work on that.

zellij-client/src/web_client/mod.rs
```diff
@@ -356,12 +358,70 @@ fn daemonize_web_server(
...
+    {
+        use std::os::windows::process::CommandExt;
+        const CREATE_NEW_PROCESS_GROUP: u32 = 0x00000200;
+        const CREATE_NO_WINDOW: u32 = 0x08000000;
+        cmd.creation_flags(CREATE_NO_WINDOW | CREATE_NEW_PROCESS_GROUP);
+    }
+
+    match cmd.spawn() {
+        Ok(_child) => {
+            let deadline = Instant::now() + Duration::from_secs(10);
```
imsnif: We have an explicit `--timeout` argument now, let's use it

zellij-client/src/lib.rs
```diff
@@ -262,17 +262,37 @@ pub fn spawn_server(socket_path: &Path, debug: bool) -> io::Result<()> {
...
 
-    if status.success() {
+    // On Unix the server daemonizes (double-fork) inside start_server(), so
+    // the intermediate child exits immediately and cmd.status() returns.
+    // On Windows there is no daemonize — we launch the server as a background
+    // process and return immediately.  We use CREATE_NO_WINDOW (not
+    // DETACHED_PROCESS) so the server gets a hidden console with valid
+    // standard handles; DETACHED_PROCESS leaves stdin/stdout/stderr as NULL,
+    // which breaks PTY creation, WASM plugin loading, and logging.
+    #[cfg(windows)]
```
imsnif: Regarding this and similar cfg(windows) blocks: could we try, as much as possible/feasible, to break them out to separate functions? I find it a bit confusing to follow when separate pieces of conditional OS logic live inside the same function.

zellij-client/src/os_input_output.rs
```diff
@@ -293,21 +323,69 @@ impl ClientOsApi for ClientOsInputOutput {
...
+        {
+            let mut stdout = self.get_stdout_writer();
+            stdout
+                .write_all(ENABLE_MOUSE_SUPPORT.as_bytes())
+                .context(err_context)?;
+            stdout.flush().context(err_context)?;
+        }
+        #[cfg(windows)]
+        {
+            // When TERM is set we're on the VT input path (terminal emulator like
```
imsnif: Could we break this into different functions, with a conditional of `is_vt_terminal` or some such? That way we can break this comment out into the different functions and I feel it will make things a bit clearer.

zellij-client/src/os_input_output_windows.rs
```diff
@@ -4,35 +4,118 @@ use async_trait::async_trait;
...
     }
 }
 
 #[async_trait]
 impl crate::os_input_output::AsyncSignals for AsyncSignalListener {
     async fn recv(&mut self) -> Option<SignalEvent> {
-        unimplemented!("Windows AsyncSignalListener not yet implemented")
+        loop {
+            tokio::select! {
+                _ = self.interval.tick() => {
```
imsnif, line 41: This is frankly not great :/ In general, if we find ourselves using static sleep timers we should take a step back and reexamine the implementation.

This is especially true in this case - windows can be resized with the mouse, producing a whole lot of consecutive resize events and this approach could make them quite sluggish.

As far as I understand it, on windows we receive these signals on STDIN and can't easily listen to them separately on a different thread like we do in unix-world. Could we bypass this by forwarding these events from the stdin thread to the signals thread (where we'll have a select to handle them)?

zellij-client/src/stdin_handler.rs
```diff
@@ -41,23 +91,114 @@ pub(crate) fn stdin_loop(
...
-                send_done_parsing_after_query_timeout(
-                    send_input_instructions.clone(),
-                    query_duration,
-                );
+                // On Windows, the ANSI terminal query mechanism (writing escape
+                // sequences to stdout and reading responses from stdin) does not
+                // work reliably — Windows Terminal may not deliver responses
+                // through the console input buffer in a way that fill_buf() can
+                // read, causing the startup to hang.  Skip it for now; pixel
+                // dimensions and color registers are nice-to-have, not critical.
```
imsnif: Respectfully disagree :)

I don't know a lot about windows, but this seems odd to me. Can an app not query the terminal for its colors/pixel-dimensions, etc? What's not reliable here?

zellij-client/src/stdin_handler.rs
```diff
@@ -41,23 +91,114 @@ pub(crate) fn stdin_loop(
...
+
+        loop {
+            match event::read() {
+                Ok(Event::Key(key_event)) => {
+                    if key_event.kind != KeyEventKind::Press {
+                        continue;
+                    }
+                    if let Some((key, bytes)) = cast_crossterm_key(key_event) {
+                        if send_input_instructions
+                            .send(InputInstruction::KeyWithModifierEvent(key, bytes, false))
```
imsnif: I guess the windows console doesn't support the kitty keyboard protocol at all? I see we seem to get quite verbose events though (press/release, etc.) - do we also get stuff like super keys, multiple modifiers, etc.? If so we might be able to translate them into kitty world and then have a win. wdyt?

zellij-server/src/plugins/mod.rs
```diff
@@ -518,6 +518,10 @@ pub(crate) fn plugin_thread_main(
                 (client_id, is_web_client),
                 completion_tx,
             ) => {
+                log::info!(
+                    "PluginInstruction::NewTab: loading plugins for tab {}",
+                    tab_id
+                );
```
imsnif: Intentional?

zellij-server/src/plugins/mod.rs
```diff
@@ -607,6 +611,7 @@ pub(crate) fn plugin_thread_main(
                         }
                     }
                 }
+                log::info!("PluginInstruction::NewTab: plugins loaded, forwarding to PTY thread");
```
imsnif: Intentional?

zellij-server/src/lib.rs
```diff
@@ -743,6 +756,7 @@ pub fn start_server(mut os_input: Box<dyn ServerOsApi>, socket_path: PathBuf) {
         err_ctx.add_call(ContextType::IPCServer((&instruction).into()));
         match instruction {
             ServerInstruction::FirstClientConnected(cli_assets, is_web_client, client_id) => {
+                info!("FirstClientConnected: loading config and layout");
```
imsnif: Intentional? (will stop adding this to every log, but let's please not add extra logs unless absolutely vital for future debugging)

zellij-server/src/lib.rs
```diff
@@ -1766,8 +1782,14 @@ fn init_session(
     let default_mode = config_options.default_mode.unwrap_or_default();
     let default_keybinds = config.keybinds.clone();
 
+    // Windows default thread stack is 1MB vs 8MB on Linux. Server worker
+    // threads do heavy work (WASM compilation, layout parsing, PTY creation)
+    // that can overflow 1MB.  Use 8MB everywhere for consistency.
+    const THREAD_STACK_SIZE: usize = 8 * 1024 * 1024;
```
imsnif: Same issue as in main. I'm a bit skeptical.

divens:
> Fantastic! On the whole looks great and was a very interesting read.
> 
> I left some comments - I did not go over everything thoroughly because I wanted to unstuck you (I'll be AFK most of today), so after your changes I'll go over the whole thing again (it'll also be easier for me considering some of the style comments) and might have more change requests.

Thanks for this first round of feedback. Will address each item in its own thread.

Following up on the CI discussion above, I've opened a PR in my own fork for E2E testing the Windows binary. Let's say it's not there yet :) I think we could also work on some broader overhaul of the CI pipelines in order to shorten cycle times. Let's discuss this separately if you're interested.

src/main.rs
divens: I did encounter a stack overflow crash in the beginning. I think it was the first or second run of my native Windows build. 

I'll revisit this now that it is more stable and proven to be working.

zellij-client/src/web_client/mod.rs
divens: Great, will rework this.

zellij-client/src/lib.rs
divens: Sure, will try to adopt the same strategy as in os_input_output.rs : have separate per platform functions where implementations differ significantly and gate at import-level.

zellij-client/src/os_input_output.rs
divens: Yes I agree, we can break down both TERM / native console to make each path clearer

zellij-client/src/os_input_output_windows.rs
divens: Will have to dig deeper. I'll see what we can do.

zellij-client/src/stdin_handler.rs
divens: Sorry I let that slip by. Seems odd to me as well.

zellij-server/src/lib.rs
divens: As for main, will have a look into this to assess how much stack is used.

zellij-server/src/lib.rs
divens: semi-intentional :) I agree this should not be sporadic like this though. Will clean them up and maybe we can discuss if / how we can improve logging by default (without having to fork/recompile) throughout the program.

zellij-server/src/plugins/mod.rs
divens: see below

zellij-server/src/plugins/mod.rs
divens: see below

imsnif:
> Following up on the CI discussion above, I've opened a PR in my own fork for E2E testing the Windows binary. Let's say it's not there yet :) I think we could also work on some broader overhaul of the CI pipelines in order to shorten cycle times. Let's discuss this separately if you're interested.

Gotcha. Let's not let this block us. Our priorities right now are, in order:
1. Being able to compile and run on windows with as few issues as possible
2. Automating a windows + windows-no-web "no-install" release binary
3. Running e2e tests on CI

I'm hoping we get 1 and 2 for the upcoming release (would really like to see them merged in the next few days if feasible, mostly so we can get people to properly rage-test them before the release), if we get 3 too - amazing but strictly nice to have. If 2 is hella problematic, we can live without it.

zellij-client/src/stdin_handler.rs
a-usr: Support for the kitty keyboard protocol has always been up to the terminal to support afaik, even on linux there are terminals where it's not supported. Microsoft folks *did* [add support to windows terminal just last week](https://github.com/microsoft/terminal/commit/c81c66b406c2f9f9320ff9ccbad71b47dafb55ae), but I'm not sure if that also adds support to conhost (or, as you call it, cmd.exe, which is about as accurate as calling bash a terminal), which is hosted in the same repository. Also worth noting is that at this time that feature isn't in any release or pre-release builds yet..

As for the events, I just tested this in neovim in WT and it did pick up me typing <kbd>Ctrl</kbd> + <kbd>Alt</kbd> + <kbd>k</kbd> in `insert-literal` mode (`:h i_CTRL-V`), so multiple modifiers definetly work. Not sure about super keys, but if you're referring to the windows key with that, the answer's obviously gonna be no. That key is reserved exclusively for windows itself and isn't passed to individual applications.

src/main.rs
divens: Ok I did some experimentation. It crashed on startup at 256KB stack size, but started normally with 512KB and above (tested 1MB and 4MB). So I guess I'll remove this part and let the default stack size be and see how it goes. I don't know how that would behave with heavy sessions (like 50+ panes you mentionned in another PR).

divens:
> Gotcha. Let's not let this block us. Our priorities right now are, in order:
> 
> 1. Being able to compile and run on windows with as few issues as possible
> 2. Automating a windows + windows-no-web "no-install" release binary
> 3. Running e2e tests on CI
> 
> I'm hoping we get 1 and 2 for the upcoming release (would really like to see them merged in the next few days if feasible, mostly so we can get people to properly rage-test them before the release), if we get 3 too - amazing but strictly nice to have. If 2 is hella problematic, we can live without it.

Sounds good to me. I'll provide an updated PR tomorrow (was a bit swamped yesterday and today). Would be nice if we can land it by Monday.

laciferin2024:
🎸 YaY , cna't wait to try it for the first time on my gaming pc.

src/main.rs
imsnif: I think 50+ or even 100+ panes and such shouldn't make a dent in the stack. Should only be deep recursions, complex calculations, data structures, etc. You could try setting the scrollback limit to 1M lines and then `cat`ing /dev/urandom (or the windows equivalent - just throwing lots of stuff into the scrollback). Lots of the viewport/scrollback structures are stack allocated for performance.

zellij-client/src/web_client/mod.rs
divens: After looking into this, I don't think we can benefit that much from the existing `web_server_status` which has the timeout argument. The http client will fail faster than the timeout on connection refused error, if the port isn't bound by any process. Actual testing shows a `ConnectionFailed` is returned in ~2s, rather than the full timeout that was passed in. What we could do is to replace the TCP polling with `web_server_status` polling, but it's a marginal improvement.

Let me know what your take is on this.

imsnif: If these have internal/implicit timeouts then I guess we can get rid of this altogether and use the `--timeout` flag as a fallback?

zellij-server/src/lib.rs
divens: Removed the custom stack size and relying on default.

zellij-server/src/lib.rs
divens: Removed

zellij-server/src/plugins/mod.rs
divens: Removed

zellij-server/src/plugins/mod.rs
divens: removed

src/main.rs
divens: Removed custom stack size, now relying on default.

zellij-client/src/os_input_output.rs
divens: Refactored to make it more readable

zellij-client/src/os_input_output_windows.rs
divens: Refactored, replacing polling with event forwarding via mpsc channel

zellij-client/src/stdin_handler.rs
divens: Now properly handled in the TERM path

zellij-client/src/lib.rs
divens: refactored to make clean platform-specific functions instead of gating inside functions.

divens:
@imsnif I've just updated the PR. It took longer than expected because I had to track down an obscure issue with ctrl+c handling. Spent my entire afternoon and evening on that yesterday actually. 

Anyway, I think I addressed most of your feedback. The web server polling to ensure proper startup is still open, I'm not sure how to improve it without getting into the --timeout implementation (which only works for the status command). Requesting your guidance on that. Also regarding the Kitty keyboard protocol, I don't have much experience, not sure how to test.

Thanks for your patience, I think we're almost there.

divens:
I've just updated the PR with CI update, including Release workflow with a regular native x64 Windows build and a no_web, static-CRT x64 Windows build. Both tested and working on my laptop.

xtask/src/flags.rs
```diff
@@ -241,3 +256,4 @@ impl Xtask {
         Self::from_vec_(args)
     }
 }
+// generated end
```
imsnif, line 259: ?

zellij-client/src/lib.rs
```diff
@@ -891,6 +898,11 @@ pub fn start_client(
     let on_force_close = config_options.on_force_close.unwrap_or_default();
     let stdin_ansi_parser = Arc::new(Mutex::new(StdinAnsiParser::new()));
 
+    // On Windows, create a channel so the stdin thread can forward resize events
+    // from crossterm to the signal handler thread (instead of polling).
+    #[cfg(windows)]
+    let (resize_sender, resize_receiver) = std::sync::mpsc::channel::<()>();
```
imsnif, line 901: Personally, I find it hard to read code with variables that conditionally exist depending on the platform. Could we please either defer their creation to a different function or create them regardless and not use them when we don't need to? (the performance-hit/resource-utilization or creating a one-shot channel is minuscule).

This is true to some other places in this PR, I hope it's alright if I don't manually list them all?

zellij-client/src/os_input_output_windows.rs
```diff
@@ -95,14 +106,42 @@ impl BlockingSignalIterator {
...
+                    Err(RecvTimeoutError::Disconnected) => {
+                        // Sender dropped — switch to poll mode
+                        break;
+                    },
+                }
+            }
+            self.resize_receiver = None;
+        }
+
+        // Poll fallback: same as the original implementation.
```
imsnif: I guess "original implementation" is one of those extremely annoying things the LLM does where it communicates with you in obscure comments inside the code? :)

In any case though, maybe we want this inside an `else` block?

zellij-server/src/route.rs
```diff
@@ -2098,23 +2098,25 @@ pub(crate) fn route_thread_main(
...
                                     .set_client_size(client_id, new_size);
-                                session_state
+                                // min_client_terminal_size() returns None when
+                                // the client hasn't been fully initialized yet
+                                // (set_client_data not called). This can happen
+                                // if a resize event arrives before the server
+                                // finishes processing the initial connection.
+                                // Skip the resize in that case — the server will
+                                // query the terminal size once setup completes.
+                                if let Some(min_size) = session_state
```
imsnif, line 2109: I'm a bit confused by this: don't we `set_client_size` just one line above? How can it be not set here?

xtask/src/flags.rs
divens: l121-259 is generated code by the xflags macro above. l6 `src "./src/flags.rs"` instructs to write the generated back in the file. As stated l123, running `env UPDATE_XFLAGS=1 cargo build` updates the generated code.

zellij-client/src/os_input_output_windows.rs
divens: Indeed "original implementation" refers to the earlier version, prior to the refactor. The comment has been updated as part of the light refactor addressing your comment regarding the `else` block. Intent is now clearer (and comment is appropriate to first time readers of this piece of code :) )

zellij-server/src/route.rs
divens: Comment has been clarified. Let me know if it addresses your question.

zellij-client/src/lib.rs
divens: So for this specific variable, I did rework it to make it uncondional.

There are 2 other places with gated variables. One is really Windows-only (`use_vt_reader`, detecting VT reader mode via TERM env variable). This is turned into `can_query_terminal` variable which is always true on non-windows, and depends on this use_vt_reader on windows. I feel this is not too distracting and actually warranted by platform specificity.

Other place is in zellij-server, regarding the `reply_listener` which is Windows-specific. This is used to create a platform-specific receiver, either with `new_client_with_reply` in Windows or regular `new_client` in non-windows. This is a few lines of code in initialization.

divens:
> I left some mostly stylistic comments on the code. Otherwise (parroting here, so take this with a grain of salt) I think we set the `crt-static` field on the windows `no-web` build but don't set it on the windows default build - so it won't be portable in the same way?

Yes, I was under the impression that it was what was needed :  a selfcontained binary for the website download and "test without installing", and a regular binary that can have some runtime dependencies. That being said, we could build both with static CRT and save ourselves the hassle of redistributing the vcrt runtime altogether. That's a few extra 100s of KB of executable, but I don't think windows people will be frowning upon that.
 
> Regarding the KKB - let's forget about it for now and optionally add it as an enhancement later. I think even if we have all the data to create the relevant events and serialize them as applications expect, there might be some windows particularities we'll at least want to consider allowing users to adjust/disable in the future.

Sure.

> Regarding the timeout - you're right, let's keep a static default timer and allow it to be overridden with a `--server-startup-timeout` flag, ok?

So I added this `--server-startup-timeout`, but it will only have an effect on Windows (there is no polling for web server startup on unix). I chose to silently ignore it on unix, rather than output a warning because it may be disruptive to scripts that don't expect warnings or errors (although that might be warranted because of explicit use of a no-op flag on this platform). Let me know if you want this changed.

imsnif:
> Yes, I was under the impression that it was what was needed : a selfcontained binary for the website download and "test without installing", and a regular binary that can have some runtime dependencies. That being said, we could build both with static CRT and save ourselves the hassle of redistributing the vcrt runtime altogether. That's a few extra 100s of KB of executable, but I don't think windows people will be frowning upon that.

Theoretically yes, but I think it would be a biut weird to have this distinction between the `web` and `no-web` binaries. If it's just a few 100s of KB, let's include it in all binaries to make it portable (no other drawbacks, right? eg. system/platform compatibility etc.?)
 
> So I added this `--server-startup-timeout`, but it will only have an effect on Windows (there is no polling for web server startup on unix). I chose to silently ignore it on unix, rather than output a warning because it may be disruptive to scripts that don't expect warnings or errors (although that might be warranted because of explicit use of a no-op flag on this platform). Let me know if you want this changed.

Sounds good, but let's mention in the CLI help that this is windows only. No need to go into too many details, just to prevent user frustration.

Otherwise - I don't have any more comments or requests. How do you feel about this? Do you think the windows support and build are robust? Are there certain parts you're unsure of?

divens:
> Theoretically yes, but I think it would be a biut weird to have this distinction between the `web` and `no-web` binaries. If it's just a few 100s of KB, let's include it in all binaries to make it portable (no other drawbacks, right? eg. system/platform compatibility etc.?)

Yes, let's simplify it an make all windows builds static-crt. We can also drop the no_web variant on windows and have a unique full build. We can always adapt later if there is demand for slimmer builds.
 
> Sounds good, but let's mention in the CLI help that this is windows only. No need to go into too many details, just to prevent user frustration.

Sure, will add some doc.
 
> Otherwise - I don't have any more comments or requests. How do you feel about this? Do you think the windows support and build are robust? Are there certain parts you're unsure of?

So I went for an extensive round of testing. I have a fix incoming for an issue with `zellij run` commands that leave panes stuck after the first command run. Also looking into a rendering issue for emojis : no emoji rendered at the moment inside Windows Terminal, black and white emoji rendered in Alacritty. Other than that, I think we're good. Will update the PR later tonight.

imsnif:
> Yes, let's simplify it an make all windows builds static-crt. We can also drop the no_web variant on windows and have a unique full build. We can always adapt later if there is demand for slimmer builds.

Just a quick note about this: the no-web build is for security reasons rather than size. Some people see Zellij being able to start a web server as an unwanted feature - I want to give them an option to get an executable without it.

divens:
> Just a quick note about this: the no-web build is for security reasons rather than size. Some people see Zellij being able to start a web server as an unwanted feature - I want to give them an option to get an executable without it.

Ok, no problem, will leave both options then.

divens:
Ok I updated the PR with 
- the fix for `zellij run` commands
- help documentation for --server-startup-timeout
- static-crt for regular build (same as no_web)

The emoji issue took me down a rabbithole. It turns out the emoji does display in Windows Terminal + zellij + pwsh but not cmd. It appears the issue is actually in crossterm, which I fixed locally by monkey patching. I'll need to double check tomorrow and maybe open an issue upstream. So, no fix for now, we may add a known issue in release notes.

I think we're good to go :)

imsnif:
Nice - thanks once again for the excellent work!

I'm going to merge this now.

Tiny nitpick: the CLI help is generally intended for users rather than developers (some happen to be both, but we cannot assume this), so in this case I'd have worded the help more in the lines of "Windows only" rather than going into explanations about TCP and such. Definitely no biggie and I'm merging it as is. Just mentioning it because I  suspect (and hope) you'll be sticking around so would like you to understand the mindset.

This has been quite a journey, I am grateful for your fantastic work, perseverance and dedication. I'm excited about releasing this at long last (I hope in the next week or so, but I find giving ETAs in OSS is bad voodoo).

As for next steps: does Windows have the concept of package managers? Something similar? How do people normally install this sort of software on windows? Should we be entering into repositories, or...?

a-usr:
Windows has its' preinstalled packet manager `winget`, and a couple of unofficial ones (`chocolatey` and `scoop`). For winget it would be necessary to have ci generate an application installer artifact (preferrably one that doesn't require the user to have administrative privileges, but thats just my preference), and then someone would have to write a package manifest to add to the [community repository](https://github.com/microsoft/winget-pkgs). For the other two the process would of course be similar.

divens:
> Nice - thanks once again for the excellent work!
> 
> I'm going to merge this now.

Thank you! 🎉🎉
 
> Tiny nitpick: the CLI help is generally intended for users rather than developers (some happen to be both, but we cannot assume this), so in this case I'd have worded the help more in the lines of "Windows only" rather than going into explanations about TCP and such. Definitely no biggie and I'm merging it as is. Just mentioning it because I suspect (and hope) you'll be sticking around so would like you to understand the mindset.

Got you. We can adjust this to make help worded consistent and more accessible to non-developers.

> This has been quite a journey, I am grateful for your fantastic work, perseverance and dedication. I'm excited about releasing this at long last (I hope in the next week or so, but I find giving ETAs in OSS is bad voodoo).

Quite a journey indeed! Thank you for supporting this effort and being open to refactoring some critical pieces of code in order to make room for a new platform. Your feedback has been invaluable and the result is much better than my initial fork.

> As for next steps: does Windows have the concept of package managers? Something similar? How do people normally install this sort of software on windows? Should we be entering into repositories, or...?

As @a-usr said, the default package manager is `winget`, and I think that should be our first goal for easy distribution. We can extend to `chocolatey` and `scoop` as a second step.
That being said, for better or worse, people on Windows are also used to downloading installers directly from official websites, so this distribution method should not be underestimated.

Muqito:
Thank you to everyone involved :) You are amazing

KudoLayton:
Congratulations. I also attempted it in the past but didn’t achieve it perfectly—yet you ultimately made it happen. Congratulations again.

sitiom:
> > As for next steps: does Windows have the concept of package managers? Something similar? How do people normally install this sort of software on windows? Should we be entering into repositories, or...?
> 
> As @a-usr said, the default package manager is `winget`, and I think that should be our first goal for easy distribution. We can extend to `chocolatey` and `scoop` as a second step. That being said, for better or worse, people on Windows are also used to downloading installers directly from official websites, so this distribution method should not be underestimated.

See https://github.com/zellij-org/zellij/issues/316#issuecomment-4111167553. I have created a PR for both Winget and Scoop.

PR #4770, opened 2026-02-26
fix: forward unbound keys in Normal mode regardless of default input mode

metaspace:
When users configure Zellij with clear-defaults=true and set Locked as their default_input_mode, unbound keys in Normal mode were being silently consumed (NoOp) instead of forwarded to the pane.

This happened because default_action_for_mode() only forwarded unbound keys when in Locked mode OR when the current mode equaled the user's default_input_mode. When default_input_mode was Locked, Normal mode fell through to NoOp.

Now Normal mode explicitly forwards unbound keys to the pane, matching the documented behavior that "input is always written to the terminal, except for the shortcuts leading to other modes."

Fixes: https://github.com/zellij-org/zellij/issues/775#issuecomment-2472786982

### DIFF
--- a/zellij-utils/src/input/keybinds.rs
+++ b/zellij-utils/src/input/keybinds.rs
@@ -76,11 +76,20 @@ impl Keybinds {
         key_is_kitty_protocol: bool,
     ) -> Action {
         match *mode {
+            // In Locked mode, unbound keys are always forwarded to the pane
             InputMode::Locked => Action::Write {
                 key_with_modifier: key_with_modifier.cloned(),
                 bytes: raw_bytes,
                 is_kitty_keyboard_protocol: key_is_kitty_protocol,
             },
+            // In Normal mode, unbound keys are always forwarded to the pane
+            // (regardless of whether Normal is the default_input_mode)
+            InputMode::Normal => Action::Write {
+                key_with_modifier: key_with_modifier.cloned(),
+                bytes: raw_bytes,
+                is_kitty_keyboard_protocol: key_is_kitty_protocol,
+            },
+            // In the user's configured default input mode, forward unbound keys
             mode if mode == default_input_mode => Action::Write {
                 key_with_modifier: key_with_modifier.cloned(),
                 bytes: raw_bytes,
--- a/zellij-utils/src/input/unit/keybinds_test.rs
+++ b/zellij-utils/src/input/unit/keybinds_test.rs
@@ -594,3 +594,42 @@ fn error_received_on_unknown_key_instruction() {
     let config_error = Config::from_kdl(config_contents, None).unwrap_err();
     assert_snapshot!(format!("{:?}", config_error));
 }
+
+#[test]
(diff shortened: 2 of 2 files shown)


imsnif:
This behavior is intentional to support the unlock-first keybinding preset: https://zellij.dev/documentation/keybinding-presets.html?highlight=preset#keybinding-presets

The expectation is that the default mode is the one that will be used to forward keys to the terminal (unless it's locked and then it's always forwarded).

metaspace:
Maybe we can make a configuration option to enable forwarding? For those of us that prefer a leader key approach similar to screen and tmux, the failure to forward these keys prevent nested zellij sessions.

Please see #775 for more discussion of the topic.

metaspace:
Actually, now does this break the unlock first preset? With this change, if keys are bound, they are still not forwarded.

imsnif:
The unlock-first preset is a leader key approach... or am I missing something?

metaspace:
> The unlock-first preset is a leader key approach... or am I missing something?

No, you are correct, and I think I get your point now. But I still do not understand why it is necessary to consume all keys in the unlocked state for this preset to work. Only the mapped keys should be consumed.

With this unlock-first preset, it is not possible to have nested sessions of zellij. Assume you have two zellij instances, one inside the other. To understand why this is desirable, consider a setup where you run zellij on a workstation and ssh into a remote machine where you also use zellij. Now, assume both instances use the "unlock-first" preset and that both sessions are in the locked state. You want to send a command to the inner session (on the remote machine). When you enter the leader key (unlock key) in the outer session, it unlocks. Now you have no way of sending the unlock key to the inner session. The only way this can work is if the unlock key on your local zellij and your remote zellij are different, which is undesirable.

But actually while writing this message it occurred to me that maybe the problem is solvable without any code chagne, and it turns out that it is:

```kdl
keybinds clear-defaults=true {
    normal {
        ...
        bind "Ctrl o" { Write 15; SwitchToMode "Locked"; }
        ...
    }
    locked {
        unbind "Ctrl g"
        bind "Ctrl o" { SwitchToMode "Normal"; }
    }
    ...
}
```

This sends `ctrl-o` to the inner session and switches the outer session to locked if `ctrl-o` is tapped twice.

Closing this PR since the change is not required.

imsnif:
> No, you are correct, and I think I get your point now. But I still do not understand why it is necessary to consume all keys in the unlocked state for this preset to work. Only the mapped keys should be consumed.

In a modal system, I think it's bad UX if you're inside a command mode (eg. anything triggered by a leader) and then have some keys trigger commands and other keys pass through. Namely for cases where you don't notice that you're in a mode and then start typing something into the terminal. Middle characters can trigger modes and cause confusion on top of damage. We used to behave like this in the early days and got a whole lot of complaints (rightfully so IMO).

As for the nested sessions - I agree this is not ideal. The plan is to automatically identify this situation and then optionally allow the user to control all sessions from the same UI. For now your workaround is sound (you could also dynamically change leader keys - also temporarily - through the `configuration` plugin).

PR #4774, opened 2026-03-01
fix: preserve argument boundaries in session resurrection

### LINKED ISSUE #4727: Session resurrection strips quotes from commands (e.g., nvim +'...')
## Issue description
When picking up a closed session (after a restart of the host machine) that was closed with a nvim +'ClaudeCode --continue', I am greeted with a message to resume the session using the same command. When I tap enter, I get the message "vim doesn't recognize --continue".

## Minimal reproduction

- Open a Zellij session.
- Start NeoVim with the command: nvim +'ClaudeCode --continue'
- Kill the session by restarting the host machine (or abruptly killing the Zellij server).
- Restart Zellij and pick up the closed session.
- Zellij will prompt you to hit Enter to run the previous command.
- Press Enter.
- The error "vim doesn't recognize --continue" is displayed instead of opening NeoVim with the intended command.

## Other relevant information

It seems like Zellij is losing or stripping the quotes during the session resurrection process. Instead of passing ClaudeCode --continue as a single string to the + argument, it appears to be evaluating it as nvim +ClaudeCode --continue, which causes nvim to interpret --continue as an invalid command-line flag.

Zellij version: 0.43.1

OS: Ubuntu on WSL2/windows 11

NeoVim version: 0.11.6

Shell: zsh

Nicknames96:
When resurrecting a session, commands with arguments containing spaces (e.g. `nvim +"ClaudeCode --continue"`) had their argument boundaries destroyed. The arg `+ClaudeCode --continue` (one arg) became `+ClaudeCode` and `--continue` (two args).

The root cause was `get_all_cmds_by_ppid` parsing `ps -ao ppid,args` output by splitting on whitespace, which cannot distinguish between inter-argument spaces and spaces within an argument.

Switched to using sysinfo `process.cmd()` which reads `/proc/<pid>/cmdline` where arguments are null-byte separated, preserving boundaries correctly.

Closes #4727

### DIFF
--- a/zellij-server/src/os_input_output.rs
+++ b/zellij-server/src/os_input_output.rs
@@ -494,43 +494,48 @@ impl ServerOsApi for ServerOsInputOutput {
     #[cfg(unix)]
     fn get_all_cmds_by_ppid(&self, post_hook: &Option<String>) -> HashMap<String, Vec<String>> {
         // the key is the stringified ppid
+        // We use sysinfo to read /proc/<pid>/cmdline which preserves argument
+        // boundaries via null-byte separation. The previous approach parsed
+        // `ps -ao ppid,args` by splitting on whitespace, which destroyed
+        // boundaries for arguments containing spaces (e.g. nvim +'Foo --bar'
+        // became ["nvim", "+Foo", "--bar"] instead of ["nvim", "+Foo --bar"]).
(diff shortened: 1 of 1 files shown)


Nicknames96:
Closing this — no longer pursuing this change. Thanks for your time.

PR #4779, opened 2026-03-02
feat: enhance shell completions for zellij(bash, zsh, fish)

ldm0:
1. Implement session completions for bash and zsh, which I frequently use. e.g. `zellij attach b<tab>` -> `zellij attach brave-lemon`
2. Optimize fish shell completion scripts.
3. Add a test script for these three shells(for behavior checking and alignment).


ldm0:
Ping @imsnif, plz take a look. This is helpful for poor bash/zsh users.

Issue #4769, opened 2026-02-26
Zellij 0.43.1: Japanese voiced and semi-voiced consonants disappear when pasting


densuke:
In Release 0.43.1, I encountered an issue where characters would be replaced when dropping a path containing Japanese voiced or semi-voiced consonants from Finder. This is probably similar to issue #3148 and is thought to be an issue with Unicode normalization (NFC, NFD).

* 「バ」(ba) → 「ハ」(ha)
* 「プ」(pu) → 「フ」(fu)

I forked it locally to confirm the cause and fixed it (the original characters were pasted properly), but since I left it entirely to Claude Opus 4.6, I'm not confident in the code.

If anyone knows the answer, could you please take a look at the validity of the code?

* macOS Tahoe 26.3 on MacBookAir M4
* zellij 0.43.1(homebrew)
* forked repo: https://github.com/densuke/zellij/tree/unicode-normalization

<img width="2838" height="288" alt="Image" src="https://github.com/user-attachments/assets/dc593621-dde6-468c-8408-0bc7fda45481" />

<img width="2844" height="324" alt="Image" src="https://github.com/user-attachments/assets/f0c196a3-698c-41d3-8461-6edbfebbe264" />

Issue #4447, opened 2025-09-29
[BUG?] Thai text rendering issue


hotuansang96:
#### Description  
Thai text does not render correctly inside Zellij. Some characters bleed into the line numbers, and some rows are duplicated or overlap with other lines. The same text renders correctly in the native terminal (for example, Ghostty). This seems to be related to handling of Unicode grapheme clusters.

#### Steps to Reproduce  
1. Open a terminal that supports Thai text rendering correctly (e.g. Ghostty).  
2. Run Zellij inside that terminal.  
3. Print some Thai text.  
4. Compare the rendering inside Zellij vs. directly in the terminal.  

#### Expected Behavior  
Thai text should render exactly the same as in the native terminal.  

#### Actual Behavior  
- In Zellij, Thai text rendering is broken/misaligned (combining characters don’t display correctly).  
- In the native terminal (e.g. Ghostty), the Thai text displays properly.  

#### Screenshots / Recordings  
<img width="1920" height="209" alt="Image" src="https://github.com/user-attachments/assets/0deead38-8a06-4fd4-b620-1f644f518fe1" />
<img width="548" height="185" alt="Image" src="https://github.com/user-attachments/assets/82ad9166-9c19-4c71-96f1-01a52363e7d1" />
<img width="526" height="173" alt="Image" src="https://github.com/user-attachments/assets/9def6a83-2fc0-4af8-a8a9-72a368b6c7ff" />

#### Environment  
- **Zellij version:** 0.43.1
- **OS:** MacOS
- **Terminal emulator:** Ghostty  
- **Font:** JetBrains Mono Nerd Font

imsnif:
We are embarrassingly aware of this long-standing issue and I can only offer my sincere apologies for the experience. @tlinford is working on fixing this and I hope we'll have something before the next release.

Peelz:
Same as #3667

dllp-beonit:
Still reproducible on **zellij 0.45.1** (current latest).

I was able to confirm the loss happens **inside zellij's own grid**, not in the
terminal emulator or the font — `zellij action dump-screen` already comes back
without the marks, so nothing downstream ever receives them.

**Environment**

- zellij 0.45.1 (x86_64 Linux, Ubuntu 26.04, kernel 7.0.0)
- `LANG=en_US.UTF-8`, `TERM=xterm-256color`
- Client: Termius on Windows over SSH (irrelevant — see below)

**Reproduction**

```sh
[...]

Issue #1538, opened 2022-06-22
Not drawing strikethrough text as strikedthrough


Mr-Andersen:
**Basic information**

`zellij --version`: 0.30.0
`tput lines`: 43
`tput cols`: 174
`uname -av` or `ver`(Windows): Linux aka-system 5.15.47 #1-NixOS SMP Tue Jun 14 16:36:28 UTC 2022 x86_64 GNU/Linux

**Reproducing**

- paste "a̶b̶c̶" in Zellij

**Expected**

- letters are strikethrough as GH renders them (hopefully)

**Actual**

- It's just "abc"

I've checked this without Zellij, terminal renders them OK

**Log**

[zellij-9.log](https://github.com/zellij-org/zellij/files/8960998/zellij-9.log)

imsnif:
I understand, hum. If you put it in an echo (outside of Zellij), do you see it in the echo? Also, which shell are you using?

Mr-Andersen:
```
> fish --version
fish, version 3.4.1
```

Copy-pasting this piece into "echo" outside of zellij works, both piece in quotes and echoed one renders correctly

Inside zellij it doesn't work

```
> echo -e "a̶b̶c̶"
a̶b̶c̶
```

Mr-Andersen:
Try copying `a̶b̶c̶` and pasting it into your zellij

raphCode:
To continue the discussion here:
- strikethrough is achieved by appending a [unicode modifier codepoint](https://codepoints.net/U+0336) to any char that should be rendered strikethrough
- this and other modifying codepoints have a width of zero
- currently, we drop zero-width codepoints
- I believe we should store them together with the previous printable character in our character representation. This would fix also future problem with some diacritics or other unicode stuff not displayed properly.

raphCode:
Maybe we can draw inspiration from tmux handling unicode stuff? I tried the striketrough thing and it worked there.

raphCode:
Ah yes, C can be awful at times, but the comment at the top explains:
`Zero width characters are just appended onto the UTF-8 data for the previous cell.`
Which is basically my initial idea too and should cover most of the use cases.

I mean the craziest UTF8 shenanigans I saw in the wild was probably these z̶̝̜̀͠ả̷̗̎ḻ̶̌̿ģ̸͔̍ȯ̷̰̠ texts like they can be generated [here](https://lingojam.com/GlitchTextGenerator). These work with tmux' approach.
[...]

PR #4800, opened 2026-03-05
Grapheme segmentation and 2027 support

tlinford:
Putting this out in draft as I think it's starting to work decently.

Would love to gather some feedback on this from anyone who can test in scenarios that currently aren't handled well, as well as from whoever has expertise in this (at least for me) complex domain and can contribute useful info/test cases etc.

I hope that this will at least start to address many issues including: #4447, #4769, #3148, #1538.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -451,15 +451,6 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "a2698f953def977c68f935bb0dfa959375ad4638570e969e2f1e9f433cbf1af6"
 
 [[package]]
-name = "castaway"
-version = "0.2.4"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "dec551ab6e7578819132c713a93c022a05d60159dc86e7a7050223577484c55a"
-dependencies = [
- "rustversion",
-]
-
-[[package]]
 name = "cc"
 version = "1.2.57"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -559,6 +550,16 @@ dependencies = [
 ]
 
 [[package]]
+name = "cold-string"
+version = "0.2.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "5c378bebe5e642bf918c089ab74a767e321cf922d74e7633c3bdcc22d0cb1fc6"
+dependencies = [
+ "rustversion",
+ "sptr",
+]
+
+[[package]]
 name = "colored"
 version = "2.0.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -584,20 +585,6 @@ dependencies = [
  "unicode-width 0.2.2",
  "zellij-tile",
  "zellij-tile-utils",
-]
-
-[[package]]
-name = "compact_str"
-version = "0.8.1"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "3b79c4069c6cad78e2e0cdfcbd26275770669fb39fd308a752dc110e83b9af32"
-dependencies = [
- "castaway 0.2.4",
- "cfg-if",
- "itoa",
- "rustversion",
- "ryu",
- "static_assertions",
 ]
 
 [[package]]
@@ -1835,7 +1822,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "334e04b4d781f436dc315cb1e7515bd96826426345d498149e4bde36b67f8ee9"
 dependencies = [
  "async-channel",
- "castaway 0.1.2",
+ "castaway",
  "crossbeam-utils",
  "curl",
  "curl-sys",
@@ -2950,9 +2937,9 @@ dependencies = [
 
 [[package]]
 name = "rustversion"
-version = "1.0.20"
+version = "1.0.22"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "eded382c5f5f786b989652c49544c4877d9f015cc22e145a5ea8ea66c2921cd2"
+checksum = "b39cdef0fa800fc44525c84ccb54a029961a8215f9619753635a9c0d2538d46d"
 
 [[package]]
 name = "ryu"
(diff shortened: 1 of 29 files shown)


boomshroom:
From what I can tell, `compact_str` takes inline 24 bytes, which is a _huge_ overhead compared to the 4 bytes of a single `char`. While I was looking into this idea, I found the relatively recent [`cold-string`](https://github.com/tomtomwombat/cold-string/), which seems to use similar tricks as `compact_str`, but uses thin pointers with the length stored with the string on the heap instead of inline. This gives it only twice the size of a `char` on 64-bit platforms, or _zero overhead_ on 32-bit platforms for the most common characters. It also only supports immutable strings, which shouldn't be much of a problem in this case; the current cell could use a full `String`, which then gets moved to a `ColdString` when starting the next character.

With 80 columns, and 10 000 lines of scrollback for 800 000 cells, that would be 12.8 MB currently, 32 MB with this PR, and ~~16 MB~~ with `cold-string`.

https://github.com/zellij-org/zellij/blob/6e521c74d3c0fa3449cd0c1f1971ee807e88d98e/zellij-server/src/panes/terminal_character.rs#L932-L933

This comment is why I wanted to make sure that the compact string used must be no larger than a single pointer.

Edit: taking a closer look `RcCharacterStyles` is a pointer, but the `width: u8` can currently be stored next to the `char`, while a either a `ColdString` or a `CompactString` would push it out and require it to get padded to a full pointer, so the actual size of a `TerminalCharacter` would be 24 bytes instead of 20, making the full scrollback 19.2 MB instead of 16 MB.

tlinford:
> From what I can tell, `compact_str` takes inline 24 bytes, which is a _huge_ overhead compared to the 4 bytes of a single `char`. While I was looking into this idea, I found the relatively recent [`cold-string`](https://github.com/tomtomwombat/cold-string/), which seems to use similar tricks as `compact_str`, but uses thin pointers with the length stored with the string on the heap instead of inline. This gives it only twice the size of a `char` on 64-bit platforms, or _zero overhead_ on 32-bit platforms for the most common characters. It also only supports immutable strings, which shouldn't be much of a problem in this case; the current cell could use a full `String`, which then gets moved to a `ColdString` when starting the next character.
> 
> With 80 columns, and 10 000 lines of scrollback for 800 000 cells, that would be 12.8 MB currently, 32 MB with this PR, and ~16 MB~ with `cold-string`.
> 
> https://github.com/zellij-org/zellij/blob/6e521c74d3c0fa3449cd0c1f1971ee807e88d98e/zellij-server/src/panes/terminal_character.rs#L932-L933
> 
> This comment is why I wanted to make sure that the compact string used must be no larger than a single pointer.
> 
> Edit: taking a closer look `RcCharacterStyles` is a pointer, but the `width: u8` can currently be stored next to the `char`, while a either a `ColdString` or a `CompactString` would push it out and require it to get padded to a full pointer, so the actual size of a `TerminalCharacter` would be 24 bytes instead of 20, making the full scrollback 19.2 MB instead of 16 MB.

Interesting, thanks. I actually have started looking into performance and memory usage,  and with very large scrollbacks the memory cost of `CompactString` vs `char` can't be ignored. I'll definitely have a look a look at `coldstring`.
I have been also looking into an alternative implementation where additional chars get stored outside of the terminal char itself... it's not pretty but could be doable.

tlinford:
@boomshroom went with `cold_string` for now, in my testing it improved memory usage significantly compared to `compact_str`, so thanks again!
In my test scenario, memory usage, as reported by macOS was roughly:
- 200MB on main
- 500MB for `compact_str`
- 300MB for `cold_str`.

@imsnif is the fact that the crate seems to be rather new and not largely depended upon a problem?

tlinford:
@divens could I kindly ask you to give this branch a quick test on windows? imsnif made me aware of the stack size issues on Windows and I'd like to know if the changes happening here could cause problems. it's up to date with main as of today.

divens:
> @divens could I kindly ask you to give this branch a quick test on windows? imsnif made me aware of the stack size issues on Windows and I'd like to know if the changes happening here could cause problems. it's up to date with main as of today.

Successfully built in debug and release modes and it launched just fine. I don't know what I'm supposed to test specifically beyond that, so I just opened a few panes and launched a few commands. All good.

tlinford:
> > @divens could I kindly ask you to give this branch a quick test on windows? imsnif made me aware of the stack size issues on Windows and I'd like to know if the changes happening here could cause problems. it's up to date with main as of today.
> 
> Successfully built in debug and release modes and it launched just fine. I don't know what I'm supposed to test specifically beyond that, so I just opened a few panes and launched a few commands. All good.

Thanks, not exactly sure either, the fact that it compiles and runs is a great start :)

boomshroom:
> @boomshroom went with `cold_string` for now, in my testing it improved memory usage significantly compared to `compact_str`, so thanks again!
> In my test scenario, memory usage, as reported by macOS was roughly:
> - 200MB on main
> - 500MB for `compact_str`
> - 300MB for `cold_str`.

I'm curious what the measurements would be on a 32 bit machine, since in that case the happy path should be exactly the same size as just a `char`. It should take the same amount of memory unless ~~[astral plane characters](https://en.wiktionary.org/wiki/astral_plane) or~~ more complex clusters are used. (Forgot that cold-string can inline 4 or 8 bytes on 32 and 64 bit platforms rather than 3 or 7. This should mean that on 32 bit platforms, cold-string should have _exactly the same memory usage for all working text._)

Faizalj:
Tested this branch on macOS (aarch64, Apple Silicon) — Thai text rendering works correctly ✅

Build: `cargo build --release` on the `grapheme-segmentation-and-2027-support` branch.

**Test results:**
- `สวัสดี` (sawasdee) — all vowels (◌ั ◌ี) and tone marks displayed correctly
- `ก้อง` — tone mark ◌้ renders fine
- `เปิ่น` — stacked vowel + tone mark (◌ิ่) correct
- `แม่น้ำ` — multiple tone marks render correctly

This also addresses the same issue I'm trying to fix in #4920 (combining characters for Thai). Your approach is more comprehensive — happy to defer to this PR. Will close mine if this gets merged.

Thanks @tlinford for the thorough work! 🙏

tlinford:
> Tested this branch on macOS (aarch64, Apple Silicon) — Thai text rendering works correctly ✅
> 
> Build: `cargo build --release` on the `grapheme-segmentation-and-2027-support` branch.
> 
> **Test results:**
> 
>     * `สวัสดี` (sawasdee) — all vowels (◌ั ◌ี) and tone marks displayed correctly
> 
>     * `ก้อง` — tone mark ◌้ renders fine
> 
>     * `เปิ่น` — stacked vowel + tone mark (◌ิ่) correct
> 
>     * `แม่น้ำ` — multiple tone marks render correctly
> 
> 
> This also addresses the same issue I'm trying to fix in #4920 (combining characters for Thai). Your approach is more comprehensive — happy to defer to this PR. Will close mine if this gets merged.
> 
> Thanks @tlinford for the thorough work! 🙏

thank you for the feedback!

Faizalj:
still not merged to main?

tlinford:
> still not merged to main?

Unfortunately had to put work on this on hold for now.

Issue #4804, opened 2026-03-05
Make the available plugins easier to find (add a `--list` option)


Zykino:
For example I know the `strider` plugin exit, I don't know how to start it other than opening a new tab with the strider layout. (Or any layout with it already included.)

I think users would be really happy to have a command like `zellij plugin --list`.

It could even be later expanded to show plugin listed in the https://github.com/zellij-org/awesome-zellij repo. I'm thinking about an option like `zellij plugin --list-online` or `zellij plugin --explore`. But this awesome repo is maybe not mature enough to propose an exploitable list yet?

PR #4808, opened 2026-03-06
feat(plugins): add --list option to plugin command

tlinford:
#4804

@Zykino seemed simple to do, wondering if we should go with a plain list with urls or something a bit fancier with two columns for name and url, what do you think?

### DIFF
--- a/src/commands.rs
+++ b/src/commands.rs
@@ -967,19 +967,51 @@ fn generate_unique_session_name_or_exit() -> String {
     unique_session_name
 }
 
-pub(crate) fn list_builtin_plugins() {
+pub(crate) fn list_builtin_plugins(opts: CliArgs) {
     use zellij_utils::consts::ASSET_MAP;
-    let mut names: Vec<String> = ASSET_MAP
+
+    // --- Built-in plugins ---
+    let mut builtins: Vec<(String, String)> = ASSET_MAP
         .keys()
         .filter_map(|p| {
             p.file_stem()
                 .and_then(|s| s.to_str())
-                .map(|s| format!("zellij:{}", s))
+                .map(|s| (s.to_string(), format!("zellij:{}", s)))
         })
         .collect();
-    names.sort();
-    for name in names {
-        println!("{}", name);
+    builtins.sort_by(|a, b| a.0.cmp(&b.0));
+
+    // --- Config aliases ---
+    let aliases = match Setup::from_cli_args(&opts) {
+        Ok((config, ..)) => config
+            .plugins
+            .aliases
+            .into_iter()
+            .map(|(name, run_plugin)| (name, run_plugin.location.display()))
+            .collect::<Vec<_>>(),
+        Err(_) => vec![],
+    };
+
+    let builtin_col = builtins.iter().map(|(n, _)| n.len()).max().unwrap_or(0);
+    let builtin_url_col = builtins.iter().map(|(_, u)| u.len()).max().unwrap_or(0);
+    println!("Built-in plugins:");
+    println!("  {:<w$}  URL", "NAME", w = builtin_col);
+    println!("  {}", "-".repeat(builtin_col + 2 + builtin_url_col));
+    for (name, url) in &builtins {
+        println!("  {:<w$}  {}", name, url, w = builtin_col);
+    }
+
+    if !aliases.is_empty() {
+        let mut aliases = aliases;
+        aliases.sort_by(|a, b| a.0.cmp(&b.0));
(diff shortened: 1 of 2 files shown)


Zykino:
I'm not sure, I want to have a list of all plugins (default or installed myself) with the name to start it (or pipe to it).

Writing about it now you gave the choice, I think seeing if the plugin is loaded from a file (`which`) or online is also useful.

After reading your PR, I now see that we should use `zellij plugin -- zellij:strider`. I am also discovering `zellij plugin -- zellij:filepicker` does not work even if `filepicker` is an existing config that should call strider… To execute config we should use `zellij plugin -- filepicker`.

So in the list/table I would really like seeing `zellij:strider`, `filepicker` and also `zjforgot` (in this community plugin is configured).

tlinford:
> I'm not sure, I want to have a list of all plugins (default or installed myself) with the name to start it (or pipe to it).
> 
> Writing about it now you gave the choice, I think seeing if the plugin is loaded from a file (`which`) or online is also useful.
> 
> After reading your PR, I now see that we should use `zellij plugin -- zellij:strider`. I am also discovering `zellij plugin -- zellij:filepicker` does not work even if `filepicker` is an existing config that should call strider… To execute config we should use `zellij plugin -- filepicker`.
> 
> So in the list/table I would really like seeing `zellij:strider`, `filepicker` and also `zjforgot` (in this community plugin is configured).

I totally missed aliases, good point! Pushed an update that gives us this:

```
❯ ./target/dev-opt/zellij plugin --list
Built-in plugins:
  NAME             URL
  ---------------------------------------
  about            zellij:about
  compact-bar      zellij:compact-bar
  configuration    zellij:configuration
  layout-manager   zellij:layout-manager
  link             zellij:link
  multiple-select  zellij:multiple-select
  plugin-manager   zellij:plugin-manager
  session-manager  zellij:session-manager
  share            zellij:share
  status-bar       zellij:status-bar
  strider          zellij:strider
  tab-bar          zellij:tab-bar

Config aliases:
  ALIAS            URL
  ---------------------------------------
  about            zellij:about
  compact-bar      zellij:compact-bar
  configuration    zellij:configuration
  filepicker       zellij:strider
  plugin-manager   zellij:plugin-manager
  session-manager  zellij:session-manager
  status-bar       zellij:status-bar
  strider          zellij:strider
  tab-bar          zellij:tab-bar
  welcome-screen   zellij:session-manager
```

Zykino:
Are configured community plugin shown too?

This looks quite long (and repetitive) of an output already, and sometimes you want to use the name of the right column, sometime the left. It is not very clear :thinking:

I hoped to have all the names `zellij plugin -- $NAME` on the same side which is apparently not the case :(

tlinford:
> Are configured community plugin shown too?
> 
> This looks quite long (and repetitive) of an output already, and sometimes you want to use the name of the right column, sometime the left. It is not very clear 🤔
> 
> I hoped to have all the names `zellij plugin -- $NAME` on the same side which is apparently not the case :(

Depends what you mean with configured community plugins, right now it would show you ones configured in `load_plugins`.

I'll try tinkering with it a bit more to make the output better.

Zykino:
> Depends what you mean with configured community plugins, right now it would show you ones configured in load_plugins.

Yeah pretty much. I'm putting my dev plugins in the `plugins` section, not the `load_plugins` one but as long as plugins from theses sections are displayed I'm happy. (I don't need them at each boot… especially since I tend to not finish them…)

> I'll try tinkering with it a bit more to make the output better.

Maybe with a table like this: on the left all the aliases we can use to start the plugin (see the strider line, the others are just copy/paste from your example), on the right the URL (file, http, …).
Not shown here, maybe sort by URL alphabetically?

Maybe later adding a `--expand` option that show the aliases like a tree and show a new column "CONFIGURATION", so we can see: "filepicker is zellij::strider + config { foo=bar }"

| ALIASES                          |  URL                      |
| -------------------------------- | ------------------------- |
| about                            | zellij:about              |
| compact-bar                      | zellij:compact-bar        |
| configuration                    | zellij:configuration      |
| filepicker, zellij::strider      | zellij:strider            |
| plugin-manager                   | zellij:plugin-manager     |
| layout-manager                   | zellij:layout-manager     |
| link                             | zellij:link               |
| multiple-select                  | zellij:multiple-select    |
| plugin-manager                   | zellij:plugin-manager     |
| welcome, zellij::session-manager | zellij:session-manager    |
| status-bar                       | zellij:status-bar         |
| share                            | zellij:share              |
| tab-bar                          | zellij:tab-bar            |

PR #4814, opened 2026-03-07
Allow subscribing to pane viewport updates (renders) from the CLI

imsnif:
This adds the `zellij subscribe` CLI command, which allows users to subscribe to pane viewport updates, receiving them in real time. Users can specify multiple pane IDs in the same command, can specify the output format (raw or json), and can elect to receive all or some of the pane scrollback on command initialization (to prevent the annoying "I need to quickly subscribe to the pane before output is generated" situations).

It's also possible to subscribe to panes from different sessions by specifying the `--session <NAME>` flag *before* the `subscribe` command.

```
zellij-subscribe
Subscribe to pane render updates (viewport and scrollback)

USAGE:
    zellij [--session <OTHER SESSION NAME>] subscribe [OPTIONS] --pane-id...

OPTIONS:
    -f, --format <FORMAT>
            Output format [default: raw] [possible values: raw, json]

    -h, --help
            Print help information

    -p, --pane-id <PANE_ID>...
            Pane ID(s) to subscribe to (e.g. terminal_1, plugin_2, or bare number like 1)

    -s, --scrollback [<SCROLLBACK>...]
            Include scrollback lines in initial delivery. Bare --scrollback = all scrollback,
            --scrollback N = last N lines
```


Zykino:
`raw` means getting raw ANSI of the scrollback? Do we need a distinction "text only" vs "ANSI"?

With just this description I'm not sure what to expect from the `json` format either. But trying is just a command away ;)

imsnif:
Fair points! It is text only for now, I intend to add a `--ansi` flag (similar to what we have now with dump screen) if there's need. I just first want to see how it goes.

About the json - I was kind of expecting people to just try as you say... I don't want to maintain a separate field list in case we add stuff.

PR #4817, opened 2026-03-08
Add cli paste action

imsnif:
This adds a `zellij action paste --pane-id <pane_id> "string_to_paste"` action to the cli. Mostly useful for pasting large amount of texts in one go (and one render cycle) rather than having them send character by character to panes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -73,6 +73,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 * feat: click to open file-paths in viewport, allow plugins to highlight viewport text (https://github.com/zellij-org/zellij/pull/4802 and https://github.com/zellij-org/zellij/pull/4809)
 * fix: better handle STDIN segmentation with latency (eg. over SSH) (https://github.com/zellij-org/zellij/pull/4623)
 * feat: `zellij subscribe` -> allow to subscribe to pane viewport/scrollback updates from the cli (https://github.com/zellij-org/zellij/pull/4814)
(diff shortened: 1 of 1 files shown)


Zykino:
Is this a `write`/`write-char` equivalent? This needs to be highlighted in the doc if so. (When to use `write` instead of `paste`?)

imsnif:
Yeah, essentially: this is like write-chars, except everything is written in bulk (also with bracketed paste, so as if the user pasted it into the terminal). Probably good for larger pieces of text (eg. paragraphs, screen dumps, etc.)

Zykino:
Isn't it always better?

Oh maybe the `write` variant can be use to send keybindings to an application (`i` to `vim`) while `paste` is better to send actual text?

imsnif:
No, that's `zellij action send-keys` (also in main) :)

```
zellij-action-send-keys
Send one or more keys to the terminal (e.g., "Ctrl a", "F1", "Alt Shift b")

USAGE:
    zellij action send-keys [OPTIONS] <KEYS>...

ARGS:
    <KEYS>...    Keys to send as space-separated strings

OPTIONS:
    -h, --help                 Print help information
    -p, --pane-id <PANE_ID>    The pane_id of the pane, eg. terminal_1, plugin_2 or 3 (equivalent
                               to terminal_3)
```

One different with `write-chars` is that it's not sent as bracketed paste. So line breaks will be interpreted as Enter, for example.

Zykino:
Oh you mean that by "bracketed paste".

~~(For documentation/ease of finding similar functions, I propose `write-block` so it is near `write` & `write-char`. Feel free to ignore, since `paste` is also a good search keyword.)~~

Forgot about `send-key` :thinking:

imsnif:
It's brand new, no worries. I can hardly keep track myself at this point

PR #4818, opened 2026-03-09
fix: handle short writes in write_to_tty_stdin to prevent paste truncation

### LINKED ISSUE #778: copy & paste in PuTTY: long lines are splitted when pasted
When I use zellij and the output is longer than the terminal and then do a copy & paste with the mouse the lines become splitted.

I use zellij 0.18.1. My config.yaml looks like this;

  no_pane_frames: true
  disable_mouse_mode: true

The scenario is like this: I use PuTTY on a Windows 10 system and SSH to a Linux VM and from there I run zellij.

When I open a PuTTY session without running zellij and do the same thing I get one single line when pasting.

### LINKED ISSUE #4120: Paste not working
`zellij --version`: 0.42.1

`stty size`: 36 69

`uname -av` or `ver`(Windows): Darwin arm64

## Issue description
- Paste does not work in zellij

## Minimal reproduction
1. Press `cmd + v` 
2. Nothing gets pasted.
3. Exit zellij.
4. Press `cmd + v`.
5. Text gets pasted.

## Other relevant information
I use `kitty` terminal but I do not know how relevant it is as I have tested the behavior on other terminals too.
Tested on:
- `kitty`
- `Wezterm`
- `Ghostty`
- `Alacritty`
Tested on different shells too:
Again I do not know how relevant this is. 
- `fish`
- `nushell`
- `zsh`

### LINKED ISSUE #4220: Copy Paste Result Unexpected
Environment:
Zellij 0.42.2
MacOS 15.5
Zsh Shell

When Copy and Paste to Zellij the result is unexpected.

[https://imgur.com/a/XJvVxQa](url)

MuncleUscles:
## Problem

On macOS the kernel PTY buffer is only 4-16 KiB. When pasting text larger than this buffer into a Zellij terminal pane, `write_to_tty_stdin` calls `unistd::write()` once and returns whatever partial count the kernel accepted. The remaining bytes are silently dropped.

In practice this means any paste over ~1 KiB gets truncated. This affects all paste operations (bracketed paste, `zellij action write` CLI, etc.) on macOS. Linux systems with larger PTY buffers are less likely to hit this but are still technically vulnerable to short writes.

## Related Issues

- Fixes #4220 — "Copy Paste Result Unexpected"
- Likely fixes #778 — "long lines are splitted when pasted"
- May fix #4120 — "Paste not working"

## Root Cause

`UnixPtyBackend::write_to_tty_stdin()` in `zellij-server/src/os_input_output_unix.rs` performs a single `unistd::write(fd, buf)` call and returns the result directly. POSIX `write()` is allowed to return fewer bytes than requested (a "short write"), especially when:
- The kernel buffer is full
- The fd is non-blocking
- A signal interrupts the call

## Fix

Wrap the write in a retry loop that handles:
- **Short writes**: advances the buffer offset and retries
- **EINTR**: retries immediately  
- **EAGAIN**: sleeps 1ms then retries (back-pressure from full buffer)

This is the standard pattern for reliable POSIX writes (analogous to `write_all` in Rust's std).

## Test

Added a unit test that:
1. Opens a real PTY pair
2. Sets `O_NONBLOCK` on the master fd to force short writes
3. Starts a delayed reader thread on the slave side
4. Writes 128 KiB through `write_to_tty_stdin`
5. Verifies all bytes are delivered with correct content

Without the fix, only ~1 KiB out of 128 KiB is delivered. With the fix, all 128 KiB arrives correctly.

## Test plan
- [x] `cargo test -p zellij-server write_to_tty_stdin_delivers_all_bytes` passes
- [ ] Manual test: paste large text (>4 KiB) into a Zellij pane on macOS — all content should arrive

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -79,6 +79,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 * fix: allow specifying --ip and --port with `zellij web --status` rather than relying on those in the config (https://github.com/zellij-org/zellij/pull/4825)
 * fix: make sure clients are focused when atomically showing floating panes (https://github.com/zellij-org/zellij/pull/4834)
 * feat: allow configuring focus_follows_mouse and mouse_click_through (https://github.com/zellij-org/zellij/pull/4836)
+* fix: handle very large writes to the pty (https://github.com/zellij-org/zellij/pull/4818)
 
 
 ## [0.43.1] - 2025-08-08


imsnif:
Hey @MuncleUscles - good catch. Thanks for this.

My only concern here is the potentially infinite busy-wait. Could we maybe add a retry counter after which we bail with a `log::error!("Failed to write to pty")` in the logs? Let's cap it at 50 retries.

MuncleUscles:
Good catch, thanks @imsnif! Went a step further — replaced the sleep-retry loop entirely with `poll(POLLOUT)` with a 5-second timeout:

- No busy-spinning (waits for kernel writability notification)
- Returns `Err` on timeout instead of `Ok(partial)` — important because `pty_writer_main` ignores the byte count, so a silent partial write would drop input
- Also fixed a flaky test (reader delay was racing with the retry cap) and added a dedicated timeout test

Let me know if the 5s timeout feels right or if you'd prefer a different value.

imsnif:
Hey @MuncleUscles - thanks for the adjustments. I'm a little concerned about the 5s timeout, given that this would mean that STDIN is blocked for all terminals in the session.

I think a proper fix would be to queue any unwritten bytes (along with a retry count) back into the PtyWriter queue. This is a bit of a bigger and more involved change though - would you like to tackle it? If not we could merge this as is with a lower timeout (eg. 1s) and add a TODO.

MuncleUscles:
Great suggestion, done! Replaced the poll loop with the re-queue approach:

- `write_to_tty_stdin` now does a single non-blocking pass — writes as much as the kernel accepts and returns the count (0 on EAGAIN)
- PtyWriter re-queues unwritten bytes back onto its own channel, so other terminals' writes can proceed between retries
- Retry count only increments on zero progress (full EAGAIN), resets when any bytes are written
- Drops bytes with `log::error!` after 50 consecutive zero-progress retries
- Added `yield_now()` before re-queue on zero progress to avoid tight spinning
- Wired up the PtyWriter's self-sender (`to_pty_writer` was `None` in the Bus init)

One thing worth noting: since re-queued writes go to the back of the channel, writes to the same terminal could theoretically arrive out of order if there are multiple pending writes for that terminal. In practice this seems unlikely (keystrokes arrive one at a time, paste is a single Write), but happy to add per-terminal ordering if you think it's needed.

imsnif:
You're right. In addition to the ordering issue, we're also still essentially busy-waiting but with a short timeout. There can be scenarios in which the child process delays reading from STDIN due to some heavy computation and we would be dropping the buffer in microseconds.

How about we do something like this:
1. We create a buffer in the thread as you suggested, something like: `HashMap<u32 /* terminal_id */, VecDeque<u8> /* bytes to write */>`.
2. Whenever we receive a write, we append it to the buffer of the relevant terminal_id - then immediately attempt to send it (as well as anything else that was in the buffer for other terminals) using the same loop in this PR for each terminal
3. Whatever wasn't accepted by the child remains in the buffer and we keep looping
4. At the beginning of the loop, if we have anything pending in our buffer, we attempt to receive a PtyInstruction with a 10ms timeout. This means that when we receive an instruction we process it and try to drain everything, if we don't within 10ms and have pending writes, we keep trying to drain them. (Critical here though: we only listen on the channel with a timeout if we have something in our buffer, otherwise it's a normal receive. If we don't do this we'll be timing out every 10ms regardless of whether we have pending instructions or writes)
5. When we append to our pending buffer, we check if the bytes are larger than some threshold (lets say 10MB). If they are, we drop everything and log an error.

I think this should take care of the ordering concerns you raised (we always append to the buffer and attempt to write it in order), the starvation concerns (we always attempt to write all buffers), the timeout concerns (we never drop on a timeout, only retry) and the potential OOM concerns (we drop the buffer if it becomes too large). Did I miss anything? :)

Wdyt? Want to implement this? If so, do you think you can get it done in the next couple of days? We're quickly approaching a point where I don't want to add anything else before release.

MuncleUscles:
Done — implemented the per-terminal buffer approach you described:

- `HashMap<u32, VecDeque<PendingWrite>>` for per-terminal write queues
- `recv_timeout(10ms)` when pending data, blocking `recv()` when idle
- 10MB per-terminal cap (drops buffer + logs on overflow)
- Terminal errors (EBADF/EIO) clear that terminal's queue
- Added `Bus::recv_timeout` via crossbeam `Select::select_timeout`
- Removed `tcdrain` from the write path

imsnif:
Hey  @MuncleUscles - great work! I made some adjustments:
1. I changed direct vec access to use accessor methods, and swapped subtraction with safe saturating_sub. While these things should not logically happen, I prefer to be defensive here since these things can crash the entire app.
2. I added some tests to the pty_writer part (the tests for the system interface were great, but I'm more comfortable also testing the retry/drop/nonblocking logic explicitly)
3. Ran `rustfmt`

Will merge once CI passes. Thanks for this!

PR #4820, opened 2026-03-09
fix(panes): clear `window_title` cache when switching tabs

### LINKED ISSUE #2614: Name passed to terminal not send on tab transition
**Basic information**

`zellij --version`: 0.37.2
`uname -av` or `ver`(Windows):`Linux zykino-pop-os 6.2.6-76060206-generic #202303130630~1685473338~22.04~995127e SMP PREEMPT_DYNAMIC Tue M x86_64 x86_64 x86_64 GNU/Linux`

List of programs you interact with as, `PROGRAM --version`: output cropped meaningful, for example:
`alacritty --version`: 0.11.0

**Further information**
I think this issue is kind of a continuation of #883 (which should be closed I think), may almost be related to #557 and #2284.

1. If I have Tab1 with a pane, and Tab2 with 2 panes. => The terminal window show "Zellij (session-name) - fish /home/zykino"
2. I rename the panes in Tab 2 with `zellij action rename-pane a` and `zellij action rename-pane b`
    1. When selecting the pane "a" => Terminal window is renamed : "Zellij (session-name) - a"
    2. When selecting the pane "b" => Terminal window is renamed : "Zellij (session-name) - b"
    3. When selecting another pane (in the same pane) I see "Zellij (session-name) - fish /home/zykino"
    4. But if I go on another tab from pane b, the name stay "Zellij (session-name) - b"
 
On the last point I expected the Terminal name to be set to the name of the other tab’s pane name. (Whoa hard to say… so "…fish…" from Tab1 pane1).

The issue can also be reproduced with named pane in layout. Or with a long running command and then going to another tab… Or going from a shell to a long running command (`htop`, `watch ls`, …)

pilgrimlyieu:
Fixes #2614.

When switching between tabs, the terminal emulator's window title is not updated to reflect the focused pane in the newly active tab. The title remains stale, showing the pane name from the previously active tab.

For example:
- Tab 1 has pane "a" focused, terminal title shows `session-name | a`
- Switch to Tab 2, terminal title stays as `session-name | a` instead of updating to Tab 2's focused pane

This PR clear the `window_title` cache in `set_force_render()` for both `TiledPanes` and `FloatingPanes`. This method is already called on the destination tab during tab switches (via `move_clients_between_tabs`), so clearing the cache here ensures the terminal title OSC sequence is always re-emitted after a tab switch.

### DIFF
--- a/zellij-server/src/panes/floating_panes/mod.rs
+++ b/zellij-server/src/panes/floating_panes/mod.rs
@@ -358,6 +358,7 @@ impl FloatingPanes {
             pane.set_should_render_boundaries(true);
             pane.render_full_viewport();
         }
+        self.window_title = None;
     }
     pub fn set_pane_frames(&mut self) -> Result<()> {
         let err_context =
--- a/zellij-server/src/panes/floating_panes/mod.rs
+++ b/zellij-server/src/panes/floating_panes/mod.rs
@@ -408,6 +408,7 @@ impl FloatingPanes {
             pane.set_should_render_boundaries(true);
             pane.render_full_viewport();
         }
+        self.window_title = None;
     }
     pub fn set_pane_frames(&mut self) -> Result<()> {
         let err_context =
--- a/zellij-server/src/panes/tiled_panes/mod.rs
+++ b/zellij-server/src/panes/tiled_panes/mod.rs
@@ -991,6 +991,7 @@ impl TiledPanes {
             pane.render_full_viewport();
         }
         self.reset_boundaries();
+        self.window_title = None;
     }
     pub fn has_active_panes(&self) -> bool {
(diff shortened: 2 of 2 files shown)


mateibarbu19:
Hello,

I have tested this PR and **it works**!

---

## Build

### Dependencies

`ubuntu:latest` container with the following dependencies

- curl
- protobuf-compiler
- build-essential
- musl-tools
- cargo & rust

### Command

According to the instructions.

```sh
cargo xtask ci e2e --build
```

But the following should work as well.

```sh
cargo build --release --target x86_64-unknown-linux-musl
```

xyven1:
Would love to see this merged, seems low risk and addresses a small pain-point that I (and I'm sure others) have been experiencing.

Issue #4745, opened 2026-02-21
Windows implementation issues


imsnif:
This issue is a continuation of the discussion stemming from https://github.com/zellij-org/zellij/issues/316#issuecomment-3920047803, namely in regards to bugs, issues and troubleshooting in the currently-being-merged windows support branch.

leet0rz:
Currently for me it seems like zellij kill-all-sessions is not working, it gets stuck forever. Here is an example:

```
✗  zellij list-sessions
brave-mountain [Created 1day 13h 42m 23s ago] (EXITED - attach to resurrect)
sensible-glockenspiel [Created 1day 13h 39m 46s ago] (EXITED - attach to resurrect)
fascinating-viola [Created 1day 13h 35m 38s ago]
charming-lake [Created 1day 12h 51m 14s ago]
delighted-crab [Created 1day 12h 51m 1s ago]
stellar-cactus [Created 1day 12h 50m 5s ago]
delighted-mouse [Created 1day 12h 47m 23s ago]
zippy-orange [Created 1day 12h 46m 16s ago]
1 [Created 1day 32m 54s ago]
2 [Created 1day 32m 43s ago]
tremendous-goose [Created 23h 12m 33s ago] (EXITED - attach to resurrect)
unique-rhinoceros [Created 16h 55m 33s ago] (EXITED - attach to resurrect)
gracious-ukulele [Created 16h 31m 43s ago] (EXITED - attach to resurrect)
auspicious-panda [Created 11h 21m 1s ago] (EXITED - attach to resurrect)
~
> zellij kill-all-sessions
WARNING: this action will kill all sessions.
Do you want to continue? yes
```

Hopefully I am doing this the right way. Currently I am on windows 10 and zellij was built with 2022.

divens:
Thanks @imsnif for creating this issue.

I'll try to recap the issues that have been reported so far.

### Build
- [ ] VS2026 / cmake issue

### Runtime
- [x] lacking mouse support (although @FelipeCybis and @leet0rz report it working in wizterm [here](https://github.com/zellij-org/zellij/issues/316#issuecomment-3935346915) and [here](https://github.com/zellij-org/zellij/issues/316#issuecomment-3936028876))
- [x] some keyboard shortcuts not working for some users (`ctrl b` + `"` / `%`, `[` `]`...)
- [x] zellij always starting `cmd` shell, ignoring config or `$SHELL`
- [x] session hang issue when restoring stale sessions
[...]

PR #4829, opened 2026-03-10
fix: restore console input mode on exit to prevent mouse sequence leak

divens:
Fix issues after exiting Zellij on Windows, reported [here](https://github.com/zellij-org/zellij/issues/4745#issuecomment-4015484270) and [here](https://github.com/zellij-org/zellij/issues/4745#issuecomment-4015890846) 

On Windows with WezTerm (VT input path), enable_vt_input() sets ENABLE_MOUSE_INPUT and ENABLE_VIRTUAL_TERMINAL_INPUT on the console handle. crossterm's disable_raw_mode() only ORs back LINE_INPUT, ECHO_INPUT and PROCESSED_INPUT — it never clears those flags.

After Zellij exits, ConPTY continues delivering mouse events as VT escape sequences (urxvt mode 1015) into the shell's stdin, causing visible garbage like [555;99;32M at the prompt.

Fix: save the original console mode before enable_vt_input() modifies it, and restore it on all exit paths via a new restore_console_mode() method on the ClientOsApi trait. Also move disable_mouse() before unset_raw_mode() on error/panic paths for correct ordering.

### DIFF
--- a/zellij-client/src/os_input_output_windows.rs
+++ b/zellij-client/src/os_input_output_windows.rs
@@ -1,4 +1,5 @@
 use crate::os_input_output::SignalEvent;
+use crate::stdin_handler_windows::restore_vt_input;
 
 use anyhow::{Context, Result};
 use async_trait::async_trait;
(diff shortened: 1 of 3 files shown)


imsnif:
Any chance we can move the windows specific static variables and imports to os_input_output_windows or some such? :)

PR #4833, opened 2026-03-10
fix: incorrect interpretation of unicode keys

renesat:
Fix incorrect interpretation of non-English symbols in some combinations. For example swap layout when use flag (Shift, Alt or Ctrl) and non-English symbol (for example cyrillic).

<details> 
  <summary>Example of problem (GIF)</summary>
 
  Example of creating few panels and after use `Ctrl+о`
  out
</details>

Fix this comment https://github.com/zellij-org/zellij/issues/1355#issuecomment-3735242646.

I'm still not sure if I removed `.to_lowercase`. But judging by the kitty keyboard protocol, it should already be in lowercase. Or does it still make sense to apply it?


Zykino:
One of the maintainers did #4800 that seems to me goes in the same direction (reading the titles). Can you try their PR and comment there if your issue is fixed by what they did or if they missed something?

tlinford:
> One of the maintainers did #4800 that seems to me goes in the same direction (reading the titles). Can you try their PR and comment there if your issue is fixed by what they did or if they missed something?

This looks like a different issue actually, @renesat this is related to not being able to type certain key combinations correctly right?

renesat:
> One of the maintainers did https://github.com/zellij-org/zellij/pull/4800 that seems to me goes in the same direction (reading the titles). Can you try their PR and comment there if your issue is fixed by what they did or if they missed something?

Yes, this is different issues.

> This looks like a different issue actually, @renesat this is related to not being able to type certain key combinations correctly right?

Yes. Not detected some combinations and start to swap windows (and write part of kitti keyboard protocol command). 
For example when I write on my native language and then try to change window using `Alt+l` without change keyboard layout.

In general, the main problem for me is not the inability to add a combination with a non-ascii character, but the annoying behavior if I forgot to switch the language.

I would also like the combinations to be independent of the layout (if set in configuration), but this is for separate PR. And current PR is more a fix than a feature.

renesat:
@Zykino, can you check it? It's just that the change is not big, but it really improves interaction in a multilingual environment.

imsnif:
Hey @renesat - this looks good. Github currently has some issues with Github Actions and so I'm unable to approve the CI run here. Once that's unblocked, I'll run the CI and merge if all is well.

Thanks!

imsnif:
@renesat - I'm still unable to run CI. I think this has to do with a Github Actions outage yesterday that is amazingly still going on in some ways. I'm not 100% sure this will work, but could you please close this PR, copy it to a different branch in your fork and then open a new PR? Please tag me there so I won't miss it. Thanks!

PR #4839, opened 2026-03-12
Use build-time git describe for CLI version output

laciferin2024:
- Add a zellij-utils build script that computes ZELLIJ_VERSION from git describe --tags --dirty --always, with env override and Cargo package version fallback.
- Wire consts::VERSION to prefer the build-injected ZELLIJ_VERSION value.
- Make clap --version use crate::consts::VERSION so runtime version output follows build-time metadata.


laciferin2024:
I find this particularly handy when raising issues against a local `build` against a commit.

PR #4843, opened 2026-03-12
Fix debug-mode stack overflow on Windows

divens:
The clap-derived augment_subcommands for CliAction (~70 variants) produces a >1 MB stack frame in debug mode, overflowing the Windows default 1 MB main-thread stack. Increase it to 8 MB via linker flag to match Linux. Only applies to debug builds; release optimizes the frame down and doesn't need it.

Stack overflow reported [here](https://github.com/zellij-org/zellij/issues/4745#issuecomment-4029203844)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,5 +1,6 @@
 [package]
 name = "zellij"
+build = "src/build.rs"
 authors = ["Aram Drevekenin <aram@poor.dev>"]
 description = "A terminal workspace with batteries included"
 homepage = "https://zellij.dev"
--- a/src/build.rs
+++ b/src/build.rs
@@ -2,8 +2,8 @@ fn main() {
     // The clap-derived `augment_subcommands` for `CliAction` (~70 variants)
     // produces a >1 MB stack frame in debug mode, overflowing the Windows
     // default 1 MB main-thread stack.  Increase it to 8 MB to match Linux.
-    // Release builds optimize the frame down, so this is only needed for debug.
-    if cfg!(target_os = "windows") && std::env::var("PROFILE").unwrap_or_default() == "debug" {
+    // Release builds optimize the frame down, so this is only needed for non-release profiles.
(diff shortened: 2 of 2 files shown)


imsnif:
So, I played around with this a little bit and actually managed to fix the stack overflow for stacks of even 700KB - however this requires a pretty massive refactor of our CLI infrastructure. While I'm not enthusiastic about playing around with stack sizes even for debug builds, I think you're right and this is the better solution for now.

We're long overdue to upgrade clap anyway - might as well push this problem down the road to when we do that.

Otherwise, I moved the file to the `src` folder so as not to clutter the repo base more than it already is. I also inverted it to a !release so that our dev-opt profile (what we get through cargo x build) will also benefit.

What do you think? Still solves the problem?

divens:
> So, I played around with this a little bit and actually managed to fix the stack overflow for stacks of even 700KB - however this requires a pretty massive refactor of our CLI infrastructure. While I'm not enthusiastic about playing around with stack sizes even for debug builds, I think you're right and this is the better solution for now.
> 
> We're long overdue to upgrade clap anyway - might as well push this problem down the road to when we do that.

I agree this is a pragmatic, acceptable short term solution. I haven't had time to dive into CLI argument parsing, and you obviously have much more experience and better judgement on this. Given the point at which we are in the release cycle, I feel, as you said, the CLI rework is better left as a future improvement.

> Otherwise, I moved the file to the `src` folder so as not to clutter the repo base more than it already is. I also inverted it to a !release so that our dev-opt profile (what we get through cargo x build) will also benefit.
> 
> What do you think? Still solves the problem?

Yes, fine with moving the build script in `src` and making it `!=release` instead of `==debug` Just rebuilt it locally in debug mode, no stack overflow. I also verified `release` build gets standard 1MB stack, and `dev-opt` gets 8MB.

imsnif:
Cool. Thanks for checking!

Issue #4500, opened 2025-11-07
Passthrough to use kitty graphics protocol


itpauker:
Can passthrough be implemented to use for example kitty graphics protocol, so that the zellij: **is not supported** statement can be removed from the following page:

[nvim snacks](https://github.com/folke/snacks.nvim/blob/main/docs/image.md)

NovaViper:
Hey this was recently completed with https://github.com/zellij-org/zellij/issues/2814#issuecomment-5145843315. This can be closed this now

Issue #4336, opened 2025-08-03
Support for yazi's image preview


r0b0ji:
In my Ghostty terminal when I use yazi, it shows the image preview. However after opening the zellij session, it doesn't work inside zellij.

From the [yazi's page](https://yazi-rs.github.io/docs/image-preview)
>Zellij currently only supports the Sixel graphics format, so you will need a terminal that also supports Sixel.

> Note that, Zellij's Sixel implementation is quite buggy and has serious performance issues at the moment, causing noticeable lagginess when quickly switching between images, and sometimes even [image tearing](https://github.com/zellij-org/zellij/issues/2576#issuecomment-1707107473) or [not working at all](https://github.com/zellij-org/zellij/issues/2814#issuecomment-2318473921).

> This situation won't improve until Zellij enhances its Sixel implementation or [provides a passthrough mode](https://github.com/zellij-org/zellij/issues/775). If the image is a stronger need to you, consider running Yazi outside of Zellij or using Überzug++:

```
# Deceive Yazi into thinking you're running in kitty,
# forcing it fallback to Überzug++ or Chafa
TERM=xterm-kitty yazi
```

imsnif:
Most sixel issues have been fixed a couple of versions ago iirc. Sixel should mostly work aside from a few glitches here and there (I followed the provided links but couldn't find a reproduction - they're mostly just user reports with a vid/screencast, which is totally fine ofc but there's not much I can do with them) - namely having to do with Zellij not supporting the aspect ratio part of the sixel protocol (at the time of implementation no-one really supported them and I considered adding them as overeager protocol implementation). If aspect ratios aren't used, it should work fine.

Otherwise - I don't want to put a lot of work into the Sixel protocol, as it seems to be going out of fashion (and honestly has some quite frustrating inherent limitations).

I have briefly experimented with implementing passthrough - but found that most kitty image implementers are heavily relying on implicit parts of the tmux passthrough protocol (eg. certain tmux-specific commands or env variables being present) and that kitty itself seems to rely on implicit behavior as well (namely *when* the passthrough bytes are sent to the terminal and at what sequence in relation to the unicode placeholders).

[...]

Issue #3954, opened 2025-01-25
Escape code passthrough


PartyWumpus:
Afaict, escape sequence are always understood by zellij and handled by zellij. This means any escape sequence zellij does not support, but the terminal emulator does support can't be used. It would be nice if there was a (probably disabled by default) option to pass through escape codes wrapped in a specific sequence, like tmux's `\ePtmux;...\e\\` ([tmux(1)#allow-passthrough(https://man.openbsd.org/OpenBSD-current/man1/tmux.1#allow-passthrough)) or screen's `\eP..\e\\` ([https://www.gnu.org/software/screen/manual/screen.html#Control-Sequences](screen#Control-Sequences))

I suppose there could be a plugin which makes zellij understand the tmux escape, but applications aren't going to be sending it because they don't think they're in tmux (unless the plugin also sets the TMUX environment variable), but I think that's a much poorer solution (also not sure if a plugin can do that).

Any plugin that implements a new sequence (say `\ePzellij;...\e\\`is going to get probably no support at all because that's a niche on a niche.

Andrew15-5:
This will unblock https://github.com/zellij-org/zellij/issues/2814 and make fixing it pretty trivial, AFAIK.

PartyWumpus:
> This will unblock [#2814](https://github.com/zellij-org/zellij/issues/2814) and make fixing it pretty trivial, AFAIK.

Sort of... It would make unicode placements viable, but I imagine anything involving any other kind of placement would be weird, as the program is trying to use a "pane-relative" position, but it's going to instead get an absolute, window-relative position.

This is okay though, because the program has to opt-in to zellij passthrough in any situation, so the developer is aware of this. Long term it would still be better for proper support, like it does have for sixel so the rest of the protocol works.

I see someone else in this thread also suggested what I am suggesting [here](https://github.com/zellij-org/zellij/issues/2814#issuecomment-1937771460) and [here](https://github.com/zellij-org/zellij/issues/2814#issuecomment-2295674844)

infogrind:
IIUC this would also fix an issue with Kitty on MacOS that Option Left/Right, which are sent as `\e[1;3D` and `\e[1;3C` by Kitty and allow to move the cursor by word in the shell prompt (at least for Fish and Bash shells on my system).

What is the reason that Zellij filters out escape sequences in terminal input?

PartyWumpus:
> What is the reason that Zellij filters out escape sequences in terminal input?

[...]

PR #4851, opened 2026-03-14
feat: add Kitty graphics protocol passthrough

### LINKED ISSUE #2814: [Feature Request] - Implement kittys terminal graphics protocol
It would be neat if zellij could use kitty terminal graphics to display images or other files. Tmux had a [similar](https://github.com/tmux/tmux/issues/1391) issue closed, it seems like they will not be implementing it. Zellij does provide support for sixels but this is a very old protocol with not much support from other applications.

varunkamath:
## Summary

Adds passthrough support for the Kitty graphics protocol, enabling image display in terminals that support it (Ghostty, Kitty, etc.) when running Kitty-graphics-aware applications inside Zellij.

- Intercept Kitty graphics APC sequences (`\x1b_G...\x1b\\`) before vte silently discards them, and re-emit them as raw VTE output
- Store combining characters (width-0 Unicode) per cell so Kitty unicode placements (`\u{10EEEE}` + diacritics) pass through correctly
- Double-buffered frame handling ensures only complete frames are emitted, with a dirty flag for efficient re-emission on tab switch

This is a passthrough-only approach (~350 lines across 3 files) — Zellij does not parse or understand the image data, it just relays it. Applications that use quiet mode (`q=2`) with unicode placements work out of the box, including image viewers and streaming video tools.

Closes #2814. Related: #3954, #4500, #4336.

## Design

**APC interception** (`terminal_pane.rs`): A pre-parser state machine intercepts `\x1b_G...\x1b\\` sequences before they reach the vte parser (which silently discards all APC content in v0.11). Batch fast paths skip the scanner entirely for non-ESC byte runs in both Normal and InKittyApc states.

**Frame management** (`grid.rs`): APC chunks are accumulated in a building buffer and promoted to a ready buffer when the frame is complete (`m=0`). A dirty flag controls emission — set on new frame arrival or full viewport re-render (tab switch), cleared after emission. This avoids redundant output while ensuring static images survive tab switches.

**Combining characters** (`grid.rs` + `output/mod.rs`): Width-0 characters are now stored as combining marks on the preceding cell instead of being dropped. These are attached to `CharacterChunk`s during change detection and emitted inline during serialization. Scroll operations adjust the combining chars HashMap via single-pass drain+collect.

**Safety bounds**: In-progress frame data is capped at 64 MB to prevent unbounded memory growth. APC parameter parsing uses proper comma-delimited key=value matching.

## What works

Applications that use **quiet mode (`q=2`) with unicode placements (`U=1`)** work correctly. This includes tools that transmit image data inline and use `\u{10EEEE}` diacritics for positioning, such as:
- Custom Kitty-graphics-aware applications
- Tools using `ratatui-image` with Kitty protocol
- Streaming video at 12+ FPS with minimal overhead vs native terminal

## Limitations

- **No reverse APC passthrough**: The Kitty protocol includes a response channel (terminal → application) for detection queries and error reporting. This implementation only handles the forward direction (application → terminal). Applications like `kitten icat` that rely on graphics query responses will not work — they need the terminal's APC responses forwarded back through the PTY. This would require changes to Zellij's client input handling path and is left as future work.
- **Floating pane artifacts**: When a floating pane partially obscures a pane displaying Kitty graphics, image fragments may appear at the visible edges of the background pane. Closing the floating pane restores correct rendering. This is inherent to the passthrough approach — the terminal renders the image wherever placeholder characters exist.
- **Combining chars not preserved in scrollback history**: Combining characters are adjusted for viewport scrolling but not stored in `lines_above`. This is fine for alternate-screen applications (the primary use case) but means images won't survive scrollback in normal-screen mode.
- **Non-Kitty APC sequences** are silently dropped (same as vte's existing behavior, no regression).

## Test plan

- [x] `cargo test` — all 862 existing tests pass
- [x] `cargo xtask format` — passes
- [x] Tested with Kitty-graphics-aware application displaying static images inside Zellij panes (Ghostty terminal)
- [x] Tested with streaming video at 12+ FPS with <0.1 FPS variance vs native Ghostty
- [x] Verified images survive tab switch and return
- [x] Verified multi-pane with different image streams (no cross-contamination)
- [x] Verified no garbage output after application exit (alternate screen cleanup)
- [x] Verified `kitten icat` fails gracefully (known limitation — no reverse passthrough)
- [ ] Test with Kitty terminal
- [ ] Test with `yazi` image previews (may need `q=2` support)
- [ ] Test with `timg`, `chafa --format=kitty`

### DIFF
--- a/zellij-server/src/output/mod.rs
+++ b/zellij-server/src/output/mod.rs
@@ -161,7 +161,7 @@ fn serialize_chunks_with_newlines(
         let mut character_styles = DEFAULT_STYLES.enable_styled_underlines(styled_underlines);
         vte_output.push_str("\n\r");
         let mut chunk_width = character_chunk.x;
-        for t_character in character_chunk.terminal_characters.iter() {
+        for (col_idx, t_character) in character_chunk.terminal_characters.iter().enumerate() {
             // Stop rendering if the next character would exceed max_size.cols
             if let Some(size) = max_size {
                 if chunk_width + t_character.width() > size.cols {
@@ -190,6 +190,11 @@ fn serialize_chunks_with_newlines(
             .with_context(err_context)?;
             chunk_width += t_character.width();
             vte_output.push(t_character.character);
+            if let Some(combiners) = character_chunk.combining_chars.get(&col_idx) {
+                for &c in combiners {
+                    vte_output.push(c);
+                }
+            }
         }
     }
     Ok(vte_output)
@@ -226,7 +231,7 @@ fn serialize_chunks(
         vte_goto_instruction(character_chunk.x, character_chunk.y, &mut vte_output)
             .with_context(err_context)?;
         let mut chunk_width = character_chunk.x;
-        for t_character in character_chunk.terminal_characters.iter() {
+        for (col_idx, t_character) in character_chunk.terminal_characters.iter().enumerate() {
             // Stop rendering if the next character would exceed max_size.cols
             if let Some(size) = max_size {
                 if chunk_width + t_character.width() > size.cols {
@@ -255,6 +260,11 @@ fn serialize_chunks(
             .with_context(err_context)?;
             chunk_width += t_character.width();
             vte_output.push(t_character.character);
+            if let Some(combiners) = character_chunk.combining_chars.get(&col_idx) {
+                for &c in combiners {
+                    vte_output.push(c);
+                }
+            }
         }
     }
     if let Some(sixel_image_store) = sixel_image_store {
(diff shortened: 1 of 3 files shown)


imsnif:
Hi - thanks for this.

This is unfortunately a bit of a naive (and protocol specific) implementation that will represent a significant long-term maintenance overhead. A proper generic protocol-agnostic passthrough solution is on my TODO. The upcoming version is quite large already. I hope to be able to dedicate time to it in the following one.

Thank you for your patience.

varunkamath:
Ah, love to hear it. Thanks for the context, a generic passthrough makes sense as the long-term solution. Happy to close this in favor of that if you prefer, otherwise will keep using my fork and periodically rebase in the meantime as it suits my needs.

imsnif:
As you wish. It is generally my policy not to close issues or PRs as that represents a significant maintenance overhead as well (I think it's one of those things that makes sense for enterprises-maintained applications and not so much for run-of-the-mill OSS projects).

varunkamath:
Closing, thanks for the quick response time.

PR #4854, opened 2026-03-15
Add windows support in strider

divens:
- Add platform-aware path handling to strider (drive letters, backslash display, UNC paths)
- Add a "Computer" virtual root view when navigating up from a drive root on Windows, listing all drives and WSL distributions
- Add `PluginCommand::ListHostEntries` (drives enumerated via `GetLogicalDriveStringsW`, WSL distros via `wsl -l -q`)
- No behavior change on Unix

Addresses issue reported [here](https://github.com/zellij-org/zellij/issues/4745#issuecomment-4049266256)

### DIFF
--- a/default-plugins/strider/src/platform.rs
+++ b/default-plugins/strider/src/platform.rs
@@ -18,15 +18,16 @@ impl Platform {
     pub fn detect(initial_cwd: &str) -> Self {
         let bytes = initial_cwd.as_bytes();
         // Drive letter: X:\ or X:/
-        if bytes.len() >= 3
-            && bytes[0].is_ascii_alphabetic()
-            && bytes[1] == b':'
-            && (bytes[2] == b'\\' || bytes[2] == b'/')
-        {
+        let is_drive_letter = matches!(
+            (bytes.get(0), bytes.get(1), bytes.get(2)),
+            (Some(c), Some(b':'), Some(b'\\' | b'/')) if c.is_ascii_alphabetic()
+        );
+        if is_drive_letter {
             return Platform::Windows;
         }
         // UNC path: \\server\share
-        if bytes.len() >= 2 && bytes[0] == b'\\' && bytes[1] == b'\\' {
+        let is_unc_path = matches!((bytes.get(0), bytes.get(1)), (Some(b'\\'), Some(b'\\')));
+        if is_unc_path {
             return Platform::Windows;
         }
         Platform::Unix
@@ -67,7 +68,11 @@ impl Platform {
     pub fn ensure_drive_root(path: PathBuf, platform: Platform) -> PathBuf {
         if platform == Platform::Windows {
             let s = path.to_string_lossy();
-            if s.len() == 2 && s.as_bytes()[0].is_ascii_alphabetic() && s.as_bytes()[1] == b':' {
+            let bytes = s.as_bytes();
+            let is_bare_drive = s.len() == 2
+                && matches!(bytes.get(0), Some(c) if c.is_ascii_alphabetic())
+                && matches!(bytes.get(1), Some(b':'));
+            if is_bare_drive {
                 return PathBuf::from(format!("{}/", s));
             }
         }
@@ -99,16 +104,19 @@ impl Platform {
         match platform {
             Platform::Unix => s == "/" || s.is_empty(),
             Platform::Windows => {
+                let bytes = s.as_bytes();
                 // Drive root: "C:/" or "C:"
-                if s.len() == 3
-                    && s.as_bytes()[0].is_ascii_alphabetic()
-                    && s.as_bytes()[1] == b':'
-                    && s.as_bytes()[2] == b'/'
-                {
+                let is_drive_root_slash = s.len() == 3
+                    && matches!(bytes.get(0), Some(c) if c.is_ascii_alphabetic())
+                    && matches!(bytes.get(1), Some(b':'))
+                    && matches!(bytes.get(2), Some(b'/'));
+                if is_drive_root_slash {
                     return true;
                 }
-                if s.len() == 2 && s.as_bytes()[0].is_ascii_alphabetic() && s.as_bytes()[1] == b':'
-                {
+                let is_bare_drive = s.len() == 2
+                    && matches!(bytes.get(0), Some(c) if c.is_ascii_alphabetic())
+                    && matches!(bytes.get(1), Some(b':'));
+                if is_bare_drive {
                     return true;
                 }
                 // UNC root: //server/share (at most 4 path components)


imsnif:
Regarding the plugin command: does it have a meaning in non-windows environments? If not, could we maybe adjust its naming to specify that it only works on windows and give a descriptive error if we're not on windows?

Otherwise, a few things off the top of my head without going too deep:
1. Unless I missed something, this new plugin command is not gated behind any permission. It needs to be gated behind the `FullHdAccess` permission.
2. I'm trying to move the plugin system away from asynchronous Event response in things that don't take Human-noticeable time (I think this qualifies?). Could we have the method return the data synchronously (there are a few other examples of this in zellij_exports).

Thanks!

divens:
> Regarding the plugin command: does it have a meaning in non-windows environments? If not, could we maybe adjust its naming to specify that it only works on windows and give a descriptive error if we're not on windows?

The semantics of this new `ListHostDirectories` command is : "list all root directories on this host". On Non-windows, this will always be `/`, given how the file system is handled on such platforms. On Windows, this will be all drives (fixed and removable) as well as all WSL virtual disks.

> Otherwise, a few things off the top of my head without going too deep:
> 
> 1. Unless I missed something, this new plugin command is not gated behind any permission. It needs to be gated behind the `FullHdAccess` permission.

I will look into this. Before adding this command to implement the virtual root and ability to navigate multiple drives, I could already navigate the whole file system with the stock Strider implementation, including system directories like `c:\windows\system32` and such. I think I saw a question about the ability to contain navigation to cwd. Is this related in any way?

> 2. I'm trying to move the plugin system away from asynchronous Event response in things that don't take Human-noticeable time (I think this qualifies?). Could we have the method return the data synchronously (there are a few other examples of this in zellij_exports).

Ok, I will dig into this.

imsnif:
> The semantics of this new `ListHostDirectories` command is : "list all root directories on this host". On Non-windows, this will always be `/`, given how the file system is handled on such platforms. On Windows, this will be all drives (fixed and removable) as well as all WSL virtual disks.

Alright. So are we getting anything out of this that we won't be getting out of something like `list_windows_drives` -> `C:\, A:\, etc.`?

> I will look into this. Before adding this command to implement the virtual root and ability to navigate multiple drives, I could already navigate the whole file system with the stock Strider implementation, including system directories like `c:\windows\system32` and such. I think I saw a question about the ability to contain navigation to cwd. Is this related in any way?

Strider is a built-in plugin, so doesn't need permission requests. I take it you're also able to navigate to other windows drives?

divens:
> Alright. So are we getting anything out of this that we won't be getting out of something like `list_windows_drives` -> `C:\, A:\, etc.`?

Sure, we can make it explicely a Windows command and remove the not-windows variant. The only call to it is within `enter_virtual_root`, which is behind `if self.platform == Platform::Windows`. Will make this change.

> Strider is a built-in plugin, so doesn't need permission requests. I take it you're also able to navigate to other windows drives?

Without the virtual root, we can't because we stop at the current drive root (c:\ in my case, but it could be any drive from the original cwd when zellij was started). With the virtual root, yes, we can navigate to any drive.

imsnif:
What's the status here?

divens:
I've renamed the new command to `list_windows_volumes` and gated it behind `FullHdAccess` permission. On non-windows, it now only ` log::error!("ListWindowsVolumes is only supported on Windows");`

I have not modified the event response part. Can do if you deem it necessary.

default-plugins/strider/src/platform.rs
```diff
@@ -0,0 +1,232 @@
...
+impl Platform {
+    /// Detect the host platform from the initial_cwd string.
+    /// Checks for drive letter (e.g. `C:\` or `C:/`) or UNC path (`\\server\`).
+    pub fn detect(initial_cwd: &str) -> Self {
+        let bytes = initial_cwd.as_bytes();
+        // Drive letter: X:\ or X:/
+        if bytes.len() >= 3
+            && bytes[0].is_ascii_alphabetic()
+            && bytes[1] == b':'
+            && (bytes[2] == b'\\' || bytes[2] == b'/')
```
imsnif: I generally like not to ever use direct array/vec accessors in this repo, as there is the possibility of a panic. Even if logically it's totally safe like the above (and even if here the only thing that will crash is the plugin) let's please use getters/setters and handle the None case. I know it can seem arbitrary, but many times thing change around the code and suddenly a logical impossibility accidentally becomes possible. Thanks for bearing with me :)

default-plugins/strider/src/platform.rs
```diff
@@ -0,0 +1,232 @@
...
+        {
+            return Platform::Windows;
+        }
+        // UNC path: \\server\share
+        if bytes.len() >= 2 && bytes[0] == b'\\' && bytes[1] == b'\\' {
```
imsnif: Same as above

default-plugins/strider/src/platform.rs
```diff
@@ -0,0 +1,232 @@
...
+        }
+    }
+
+    /// Ensure bare drive letters like `C:` become `C:/`.
+    /// WASM PathBuf's `parent()` strips the trailing slash, but on Windows
+    /// `C:` means "current directory on drive C", not the drive root.
+    pub fn ensure_drive_root(path: PathBuf, platform: Platform) -> PathBuf {
+        if platform == Platform::Windows {
+            let s = path.to_string_lossy();
+            if s.len() == 2 && s.as_bytes()[0].is_ascii_alphabetic() && s.as_bytes()[1] == b':' {
```
imsnif: Same, etc. Won't comment anymore from here and will trust you to get them all

ofek:
Thanks for all the hard work on the effort to support Windows! I'm very excited to become a user 🙂 

> UNC paths

I haven't read the diff but I saw this in the description. Please do everything you can to avoid using that format as the internal representation. You can take a look at https://github.com/openai/codex/issues/10347 as an example of how many issues can be caused by doing so.

divens:
> Thanks for all the hard work on the effort to support Windows! I'm very excited to become a user 🙂
> 
> > UNC paths
> 
> I haven't read the diff but I saw this in the description. Please do everything you can to avoid using that format as the internal representation. You can take a look at [openai/codex#10347](https://github.com/openai/codex/issues/10347) as an example of how many issues can be caused by doing so.

Thanks for bringing this to my attention. UNC paths are only used to access WSL file systems (e.g. `\\wsl.localhost\Ubuntu-24.04`) within strider. Opening a file located in a WSL fs by selecting it in strider actually works (at least with `$EDITOR=edit`), which is reassuring.

divens:
@imsnif Just rebased and fixed recent conflicts. It should be good to go.

PR #4882, opened 2026-03-22
fix(stdin): preserve non-zellij ANSI bytes during startup parsing (#4093)

### LINKED ISSUE #4093: SSH hangs when using Kitty's SSH kitten
# zellij makes ssh imposssible with kitty and ssh kitten 

## Issue description

Running the SSH kitten (not to worry about terminfo on remote servers when using Kitty) from inside Zellij
makes the SSH connection hang and never complete, you're not able to interact with the host, despite as when running with `-v` you can see all the key exchange steps succeed.

The process just hangs, running the same command outside zellij works.

I found https://github.com/kovidgoyal/kitty/issues/5900 with a quick search, it seems on the zellij side.

## Minimal reproduction

 `kitty +kitten ssh <somehost>`
or `alias ssh="kitty +kitten ssh"` followed by `ssh <somehost>`.

## Other relevant information

zellij version: 0.42.0

krapcys1-maker:
Fixes #4093

## Summary

Prevent SSH kitten startup hangs by preserving non-zellij ANSI bytes while still parsing zellij startup queries.

## Problem

When stdin startup ANSI parsing is enabled, unknown/non-zellij ANSI bytes can be dropped, which can stall kitty +kitten ssh protocol exchanges.

## What changed

- add parse_with_leftovers() in stdin_ansi_parser to return unrecognized bytes instead of dropping them
- forward leftover bytes through normal input parsing path in stdin_handler
- add unit tests for preserving non-zellij sequences and parsing known zellij sequences

## Solution

Make startup ANSI parsing extraction non-destructive: consume known zellij responses, keep and forward unknown bytes.

## Why

This keeps zellij startup capability detection intact while avoiding data loss for external terminal protocol traffic used by kitty ssh kitten.

## Tests

- `cargo test -p zellij-client parse_with_leftovers -- --nocapture`
- `cargo test -p zellij-client --lib stdin_ansi_parser -- --nocapture`
- `cargo check -p zellij-client`

### DIFF
--- a/zellij-client/src/stdin_ansi_parser.rs
+++ b/zellij-client/src/stdin_ansi_parser.rs
@@ -105,6 +105,19 @@ impl StdinAnsiParser {
         }
         self.drain_pending_events()
     }
+    pub fn parse_with_leftovers(
+        &mut self,
+        mut raw_bytes: Vec<u8>,
+    ) -> (Vec<AnsiStdinInstruction>, Vec<u8>) {
+        let mut leftover_bytes = vec![];
+        for byte in raw_bytes.drain(..) {
+            if let Some(leftovers) = self.parse_byte(byte) {
+                leftover_bytes.extend(leftovers);
+            }
+        }
+        leftover_bytes.append(&mut self.raw_buffer);
+        (self.drain_pending_events(), leftover_bytes)
+    }
     pub fn read_cache(&self) -> Option<Vec<AnsiStdinInstruction>> {
         match OpenOptions::new()
             .read(true)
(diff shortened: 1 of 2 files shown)


kika:
@imsnif hola

imsnif:
Hi @kika - thanks for the ping.

I understand the use-case and would definitely like to support it, but this is unfortunately a bit of a naive implementation. Forwarding all unrecognized bytes unconditionally from STDIN to Zellij can cause unexpected and even dangerous side-effects. Furthermore, this doesn't address the other direction (which I know kitty relies on in other places) of communicating using private sequences from the app to the terminal emulator.

On my TODO list for the next version is implementing a passthrough protocol (similar to what tmux does) that will allow this behavior to be explicit. The Kitty maintainers have indicated they will accept a PR to implement this behavior when it exists: https://github.com/kovidgoyal/kitty/issues/5900#issuecomment-1387420845

PR #4890, opened 2026-03-23
fix: resolve Windows compilation warnings

divens:
- Consume unused `CompilationResult` from `embed_resource::compile` in `build.rs`
- Gate `std::io::prelude` import with `#[cfg(unix)]` since it's only used by `daemonize_web_server` which is unix-only


divens:
@imsnif Can you please merge this? Thanks!

PR #4892, opened 2026-03-23
fix: user themes from `~/.config/zellij/themes` not loaded on session start

### LINKED ISSUE #4889: [zellij 0.44.0] No custom theme applied at launch
## Issue description

I'm having trouble loading a theme when Zellij launches.

The theme in question, located in the `.config/zellij/themes` folder, is called `current.kdl` and is a symlink. The original file has `themes.current`.

When I tap or refresh config.kdl, the theme is applied to the current terminal.
But when I launch a new Alacritty window with Zellij, it doesn't.

## Minimal reproduction

I'm using the default config.kdl for testing.

```kdl
// Choose the theme that is specified in the themes section.
// Default: default
//
// theme "dracula"
theme "current"
```

## Other relevant information

I'm on Ubuntu (24.04), Zellij version is 0.44.0 and Alacritty 0.16.1.

Kasui92:
User-defined themes placed in `~/.config/zellij/themes/` were silently ignored when starting a new session, causing `theme "name"` in `config.kdl` to always fall back to the built-in default palette.

Close #4889

### DIFF
--- a/zellij-utils/src/input/cli_assets.rs
+++ b/zellij-utils/src/input/cli_assets.rs
@@ -72,7 +72,13 @@ impl CliAssets {
             .as_ref()
             .and_then(|o| o.theme_dir.clone())
             .or_else(|| {
-                get_theme_dir(self.config_dir.clone().or_else(find_default_config_dir))
+                config_with_merged_layout_opts
+                    .options
+                    .theme_dir
+                    .clone()
+                    .or_else(|| {
+                        get_theme_dir(self.config_dir.clone().or_else(find_default_config_dir))
+                    })
                     .filter(|dir| dir.exists())
             });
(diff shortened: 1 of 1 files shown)


imsnif:
Hey, good catch! I made a small adjustment so that we won't lose the `theme_dir` if configured in a layout file. Could you check if this still solves the problem for you?

Kasui92:
I've tried several times and it doesn't seem to work. 
I'm changing my approach and trying to fix it while preserving the layouts.

Kasui92:
> I've tried several times and it doesn't seem to work. I'm changing my approach and trying to fix it while preserving the layouts.

I have to think again, I probably didn't build it properly or was using the wrong compiled. 

The fix works great! 💪🏻

imsnif:
Nice. Thank you!

Issue #2111, opened 2023-01-22
Does not work on RISC-V 64, unsupported cranelift ISA


FSMaxB:
**Basic information**

`zellij --version`: `zellij 0.34.4`
`stty size`: `91 384`
`uname -av` : `Linux starfive 5.15.0-starfive #1 SMP Wed Dec 21 03:56:12 EST 2022 riscv64 GNU/Linux`

**Further information**
I recently received a Starfive VisionFive 2 RISC-V devboard, installed rust on it and installed zellij using `cargo install zellij`.

When trying to run `zellij`, I get the following error:
```

Error occurred in server:

  × Thread 'wasm' panicked.
  ├─▶ Originating Thread(s)
  │   	1. ipc_server: NewClient
  │   	2. screen_thread: NewTab
  │   	3. plugin_thread: NewTab
  │
  ├─▶ At /home/max/.cargo/registry/src/github.com-1ecc6299db9ec823/wasmer-compiler-cranelift-2.3.0/src/config.rs:73:45
  ╰─▶ construct Cranelift ISA for triple: Unsupported
  help: If you are seeing this message, it means that something went wrong.
        Please report this error to the github issue.
        (https://github.com/zellij-org/zellij/issues)

        Also, if you want to see the backtrace, you can set the `RUST_BACKTRACE` environment variable to `1`.
```

I expect that this is a result from plugins being WebAssembly and whatever webassembly runtime (probably `wasmtime`) you use relies on cranelift, which doesn't have support for RISC-V yet.

According to https://docs.wasmtime.dev/stability-platform-support.html#interpreter-support, `wasmtime` doesn't even have an interpreter, so I guess the only way to get `zellij` to run on RISC-V is if it were possible to completely disable plugins, so it doesn't even try to run webassembly?

FSMaxB:
Although this changelog entry confuses me: https://github.com/bytecodealliance/wasmtime/blob/main/RELEASES.md#200

Apparently `wasmtime` *does* have support for `riscv64` :thinking:

FSMaxB:
Nevermind, you're using `wasmer`, not `wasmtime`.

valpackett:
riscv64 is supported starting with wasmer 3.2.0. zellij is way behind on wasmer updates…

ifsheldon:
I got the same/very similar issue when I tried to compile zellij from source on my RISC-V board.

ToxicFrog:
[...]

PR #4913, opened 2026-03-25
ci: add riscv64 to release binaries

gounthar:
## What does this PR do?

Adds `riscv64gc-unknown-linux-gnu` to the release workflow matrix so that each release includes pre-built riscv64 binaries (both normal and no-web variants).

## Why

Since #3349 switched from Wasmer to Wasmtime, zellij runs on riscv64. However, users on RISC-V boards must build from source. Adding riscv64 to the release matrix removes that requirement.

Relates to #2111.

## How

Adds a `linux gnu riscv64` entry to both `build-release-normal` and `build-release-noweb` matrices in `release.yml`. Uses `cargo xtask ci cross riscv64gc-unknown-linux-gnu`, consistent with how aarch64 is already handled.

The target is `riscv64gc-unknown-linux-gnu` (not musl) because musl support for riscv64 in cross-rs is less mature.

## Build evidence

Cross-compilation test (8 minutes on x86_64 runner): https://github.com/gounthar/zellij/actions/runs/23536356769
Native riscv64 build (95 minutes): https://github.com/gounthar/zellij/actions/runs/23497612353


imsnif:
Hey, thanks for this. A couple of things:
1. I think we don't need the separate build file and the addition to release is enough
2. I looked into it and it seems that musl `riscv64` is now Tier2: https://doc.rust-lang.org/nightly/rustc/platform-support/riscv64gc-unknown-linux-musl.html - maybe things have changed? wdyt?

gounthar:
1. Removed the separate cross-compile workflow file. Only the release.yml change remains.
2. Switched to `riscv64gc-unknown-linux-musl`, consistent with the other Linux targets. Rust tier 2 support for musl riscv64 has been stable since 1.76. My bad. 🤷

imsnif:
Nice, thank you! Will be releasing a patch version with some fixes in the coming days so we'll have a chance to test this.

gounthar:
Hi @imsnif,

I noticed the HOTFIX commit that commented out riscv64 before the v0.44.1 release. I looked at the failed job logs to understand why.

The root cause seems to be that `cross-rs`'s Docker image for `riscv64gc-unknown-linux-musl` does not include `riscv64-linux-musl-gcc`, so any crate with a C build script fails:

```
error occurred in cc-rs: failed to find tool "riscv64-linux-musl-gcc": No such file or directory
```

I may be wrong, but one possible way to fix this without depending on cross-rs for riscv64 would be to build natively on a RISC-V runner. The RISE project provides free GitHub Actions runners with the label `ubuntu-24.04-riscv`. I have been using them for other projects and they seem to work well for Rust builds.

The matrix entry might look something like:

```yaml
- build: linux musl riscv64
  os: ubuntu-24.04-riscv
  target: riscv64gc-unknown-linux-musl
```

With a setup step to install the musl toolchain on the native runner:

```bash
sudo apt-get install -y musl-tools
rustup target add riscv64gc-unknown-linux-musl
```

Because the runner is native riscv64, cc-rs finds the system compiler without extra configuration. That said, the xtask's `cross_compile` function calls the `cross` binary, so it would need a small adjustment to use `cargo` directly when building natively, or the workflow could add a separate native build path for riscv64. I am not sure which approach fits better with the existing structure.

Happy to open a follow-up PR if that would help, but I wanted to check with you first in case there is something I am missing.

imsnif:
Hey, thanks for your vigilance. I was intending to tag you on this issue.

The failure of these builds failed the entire release and caused me to have to delete the tag and redo it, introducing failed state window for the entire Zellij installation path for all systems. While this dependency was caused by an issue in our build pipeline that I already fixed, this is just to show how critical of a path it is that we're dealing with here.

Admittedly, I don't know a lot about riscv - but what I'm seeing is that a small dependency change on our part before the release caused this to break. Even if we mitigate this and similar issues as you suggest, I'm worried about similar stability issues in the future. Even if the only thing that will fail during release is the riscv build, we'll have already put ourselves in a position where we have users expecting us to build this binary, and we have maintainers chasing pieces of infrastructure that might not yet be very stable.

All this is to say: I'm not saying an outright "no" on having riscv builds in our release, but I will need to somehow get a higher confidence level in the stability of such builds. Could you maybe do a short investigation to find other projects, similar to Zellij (in the sense that they're pre 1.0 fast-moving, dependency-changing programs) that offer such builds? See what they're doing, find out whether they encountered similar problems and how they solved them? I'm normally happy to be a trailblazer in such things, but when the stove burns my fingers, I will be hesitant before cooking on it again.

gounthar:
Sorry about the disruption to the release — forcing a tag delete is exactly the kind of thing I was hoping to avoid, and I should have been more careful about testing the release path end-to-end before the PR landed.

Your concern is fair. I looked at a few comparable Rust TUI projects to answer your question honestly.

[bottom](https://github.com/ClementTsang/bottom) (~13k stars, active, pre-1.0 feel) does ship riscv64 binaries via cross in its release workflow. I checked the latest release: it ships `bottom_riscv64gc-unknown-linux-gnu.tar.gz`. However, that is a glibc build, not musl. I could not find a project similar to Zellij that ships musl+riscv64 static binaries — that combination is less common, and it is where the cross-rs toolchain gap bit us.

So to be straight with you: the musl requirement makes riscv64 harder here than it would be for a glibc build. The approach I had proposed (RISE native runner) would sidestep cross-rs entirely and avoid that class of problem, but it does introduce a different dependency on third-party runner infrastructure, which is a reasonable thing to be cautious about.

I am happy to keep investigating if it would help, or to leave this aside until the riscv64 ecosystem matures a bit more. No pressure either way.

Issue #4903, opened 2026-03-23
Tab-bar rendering lags ~400-700ms when using custom layouts (0.44.0 regression)


CVirus:
> Disclaimer: This bug was detected by a human programmer, but the report itself is AI-generated. I've reviewed it thoroughly, though my Zellij experience is limited. I don't want to waste open-source maintainers' time by asking you to review an AI-written report — so please let me know if more effort is needed from my side for you to look into this.

## Issue description

After upgrading from 0.43.1 to 0.44.0, switching tabs has a ~400-700ms delay before the tab-bar active indicator updates. The pane content renders instantly, only the tab-bar plugin rendering lags behind.

This ONLY occurs when using a custom layout file. A default session (no layout) with 10+ manually added tabs has no lag at all.

## Minimal reproduction

1. Save this as `test.kdl`:

    ```kdl
    layout {
        default_tab_template {
            pane size=1 borderless=true {
                plugin location="zellij:tab-bar"
            }
            children
            pane size=1 borderless=true {
                plugin location="zellij:status-bar"
            }
        }
        tab name="one" focus=true
        tab name="two"
    }
    ```

2. Start with a clean config to rule out config issues: `zellij --config-dir /tmp/zellij-clean -n test.kdl -s test`
3. Switch between tabs (Alt+Left / Alt+Right)
4. Observe: pane content switches instantly, tab-bar highlight updates ~400-700ms later

**Compare with:** `zellij -s default`, manually add 10+ tabs, switch between them — tab-bar updates instantly.

Also tested with a named `tab_template` instead of `default_tab_template` — same result. The issue is any custom layout vs no layout.

## Other relevant information

- **zellij version:** 0.44.0 (Homebrew)
- **OS:** macOS, Apple Silicon, Darwin 25.3.0
- **Not present in 0.43.1**
- Logs show `Loading plugins using Wasmi interpreter` — possibly related to the wasmtime to wasmi migration

imsnif:
Hey, I'm not reproducing this. A guess would be that something here wasn't compiled with `--release` one way or another?

Either that or this is mac-specific - maybe @tlinford can try to reproduce if he has a moment?

dastein1:
I'm also facing severe performance issues with zellij 0.44 on Mac. 
Switching tabs gets slow, and the session-manager gets very slow as well, to a point where I can hardly select any sessions.

imsnif:
@CVirus , @dastein1 - how did you install Zellij?

[...]

PR #4918, opened 2026-03-26
fix: performance regression when switching tabs

imsnif:
This fixes an issue reported in https://github.com/zellij-org/zellij/issues/4903 where in some systems switching tabs would lag in 0.44.0 compared to 0.43.1.

This was caused by the migration to wasmi from wasmtime exposing a long-standing performance issue where we would serialize an excessive amount of data in an often-sent `Event` to plugins.

This fixes it by:
1. Only sending `ModeUpdate`, `TabUpdate`, `PaneUpdate` and `InputReceived` `Event`s to plugins that are either in the active tab or are background plugins. Other plugins receive the relevant events once they become active.
2. Reducing the size of the `ModeUpdate` `Event`, namely removing the large keybindings `BTreeMap` from it in all instances that it's sent unless it has actually changed (eg. when the keybindings were reconfigured). To maintain backwards compatibility, we now allow plugins to register to the new `InitialKeybindings` `Event`. Subscribing to this `Event` signals to Zellij that the plugin caches its own keybindings and they do not need to be sent explicitly every `ModeUpdate` `Event`. Third-party plugins are encouraged to use this method so as not to hurt performance, but this is not a must.

An additional fix included in this PR is to only send pane content updates for panes that have actually changed. This was an initial attempt at mitigating this issue which was not effective enough, but is still worth including.

### DIFF
--- a/default-plugins/fixture-plugin-for-tests/src/main.rs
+++ b/default-plugins/fixture-plugin-for-tests/src/main.rs
@@ -67,8 +67,14 @@ impl ZellijPlugin for State {
             PermissionType::RunActionsAsUser,
             PermissionType::ReadSessionEnvironmentVariables,
         ]);
-        let should_subscribe_initial_keybinds = configuration.get("subscribe_initial_keybinds").map(|v| v == "true").unwrap_or(false);
-        let should_subscribe_mode_update = configuration.get("subscribe_mode_update").map(|v| v == "true").unwrap_or(false);
+        let should_subscribe_initial_keybinds = configuration
+            .get("subscribe_initial_keybinds")
+            .map(|v| v == "true")
+            .unwrap_or(false);
+        let should_subscribe_mode_update = configuration
+            .get("subscribe_mode_update")
+            .map(|v| v == "true")
+            .unwrap_or(false);
         self.configuration = configuration;
         subscribe(&[
             EventType::InputReceived,
--- a/src/tests/e2e/cases.rs
+++ b/src/tests/e2e/cases.rs
@@ -1148,7 +1148,8 @@ pub fn detach_and_attach_session() {
             name: "Wait for session to be attached",
             instruction: |remote_terminal: RemoteTerminal| -> bool {
                 let mut step_is_complete = false;
-                if remote_terminal.ctrl_plus_appears() && remote_terminal.cursor_position_is(77, 2)
+                if remote_terminal.ctrl_plus_appears()
+                    && remote_terminal.cursor_position_is(77, 2)
                     && !remote_terminal.snapshot_contains("Detach")
                 {
                     // we're back inside the session
--- a/zellij-server/src/plugins/unit/plugin_tests.rs
+++ b/zellij-server/src/plugins/unit/plugin_tests.rs
@@ -11,8 +11,8 @@ use zellij_utils::data::{
     BareKey, Event, InputMode, KeyWithModifier, ModeInfo, PermissionStatus, PermissionType,
     PluginCapabilities,
 };
-use zellij_utils::input::actions::Action;
 use zellij_utils::errors::ErrorContext;
+use zellij_utils::input::actions::Action;
 use zellij_utils::input::keybinds::Keybinds;
 use zellij_utils::input::layout::{
     Layout, PluginAlias, PluginUserConfiguration, RunPlugin, RunPluginLocation, RunPluginOrAlias,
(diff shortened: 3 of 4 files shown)


Zykino:
>     2. Reducing the size of the `ModeUpdate` `Event`, namely removing the large keybindings `BTreeMap` from it in all instances that it's sent unless it has actually changed (eg. when the keybindings were reconfigured). To maintain backwards compatibility, we now allow plugins to register to the new `InitialKeybindings` `Event`. Subscribing to this `Event` signals to Zellij that the plugin caches its own keybindings and they do not need to be sent explicitly every `ModeUpdate` `Event`. Third-party plugins are encouraged to use this method so as not to hurt performance, but this is not a must.

Either I don't understand this, or it seems backward to me: if my plugin does not care about key-bindings, I should subscribe to `InitialKeybindings` to prevent bad performances.

I understand the API breakage issue…  
Maybe deprecate current `ModeUpdate` (to remove it next version 0.45) and provide a new `CoreConfigUpdate` and separate `Keybindings` Event?  
Or ask the dev to subscribe to `KeybindingsChanges` to have them populated in the current `ModeUpdate`?

imsnif:
> Either I don't understand this, or it seems backward to me: if my plugin does not care about key-bindings, I should subscribe to `InitialKeybindings` to prevent bad performances.

Yeah, it's a bit involved. `ModeUpdate` is kind of a generic catch-all `Event` that's been with us since the tool's inception. It is an unfortunate design choice that was written into the core of both our internal logic and the plugin API. It includes all the Keybindings as well as lots of other stuff.

Since we don't (yet?) have a mechanism of deprecating plugin APIs, we have to keep supporting it both in the serialization protobuf contract and in its behavior. Meaning that plugins expecting `ModeUpdate` to always contains the keybindings should still receive them. To get around this, I created a new `Event` - subscribing to which signals to Zellij that instead of sending the keybindings through `ModeUpdate`, it should send them through the `InitialKeybindings` `Event` and only do so when they changed or on plugin load.

This gets around the performance problem without breaking the plugin API.

To fix this and other problems, I want to do the same thing done in this version with the client-server API: create a new contract and deprecate the old one. But since this will break all old plugins (both the already compiled wasm blobs and their source code), I want to do this deliberately - fix all of the design problems we have that stem from the plugin API and formalize this for future breaking changes.

I didn't want the performance problem to depend on this though.

Makes sense?

imsnif:
I'm going to merge this. If you have concerns, please let me know and we'll figure it out.

Zykino:
Makes sense 👍 .

It could be nice to annotate things we want to change the API but can't for now. I know Qt added this sort of comments (`// ### Qt 6: <explanation>`) in their codebase to find and fix them all on their next major release (whenever it happens):

```c++
     // ### Qt 6: make this constructor constexpr, after the destructor is made trivial
    /* … */

#if QT_VERSION < QT_VERSION_CHECK(6,0,0)
    // ### Qt 6: remove all of these, the implicit ones are fine
    /* … */
#endif // Qt < 6.0.0
```

They also added deprecation "feature-gate" which can help peoples compile and disable the deprecated feature ahead of time to check if their code is ready for the breaking version:
```c++
#if QT_DEPRECATED_SINCE(5, 15)
    // ### Qt 6: replace with Qt:: version
    /* … */
#endif // 5.15 deprecations
```

I think having some sort of greppable comments will help to find every occurrences of breakage and do them all at once without costing much.

When you are ready, having a way to disable deprecated APIs may let us compile & run the "awesome list of plugin" against this target and estimate if the changes will break a lot of plugins or not. Maybe even proactively warn them of the deprecation.

imsnif:
Sounds like a cool idea. I'm not really sure how to go about it though. Most of these things are usually in my head until I sit down and write a plan for it.

Zykino:
I usually try to let `TODO` comments in my code.

At work I had to code a legacy feature in a new code to then use it as basis for a new one. Since the original feature was never used in the first place, it was never tested and not in a "known working" state.
So to make the new dev, I placed "VAMPIRISM feature-name" comments all over the code each time I knew "there I'm taking on the code too much, I'm sure it will be an issue if we want to use the old feature". Since the base feature wasn't tested, keeping its code intact and doing some interfaces would have been wrong.
Now (5 years later) I don't remember exactly how the 2 features are intricate, and we may have a client wanting the old feature. I'm on another part of the code but I told a new hire: "if they confirm they want the old feature, you should check all instance of my comment and check how to properly dissociate the old and new feature".

All that to say: at the very least, don't hesitate to add little comments all over the place as they come. Maybe in the form: "Deprecate: blah would be better now". That way down the line searching for "Deprecate:" will bring all the places without missing one not seen from a long time.

Issue #4423, opened 2025-09-15
Web client on android typing is broken


HKalbasi:
Everything gets typed twice. I saw in reddit that you are aware of this issue, just wanted to have some github issue to track the progress.

By the way, thank you for zellij. The web feature is really nice and can become a game changer for me if its mobile version becomes usable.

hkalbasi:
An idea: zellij on termux works good, so a way to connect to a zellij web server by a zellij terminal client instead of a browser would fix this issue somehow.

OwenTrokeBillard:
I, too, would love some kind of mobile support. The main reason I installed Zellij web is so I can check in on my phone.

hamannju:
I tried to get Zellij Web to run on an Ubuntu 24.04 host and connect via my Android tablet with a bluetooth keyboard and ran into this issue. I had Codex hack away at it for a few hours on the side and it seems that it actually managed to fix it. I had it draft a comment + PR. 

The issue seems to have been two-fold: The upstream xterm has a fix that fixes the characters appearing twice on Android, so updating xterm fixed that part. Then there is another issue with the input being sent multiple times upon pressing "enter" and for this codex created a special handler that is triggered on android and prevents this from happening. I have used these fixes for about an hour now and it works and I have not noticed any regressions yet.

------------------------- BEGIN AI SLOP -------------------------

I reproduced this on `main` with Firefox on Android and a Bluetooth keyboard.

What I observed locally:

- With the current vendored xterm bundle, characters could duplicate while typing in the web client on Android.
- After updating the vendored xterm assets, the immediate per-keystroke duplication stopped, but there was still a second bug: typing `clear` looked normal, then pressing `Enter` replayed the word multiple times.

I instrumented the web client and saw that on Android + external keyboard the input path goes through IME composition rather than normal key events:

- `key=Process`
- `compositionupdate data=c`
- `compositionupdate data=cl`
- `compositionupdate data=cle`
- ...

Then on submit:

- the composed text was still present in xterm's hidden textarea
- `Enter` would commit/replay that buffered text before the newline

I also found that `zellij-client/assets/index.js` currently calls `setupInputHandlers(term, sendAnsiKey)` twice, once before the websocket send function is real and once after. That duplicates the client-side input handlers and is worth fixing independently.

The local fix that resolved Android typing for me was:

1. Remove the duplicate `setupInputHandlers(...)` registration in `zellij-client/assets/index.js`
2. Update the vendored xterm web assets
[...]

PR #4922, opened 2026-03-26
Fix Android web client typing with external keyboards

hamannju:
## Summary

Fix Android web-client typing in two parts:

1. remove the duplicate web input-handler registration
2. refresh the vendored xterm web assets and handle Android `Enter` more defensively

## Root Cause

I reproduced issue #4423 on Firefox for Android with a Bluetooth keyboard.

There were two separate problems:

- `zellij-client/assets/index.js` called `setupInputHandlers(term, sendAnsiKey)` twice, which duplicated the web-client input handlers
- on Android, the external keyboard path goes through IME composition (`key=Process` / `compositionupdate`) in xterm's hidden textarea, and `Enter` could replay the buffered composed text before the newline

On top of that, the currently vendored xterm bundle is older and still exhibits Android physical-keyboard problems that improved after updating the vendored assets.

### Commit 1

- remove the extra `setupInputHandlers(...)` call in `zellij-client/assets/index.js`
- on Android user agents, intercept plain `Enter` in `zellij-client/assets/input.js`
- clear `term.textarea.value`
- send a single `\"\\r\"` directly

### Commit 2

- update the vendored xterm frontend assets:
  - `xterm.js`
  - `xterm.css`
  - `addon-fit.js`
  - `addon-clipboard.js`
  - `addon-web-links.js`
  - `addon-webgl.js`

## Repro

1. Open Zellij Web in Firefox on Android
2. Connect a Bluetooth keyboard
3. Type `clear`
4. Press `Enter`

Before:

- text could duplicate while typing
- after partial fixes, pressing `Enter` still replayed the composed word multiple times

After:

- typing is normal
- `Enter` submits exactly once

## Notes

- This is scoped to the Android web-client path only
- Desktop browser behavior is unchanged
- The Android `Enter` handling is deliberately narrow: only plain `Enter` on Android user agents is intercepted


tlinford:
Hey, this could be a good fix, but I think we should find a better way of handling the vendored xterm code. I think it's a bit difficult to accept blobs of minified js like this.

What version of xterm did you use here?

hamannju:
Hi,

I am currently not able to reproduce this because I use that zellij instance as my production setup. Please don't waste any time on this right now, because the pull request is AI generated and I am not 100% sure that the xterm.js asset update was even required. I will try to properly reproduce this next week and then determine what change actually fixed the issue.

PR #4931, opened 2026-03-28
Support OSC-99 desktop notifications

imsnif:
This adds support for OSC99, which means Zellij will now forward desktop notifications that arrive from applications running inside panes.

Furthermore, when notifications are clicked, Zellij will also focus the relevant pane that sent the notification. If the pane requested a response, Zellij will also forward the response to the pane as well.

To support this, we're performing a minor rewrite of these notifications, namely:
1. We namespace the notifications to make sure they're unique across a Zellij session (adding the pane id to them) - we strip this when sending the notification back to the pane (stripping the `i` id field entirely if it wasn't originally sent).
2. We unconditionally add the `report` directive to the notifications so that the response will always be sent to Zellij regardless. We do this so that we can be sure to focus the originating pane. If the application did not itself request a report, we drop it.

Instructions querying support for OSC-99 are sent to the host terminal (with the same namespacing override described in `1` above) and returned to the requesting pane.


yinaoxiong:
So excited for this feature to be added!

PR #4941, opened 2026-03-29
Preserve scrollback for partial scroll regions starting at top

nitingupta910:
When a partial scroll region starts at row 0 (the top of the viewport), lines scrolled off the top were silently discarded via viewport.remove(0) instead of being transferred to lines_above (the scrollback buffer). This affected both add_canonical_line and rotate_scroll_region_down.

Applications like Codex CLI use partial scroll regions to push rendered content into terminal scrollback (via DECSTBM + newlines). In tmux, these lines are preserved in the scrollback buffer and accessible via copy-mode scrolling. In Zellij, they were lost, causing mouse wheel scrolling to show only stale pre-application content.

Transfer lines to lines_above when scroll_region_top == 0 and alternate screen is not active, matching the behavior of tmux.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 * fix: compilation warnings in windows (https://github.com/zellij-org/zellij/pull/4890)
 * feat: add --layout-string to allow for in-line layouts, add --tab-id to all new-pane CLI commands (https://github.com/zellij-org/zellij/pull/4956)
 * fix: proper error on socket paths that are too long (https://github.com/zellij-org/zellij/pull/4957)
+* fix: scrollback with partial scroll region (eg. Codex bug) (https://github.com/zellij-org/zellij/pull/4941)
 
 ## [0.44.0] - 2026-03-23
 * fix: meta key handling in web client (https://github.com/zellij-org/zellij/pull/4376)


imsnif:
Hey @nitingupta910 - does this actually solve the issue? Last I checked the issue happened because Codex were relying on the off-spec scrollback in alternate screen.

I'm guessing they fixed the former and are now hitting this? (this btw is also off-spec behavior, but one that I'd be willing to accept considering it's not directly opposed to the spec - I just want to make sure it actually solves the problem).

nitingupta910:
Yes it does solve the problem. I'm running zellij from this branch since a few days now with codex.

imsnif:
Hey @nitingupta910 - much thanks for pointing this out. Had I been aware that Codex no longer relied on scrollback in the alternate screen and that this was their only current issue, I'd have happily issued this fix much earlier.

As I mentioned above, this new behavior is a gray area (some emulators behave like this, others don't), but since it's not explicitly against the spec as the previous Codex behavior was, I'm willing to accept this change.

I added handling for mouse selection (eg. when the user is trying to select text that scrolls upwards) and also for `CSI M` which was missing. I also added tests. Will merge once CI passes.

JiehoonKwak:
This PR fixes codex scrollback issue, but same issue occurs after applying this PR in Claude Code (v 2.1.87 or 2.1.89) - Wondering is anyone has similar issues

nitingupta910:
@JiehoonKwak I tested claude 2.1.89 today and the scrollback is fine within zellij (running ng/codex_scroll_fix branch). The current main branch has more changes than this branch, so maybe a new regression in the main branch?

imsnif:
I did a brief test in the current `main` and I also can't reproduce. Seems like there are a lot of scrollback issues in the Claude repo recently, so might be something particular on their side.

JiehoonKwak:
Thanks for kind reply. Maybe related to recent Claude Code v2.1.89 ink rendering. In my case, it only happens within zellij but it seems like many people are experiencing various scrollback issues not even inside zellij. I cannot live without zellij so i'm falling back to claude code v2.1.87 and it seems to solve issue for now

PR #4942, opened 2026-03-29
fix: Fix package upgrade lost sessions (#3420)

### LINKED ISSUE #3420: Fix package upgrade lost sessions
**Basic information**

`zellij --version`: `zellij 0.40.1`

`stty size`: `66 104`

`uname -av` or `ver`(Windows): `Linux saturn 6.1.70 #4 SMP PREEMPT_DYNAMIC Tue Jan  2 17:33:33 UTC 2024 x86_64 Intel(R) Xeon(R) CPU E5-2650 0 @ 2.00GHz GenuineIntel GNU/Linux`

## Further information
Regular Gentoo updates migrated Zellij from 0.39.2 to 0.40.1.

When I ssh'd back in to the server, I expected `ssh -t saturn zellij a session_name` to re-attach to the existing session.  The existing session was started under 0.39.2.  No session was reported to exist, despite `ps aux` showing otherwise.

I found existing issue #3371 , with similar behavior.  I attempted to follow the instructions listed as a temporary workaround.  However, I was greeted with some form of `empty message received` error from the server.  I've since reconstructed the session since I needed to keep working.

## Other relevant information

I understand the client - server protocol, as well as the on-disk format, are subject to change.  However, this is a huge PITA.  Definitely kicks me out of flow state. :-(

## Proposed solution

Could we add a `zellij upgrade-session 0.39.2/session_name` or similar to cleanly shut down the old session, migrate it, and spawn a new session with the new version of the server process?  That would be a **huge** help.

Thanks!

krapcys1-maker:
Fixes #3420

## Summary

Fixes issue #3420 in zellij-org/zellij.

## Problem

Fix package upgrade lost sessions

## What changed

- zellij-utils/src/sessions.rs

## Solution

Apply focused code changes in `zellij-utils/src/sessions.rs` to address the reported behavior.

## Why

Keeps behavior aligned with issue expectations while preserving existing flow and validation checks.

## Tests

- `env CARGO_TARGET_DIR=target cargo xtask build --no-web`

### DIFF
--- a/zellij-utils/src/sessions.rs
+++ b/zellij-utils/src/sessions.rs
@@ -5,8 +5,10 @@ use crate::{
     },
     envs,
     input::layout::Layout,
-    ipc::{ClientToServerMsg, IpcReceiverWithContext, IpcSenderWithContext, ServerToClientMsg},
+    ipc::{ClientToServerMsg, IpcSenderWithContext},
 };
+#[cfg(windows)]
+use crate::ipc::{ClientToServerMsg, IpcReceiverWithContext, ServerToClientMsg};
 use anyhow;
 use humantime::format_duration;
 use std::collections::HashMap;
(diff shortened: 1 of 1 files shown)


Zykino:
What are you trying to do here? Accepting whatever the server sends even if the client does not have a clue what it is? It is all gonna end up well…

The work for what you want has been done in #4439, it just needed an official version to prepare the upgradability of session. The version including this is 0.44.0, released 23/03/2026 (less than a week ago). So we can all verify that it works when 0.44.1 will land (once early reports and fixes lands).

krapcys1-maker:
Thanks for the callout — you are right.

I re-checked this against #4439 and the 0.44.0 forward-compatibility work.
My patch in #4942 is too broad (it treats any connectable socket as alive), which is weaker than the protocol-contract approach introduced in #4439.

So this PR is not the right direction. I’ll drop this patch and keep validation focused on the official upgrade path (0.44.0 -> 0.44.1) with regression evidence instead of weakening socket probe semantics.

Thanks for catching this.

PR #4920, opened 2026-03-26
fix: support Unicode combining characters (Thai, Arabic, Hebrew, etc.)

Faizalj:
## Summary
Fix rendering of scripts that rely on Unicode combining marks — currently **all zero-width characters are dropped**, making many languages unreadable in zellij.

### Languages fixed
- 🇹🇭 **Thai** — vowels (◌ิ ◌ี ◌ุ ◌ู), tone marks (◌่ ◌้ ◌๊ ◌๋)
- 🇸🇦 **Arabic** — diacritics/harakat (◌َ ◌ُ ◌ِ ◌ّ ◌ْ)
- 🇮🇱 **Hebrew** — nikkud vowel points (◌ַ ◌ָ ◌ִ ◌ֵ ◌ֶ)
- 🇮🇳 **Hindi/Devanagari** — matras and virama (◌ा ◌ि ◌ी ◌ु ◌ू ◌्)
- 🇮🇳 **Bengali** — vowel signs and virama
- 🇱🇦 **Lao** — vowels and tone marks
- 🌍 **Latin diacritics** — combining accents (◌̀ ◌́ ◌̂ ◌̃)
- **Syriac, Cyrillic** combining marks
- **Variation Selectors Supplement** (emoji sequences)

## Problem
Zellij drops all zero-width Unicode codepoints (see #1538). This includes combining characters that are essential for rendering many world scripts. These characters modify the preceding base character (e.g., adding a vowel above/below a consonant) and must be preserved.

Example: Thai "สวัสดี" (sawasdee) → rendered as "สวสด" with vowels and tone marks stripped.

## Changes
- **`terminal_character.rs`**: Added `combining_chars: Option<Box<Vec<char>>>` to `TerminalCharacter`, with `append_combining()` method
- **`grid.rs`**: Added `is_combining_character()` function covering all major Unicode combining mark ranges; modified `add_character()` to attach combining marks to preceding cell instead of dropping them
- **`output/mod.rs`**: Modified both `serialize_chunks` and `serialize_chunks_with_newlines` to output combining characters after the base character

## Test plan
- [x] Built and tested on macOS (aarch64) with Thai text input
- [x] Verified Thai vowels and tone marks display correctly
- [x] Verified existing non-combining character behavior unchanged
- [x] Covers Unicode ranges: Combining Diacritical Marks, Thai, Lao, Arabic, Hebrew, Syriac, Devanagari, Bengali, Cyrillic, and more

Relates to: #1538

PR #4943, opened 2026-03-29
fix(ime): correct cursor position for IME during Japanese input in Claude Code

jedipunkz:
## Summary
- Fixes an issue where intermediate conversion characters are misaligned when inputting Japanese text in Claude Code

## Problem
- When typing Japanese text using IME in Claude Code, the conversion characters (変換中の文字) appear misaligned from the input position
- This happens because the cursor position used for IME positioning was being captured at the wrong time during the BSU/ESU render cycle

## What Changed
- Modified `cursor_position_for_ime()` in `zellij-server/src/panes/grid.rs` to use the current cursor position with bounds checking instead of the ESU-time saved position
- Removed the `cursor_position_at_last_esu` field and its associated save logic in `unlock_renders()`

## Solution
- Uses the current cursor position with bounds checking (`if self.cursor.x < self.width && self.cursor.y < self.height`) to ensure the IME always receives a valid cursor location for displaying conversion candidates
- This ensures the cursor position is correctly reported to the OS IME at the time of the query, rather than relying on a stale cached position

## Why
- When the terminal hides the cursor during BSU/ESU render cycles (like in ink/React terminal UIs), by the time zellij's background render fires, the cursor may appear at an incorrect position even though its logical position is valid
- The previous approach of saving the cursor position at ESU time was intended to solve this, but it introduced the opposite problem where the position could become stale or incorrect
- Using the current position directly ensures the IME receives the most up-to-date and accurate cursor location

### DIFF
--- a/zellij-server/src/panes/grid.rs
+++ b/zellij-server/src/panes/grid.rs
@@ -332,7 +332,6 @@ pub struct Grid {
     pub changed_colors: Option<[Option<AnsiCode>; 256]>,
     pub should_render: bool,
     pub lock_renders: bool,
-    cursor_position_at_last_esu: Option<(usize, usize)>, // saved at ESU (End Synchronized Update)
     pub cursor_key_mode: bool, // DECCKM - when set, cursor keys should send ANSI direction codes (eg. "OD") instead of the arrow keys (eg. "[D")
     pub bracketed_paste_mode: bool, // when set, paste instructions to the terminal should be escaped with a special sequence
     pub erasure_mode: bool,         // ERM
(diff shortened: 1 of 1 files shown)


Zykino:
Hello,

I'm a bit of an outsider considering Unicode graphemes. One of the zellij members drafted this #4800 which hope to fix all this kinds of issues. Can you checkout their branch and give them feedbacks on it?

[Their exact reply in another similar PR 3 days ago](https://github.com/zellij-org/zellij/pull/4920#issuecomment-4135497627):
> Author of #4800 here, would really appreciate if you could give it a try and let me know if it fixes your issues as well :).
> 
> EDIT: just for some context, I'm hoping to wrap this work up in the next weeks, and as you can see in the changes it's more comprehensive than handling combining marks.

jedipunkz:
@Zykino 
Hello,

I checked out PR: https://github.com/zellij-org/zellij/pull/4800 branch and built and tested it, but the characters being converted during Japanese input in Cloud Code appeared outside the terminal window.

Therefore, I may need to modify my own PR.

tlinford:
#4800 is about handling unicode graphemes properly, this IME issue might be related but it's probably not covered yet.

Issue #4878, opened 2026-03-20
Single character IME confirmation does not work on macOS (Japanese input)


Okiichi1122:
## Summary

When using Japanese IME (macOS built-in) in Zellij, confirming a single character input with **Control+M** does not work. The character remains in the pre-edit (uncommitted) state and is not confirmed. 

**Important distinction:** Pressing the **Enter key** works correctly for single character confirmation. Only **Control+M** fails, even though they should send the same keycode (0x0D). This suggests Zellij may be handling Control+M differently from Enter when IME is in composing state.

Inputting two or more characters and pressing Control+M also works fine — only the single character case fails.

## Environment

- **OS:** macOS (Darwin 25.3.0, Apple Silicon)
- **Terminal emulator:** Ghostty
- **Zellij version:** 0.43.1 (installed via Homebrew)
- **IME:** macOS built-in Japanese input (Hiragana)
- **Shell:** zsh
- **Application running inside Zellij:** Claude Code (Node.js CLI app)

## Steps to reproduce

1. Launch Ghostty with Zellij
2. Open any pane running an interactive CLI app (e.g. Claude Code)
3. Switch to Japanese IME (Hiragana mode)
4. Type a single character, e.g. "あ" (press "a" key)
5. Press **Control+M** to confirm/commit the character

**Expected:** The character "あ" is confirmed and submitted as input (same behavior as pressing Enter key).  
**Actual:** The character is not confirmed. It stays in the pre-edit/composing state, or disappears.

## Additional observations

- Pressing the **Enter key** to confirm a single character works correctly.
- **Control+M** fails only for single character confirmation — two or more characters work fine with Control+M.
- This issue does **not** occur when running the same CLI app directly in a terminal (without Zellij).
- Setting `support_kitty_keyboard_protocol false` in Zellij config does **not** resolve the issue.

## Current workaround

Type an extra character (e.g. "あ1"), press Enter to confirm, then press Backspace to delete the extra character. This is functional but tedious.

## Possibly related

- #4494 (Chinese input duplication in web mode) — different issue but also CJK IME related

PR #4951, opened 2026-03-30
fix: position host cursor for CJK IME even when app hides cursor

satoruhiga:
## Summary

When apps like Claude Code (which uses the ink/React library) run inside Zellij,
IME composition candidates appear at the wrong position. This makes CJK input
(Japanese, Chinese, Korean) effectively broken in these apps.

The same apps work correctly in a bare terminal or under tmux.

## Root cause

ink-based TUIs hide the cursor (`CSI ? 25 l`) during each render cycle, then
show it again and reposition it after rendering. Zellij's `render_cursor()`
treats a hidden cursor as "no cursor position to report" — it sends
`ESC[?25l` to the host terminal but does **not** send a CUP sequence to
position it. After a content render (which moves the cursor around to draw
character chunks), the host terminal's cursor ends up at an arbitrary position.

The host terminal's IME system reads the cursor position to place the
composition window, so it appears in the wrong place.

## Fix

In the `None` branch of `render_cursor()` (cursor hidden by app), still send a
CUP escape sequence to position the host terminal cursor at the correct
location, then hide it. The IME subsystem reads the cursor position regardless
of visibility.

Changes:
- Add `cursor_position_for_ime()` to `Grid` and `Pane` trait — returns cursor
  position ignoring `cursor_is_hidden`
- Add `is_cursor_not_occluded()` to `Output` — floating pane visibility check
  without the side effect of `cursor_is_visible()` (which writes to
  `cursor_coordinates` and would corrupt the crop check in `serialize()`)
- Add `ime_cursor_positions` cache to `Tab` — change detection to avoid
  sending CUP every frame
- Send hide-cursor before CUP to prevent momentary cursor flash

## Testing

https://github.com/user-attachments/assets/9fe1a121-6f9b-44f3-8b74-df0396f97c05

Tested on macOS (Apple Silicon, Darwin 25.3.0) with Ghostty + Zellij.

IME input confirmed working:
- Claude Code (v2.1.93) with macOS default Japanese IME and macSKK
- nvim — no regression
- zsh — no regression

## Related issues

- zellij-org/zellij#4878 (Single character IME confirmation on macOS)
- anthropics/claude-code#1547 (IME input causes issues — multiple CJK users
  report Zellij-specific breakage in comments)

---

This change was investigated and implemented with the help of Claude Code.

### DIFF
--- a/src/tests/e2e/remote_runner.rs
+++ b/src/tests/e2e/remote_runner.rs
@@ -337,8 +337,10 @@ fn read_from_channel(
                         Ok(0) => {
                             let current_snapshot = take_snapshot(&mut terminal_output);
                             let mut last_snapshot = last_snapshot.lock().unwrap();
-                            *cursor_coordinates.lock().unwrap() =
-                                terminal_output.cursor_coordinates().map(|(x, y, _)| (x, y)).unwrap_or((0, 0));
+                            *cursor_coordinates.lock().unwrap() = terminal_output
+                                .cursor_coordinates()
+                                .map(|(x, y, _)| (x, y))
+                                .unwrap_or((0, 0));
                             *last_snapshot = current_snapshot;
                             should_sleep = true;
                         },
@@ -348,8 +350,11 @@ fn read_from_channel(
                             }
                             let current_snapshot = take_snapshot(&mut terminal_output);
                             let mut last_snapshot = last_snapshot.lock().unwrap();
-                            *cursor_coordinates.lock().unwrap() =
-                                terminal_output.grid.cursor_coordinates().map(|(x, y, _)| (x, y)).unwrap_or((0, 0));
+                            *cursor_coordinates.lock().unwrap() = terminal_output
+                                .grid
+                                .cursor_coordinates()
+                                .map(|(x, y, _)| (x, y))
+                                .unwrap_or((0, 0));
                             *last_snapshot = current_snapshot;
                             should_sleep = true;
                         },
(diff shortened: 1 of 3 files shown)


satoruhiga:
Confirmed working on Windows as well.

Tested with Windows 11 + Windows Terminal + Zellij (this branch) + Claude Code.
The IME cursor positioning issue was also occurring on Windows, and this fix resolves it.

JiehoonKwak:
I also tested in macOS 26.4 with ghostty, and it works great for me! Hope this PR to me merged soon

imsnif:
@tlinford - will this be solved by grapheme segmentation work, or is it unrelated?

tlinford:
> @tlinford - will this be solved by grapheme segmentation work, or is it unrelated?

this is a different issue

imsnif:
Hey @satoruhiga - thanks for this. I'd be happy to merge this, but I'd like to ask for some changes:
1. While I don't mind code written by LLMs (with the usual "as long as it's not copyrighted" caveat), Claude Code is a commercial product. I don't want to give a commercial company free ad-space in our commit logs. So please remove the commits from Claude Code and change them to be under your name. Thanks for understanding.
2. Regarding the technical aspect: it seems here that we're going through a second path during render in the case where we receive None from an occluded cursor. Instead of doing that, could we change the functions returning cursor coordinates to return them unconditionally (assuming the cursor is in the pane, that is) and then also return a bool of whether the cursor is shown or hidden?

Let me know if you have any questions or if I wasn't clear enough.

satoruhiga:
@imsnif Thanks for the review!

1. Removed the Co-Authored-By line.

2. Refactored as suggested: `cursor_coordinates()` now returns the position
   unconditionally (within bounds) along with a `bool` for visibility.
   `render_cursor()` uses this to always position the host cursor via CUP,
   even when the app has hidden it.

   All the IME-specific code from the previous revision
   (`cursor_position_for_ime`, `is_cursor_not_occluded`, `ime_cursor_positions`)
   has been removed — the logic is now integrated into the existing path.

   Test files are updated to match the new return type. Existing test behavior
   is preserved (snapshots and plugin API only use coordinates when cursor is
   visible).

Tested on macOS — IME still works correctly. `cargo test -p zellij-server`
passes (938 tests).

imsnif:
Thanks for the changes! I added a test and ran `rustfmt`. Once CI passes I'll merge.

satoruhiga:
Thank you for the quick review and merge, @imsnif!

Zellij is an amazing piece of software — this fix will make a lot of
CJK developers very happy. Appreciate all the work you and the team
put into it.

Issue #4930, opened 2026-03-27
[Windows] cross-login session visibility


divens:
## Issue description

When the same Windows user creates a Zellij session from one login context (e.g., SSH) and lists sessions from a different login context (e.g., interactive desktop / RDP), `list-sessions` incorrectly treats the session as stale and deletes its marker file. The Zellij server process continues running but becomes invisible and unattachable from any context, effectively a zombie session.

##  Root cause

`assert_socket` in `sessions.rs` calls `OpenProcess(PROCESS_QUERY_LIMITED_INFORMATION)` to check if a session's server process is alive. On Windows, processes spawned in different logon sessions have different security tokens. `OpenProcess` can return `ERROR_ACCESS_DENIED` for a live process when the caller's logon session differs from the target's. The current code treats any `OpenProcess` failure as "process is dead" and deletes the marker file.

## Minimal reproduction

  1. Enable OpenSSH server on your Windows machine
  2. SSH into your own machine: `ssh youruser@localhost`
  3. In the SSH session, create and detach a Zellij session: `zellij -s test-session` (Ctrl+O, D to detach)
  4. From SSH, verify the session is listed: `zellij list-sessions` → shows `test-session`
  5. From the interactive desktop (e.g., Windows Terminal), list sessions: `zellij list-sessions`
  6. Result: `test-session` does not appear. The marker file has been deleted.
  7. From SSH, list sessions again: `zellij list-sessions` → session is gone
  8. The server process is still running (`tasklist | Select-String "zellij"`) but is now unreachable

## How to fix
Imrove error handling on ` OpenProcess()` and display "foreign" sessions as such.
**PR incoming**

laciferin2024:
Additionally @divens 

In SSH session, 

Detaching : kills the session !

But still lists the session, so upon attaching to the session and then switching to RDP to list the session. Session magically appears back in the list but guess what attaching results in infinite hang.

divens:
@laciferin2024 I have not run into this issue. Can you give a step by step repro please?

laciferin2024:
Sure , 


Try the same reproducer of yours , but stay in the session (wth ssh) and try list the session in RDP (its there) but u can't attach.

PR #4954, opened 2026-03-30
fix(windows): allow session access across logon sessions

divens:
On Windows, sessions created in one logon context (e.g., SSH) could not be listed or attached from a different logon context (e.g., interactive desktop) for the same user. Two issues:

1. assert_socket treated OpenProcess ERROR_ACCESS_DENIED as "process dead" and deleted the marker file, orphaning the session. Now checks GetLastError to distinguish access denied (alive, different logon session) from invalid parameter (genuinely dead).

2. Named pipes inherited the creating logon session's default DACL, blocking cross-session connections. Now sets an explicit DACL granting access to the current user's SID, allowing the same user to connect from any logon session without exposing pipes to other users.

Fixes https://github.com/zellij-org/zellij/issues/4930

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4996,6 +4996,7 @@ dependencies = [
  "unicode-width 0.1.10",
  "url",
  "uuid",
+ "widestring",
  "winapi",
  "windows-sys 0.52.0",
 ]
--- a/zellij-utils/Cargo.toml
+++ b/zellij-utils/Cargo.toml
@@ -62,10 +62,13 @@ nix = { workspace = true }
 crossterm = { workspace = true, features = ["events"] }
 windows-sys = { workspace = true, features = [
     "Win32_Foundation",
+    "Win32_Security",
+    "Win32_Security_Authorization",
     "Win32_System_Threading",
 ] }
 winapi = { version = "0.3.9", default-features = false }
 dunce = "1.0"
+widestring = "1.0"
 
 [dev-dependencies]
 insta = { version = "1.6.0", features = ["backtrace"] }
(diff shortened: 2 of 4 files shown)


divens:
@imsnif Can you have a quick look please? Also, the E2E test failed for unrelated reasons. Same pipeline passed 5 minutes before with sames changes except format fix.

imsnif:
Hey, I re-ran the e2e test. I don't know a lot about windows, but any block that includes an `unsafe` and includes a seemingly involved way of figuring out a specific error makes me suspicious that there might be a more straightforward way of doing this. But maybe there isn't.

divens:
@imsnif So I took some time to do more research on this. There's no shortcut (that I could find) to get the current user identity and craft an adequate permission to allow multiple logon sessions of a user accessing the same pipe.  We have to go through  `OpenProcessToken` → `GetTokenInformation` → `ConvertSidToStringSidW`. Each step crosses FFI boundary, uses out-parameters and raw buffers cast to `TOKEN_USER`, and requires manual `LocalFree` / `CloseHandle`.

imsnif:
Alright. Thanks for looking into this. This still seems very brittle to me. Are there no other applications that require this sort of functionality? Are we the first?

divens:
@imsnif I understand your concerns. I would say that our use-case is certainly not mainstream in the general application development space, but within our userbase, we've seen a number of people setting up zellij sessions and accessing them by mixing ssh and RDP or interactive desktop sessions. 

I see two ways we can move forward :
- we can consider this usecase legitimate and try to support it the best way we can given the platform specifics
- we can also document it as a limitation and not support it until we find a better solution (haven't found one yet, but it doesnt mean it doesnt exist). In this case we should still at a minimum keep the first part of this PR which recognizes the permission issues and avoid orphaning sessions when such issues occur.

Let me know how you want to proceed.

imsnif:
Is there some sort of manual operation the user can do in such a case to fix it? Something similar to the unix `chmod` and the like?

laciferin2024:
@divens 

What if we leverage user group like docker/podman : create a user group `zellij-users` ?

> Docker Desktop configures the ACL on its named pipe/service so that members of the docker-users group (and admins) can access it, instead of granting access only to a single user account.

Issue #4961, opened 2026-03-30
Ansi code for resetting bold should not reset underline


Kazimazi:
# 1. Graphical issue inside a terminal pane (eg. something does not look as it should)

I skip attaching logs, because it's a relatively simple issue.

**Basic information**

`zellij --version`: zellij 0.44.0

`stty size`: 63 280

`uname -av` or `ver`(Windows): Linux ryangosling 6.19.9 #1-NixOS SMP PREEMPT_DYNAMIC Thu Mar 19 15:15:33 UTC 2026 x86_64 GNU/Linux

## Further information

Image with the issue (underline breaks after bold char):

<img width="1736" height="234" alt="Image" src="https://github.com/user-attachments/assets/080edc0d-6394-4ac5-8791-0c587589440d" />

How it looks like without zellij (or in tmux):

<img width="1607" height="227" alt="Image" src="https://github.com/user-attachments/assets/ca1e7fef-2fa6-4bfa-a964-53c612d4815d" />

## Issue description
This issue had been grinding my gears for a long while. I first noticed it in neovim, that an underlined line was broken. It happens when the line has bold character as well.

## Minimal reproduction
This is a somewhat minimal, yet descriptive snippet, to show off the issue:
```sh
PREFIX="\x1b"
BOLD="$PREFIX[1m"
UNDLN="$PREFIX[4m"
BOLD_RES="$PREFIX[22m"
printf "${UNDLN}${BOLD}{${BOLD_RES} should be underlined"

# or as a 1liner
printf "\x1b[4m\x1b[1m{\x1b[22m should be underlined"
```
The " should be underlined" text is not underlined, but it should be.

## Other relevant information

I'm also opening a PR with a fix.

PR #4962, opened 2026-03-30
fix(ansi): reset bold should not entail reset underline

Kazimazi:
Fixes: https://github.com/zellij-org/zellij/issues/4961

According to [ANSI Escape Sequences cheatsheet](https://gist.github.com/ConnerWill/d4b6c776b509add763e17f9f113fd25b#colors--graphics-mode):

- the escape code for reset bold is `ESC[22m` and shouldn't entail `ESC[24m` reset underline
- my change fixes that behavior
- fixing that those comments are not needed.

I tested my change. Maybe I also need to adjust unit tests?


imsnif:
Good catch! Thanks for fixing this necrobug :)

Issue #4166, opened 2025-04-28
Configurable semantic escape characters for double-click selection


andre-ben:
Hi,

It would be great to have the possibility to configure which characters are treated as "word boundaries" during double-click selection, similar to Alacritty’s semantic_escape_chars option.

In Alacritty, this is configurable like this:

```yaml
selection:
  # This string contains all characters that are used as separators for
  # "semantic words" in Alacritty.
  semantic_escape_chars: ",│`|:\"' ()[]{}<>\t"
```

I attached a video showing an example: when double-clicking on a word enclosed in double quotes ("), I would like the selection to only include the word itself — excluding the double quotes.

https://github.com/user-attachments/assets/92c2d7c8-4075-4d8b-befe-f2203b8eb5c1

Having such an option would allow users to fine-tune how double-click selection behaves, especially for code, logs, or other structured text formats where the default word boundaries might not be ideal.

Thanks for considering!

azeveco:
That would be incredible. I started using Zellij not too long ago and this is driving me crazy lol

Allow us to change this behaviour, please

![Image](https://github.com/user-attachments/assets/3546e45b-b31b-4a02-9128-122f82d18bdd)

spacerodent:
I was searching how to change this and found this issue. I too want what is in the "" and not the " themselves.

awsms:
> I was searching how to change this and found this issue. I too want what is in the "" and not the " themselves.

I can't find a single case where I would actually want `"`, or even the comma that might come after the second `"`.

Tubbles:
Additionally, I would like to exclude colons `:` and number sign `#` from being selected with double click

dfontana:
👋 I made an attempt at addressing this after stumbling into it myself. Have a look and try it yourself 🙇 https://github.com/zellij-org/zellij/pull/4966

Issue #1258, opened 2022-03-23
Support double clicking to select on word boundaries


raphCode:
I see myself prefer the shift+click selection mode of my terminal emulator, because I can use double-click to select whole words. Double clicking and dragging selects the word under the cursor and extends the selection word-wise.
This reduces the need to aim the cursor at character precision, which is very comfortable.

I need to reach out to zellij's selection only when I want to copy multiple lines, where the pane borders get in the way would I use the terminal's selection.
It would be nice for zellij to support the described double-clicking behaviour as well.

vebodev:
vote +1

qiaoboxy:
vote +1

apbyte:
vote +1

thomasaarholt:
Please just give thumbs up on the top issue to prevent spam :)

Shouldn't this behaviour be entirely controlled by the terminal emulator (app), rather than by zellij?

etienne-k:
> Please just give thumbs up on the top issue to prevent spam :)
> 
> Shouldn't this behaviour be entirely controlled by the terminal emulator (app), rather than by zellij?

While that would be nice, I fear it would no longer be possible to click and drag the selection beyond the currently visible area (i.e. drag up and have Zellij scroll up while selecting text)...

barrykp:
> Please just give thumbs up on the top issue to prevent spam :)
> 
> Shouldn't this behaviour be entirely controlled by the terminal emulator (app), rather than by zellij?

The terminal app does normally provide this functionality (and every terminal I've used has it), but you have to disable zellij's `mouse_mode` to get it back (which I'd rather not do because I like the feature).

raphCode:
The terminal emulator is not aware of zellij's panes and using its selection is unsuited when more than one line should be selected:
![image](https://user-images.githubusercontent.com/15750438/198570394-65c0762e-cf1e-4bb2-81bd-d3ebf5137793.png)

emosenkis:
Triple click to select an entire line is also useful for the same reasons

dominz88:
If this is hard to implement, it would be a good idea to make the fullscreen borderless, this will allow the use of the terminal selection (typically with shift) with double and triple click without the border characters.

am-zest:
Yes, we definitely need Zellij to support double-click to select words, and drag to extend the selection.

As part of that, we would need the ability to define the character set for delimiting 'word' boundaries.

avila-m-6:
[...]

PR #4966, opened 2026-03-31
feat: configurable word boundary characters for text selection

dfontana:
## Summary
This PR adds a new configuration option `select_by_word_characters` that allows users to customize which characters are considered part of a word during double-click text selection. Previously, word boundaries were hardcoded to specific punctuation characters. The original implementation was https://github.com/zellij-org/zellij/pull/3996 motivated by https://github.com/zellij-org/zellij/issues/1258. This PR builds on the original implementation, but is motivated by https://github.com/zellij-org/zellij/issues/4166

| Example | Before | After|
|:-|:-|:-|
| `*` is not a word character | 2026-03-30T22:01:15,257409093-04:00 | 2026-03-30T22:22:20,549720063-04:00 |
| `"` is not a word character | 2026-03-30T22:24:22,986053316-04:00 | 2026-03-30T22:22:50,017576515-04:00 |
|Can use entirely custom list | 2026-03-30T22:38:15,758352179-04:00 | 2026-03-30T22:37:24,749081879-04:00 |

## Key Changes
`is_selection_boundary_character()` in `grid.rs` now uses a configurable character set instead of the previously hardcoded list. This character set is stored in a global singleton `SELECT_BY_WORD_CHARACTERS`. 

- The default I've taken from kitty: `@-./_~?&=%+#`
- The new selection boundary logic checks characters in this order:
  1. Whitespace characters are always boundaries
  2. Alphanumeric characters are never boundaries
  3. Other characters are boundaries only if they're NOT in the configured `select_by_word_characters` set

#### Configuration Semantic?

I copied the kitty style where the characters you provide are the ones that _are words_ but the issue's example & the way this function is implemented now are describing character sets that _are boundaries_. Is it preferred to define a boundary list instead?

#### Breaking Change?
The new default is slightly more restrictive than before, do I need to maintain existing behavior or can this be "breaking"?

Before it was whitespace + `[](){}<>` that made boundaries, now it's whitespace  + the defaults above. So while `[](){}<>` are not part of a word (and thus form a boundary), many other characters now create a boundary. I assume it's reasonable to become more restrictive, but open to thoughts. It may be a little trickier to maintain existing behavior, but if paramount could be done by changing the behavior of `is_selection_boundary_character` when the configuration is _not_ set.

### DIFF
--- a/zellij-server/src/lib.rs
+++ b/zellij-server/src/lib.rs
@@ -54,7 +54,8 @@ use route::{route_thread_main, NotificationEnd};
 use zellij_utils::{
     channels::{self, ChannelWithContext, SenderWithContext},
     consts::{
-        DEFAULT_SCROLL_BUFFER_SIZE, SCROLL_BUFFER_SIZE, ZELLIJ_SEEN_RELEASE_NOTES_CACHE_FILE,
+        DEFAULT_SCROLL_BUFFER_SIZE, DEFAULT_SELECT_BY_WORD_CHARACTERS, SCROLL_BUFFER_SIZE,
+        SELECT_BY_WORD_CHARACTERS, ZELLIJ_SEEN_RELEASE_NOTES_CACHE_FILE,
     },
     data::{
         ConnectToSession, InputMode, KeyWithModifier, LayoutInfo, LayoutWithError, Style,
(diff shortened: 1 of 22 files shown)


dfontana:
hi @imsnif , apologies if it's inappropriate to tag you directly for a review (not sure what's the right standard). Understood priority is given to ROADMAP.md items. Would you be willing to review this smaller improvement? I'm also happy to take this back to https://github.com/zellij-org/zellij/issues/4166 if you think we need to change the approach altogether or re-establish the worth of this feature 🙇

Issue #4927, opened 2026-03-27
[Windows] Even when in 'locked' mode modifier keys aren't making it to downstream apps


jgannon65:
## Issue Description
Modifier Keys i.e. 'alt', 'shift', etc aren't being passed to downstream applications, even when zellij is in locked mode. This issue only occurs on Windows. ( I verified it didn't happen on Linux with Ghostty and Nushell).

## Minimal reproduction

Versions:
- Terminal Emulator: Windows Terminal 1.24.10621.0
- Shell: Nushell 0.111.0 
- Zellij: 0.44.0

I originally observed the problem bc broot wasn't able to 'alt + enter' into directories.  I then setup Nushell to listen for input with `input listen --types [key]`.  When listening outside of zellij and I type `alt + enter` I see 
modifiers = keymodifiers(alt). But in zellij there are no modifiers being passed to the program when I do the exact same thing.

divens:
@jgannon65 Thanks for the report, I'll look into it.

PR #4967, opened 2026-03-31
fix(windows): encode modifiers in raw bytes for non-Char keys

divens:
On the native console input path, crossterm captures modifier flags from INPUT_RECORD but the raw bytes written to the PTY don't include them. This causes inner applications to never see ALT, SHIFT, or CTRL modifiers for keys like Enter, Tab, Backspace, arrows, and function keys.

Encode modifiers using standard VT conventions:
- Simple keys (Enter, Tab, Backspace): ALT uses ESC prefix, other modifiers use CSI u encoding
- CSI letter-final keys (arrows, Home, End): xterm modifier parameter
- CSI tilde keys (PageUp, Delete, F5+): modifier before tilde
- SS3 keys (F1-F4): convert to CSI with modifier

Fixes https://github.com/zellij-org/zellij/issues/4927

### DIFF
--- a/zellij-utils/src/input/mod.rs
+++ b/zellij-utils/src/input/mod.rs
@@ -402,6 +402,57 @@ mod not_wasm {
             _ => return None,
         };
 
+        // Encode modifiers in VT raw bytes for non-Char keys.
+        //
+        // On the native console path (Windows), crossterm captures modifier
+        // flags from INPUT_RECORD but the raw bytes don't include them.
+        // Without encoding, the inner application never sees the modifiers.
+        let has_shift = modifiers.contains(&KeyModifier::Shift);
+        let has_any_modifier = has_alt || has_ctrl || has_shift;
+        let raw_bytes = if has_any_modifier && !raw_bytes.is_empty() {
+            let modifier_code = 1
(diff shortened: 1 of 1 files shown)


divens:
@imsnif Ready for review. Tested in Windows Terminal + nushell `input listen --types [key]` to verify how special keys and modifiers reach inner application through zellij.

imsnif:
Hey @divens - since this only happens in the windows path, I don't have a lot of context on this. Happy to merge this if you think it's good. I'll just say that `CSI u` can mean a few different things. It could be https://www.leonerd.org.uk/hacks/fixterms/ or https://sw.kovidgoyal.net/kitty/keyboard-protocol/. As far as I remember there is some overlap between them, but the latter has more features and capabilities. Zellij generally works with the kitty keyboard protocol. Is this what you were going for?

divens:
@imsnif 3 out of the 4 arms are identical in fixterms and KKP : CSI letter, CSI tilde and SS3. The CSI byte;mod u is fixterms style, but only concerns Alt Enter Backspace. I have tested with nushell that I do get the proper key and modifier down in the inner app. So we're good to go!

PR #4968, opened 2026-03-31
Add default keybinding for ToggleMouseMode in scroll/search modes

### LINKED ISSUE #1604: [feature request] switch between mouse_mode enabled/disabled
Hello,
as mentioned in issue [1330](https://github.com/zellij-org/zellij/issues/1330), would it be possible to add a new feature that would allow switching between mouse_mode enabled/disabled in live ? Maybe like the scroll mode but with another keybinding (CTRL + m), without having to change the config.yml every time ?
That would be great, thanks !

MavenRain:
Bind "m" to ToggleMouseMode in both scroll and search modes so users can toggle mouse mode at runtime without custom config.  Closes #1604.


Zykino:
Isn't it already done by just maintaining `Shift` while using the mouse?

PR #4970, opened 2026-03-31
feat: add default_shell args support

divens:
Adds support for passing arguments to the default shell via config:

    default_shell "pwsh" {
        args "-NoLogo"
    }

Introduces DefaultShell struct (path + args) replacing the previous Option<PathBuf>. Backward compatible - default_shell without args block still works, as does --default-shell CLI flag.

### DIFF
--- a/zellij-server/src/lib.rs
+++ b/zellij-server/src/lib.rs
@@ -391,7 +391,12 @@ impl SessionMetaData {
                         .theme_config(new_config.options.theme.as_ref())
                         .unwrap_or_else(|| default_palette().into()),
                     simplified_ui: new_config.options.simplified_ui.unwrap_or(false),
-                    default_shell: new_config.options.default_shell.map(|s| s.path),
+                    default_shell: new_config
+                        .options
+                        .default_shell
+                        .as_ref()
+                        .map(|s| s.path.clone()),
+                    default_shell_args: new_config.options.default_shell.map(|s| s.args),
                     pane_frames: new_config.options.pane_frames.unwrap_or(true),
                     copy_command: new_config.options.copy_command,
                     copy_to_clipboard: new_config.options.copy_clipboard,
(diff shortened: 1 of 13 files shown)


imsnif:
Hey, looks good on the whole. Only things that are missing is that we need to also serialize these args when serializing the config (in the `to_kdl` path) and we also need to send these to the plugins (iirc we send the default shell in`ModeInfo`).

divens:
> Hey, looks good on the whole. Only things that are missing is that we need to also serialize these args when serializing the config (in the `to_kdl` path) and we also need to send these to the plugins (iirc we send the default shell in`ModeInfo`).

Thanks for the review. Will work on this.

divens:
@imsnif Added config serialization and passing shell arguments to plugins

imsnif:
Hey @divens - thanks for the changes, but the more I look into this change the more I find that it's not currently sustainable. Additional things that this breaks (mostly silently):

- tab/mod.rs:1434 — normalization ignores args (which will break session resurrection and likely other stuff)
- session_layout_metadata.rs:201 — requires args empty (will also break session resurrection)
- pty.rs:942 — fallback drops args (likely harmless, but still problematic for future features/refactoring)
- zellij_exports.rs:1694,1728,1771,1809,1846,1885 — plugin fallbacks drop args (same as pty)
- route.rs:480 — DumpLayout discards args (not sure what the right behavior here should be, tbh)

While I definitely want to support this functionality, I don't think this fits in the box of a small change for a patch. Can we maybe work around this problem currently by inviting users to run a small script as their default shell that will include their desired args? We can then look into this more thoroughly after this coming critical patch.

divens:
@imsnif Thanks for your feedback. I definitely see that this has more impact than I expected initially. Agreed, it is better to delay it and avoid shooting ourselves in the foot by creating future issues to be discovered later on.

Issue #4955, opened 2026-03-30
remote terminal attach didn't work


scottgai:
Latest version 0.44.0. 
Tried to attach to remote terminal and got different error on MacOS and Ubuntu.
MacOS:
```
$ zellij attach https://10.10.0.1/fascinating-mountain --token #### --ca-cert localhost+3-jb-ca.pem
Connection failed: TLS error: native-tls error: bad protocol version
```

Ubuntu:
```
$ zellij -d attach https://10.10.0.1/fascinating-mountain --token #### --ca-cert localhost+3-jb-ca.pem 

Bye from Zellij!
```

Loading "https://10.10.0.1/fascinating-mountain" on browser reported the below error after entering the token:
Web clients are not allowed to attach to this session. 

However, curl command could return 200 with content from remote server. 
```
$ curl -vvv  https://10.0.0.1/fascinating-mountain --cacert localhost+3-jb-ca.pem
*   Trying 10.0.0.1:443...
* Connected to 10.0.0.1 (10.0.0.1) port 443
* ALPN: curl offers h2,http/1.1
* (304) (OUT), TLS handshake, Client hello (1):
*  CAfile: localhost+3-jb-ca.pem
*  CApath: none
* (304) (IN), TLS handshake, Server hello (2):
* (304) (IN), TLS handshake, Unknown (8):
* (304) (IN), TLS handshake, Certificate (11):
* (304) (IN), TLS handshake, CERT verify (15):
* (304) (IN), TLS handshake, Finished (20):
* (304) (OUT), TLS handshake, Finished (20):
* SSL connection using TLSv1.3 / AEAD-CHACHA20-POLY1305-SHA256 / [blank] / UNDEF
* ALPN: server accepted h2
* Server certificate:
*  subject: O=mkcert development certificate; OU=user1@jumpbox
*  start date: Mar 30 04:56:03 2026 GMT
*  expire date: Jun 30 04:56:03 2028 GMT
*  subjectAltName: host "10.0.0.1" matched cert's IP address!
*  issuer: O=mkcert development CA; OU=user1@jumpbox; CN=mkcert user1@jumpbox
*  SSL certificate verify ok.
* using HTTP/2
* [HTTP/2] [1] OPENED stream for https://10.0.0.1/fascinating-mountain
* [HTTP/2] [1] [:method: GET]
* [HTTP/2] [1] [:scheme: https]
* [HTTP/2] [1] [:authority: 10.0.0.1]
[...]

PR #4972, opened 2026-03-31
Move from native-tls to rustls in the remote client (mirrors the web server)

imsnif:
This should fix https://github.com/zellij-org/zellij/issues/4955

Essentially, we were using native-tls, which on macOS (apparently) only supports tls1.2, while our webserver requires tls1.3. To solve this, I moved our https client from native-tls to rustls (we have it as a dependency for our web server anyway).


johnstegeman:
Same issue here - web UI works, cannot attach from macos zellij. Built this branch, and all is good

imsnif:
> Same issue here - web UI works, cannot attach from macos zellij. Built this branch, and all is good

Nice, thank you for checking!

PR #4977, opened 2026-03-31
fix(web): allow --token-name with --create-token/--create-read-only-token

### LINKED ISSUE #4976: zellij web --create-token --token-name fails due to incorrect exclusive(true) flag
## Bug

Cannot use `--token-name` with `--create-read-only-token`, despite the [official documentation](https://zellij.dev/documentation/web-client.html#read-only-tokens) showing this as a valid command:

```bash
$ zellij web --create-read-only-token --token-name "observer-token"
error: the argument '--create-read-only-token' cannot be used with '--token-name <TOKEN_NAME>'
```

## Root cause

In `zellij-utils/src/cli.rs:208,214`, both `create_token` and `create_read_only_token` use `exclusive(true)`, which prevents them from being used with any other optional arguments, including `--token-name`.

## Fix

Replace `exclusive(true)` with `conflicts_with_all` to explicitly list conflicting options while allowing `--token-name`.

TMYTiMidlY:
## Summary

Fixes #4976

Replace `exclusive(true)` with `conflicts_with_all()` for token management flags to allow `--token-name` to be used together with token creation flags.

## Changes

Previously, `exclusive(true)` prevented `--create-token` and `--create-read-only-token` from being used with any other optional arguments, including `--token-name`, despite the [official documentation](https://zellij.dev/documentation/web-client.html#read-only-tokens) showing this as valid usage.

Now uses `conflicts_with_all()` (already the pattern used elsewhere in this file) to explicitly list conflicting operations while allowing `--token-name` to work as documented.

## Testing

```bash
$ zellij web --create-read-only-token --token-name "observer-token"
Created token successfully

observer-token: <token> (read-only)

$ zellij web --create-token --token-name "my-token"
Created token successfully

my-token: <token>
```

### DIFF
--- a/zellij-utils/src/cli.rs
+++ b/zellij-utils/src/cli.rs
@@ -205,28 +205,83 @@ pub struct WebCli {
     pub server_startup_timeout: Option<u64>,
     /// Create a login token for the web interface, will only be displayed once and cannot later be
     /// retrieved. Returns the token name and the token.
-    #[clap(long, value_parser, exclusive(true), display_order = 7)]
+    #[clap(
+        long,
+        value_parser,
+        conflicts_with_all(&[
+            "stop",
+            "status",
+            "create-read-only-token",
+            "revoke-token",
+            "revoke-all-tokens",
+            "list-tokens"
+        ]),
+        display_order = 7
+    )]
     pub create_token: bool,
     /// Optional name for the token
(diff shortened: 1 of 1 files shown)


TMYTiMidlY:
Hi 👋 Just checking if this PR is on the radar for review. Happy to make any changes if needed. Thanks!

surma:
I’d love to see this landed!

Issue #4973, opened 2026-03-31
Name flag and name option does not work on windows


Rahmyeon:
## Minimal reproduction
On windows 11 run a clean instance of zellij and the use the following command
```bash
zellij action new-pane --name flame -f
```

The pane will not be renamed, it will show the process name instead. This bug also affects the name option in keymaps and zellij run (have not tested layouts)
## Other relevant information
Tested on windows terminal and alacritty

divens:
Confirmed, thanks for reporting. Working on a fix.

PR #4979, opened 2026-03-31
fix: respect --name flag when creating new panes

divens:
The user-provided pane name (`--name`) was being discarded for default shell panes because `use_terminal_title` was set to true, which caused `pane_title` to be set to `None`. The name was also never passed as `pane_name` to `TerminalPane` — it was always hardcoded to empty string.

Three issues fixed:

1. Pane creation: `--name` was silently discarded for default shell panes because `use_terminal_title` was true. Now user-provided names take priority. Also pass `initial_pane_title` as `pane_name` to `TerminalPane` constructor (was hardcoded to empty string).

2. CLI rename without `--pane-id`: was routed through `PaneNameInput` which appends to existing name. Now uses a dedicated `RenameActivePane` action that fully replaces the name, working for any name length including single characters.

3. Interactive rename: entering rename mode preserves the existing name for editing (append/backspace). Esc reverts to the name before rename mode was entered.

Adds 28 tests covering rename semantics, interactive editing, undo, OSC title interaction, and a wire-level IPC round-trip test.

Fixes https://github.com/zellij-org/zellij/issues/4973

### DIFF
--- a/zellij-server/src/route.rs
+++ b/zellij-server/src/route.rs
@@ -939,15 +939,6 @@ pub(crate) fn route_action(
                 ))
                 .with_context(err_context)?;
         },
-        Action::RenameActivePane { name } => {
-            senders
-                .send_to_screen(ScreenInstruction::RenameActivePane(
-                    name,
-                    client_id,
-                    Some(NotificationEnd::new(completion_tx)),
-                ))
-                .with_context(err_context)?;
-        },
         Action::Run {
             command,
             near_current_pane,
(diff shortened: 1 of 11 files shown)


divens:
@imsnif this was reported on Windows but is also happening on linux. The fix works on both.

Rahmyeon:
Just tested it just now, has a weird quirk where if you use rename pane on a named pane the names are combined without a space ex. name pane hello rename pane world you would get helloworld. (thank you so much for your hard work). I'll try and help you look for a solution for this. But this solution does fix the first problem

Rahmyeon:
For the problem I ecountered on line I got i solved via changing line 6207 of
~/zellij/zellij-server/src/screen.rs
```rust
                    |tab: &mut Tab, client_id: ClientId| tab.update_active_pane_name(c, client_id), ?
```

to

```rust
                     |tab: &mut Tab, client_id: ClientId| tab.set_active_pane_name(c, client_id), ?
```

and added this function to

```rust
    pub fn set_active_pane_name(&mut self, buf: Vec<u8>, client_id: ClientId) -> Result<()> {
        self.get_active_pane_mut(client_id)
            .with_context(|| format!("no active pane found for client {client_id}"))
            .map(|active_pane| {
                active_pane.rename(buf);
            })
    }
```

~zellij/zellij-server/src/tab/mod.rs

I hope this helps (sorry I did not make my own pull request I  do not know rust I but I had claude and went down a rabbit hole of testing this) hope this helps thank you so much for your hard work

divens:
@Rahmyeon Thanks for the additional testing, looking into it.

divens:
@imsnif ready for review

imsnif:
Hey, good catch. A few things:
1. Can't we use the existing action but fix the routing rather than the very verbose change of adding another action?
2. Can we please not include [LLM generated comments](https://github.com/zellij-org/zellij/pull/4979/changes#diff-f3ece84b1e7e730041b2588c9a57105fa08be712dab86d215df05b317a565e74R15592) that refer to non-existing specs?

divens:
@imsnif replaced the new `RenameActivePane` action with `Option<PaneId>` on the existing `RenamePaneByPaneId`. Also removed the spec-reference comment.

imsnif:
Great, thank you!

PR #4981, opened 2026-03-31
fix(web-client): use base URL when switching sessions

### LINKED ISSUE #4980: Web client: session switch ignores base_url behind reverse proxy
When switching sessions via the session manager, the web client redirects to `/{session_name}` instead of `{base_url}/{session_name}`, ignoring the configured `base_url`.

**Steps to reproduce:**
1. Configure `base_url = "/zellij/"` in `web_client` config
2. Set up a reverse proxy (e.g. Caddy `handle_path /zellij*`)
3. Connect to the web client, create multiple sessions
4. Switch sessions using the session manager

**Expected:** Browser navigates to `/zellij/my-session`
**Actual:** Browser navigates to `/my-session` (404)

**Root cause:** `websockets.js` line 244 hardcodes `window.location.pathname = \`/\${new_session_name}\`` without using the base URL.

TMYTiMidlY:
## Summary

Fixes #4980

When switching sessions via the session manager, the web client redirects to `/{session_name}`, ignoring the configured `base_url`. This breaks all reverse proxy setups that serve zellij under a subpath.

## Changes

Use `getBaseUrl()` to construct the redirect URL in `websockets.js`, consistent with how `auth.js` and `connection.js` already handle the base URL.

## Testing

Tested with Caddy reverse proxy (`handle_path /zellij-dev*` → `localhost:8084`) and `base_url = "/zellij-dev/"`. Session switching now correctly navigates to `/zellij-dev/{session_name}`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,1272 +5,1198 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 
 ## [Unreleased]
-
-- fix: don't display default ports as offline in `share` plugin (https://github.com/zellij-org/zellij/pull/4908)
-- fix: theme folder not being picked up on startup (https://github.com/zellij-org/zellij/pull/4892)
-- fix: do not reset selection in session-manager on SessionInfo update (https://github.com/zellij-org/zellij/pull/4919)
-- fix: performance regression when switching tabs (https://github.com/zellij-org/zellij/pull/4918)
-- feat: add "focus-pane-with-id" and "are-floating-panes-visible" cli commands (https://github.com/zellij-org/zellij/pull/4926)
-- build: add risc64 binaries to releases (https://github.com/zellij-org/zellij/pull/4913)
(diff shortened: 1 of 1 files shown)


zellij-client/assets/websockets.js
```diff
@@ -241,7 +241,8 @@ function startWsControl(wsControl, term, fitAddon, ownWebClientId, userConfig) {
             }
         } else if (msg.type === "SwitchedSession") {
             const { new_session_name } = msg;
-            window.location.pathname = `/${new_session_name}`;
+            const baseUrl = getBaseUrl();
+            window.location.href = `${baseUrl}/${new_session_name}`;
```
tlinford: just a small thing, should use `${encodeURIComponent(new_session_name)`.
href and pathname behave a bit differently and we could have problems if the session name contains special characters

TMYTiMidlY: Done, thanks for catching this!

tlinford: Merged, thanks for the fix!

Issue #694, opened 2021-09-05
feature: Allow to hide help bar and tab bar


phaazon:
Both bar should be allowed to be hidden, potentially even dynamically (my workflow in tmux relies a lot on hiding the statusbar to use all the screen estate).

I will have a look and try to come up with PR to fix this.

a-kenji:
So for a little bit more context. 
If we fullscreen a pane, it is inside of the tab boundary.
We want to be able to toggle the things surrounding this boundary.

roland-rollo:
Could I add idea for this? I have small screen at this moment and any pixel is life important. In sway/i3 windows manager I can see system bar pressing key (in my case mod4) and after toggle with key, this bar is on layer above apps on screen (top layer). Would be in future option to have mode starting "naked" like normal session in any terminal and after key combination (let say alt+y) in zellij it will show two bars we have (top and bottom) in layer above zellij window (so that both panels would cover the window fragments, obscuring the contents) and then we could normally use and see active mode/option (like ctrl+p show option for panes and ctrl+t for tabs) and pressing again the same key combination (alt+y) will hide everything again.

ladyrick:
+1

jeffschwartz:
Is this still a thing? I'd love to see this implemented and if I knew Rust I'd take a crack at it but I don't know Rust unfortunately. It would be so nice to be to run terminal apps, like (N)Vim for instance, in full screen and regain all that screen real estate.

spfrommer:
+1

leitao:
This is going to be very helpful. When looking at Zellij, that is the first thing I was looking for, and unfortunately this seems not possible.

That said, thanks for the project. It is great.

jeremyjjbrown:
Once this is added zellij will be perfect and I will uninstall tmux.

multivac61:
This is a killer feature 🔥

Zik42:
+1

jorge-menjivar:
For now, you can create a layout that hides the status bar and border, but you won't be able to switch dynamically between layouts.

Create a new layout somewhere (for example ~/.config/zellij/layouts/naked.kdl)
```
layout {
    pane
}
pane_frames false
```

Then execute `zellij --layout ~/.config/zellij/layouts/naked.kdl`, which will open a new pane with that layout. You can set the default layout in ~/.config/zellij/config.kdl so all sessions open with that layout `default_layout "compact"`.

[...]

PR #4983, opened 2026-04-01
add ToggleView action to show/hide panes/tab bars

ddppt-yy:
In issue #694, many people need the ability to dynamically hide unnecessary UI to expand usable screen space. I have this need too, so I modified the main code.

My modifications might not be perfect, but I still hope this feature can be merged into the main branch soon. 

Currently, I added a configuration parameter `lock_hide_toggle`, which is enabled by default. When enabled, entering lock mode with `ctrl+g` hides UI elements like panes, tabs, etc., and the freed space can be used by the user.
When the user exits lock mode (again with `ctrl+g`), the previous UI is restored.

FIX https://github.com/zellij-org/zellij/issues/694

### DIFF
--- a/zellij-server/src/panes/tiled_panes/mod.rs
+++ b/zellij-server/src/panes/tiled_panes/mod.rs
@@ -511,6 +511,18 @@ impl TiledPanes {
             })
             .collect()
     }
+    pub fn non_selectable_pane_ids(&self) -> Vec<PaneId> {
+        self.panes
+            .iter()
+            .filter_map(|(id, p)| {
+                if !p.selectable() {
+                    Some(*id)
+                } else {
+                    None
+                }
+            })
+            .collect()
+    }
     pub fn first_selectable_pane_id(&self) -> Option<PaneId> {
         self.panes
             .iter()
--- a/zellij-server/src/screen.rs
+++ b/zellij-server/src/screen.rs
@@ -1310,6 +1310,8 @@ pub(crate) struct Screen {
     default_mode_info: ModeInfo, // TODO: restructure ModeInfo to prevent this duplication
     style: Style,
     draw_pane_frames: bool,
+    draw_pane_frames_before_lock: Option<bool>, // Save pane frames state before entering lock mode
+    lock_hide_toggle: bool, // Whether to hide UI elements in lock mode
     auto_layout: bool,
     session_serialization: bool,
     serialize_pane_viewport: bool,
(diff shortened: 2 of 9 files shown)


imsnif:
Why not use [`OverrideLayout`](https://zellij.dev/documentation/keybindings-possible-actions.html#overridelayout)? All you need is some external place to hold the state of "which layout I'm in right now", you can potentially even query this with `zellij action list-panes`. With some creativity you can also use an external layout file and adjust it in the same keybinding to switch between the layouts you want. This was the main reason I introduced those capabilities.

I understand it's easiest to have this completely built-in, I can totally relate. But for some levels of custom behavior we have to rely on users using the app's configuration capabilities.

ddppt-yy:
> Why not use [`OverrideLayout`](https://zellij.dev/documentation/keybindings-possible-actions.html#overridelayout)? All you need is some external place to hold the state of "which layout I'm in right now", you can potentially even query this with `zellij action list-panes`. With some creativity you can also use an external layout file and adjust it in the same keybinding to switch between the layouts you want. This was the main reason I introduced those capabilities.
> 
> I understand it's easiest to have this completely built-in, I can totally relate. But for some levels of custom behavior we have to rely on users using the app's configuration capabilities.

Hello,
I tried the method you suggested, and it does achieve the function of hiding the UI. 

However, there is a problem:  Zellij does not seem to remember the distribution of panes and tabs before hiding the UI; it can only switch between two predefined layouts (or maybe I didn't set something up correctly?).

Therefore, I believe this does not fully implement the functionality I modified.

Or could you provide me with a solution to remember the previous layout? For example, write the current layout to a temporary file before hiding, and then reload it afterward?

imsnif:
Glad it worked out. You can dump the current layout with: https://zellij.dev/documentation/cli-actions.html#dump-layout

If you need support, please don't hesitate to ask on our Discord or Matrix.

ddppt-yy:
very thx @imsnif

ddppt-yy:
> Glad it worked out. You can dump the current layout with: https://zellij.dev/documentation/cli-actions.html#dump-layout
> 
> If you need support, please don't hesitate to ask on our Discord or Matrix.

Sorry to bother you again. I’ve found another issue that needs to be addressed: when I have two panes stacked vertically in a tab, with the left pane running a program (e.g., I’m synthesizing chip RTL code and data keeps streaming to the screen, or something else that requires constant monitoring), and the right pane where I’m using vim to debug a script that the left pane uses. At this point, if I load a new layout without tab bar or status bar, I’m taken to a new tab instead of continuing my work in the original tab. Is there a way to solve this?

I sincerely hope you can understand my struggle with a small screen—hiding three lines of space and removing extra UI gives me a clean screen, which greatly helps me scan and find information on screen. At the same time, I’d still appreciate it if you could reconsider the possibility of merging my PR.

Even if my writing isn’t great, I still hope this feature can be implemented. Thank you so much.

Maybe implementing this as a plugin is the better approach.

PR #4984, opened 2026-04-01
Add permission grants for plugins in load_plugins config

### LINKED ISSUE #4982: Background plugins (load_plugins) cannot receive permission grants
## Description

The symptom is that no permission dialogs are shown for me for a background plugin that is added in `config.kdl` via 

```
load_plugins {
   "file:/Users/user/.config/zellij/plugins/cc_deck_controller.wasm" {
       mode "controller"
   }
}
```

The following analysis is an attempt to find the root cause, but I'm not sure if it is appropriate (generated via Claude Opus 1M). 

<details>
  <summary>AI assessment of the flow</summary>
Plugins loaded via `load_plugins` in `config.kdl` start with `start_suppressed = true` and are not associated with any tab. When such a plugin calls `request_permission()` in its `load()` method, the permission request is handled by `ScreenInstruction::RequestPluginPermissions`, which iterates over all tabs looking for one that contains the plugin via `tab.has_plugin(plugin_id)`. Since background plugins are not in any tab at the time `load()` runs, the request is pushed to `pending_events_waiting_for_client` but never successfully processed, meaning the user never sees the permission dialog.

## Root Cause Analysis

The permission request flow works as follows:

1. **Plugin calls `request_permission()`** (`zellij-tile/src/shim.rs:89-94`), which sends a `PluginCommand::RequestPluginPermissions` to the host.

2. **Server-side handler** (`zellij-server/src/plugins/zellij_exports.rs:885-913`) first checks the permission cache. If permissions are already cached, it immediately grants them and returns (this is why pre-populating `permissions.kdl` works as a workaround). Otherwise, it sends `ScreenInstruction::RequestPluginPermissions` to the screen thread.

3. **Screen handler** (`zellij-server/src/screen.rs:7734-7750`) iterates over all tabs calling `tab.has_plugin(plugin_id)`. While `has_plugin()` (`zellij-server/src/tab/mod.rs:2590-2599`) does check tiled panes, floating panes, and suppressed panes, the background plugin isn't in any tab at this point due to a timing issue: `load()` fires before the `ScreenInstruction::AddPlugin` for the background plugin has been processed by the screen thread.

4. **Caching and replay**: The unmatched request is pushed to `pending_events_waiting_for_client`, which is replayed at three points (after layout applied, after client added, after tabs overridden) in `screen.rs:6395-6397, 6651-6653, 7027-7029`. However, if the plugin is still not associated with any tab at replay time, the request fails again in the same way.

5. **The dialog mechanism itself works for suppressed panes**: `request_plugin_permissions()` (`zellij-server/src/tab/mod.rs:5385-5416`) already handles suppressed panes by temporarily promoting them to floating panes via `focus_suppressed_pane_for_all_clients()`, showing the permission dialog, and re-suppressing after the user responds. The problem is that the request never reaches this code because the plugin isn't in any tab to begin with.
</details>

## Use Case

I'm building a two-plugin architecture for [cc-deck](https://github.com/rhuss/cc-deck) (a Claude Code session manager):

- **Controller** (background plugin via `load_plugins`): Owns all session state, processes heavyweight events (PaneUpdate, TabUpdate, Timer), broadcasts pre-computed display data to sidebars
- **Sidebar** (one per tab via layout template): Thin renderer that displays cached data and handles local user interaction

The controller needs some permissions, but never runs in the foreground (so has no pane attached for asking for permissions)

## Current Workaround

Pre-populating `permissions.kdl` in the Zellij cache directory (`~/.cache/zellij/permissions.kdl`) during the CLI install step. This works because the cache check in step 2 above short-circuits before the broken screen instruction path is reached. The format is:

```kdl
"file:~/.config/zellij/plugins/my_background_plugin.wasm" {
    ReadApplicationState
    ChangeApplicationState
    RunCommands
    ReadCliPipes
    MessageAndLaunchOtherPlugins
    Reconfigure
}
```

This is reliable but depends on the internal cache file format and location.

### Option A: Handle tab-less plugins in the screen handler (smallest fix)

In the `ScreenInstruction::RequestPluginPermissions` handler (`screen.rs:7734-7750`), when no tab contains the plugin, instead of just caching the event, add the plugin to the active tab's suppressed panes and call `request_plugin_permissions()` there. The existing suppressed-to-floating promotion logic already handles the UI correctly.

### Option B: Allow declaring permissions in `load_plugins` config

Add a config option to auto-grant permissions at load time:

```kdl
load_plugins {
    "file:plugin.wasm" {
        grant_permissions true
    }
}
```

This bypasses the dialog entirely, which is appropriate for plugins the user has explicitly configured.

### Option C: CLI command for non-interactive permission grants

```
zellij action grant-plugin-permissions "file:plugin.wasm"
```

This would allow scripts and installers to grant permissions without knowledge of the cache file internals.

## Reproduction

```kdl
// config.kdl
load_plugins {
    "file:~/.config/zellij/plugins/my_background_plugin.wasm" {
        mode "background"
    }
}
```

The plugin's `load()` calls `request_permission(&[...])` but `PermissionRequestResult` is never delivered. The plugin stays in a "waiting for permissions" state indefinitely.

## Environment

- Zellij 0.44.0
- macOS (also reproducible on Linux)
- zellij-tile 0.44

rhuss:
## Summary

- Add `grant_all_permissions` and `grant_permissions` options to the `load_plugins` config block
- Pre-populates the permission cache at config parse time so plugins start with permissions already granted
- Includes unit tests for all three scenarios (grant all, grant specific, no grant)

Closes #4982

## Motivation

Background plugins loaded via `load_plugins` start with `start_suppressed = true` and are not associated with any visible pane. When such a plugin calls `request_permission()` during `load()`, the permission dialog cannot be shown because it relies on promoting the suppressed pane to a floating pane, which has timing and rendering issues at startup (the pane is created before clients connect, so the dialog's client grid is empty and never renders).

The current workaround is to manually pre-populate the `permissions.kdl` cache file, which works but depends on internal cache file format and location.

This PR provides a first-class config option instead. Since `load_plugins` entries are explicitly added by the user to their own config file, granting permissions there is a deliberate, conscious decision, no different from the user pressing "y" on the permission dialog. The security model is preserved: only the config file owner can grant permissions, and only for plugins they choose to load.

## Usage

**Grant all permissions** (for trusted plugins where you don't want to track individual permissions):

```kdl
load_plugins {
    "file:~/.config/zellij/plugins/my_controller.wasm" {
        grant_all_permissions true
    }
}
```

**Grant specific permissions** (for tighter control):

```kdl
load_plugins {
    "file:~/.config/zellij/plugins/my_sidebar.wasm" {
        grant_permissions {
            ReadApplicationState
            ChangeApplicationState
            RunCommands
            ReadCliPipes
            MessageAndLaunchOtherPlugins
            Reconfigure
        }
    }
}
```

**No grant** (existing behavior, unchanged):

```kdl
load_plugins {
    "file:~/.config/zellij/plugins/other_plugin.wasm" {
        config_key "config_value"
    }
}
```

## How it works

During config parsing in `load_plugins_from_kdl()`, if `grant_all_permissions` or `grant_permissions` is set, the specified permissions are written to the `PermissionCache` file. When the plugin later calls `request_permission()`, the cache check in `zellij_exports.rs` finds the permissions already granted and immediately returns `PermissionRequestResult::Granted` without ever reaching the screen thread or requiring a dialog.

## Test plan

- [x] `load_plugins_grant_all_permissions` - verifies all permission types are returned
- [x] `load_plugins_grant_specific_permissions` - verifies only listed permissions are returned
- [x] `load_plugins_no_grant_permissions` - verifies no permissions when option is absent
- [ ] Manual: configure a background plugin with `grant_all_permissions true`, verify it starts with permissions granted

🤖 Assisted-By [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/example/default.kdl
+++ b/example/default.kdl
@@ -245,10 +245,31 @@ plugins {
 }
 
 // Plugins to load in the background when a new session starts
+// Background plugins cannot show the interactive permission dialog,
+// so permissions can be granted directly here.
+//
+// grant_all_permissions: grants every permission type
+// grant_permissions:     grants only the listed permission types
+//   Available permissions:
+//     ReadApplicationState, ChangeApplicationState, OpenFiles,
+//     RunCommands, OpenTerminalsOrPlugins, WriteToStdin, WebAccess,
+//     ReadCliPipes, MessageAndLaunchOtherPlugins, Reconfigure,
+//     FullHdAccess, StartWebServer, InterceptInput, ReadPaneContents,
+//     RunActionsAsUser, WriteToClipboard, ReadSessionEnvironmentVariables
 load_plugins {
     "zellij:link"
     // "file:/path/to/my-plugin.wasm"
     // "https://example.com/my-plugin.wasm"
+    // "file:/path/to/my-plugin.wasm" {
+    //     grant_all_permissions true
+    // }
+    // "file:/path/to/my-plugin.wasm" {
+    //     grant_permissions {
+    //         ReadApplicationState
+    //         ChangeApplicationState
+    //         RunCommands
+    //     }
+    // }
 }
 
 // Choose what to do when zellij receives SIGTERM, SIGINT, SIGQUIT or SIGHUP
--- a/zellij-utils/assets/config/default.kdl
+++ b/zellij-utils/assets/config/default.kdl
@@ -245,10 +245,31 @@ plugins {
 }
 
 // Plugins to load in the background when a new session starts
+// Background plugins cannot show the interactive permission dialog,
+// so permissions can be granted directly here.
+//
+// grant_all_permissions: grants every permission type
+// grant_permissions:     grants only the listed permission types
+//   Available permissions:
+//     ReadApplicationState, ChangeApplicationState, OpenFiles,
+//     RunCommands, OpenTerminalsOrPlugins, WriteToStdin, WebAccess,
+//     ReadCliPipes, MessageAndLaunchOtherPlugins, Reconfigure,
+//     FullHdAccess, StartWebServer, InterceptInput, ReadPaneContents,
+//     RunActionsAsUser, WriteToClipboard, ReadSessionEnvironmentVariables
 load_plugins {
     "zellij:link"
     // "file:/path/to/my-plugin.wasm"
     // "https://example.com/my-plugin.wasm"
+    // "file:/path/to/my-plugin.wasm" {
+    //     grant_all_permissions true
+    // }
+    // "file:/path/to/my-plugin.wasm" {
+    //     grant_permissions {
+    //         ReadApplicationState
+    //         ChangeApplicationState
+    //         RunCommands
+    //     }
+    // }
 }
 
 // Choose what to do when zellij receives SIGTERM, SIGINT, SIGQUIT or SIGHUP


rhuss:
Closing this PR as it's not really the direction that permission guardrails are heading for. See discussion in #4982 .

PR #4988, opened 2026-04-01
Fix delete-all-sessions silently doing nothing on Windows

ci4ic4:
On Windows, folder_name.display().to_string() returns the full absolute path, which when passed to session_layout_cache_file_name() produces a malformed path (the base cache dir joined with the full path again).

Extract just the folder name component using file_name() before passing it to session_layout_cache_file_name(). This fixes both get_resurrectable_sessions() and get_resurrectable_session_names().


imsnif:
@divens - what do you think? (of course if this is the direction we need to make sure there are no changes for non-windows)

divens:
@imsnif I tested yesterday on stock zellij v0.44.0 and could not reproduce the issue. I need to spend a bit more time to figure out in which circumstances this happens.

PR #5004, opened 2026-04-05
Avoid unneeded allocations with `to_string`

mo8it:
I am working on a Zellij integration in Rustlings and wanted check in the source code whether the environment variable `ZELLIJ` is always set. Then I saw that it leads to unneeded allocations and I removed some of the related ones. Are you open to such contributions?

### DIFF
--- a/zellij-utils/src/vendored/termwiz/input.rs
+++ b/zellij-utils/src/vendored/termwiz/input.rs
@@ -1943,19 +1943,19 @@ mod test {
 
         assert_eq!(
             KeyCode::Enter.encode(Modifiers::CTRL, mode, true).unwrap(),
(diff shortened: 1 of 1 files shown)


imsnif:
Much appreciated and thanks for your vigilence. Generally my policy about this stuff is that code changes need to either be a behavior improvement, a resource utilization / performance improvement or a massive simplification on the level of removing internal subsystems, unneeded dependencies, etc. The latter part is usually part of a bigger plan that's made beforehand.

So, if these allocations were in a critical and non-I/O blocked path - there would definitely be room to discuss.

mo8it:
Probably not significant enough for your policy. So feel free to close it :)

PR #5006, opened 2026-04-06
fix(windows): resolve named pipe path from marker file after session rename

### LINKED ISSUE #4998: Cannot attach to a session after renaming it on Windows
# Issues with the Zellij UI / behavior / crash

## Issue description

**Session rename breaks IPC on Windows — named pipes retain old name while marker file is renamed**

On Windows, renaming a session (e.g. from `splendid-rhinoceros` to `agent`) only renames the marker file on disk. The underlying named pipes (`\\.\pipe\...`) are kernel objects and cannot be renamed via `std::fs::rename`. After renaming, all session operations (`attach`, `kill-session`, etc.) fail because they try to connect to a pipe name derived from the new marker file, which doesn't exist.

### Root Cause

In `zellij-server/src/screen.rs:7623`:

```rust
// rename socket file
let old_socket_file_path = ZELLIJ_SOCK_DIR.join(&old_session_name);
let new_socket_file_path = ZELLIJ_SOCK_DIR.join(&name);
if let Err(e) = std::fs::rename(old_socket_file_path, new_socket_file_path) {
    log::error!("Failed to rename ipc socket: {:?}", e);
}
```

On Unix this works correctly because the socket **is** the file — renaming the file renames the socket.

On Windows, the architecture is different:
- **Marker file** on disk (e.g. `C:\...\contract_version_1\splendid-rhinoceros`) — contains the server PID, used for session discovery
- **Named pipes** in the kernel namespace (e.g. `\\.\pipe\C:\...\contract_version_1\splendid-rhinoceros` and `...-reply`) — used for actual IPC

`std::fs::rename` only renames the marker file. The named pipes are kernel objects and retain their original names. Windows does not support renaming named pipes after creation.

### Symptoms

- `zellij attach <new-name>` hangs indefinitely (infinite retry loop in `connect_to_server`)
- `zellij kill-session <new-name>` fails with:
  ```
  Error occurred: Os { code: 2, kind: NotFound, message: "The system cannot find the file specified." }
  ```
- `zellij list-sessions` still shows the session as alive (marker file exists, PID is alive)
- The session is fully functional internally — the server process and all pipes are intact, just unreachable under the new name
- Manually renaming the marker file back to the original name immediately restores connectivity

## Minimal reproduction

1. On Windows, start a new session:
   ```
   zellij
   ```
   (Zellij auto-generates a name like `splendid-rhinoceros`)

2. Rename the session from within Zellij (via the session-manager plugin or `RenameSession` action), e.g. rename to `my-session`

3. Detach from the session (Ctrl+O, d)

4. Try to re-attach:
   ```
   zellij attach my-session
   ```
   → **Hangs forever**

5. Try to kill:
   ```
   zellij kill-session my-session
   ```
   → **Error: Os { code: 2, kind: NotFound }**

6. Verify the pipes still use the old name:
   ```powershell
   [System.IO.Directory]::GetFiles("\\.\pipe\") | findstr zellij
   ```
   → Shows pipes with the **original** name (`splendid-rhinoceros`), not the new name (`my-session`)

7. Workaround — rename the marker file back:
   ```powershell
   Rename-Item "$env:TEMP\zellij\contract_version_1\my-session" "splendid-rhinoceros"
   zellij attach splendid-rhinoceros
   ```
   → **Works immediately**

## Other relevant information

**`zellij --version`**: 0.44.0 (release)

**`ver`**: Microsoft Windows [Version 10.0.26100.7840]

### Suggested Fix

**Option A** (recommended): Recreate the named pipes with the new name during rename — close the old listeners, create new listeners with `ipc_bind()` / `ipc_bind_reply()` using the new path.

**Option B**: Disable rename on Windows — return an error like "Session rename is not supported on Windows" until a proper fix is implemented.

**Option C**: Store the original pipe name in the marker file — instead of just the PID, write both PID and original pipe name, and modify `ipc_connect` to read the pipe name from the marker file rather than deriving it from the file path.

**This issue was analyzed with the assistance of Claude Code by stepping through the Zellij source code to identify the root cause**

divens:
On Windows, session rename only moves the marker file. The named pipes are kernel objects that cannot be renamed. Store the original pipe source path in the marker file (PID\n<path>) so clients can connect to the correct pipe even after the session has been renamed.

Fixes https://github.com/zellij-org/zellij/issues/4998

### DIFF
--- a/zellij-utils/src/consts.rs
+++ b/zellij-utils/src/consts.rs
@@ -182,10 +182,30 @@ pub fn ipc_connect(path: &std::path::Path) -> std::io::Result<interprocess::loca
     LocalSocketStream::connect(fs_name)
 }
 
+/// Read the pipe source path from a Windows marker file.
+///
+/// The marker file format is `PID\n<pipe_source_path>`. If the second line
+/// is present, returns that path (the original path used when the pipe was
+/// created). Otherwise falls back to the marker file path itself. This
+/// allows session rename to work: the marker file moves but its content
+/// still points to the original pipe name.
+#[cfg(windows)]
+fn read_pipe_path(marker_path: &std::path::Path) -> std::path::PathBuf {
+    if let Ok(contents) = std::fs::read_to_string(marker_path) {
+        if let Some(pipe_path) = contents.lines().nth(1) {
+            if !pipe_path.is_empty() {
+                return std::path::PathBuf::from(pipe_path);
+            }
+        }
+    }
+    marker_path.to_path_buf()
+}
+
 #[cfg(windows)]
 pub fn ipc_connect(path: &std::path::Path) -> std::io::Result<interprocess::local_socket::Stream> {
     use interprocess::local_socket::{prelude::*, GenericNamespaced, Stream as LocalSocketStream};
-    let name = path.to_string_lossy().to_string();
+    let pipe_path = read_pipe_path(path);
+    let name = pipe_path.to_string_lossy().to_string();
     let ns_name = name.to_ns_name::<GenericNamespaced>()?;
     LocalSocketStream::connect(ns_name)
 }
@@ -208,7 +228,8 @@ pub fn ipc_bind(path: &std::path::Path) -> std::io::Result<interprocess::local_s
     let name = path.to_string_lossy().to_string();
     let ns_name = name.to_ns_name::<GenericNamespaced>()?;
     let listener = ListenerOptions::new().name(ns_name).create_sync()?;
-    std::fs::write(path, std::process::id().to_string())?;
+    let marker = format!("{}\n{}", std::process::id(), path.to_string_lossy());
+    std::fs::write(path, marker)?;
     Ok(listener)
 }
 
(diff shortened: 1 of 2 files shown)


divens:
@imsnif Ready for review

guptaanmol184:
@divens  I tested the fix on my local and session renaming is working and I can attach to existing sessions. This is awesome!

guptaanmol184:
I have been using this (building using devins fix). @imsnif could we please have it in the next release? It's a small but significant friction point.

Cujuju:
Please merge this.

hku:
Tested this on **Windows 11**, cherry-picked onto the `v0.45.1` tag. 
Rename + detach + re-attach works; 
`kill-session` on the renamed session also succeeds.

Verified the marker file format directly -- after renaming `abc` -> `xyz`,
`%TEMP%\zellij\contract_version_1\xyz` contains `PID\n<original pipe path>`,
so `read_pipe_path()` resolves to the pipe created under the old name.

Details in #4998.

guptaanmol184:
@imsnif @divens Can we merge this ? It's been almost 5 months since this PR was raised. 
Renaming a session is a key functionality for zellij users, and would be great to have in the next release.

divens:
@hku @guptaanmol184 This PR will likely not be merged. As explained [here](https://github.com/zellij-org/zellij/issues/4998#issuecomment-5371565616), this PR received feedback asking for a broader solution which I implemented at first in #5103 and recently overhauled with a more robust solution in #5513. 

I'm aware that session renaming is important, which is why I wrote 3 PRs for it already. It is as important for the long term stability of the project that we choose the right approach when solving issues. Aram said he would review the most recent PR in the coming weeks, and I'm confident we can work any outstanding issues to have it land in the next release.

PR #5039, opened 2026-04-11
feat(link): make floating pane size configurable via plugin config

### LINKED ISSUE #5038: link plugin: make floating pane size configurable
## Feature Request

When alt+clicking a file path in the terminal, the link plugin opens it in a floating pane. The floating pane size is hardcoded — there's no way to configure it.

**Specifically:** I'd like the default floating pane size used by the link plugin to be user-configurable (e.g. 80% of the window). Right now `open_file_floating` is called with `None` for coordinates and the plugin's `configuration` map in `load()` is ignored (`_configuration`).

## Desired behavior

Allow configuring the floating pane size via `config.kdl`:

```kdl
load_plugins {
    zellij:link {
        x "10%"
        y "10%"
        width "80%"
        height "80%"
    }
}
```

## Current behavior

`load()` receives but ignores its `BTreeMap<String, String>` config. `open_file_floating` is always called with `None` for `FloatingPaneCoordinates`, so the pane opens at zellij's hardcoded default size (~50% centered).

## Implementation note

The fix is small — wire the ignored `configuration` map in `load()` into `FloatingPaneCoordinates::new()` and pass the result to `open_file_floating`. `FloatingPaneCoordinates::new()` already returns `None` when all args are `None`, so default behaviour is preserved for users who don't configure it.

KalebJS:
## Summary

The link plugin opens files in a floating pane on alt+click, but hardcodes `None` for `FloatingPaneCoordinates`, making the pane size uncontrollable by users.

This PR wires the existing (but ignored) `configuration: BTreeMap<String, String>` in `load()` into `FloatingPaneCoordinates::new()` and passes the result to `open_file_floating()`. When no config keys are provided, `new()` returns `None` and behaviour is completely unchanged.

## Usage

```kdl
load_plugins {
    zellij:link {
        x "10%"
        y "10%"
        width "80%"
        height "80%"
    }
}
```

All four dimensions accept either a percentage (`"80%"`) or a fixed character count (`"40"`). `pinned` and `borderless` are also supported (`"true"`/`"false"`). Any omitted key falls back to zellij's default.

## Changes

- `State` struct: add `floating_pane_coordinates: Option<FloatingPaneCoordinates>` field
- `load()`: parse `x`/`y`/`width`/`height`/`pinned`/`borderless` from config map
- `handle_highlight_clicked()`: pass `self.floating_pane_coordinates.clone()` instead of `None`

Closes #5038

### DIFF
--- a/default-plugins/link/src/main.rs
+++ b/default-plugins/link/src/main.rs
@@ -17,12 +17,15 @@ struct State {
     /// Session environment variables, fetched once on load.
     /// Used for `~` and `$VAR` expansion in clicked paths.
     env_vars: BTreeMap<String, String>,
+    /// Optional floating pane coordinates parsed from plugin configuration.
+    /// When `None`, zellij uses its default floating pane size.
+    floating_pane_coordinates: Option<FloatingPaneCoordinates>,
 }
 
 register_plugin!(State);
 
 impl ZellijPlugin for State {
-    fn load(&mut self, _configuration: BTreeMap<String, String>) {
+    fn load(&mut self, configuration: BTreeMap<String, String>) {
         subscribe(&[
             EventType::PaneUpdate,
             EventType::HighlightClicked,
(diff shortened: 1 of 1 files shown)


Zykino:
The built-in plugins are here to serve the bare-minimum most basic uses of the average user. The philosophy is that if you want extra pieces of UI, you can always roll your own (even by forking the existing plugins and changing their behaviour).

And since you already made the change you want to see you are mostly done publishing your own. If you do, don't hesitate to add it in https://github.com/zellij-org/awesome-zellij :)

KalebJS:
Closing — the implementation approach needs more work. Keeping the issue open to track the feature request.

Issue #4129, opened 2025-04-10
Inconsistent Session Serialization / Resurrection


Aequitosh:
As the title says, `zellij` is sometimes inconsistent when it comes to session serialization / resurrection.

Version: `0.42.1`, installed via `cargo install zellij`

To describe my specific case:
- Occasionally tabs will be missing or not be restored to their original state.
  - For example, I usually have a couple Neovim instances open, because `auto-session` + `zellij` allows me to instantly return to whatever I was working on (zellij is awesome in that regard btw, absolutely love it).
  - *Sometimes* those tabs will just take me to my home directory instead, while some other tabs do take me back into Neovim or my mail client.
- Panes seem to be more consistent in this regard, *except* the first pane in a tab.
  - The first pane will often also just take me to my home directory, while the second pane (and onwards) will be restored correctly.
  - To give a more concrete example, when working with `git`, I'll often have two panes (left and right) in which I'll run `git` commands. Upon resurrection, the left pane will often just take me to my home dir, while the right pane will drop me into the dir of the repository I was working in (as expected).
- The directory in which my shell runs is inconsistent.
  - In my `default` session, I'm always put into my home dir whenever I open a new tab.
  - In a different, brand new session, when I go to `~/Documents` and open a new tab, I am also put into `~/Documents`. When I now `cd ~/Documents/Notes` and open a new tab, I am now put into `~/Documents/Notes`.
  - I have made a new session named `debug` in order to reproduce this---in this session, whenever I open a new tab, I will be dropped into `~/Documents` for some reason, no matter what directory I am in currently. That was the first directory I ever `cd`ed into in this session, so perhaps that's why..?
  - This behaviour (in the session named `debug`) only started to occur *after* the session was killed for the first time.
  - Panes seem to not be affected by this and will always drop me into the directory of the pane was previously in focus.
- Certain commands are not executed in the same manner as they were before resurrecting the session.
  - For example, when I run `git log` in some repository, detach, kill the session (after it was serialized), re-attach to the session again, `less` will complain that it hasn't received any input (since I use `less` as my pager for `git`).
  - `git branch` will also have a different output format upon resurrection. I have the following in my `~/.gitconfig`:
    ```
    [column]
            ui = auto
    ```
    This causes my branches to be displayed as columns instead of one huge list. However, when resurrecting a session, this isn't respected; `git branch` shows a list of branches instead.

[...]

PR #5046, opened 2026-04-13
fix(serialization): preserve single-pane tabs in session layout serialization

skmtkytr:
## Summary

- Fixes a bug where tabs with exactly one tiled pane were serialized as empty `tab name="..." {}` in `dump-layout` and session serialization
- This caused a cascading resurrection failure: once a tab lost its plugin pane (e.g. zjstatus), it became a single-pane tab → serialized as empty → resurrected bare → repeating forever
- Root cause: `serialize_tab` always extracted `tiled_panes_layout.children` when split direction was the default (Horizontal), but for single-pane tabs the pane data lives on the root node with empty children

## Details

In `serialize_tab()` (`zellij-utils/src/session_serialization.rs`), the condition at line 98 decided whether to wrap `tiled_panes_layout` or use its `.children`:

```rust
// Before (buggy): single-pane tabs hit the else branch → empty children
let tiled_panes = if &tiled_panes_layout.children_split_direction
    != &SplitDirection::default()
    || tiled_panes_layout.children_are_stacked
{
    vec![tiled_panes_layout]
} else {
    tiled_panes_layout.children  // empty for leaf nodes!
};
```

For a leaf `TiledPaneLayout` (returned by `get_tiled_panes_layout_from_panegeoms` for a single pane), `children` is an empty `Vec` while the pane's `run`, `borderless`, etc. are on the root. The fix adds a guard:

```rust
// After: wrap the root node when it's a leaf with actual pane data
let tiled_panes = if (tiled_panes_layout.children.is_empty() && has_tiled_panes)
    || &tiled_panes_layout.children_split_direction != &SplitDirection::default()
    || tiled_panes_layout.children_are_stacked
{
    vec![tiled_panes_layout]
} else {
    tiled_panes_layout.children
};
```

`has_tiled_panes` distinguishes truly empty tabs (0 input panes → keep empty) from single-pane tabs (1 input pane → wrap the root node).

Related: #4129

## Test plan

- [x] Added `single_pane_tab_is_not_serialized_as_empty` — single terminal pane tab
- [x] Added `single_plugin_pane_tab_is_not_serialized_as_empty` — single borderless plugin pane (zjstatus-like)
- [x] Added `terminal_plus_plugin_status_bar_tab_serialization` — 2-pane tab (terminal + zjstatus plugin)
- [x] Added `multi_tab_with_single_and_multi_pane_tabs` — mixed session with 2-pane, single-pane, and empty tabs
- [x] Updated `can_serialize_multiple_tabs` snapshot (its "First tab" single-pane now correctly includes `pane`)
- [x] All 18 session_serialization tests pass
- [x] Full zellij-utils test suite (374 tests) passes

### DIFF
--- a/zellij-utils/src/session_serialization.rs
+++ b/zellij-utils/src/session_serialization.rs
@@ -90,13 +90,14 @@ fn serialize_tab(
     floating_panes: &Vec<PaneLayoutManifest>,
     pane_contents: &mut BTreeMap<String, String>,
 ) -> Option<KdlNode> {
+    let has_tiled_panes = !tiled_panes.is_empty();
     let mut serialized_tab = KdlNode::new("tab");
     let mut serialized_tab_children = KdlDocument::new();
     match get_tiled_panes_layout_from_panegeoms(tiled_panes, None) {
         Some(tiled_panes_layout) => {
             let floating_panes_layout = get_floating_panes_layout_from_panegeoms(floating_panes);
-            let tiled_panes = if &tiled_panes_layout.children_split_direction
-                != &SplitDirection::default()
+            let tiled_panes = if (tiled_panes_layout.children.is_empty() && has_tiled_panes)
+                || &tiled_panes_layout.children_split_direction != &SplitDirection::default()
                 || tiled_panes_layout.children_are_stacked
             {
                 vec![tiled_panes_layout]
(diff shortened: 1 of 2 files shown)


stefanpartheym:
I think what currently also isn't covered yet, that brand new sessions, that never got serialized and also never changed layout (no new split, no new tab, etc.) do not trigger the `session_layout_metadata.is_dirty()` check in `pty.rs` (line 755).
It's just a minor case, since usually session layouts change. But I have this frequently with my "default" session, that I use as kind of a scratchpad session to just run some command, but still kinda annoying.

PR #5055, opened 2026-04-16
Fix session-manager and tab issue

### LINKED ISSUE #3834: Session manager focuses tab that last ran it

## Issue description
When opening the session manager more than once, zellij will focus whatever tab that last ran it.
this occurs even when switching sessions.

## Minimal reproduction
- create a new session or open an existing one.
- have at least 2 tabs in the layout. i will refer to them as `tab A` and `tab B`.
- navigate to `tab A`
- open session manager
- do anything with it. switch sessions, or back out of it with `Esc`.
- navigate to `tab B`.
- open session manager again

it is at this point that zellij will unexpectedly navigate to `tab A`, since that was the last tab that the session manager was attached to. not only that, but the session manager will also not be opened at this point either, and required the user to hit the keybind to open it again.

## Other relevant information

after consulting with others in the discord, we determined the root cause is that the `Esc` key **_hides_** the session manager instead of closing it altogether. opening the session manager and then hitting the `close pane` keybind actually closes the manager, so this issue doesnt occur again (until you use `Esc` to back out of the session manager again, of course)

we believe the fix would be to just close the session manager instead of hiding it.

relevant code: https://github.com/zellij-org/zellij/blob/main/default-plugins/session-manager/src/main.rs#L394

Edit: this behaviour has been confirmed to occur on zellij version 0.41.2 and with a default `config.kdl`

cossio:
Fix #3834.

### DIFF
--- a/default-plugins/session-manager/src/main.rs
+++ b/default-plugins/session-manager/src/main.rs
@@ -607,7 +607,7 @@ impl State {
                         self.reset_selected_index();
                     } else if !self.is_welcome_screen {
                         self.reset_selected_index();
-                        hide_self();
+                        close_self();
                     }
                     should_render = true;
                 },
@@ -620,7 +620,7 @@ impl State {
                         self.renaming_session_name = None;
                         should_render = true;
                     } else if !self.is_welcome_screen {
-                        hide_self();
+                        close_self();
                     }
                 },
                 BareKey::Char('a') if key.has_modifiers(&[KeyModifier::Ctrl]) => {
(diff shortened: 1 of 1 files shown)


imsnif:
Hi, thanks for noting this. I'd be happy to accept the fix. However, please remove copilot from the commit log. As a policy we avoid advertising commercial products who do not directly sponsor the project.

cossio:
@imsnif done.

cossio:
@imsnif friendly bump :)

imsnif:
Thanks for the reminder!

PR #5057, opened 2026-04-16
fix: fall back to modified time when creation time is unavailable on Linux

Kai-Liu001:
## Problem

On Linux with ext4 filesystem, `std::fs::Metadata::created()` returns `ErrorKind::Unsupported` because creation time (btime) is not exposed through the standard Rust API on this platform.

In `background_jobs.rs`, the resurrection file listing function calls `metadata.created()` in a loop. When it fails, the error is logged but the function continues normally — however, **the caller re-invokes this function repeatedly (~1s interval)**, causing the error to be logged indefinitely.

In a long-running session (17 days, 11 tabs, 13 panes), this contributed to severe degradation:

- **193% CPU** (2 cores saturated — `screen` and `plugin-exec-1` threads stuck in Running state)
- **1.4 GB RSS** (vs ~50 MB for healthy sessions), peak 3.9 GB
- 17,307 "Broken pipe" errors in 25 minutes
- `MouseEvent` action timeouts every second
- UI completely frozen, required SIGKILL

Log excerpt (repeated ~14,000 times over ~4 hours before rotation):
```
ERROR | zellij_server::background | background_jobs.rs:767:
  Failed to read created stamp of resurrection file:
  Error { kind: Unsupported, message: "creation time is not available on this platform currently" }
```

## Fix

- Fall back to `metadata.modified()` when `metadata.created()` is unavailable, instead of logging an error and returning `None`
- This provides a reasonable approximation of session age on platforms where creation time is not supported
- Also separates the `metadata()` call from `created()` to properly handle metadata read failures

## Environment where the bug was observed

- **Zellij:** 0.44.0 (verified same code in 0.44.1)
- **OS:** Ubuntu 22.04.5 LTS, kernel 5.15.0-173-generic
- **Filesystem:** ext4

### DIFF
--- a/zellij-server/src/background_jobs.rs
+++ b/zellij-server/src/background_jobs.rs
@@ -755,23 +755,25 @@ fn find_resurrectable_sessions(
                         return None;
                     }
                     let layout_file_name = session_layout_cache_file_name(&session_name);
-                    let ctime = match std::fs::metadata(&layout_file_name)
-                        .and_then(|metadata| metadata.created())
-                    {
-                        Ok(created) => Some(created),
+                    let metadata = match std::fs::metadata(&layout_file_name) {
+                        Ok(m) => m,
                         Err(e) => {
                             if e.kind() == std::io::ErrorKind::NotFound {
-                                return None; // no layout file, cannot resurrect session, let's not
-                                             // list it
-                            } else {
-                                log::error!(
-                                    "Failed to read created stamp of resurrection file: {:?}",
-                                    e
-                                );
+                                return None; // no layout file, cannot resurrect session
                             }
-                            None
+                            log::error!(
+                                "Failed to read metadata of resurrection file: {:?}",
+                                e
+                            );
+                            return None;
                         },
                     };
+                    // metadata.created() is not available on some Linux
+                    // filesystems (e.g. ext4), fall back to modified time
+                    let ctime = metadata
+                        .created()
+                        .or_else(|_| metadata.modified())
+                        .ok();
                     let elapsed_duration = ctime
                         .map(|ctime| {
                             Duration::from_secs(ctime.elapsed().ok().unwrap_or_default().as_secs())


alunduil:
Adding another affected platform: Crostini (Debian guest, kernel 6.6.99, ext4). Same `ErrorKind::Unsupported` from `metadata.created()` and same ~1 Hz spam in `/tmp/zellij-1000/zellij-log/zellij.log`. Independent of your Ubuntu/5.15 setup so this isn't just a one-distro thing.

Worth noting the sibling path in the same file at `background_jobs.rs:754-758` already does what your fix does — `created().ok().or_else(|| modified().ok())` — so this brings `find_resurrectable_sessions` in line with existing precedent.

One amplification mode I hit, not really a zellij issue but maybe useful context: my Claude Code hook fires `zellij pipe --name zellaude` on every tool call, and when the daemon is busy in this loop each invocation blocks in `unix_stream_recvmsg` and never returns. I ended up with around 1,300 orphan `zellij pipe` clients and 2,600 orphan hook parents over about five hours before new `zellij` invocations queued behind them and looked frozen. Removing the trigger removes the whole cascade for me.

On the merge conflict, I think it's just #5063 (`aa29cbaf`) reshuffling surrounding code in `background_jobs.rs` — the offending block looks structurally unchanged so your fix should rebase cleanly. I'm happy to rebase and push to a successor branch if you'd like, just let me know. Otherwise I'll plan to open a superseding PR crediting you in a couple of weeks if there's no movement here.

jwon:
Confirming this reproduces on another environment:

- **Zellij:** 0.44.3 (static musl build)
- **OS:** Linux (CBL-Mariner / Azure Linux), kernel 5.15
- **Filesystem:** ext4

Same symptom as the PR description / #4165 — `background_jobs.rs:808` logs

```
Failed to read created stamp of resurrection file:
Error { kind: Unsupported, message: "creation time is not available on this platform currently" }
```

at ~60–90 lines/sec continuously. `/tmp/zellij-<uid>/zellij-log/` grew to ~16 MB (plus a rotated `.old.0`), and several long-lived servers were each sitting at 7–22% CPU.

Worth noting this is a **static musl** build, so it isn't covered by #4645 (which fixed `zellij ls`) — the resurrection listing path still calls `created()` and spams. The `modified()` fallback here would resolve it. 👍 Would love to see this merged.

Zykino:
With which rust version did you compile rust? **At least 1.87** as explained in the thread linked by jwon (just above)?

Or did you use the provided binary from the release page? Both this options should fix the issue without any code modification. So I don't think code change will be accepted to fix this issue.

PR #5065, opened 2026-04-17
fix: address path traversal, server DoS, permission catch-all, and silent error vulnerabilities

pimentelleo:
## Summary

This PR addresses 4 security vulnerabilities discovered during a comprehensive codebase audit.

### 1. Path Traversal in Plugin Sandbox (HIGH)
**File:** `zellij-server/src/plugins/zellij_exports.rs` — `translate_plugin_path()`

Plugins could escape their sandbox by using `..` path components (e.g. `/host/../../etc/passwd`). The function now canonicalizes resolved paths and verifies they remain within the expected base directory, mirroring the existing guard in `scan_host_folder()`.

### 2. Server DoS via IPC Connection Error (HIGH)
**File:** `zellij-server/src/lib.rs`

A `panic!()` in the IPC listener loop meant that any malformed or abruptly-closed client connection would crash the entire server, killing all sessions for all users. Replaced with `log::error!` + `continue` so the server keeps running.

### 3. Overly Permissive Plugin Permission Catch-All (MEDIUM)
**File:** `zellij-server/src/plugins/zellij_exports.rs` — `check_command_permission()`

The `_ => Granted` catch-all meant any new `PluginCommand` variant added without updating the permission check would be silently granted access. Now explicitly enumerates the safe self-management commands (Subscribe, SetTimeout, ShowSelf, etc.) and denies unrecognized commands by default with a warning log.

### 4. Silent Permission Error on Token DB (LOW)
**File:** `zellij-utils/src/web_authentication_tokens.rs`

`let _ = std::fs::set_permissions(...)` silently discarded errors when restricting the token database to owner-only access (mode 0600). Replaced with `log::warn` so operators are alerted if the security-critical permission cannot be set.

## Changes

- 3 files changed, 69 insertions, 10 deletions
- All changes compile cleanly (`cargo check`)

## Out of Scope (Future Work)

The audit also identified larger-scope issues that warrant separate PRs:
- Unmaintained dependencies: `nix` 0.23.1 (RUSTSEC-2024-0064), `isahc` (RUSTSEC-2024-0444), `ansi_term` (RUSTSEC-2021-0139)
- `inherit_env()` leaking all host environment variables to plugins
- No checksum verification for downloaded plugin WASMs
- Deprecated `serde_yaml` dependency

### DIFF
--- a/zellij-server/src/lib.rs
+++ b/zellij-server/src/lib.rs
@@ -760,7 +760,8 @@ pub fn start_server(mut os_input: Box<dyn ServerOsApi>, socket_path: PathBuf) {
                                 .unwrap();
                         },
                         Err(err) => {
-                            panic!("err {:?}", err);
+                            log::error!("Failed to accept IPC connection: {:?}", err);
+                            continue;
                         },
                     }
                 }
--- a/zellij-server/src/plugins/zellij_exports.rs
+++ b/zellij-server/src/plugins/zellij_exports.rs
@@ -140,19 +140,46 @@ macro_rules! apply_action {
 }
 
 fn translate_plugin_path(env: &PluginEnv, path: PathBuf) -> PathBuf {
-    if let Ok(stripped) = path.strip_prefix("/host") {
-        env.plugin_cwd.join(stripped)
+    let (base_dir, joined) = if let Ok(stripped) = path.strip_prefix("/host") {
+        (Some(&env.plugin_cwd), env.plugin_cwd.join(stripped))
     } else if let Ok(stripped) = path.strip_prefix("/data") {
-        env.plugin_own_data_dir.join(stripped)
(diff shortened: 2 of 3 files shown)


pimentelleo:
What is the actual status of this? It will be implemented?
It is a lot insecure to maintain our collaborative work on a such vunerable tool.

imsnif:
None of these are actual security vulnerabilities. All the commands in 1 are behind relevant permissions (and so this functionality is actually desired), the assertion in 2 is incorrect (and even if it were, it's most definitely within the security model of our client/server model). etc. etc.

Thanks for the effort, if you have a specific scenario you'd like to discuss, you are welcome to do so. If you don't like the tool, by all means you can use another.

pimentelleo:
Sure! Thanks for your attention.

PR #5068, opened 2026-04-18
fix(screen): prevent pane flash when process exits before NewPane is processed

cscnk52:
When a spawned process exits near-instantaneously (e.g. a shell that fails to start under ConPTY on Windows), the exit-monitoring thread fires quit_cb and sends ClosePane to the screen before the PTY thread has had a chance to send NewPane.

93c7ef52 (Fix pty race when opening new panes, #5010) introduced pending_events_waiting_for_pane to avoid zombie panes in this case: the ClosePane is queued, then replayed after NewPane arrives. However, the replay re-sends the event back onto the screen channel, so it is processed in the *next* event-loop iteration – after a render has already shown the pane to the user, causing a visible flash.

Fix this by handling a pending ClosePane synchronously inside the NewPane handler, before the render call. The pane is created and closed within the same iteration so it is never visible. Buffered PtyBytes for the same pane are discarded since the process is already dead and the bytes will never be displayed. HoldPane and all other pending event types continue to be re-queued as before.

Resolve: #5054

AI disclosure: The work is done by Claude Code, and review by me.

Issue #5054, opened 2026-04-16
Windows: stack pane can't recreate when closed


cscnk52:
# 1. Graphical issue inside a terminal pane (eg. something does not look as it should)

1. Delete the contents of `/tmp/zellij-1000/zellij-log`, ie with `cd /tmp/zellij-1000/` and `rm -fr zellij-log/` (`/tmp/` is `$TMPDIR/` on OSX)
2. Run `zellij --debug`
3. Run `stty size`, copy the result and attach it in the bug report
4. Recreate your issue.
5. Quit Zellij immediately with ctrl-q (your bug should ideally still be visible on screen)

Please attach the files that were created in `/tmp/zellij-1000/zellij-log/` to the extent you are comfortable with.

**Basic information**

`zellij --version`: 0.45.0 (the latest dev version)

`stty size`:

`uname -av` or `ver`(Windows): Windows 11 25H2 Build 26200.8246

## Further information
Reproduction steps, noticeable behavior, related issues, etc

# 2. Issues with the Zellij UI / behavior / crash

## Issue description

When i use stack pane on windows, when the stack pane was closed, i can't recreate stack pane again. But it work as expect in my WSL, i think it's a bug related to windows.

## Minimal reproduction

https://github.com/user-attachments/assets/4710a5fc-de9d-4dde-bf8c-e4953fc9b6a6

## Other relevant information

<details>
<summary> log </summary>

```log
INFO   |zellij_client            | 2026-04-16 11:20:17.871 [main      ] zellij-client\src\lib.rs:726: Starting Zellij client! 
INFO   |zellij_server            | 2026-04-16 11:20:17.912 [main      ] zellij-server\src\lib.rs:655: Starting Zellij server! 
ERROR  |zellij_client::stdin_ansi| 2026-04-16 11:20:17.929 [stdin_handler] zellij-client\src\stdin_ansi_parser.rs:125: Failed to open STDIN cache file: Os { code: 2, kind: NotFound, message: "The system cannot find the file specified." } 
INFO   |zellij_server            | 2026-04-16 11:20:17.984 [main      ] zellij-server\src\lib.rs:811: FirstClientConnected: initializing session 
INFO   |zellij_server            | 2026-04-16 11:20:17.986 [main      ] zellij-server\src\lib.rs:2211: Loading plugins using Wasmi interpreter 
[...]

PR #5082, opened 2026-04-21
fix(windows): reuse of terminal pane ids breaks new panes from opening

Simoyd:
Hello,

I have no idea how this code works so I used AI to fix a critical issue I was having (described below). to be clear this code change does fix my issue, but this is **not necessarily intended to be accepted** verbatim. It is to provide insight to a real dev who knows what they're doing to make a real fix.

below is mostly AI generated but I read it all multiple times and it seems okay. my intention here is to help not be annoying with AI stuff. I'm sorry if I've offended anyone.

## Summary
Fix a Windows-specific pane reopen bug caused by terminal pane ids being reused after a pane is closed.

## Reproduction
using version 0.44.1 on windows. either cmd or powershell.
1. Open a new pane with `Alt+n`
2. Close that pane with `Ctrl+p`, then `x`
3. Open a new pane again

## Expected
A new pane opens normally.

## Actual
On Windows, the reopened pane is immediately closed / fails to stay open.

## Fix
Use a monotonic terminal id allocator on Windows so a stale deferred close event cannot target a newly created pane.

## Reasoning
On Windows, pane ids could be reused after a pane was closed. A late `ClosePane` event from the old pane could then be delivered after a new pane had already been created with the same id, causing the new pane to be closed immediately. The fix is to stop reusing terminal ids on Windows and allocate them monotonically instead, matching the Unix implementation.

## Validation
- Verified manually on Windows that pane reopen now works correctly after the fix
- `cargo xtask build` passed locally
- `cargo xtask format --check` passed locally
- `cargo xtask test` did not complete locally due to unrelated pre-existing Windows plugin-test linker failures (`host_run_plugin_command` unresolved in multiple plugin test targets)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 ## [Unreleased]
 * fix: idle CPU + disk i/o, CommandChanged plugin Event, GetSessionList plugin command (https://github.com/zellij-org/zellij/pull/5063)
 * fix: restore kitty keyboard mode on error, handle DECRPM 2026 state 3, clean up client terminal teardown sequences (https://github.com/zellij-org/zellij/pull/5058)
+* fix: stop reusing terminal pane ids on windows so reopened panes are not closed by stale events (https://github.com/zellij-org/zellij/pull/5082)
 
 ## [0.44.1] - 2026-04-07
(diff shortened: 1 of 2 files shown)


imsnif:
@divens ?

divens:
Can reproduce. Thanks for reporting. I'll look into the suggested fix.

cscnk52:
May this could close #5068 and #5054.

Issue #1988, opened 2022-12-03
[Feature Request] open a new pane or tab without moving focus


f3fora:
Actions/commands `zellij-action-new-pane`, `zellij-action-new-tab`, `zellij-run` should allow to keep the focus on the current instance and not to move it to the newly created one.

Tmux implements this feature with the `-d` flag:
`If -d is given, the session does not make the new window the current window.`

PR #5095, opened 2026-04-24
Feat/cli new tab no focus

wkrettek:
I created a `--no-focus` flag when creating new tabs that won't change the focus when using the cli. This is related to issue #1988. It looks like the hooks were already there in the server, so I just plugged into the existing functionality and exposed it. 

Thanks for reviewing!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 
 ## [Unreleased]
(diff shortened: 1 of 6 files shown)


davidlt:
It would be nice to have this one with all those agents around, doing something like:
`zellij action new-tab -d --name "claude-feature-a" --cwd $PWD/blah -- claude -n add-feature-a`

You stay on the same tab/pane with existing agent session, then sub-agent runs in a new tab and you can message between with it from the current tab/pane. If needed open that tab to talk directory to that sub-agent/agent.

wkrettek:
Closing because it looks like https://github.com/zellij-org/zellij/pull/5346 addresses this.

PR #5099, opened 2026-04-26
Implement OSC 9 and OSC 777 desktop notification protocols

### LINKED ISSUE #3954: Escape code passthrough

DaveDev42:
This is a follow-up to https://github.com/zellij-org/zellij/pull/4931 and implements OSC 9 and OSC 777, the two de-facto desktop-notification protocols that predate OSC 99 and that most modern host terminals (WezTerm, iTerm2, Kitty, Alacritty) still recognize. Without this, programs running inside a Zellij pane that emit these sequences see them silently dropped.

Fixes https://github.com/zellij-org/zellij/issues/3954.

This is **not** a raw escape-code passthrough. Each notification is parsed at the grid layer, every field is sanitized, and a fresh sequence is rebuilt from a fixed template before it is emitted to clients — user-supplied bytes never reach the host terminal as part of an unescaped sequence. The forward path itself mirrors OSC 99: notifications are queued at the grid level and re-emitted to all connected clients of the app. Because OSC 9 / 777 are fire-and-forget — there is no host response and no identifier to route — we don't need the namespacing dance from the OSC 99 work, and we can apply stronger input handling instead:

1. Each notification is parsed into a typed `LegacyDesktopNotification` enum with an `Osc9 { text }` variant and an `Osc777Notify { title, body }` variant. The host sequence is rebuilt from a fixed template (`to_canonical_bytes`), so user-supplied bytes never reach a format string unescaped.
2. Every field is run through `sanitize_legacy_notification_field`, which strips C0, DEL, and C1 control characters and caps the length at 1024 chars. This prevents a program inside a pane from injecting additional escape sequences (further OSC, CSI, title-set, file download) into the host terminal via the notification body.
3. OSC 777 is restricted to the `notify` subcommand. Other subcommands like `preexec`, `precmd`, and `cwd` are intentionally dropped — they are not notifications and forwarding them would leak shell hooks across the multiplexer boundary.
4. The terminator is normalized to BEL on emit; the input's ST is ignored. All common host terminals accept BEL.

Tests cover OSC 9 and OSC 777 parsing, control-character stripping, Unicode and tab preservation, the length cap, the canonical re-emit, and the silent drop of non-`notify` OSC 777 subcommands.


matteo-cavallo:
What's the status? Is it planned to implement it soon?

Dreaming-Codes:
@imsnif any way you can look into this, a lot of TUI uses this

imsnif:
Implemented today actually, as part of a bigger change: https://github.com/zellij-org/zellij/pull/5485

My apologies for not having seen this through, thanks to the implementers regardless.

Issue #3590, opened 2024-09-04
OSC 10 / OSC 11 queries report incorrect colors


bash:
## Issue description
When querying for the foreground or background color using the corresponding OSC sequences, zellij always answers with black.

For instance, when I send `OSC 11 ; ? ST` (query background) I receive back `OSC 11 ; rgb:0000/0000/0000 ST` even though I'm in a light terminal. Similarly in a dark terminal `OSC 10 ; ? ST` (query foreground) is responded to with `OSC 10 ; rgb:0000/0000/0000 ST`.

Example:
```shell
$ printf '\e]11;?\e\\' && cat -v
^[]11;rgb:0000/0000/0000^[\
```

## Collected information

**Basic information**

`zellij --version`: `zellij 0.40.1`

`stty size`: `26 102`

`uname -av` or `ver`(Windows): `Linux taus-fedora 6.8.1-300.fc40.x86_64 #1 SMP PREEMPT_DYNAMIC Wed Mar 20 04:39:30 UTC 2024 x86_64 GNU/Linux`

Terminal: Alacritty and vte (Ptyxis) 

**Logs**
[zellij-12.log](https://github.com/user-attachments/files/16870776/zellij-12.log)
[zellij.log](https://github.com/user-attachments/files/16870777/zellij.log)

## Further information

I noticed that when I clean the cache directory (`~/.cache/zellij`) I get correct results for the first run of zellij.

Originally this issue was reported in delta (https://github.com/dandavison/delta/issues/1663). Reporting this here totally fell off my radar (sorry).

Nikhiludupa:
try this
https://gofile.io/d/yMHabA
select the desired compiler in the installer

bash:
@Nikhiludupa This feels like spam?

jmodrako:
Similar behaviour here. I'm using Omakub and when I've been switching themes, the NeoVim has had background=light set for some dark themes. I've checked the OSC11 response and it seems that it was wrong. After clearing the cache, all started working properly :)

linuxmobile:
I did some digging on this and found the root cause.

**TL;DR: Zellij caches the OSC 10/11 response and never invalidates it.**

When Zellij starts, it queries the terminal for fg/bg colors and saves the response to `~/.cache/zellij/`. On subsequent sessions, it returns the cached values instead of asking the terminal again.

The problem is this cache persists even when:
- You switch between dark/light themes
- You open Zellij in a different terminal
- The terminal's color scheme changes

[...]

PR #5105, opened 2026-04-28
Support CSI2031 (dark/light theme switching) + remove STDIN cache to avoid stale color queries

### LINKED ISSUE #3831: Allow automatic light/dark mode detection from applications running within zellij (CSI 2031)
Several terminal emulators now allow for automatic switching between light and dark themes, and allow applications running within the terminal to subscribe to theme changes. Automatic switching off all applications together is important to prevent situations where bright white and black windows end up side by side on a screen.

It would be great if zellij would understand CSI 2031, so that it could proxy these requests from applications to the outer terminal.

A few other terminal emulators and applications have implemented a simple protocol for dark/light mode detection and update notifications. For documentation, see: https://github.com/contour-terminal/contour/blob/f3c3334aa5c861348c5bbe8ffe572c872eef2e08/docs/vt-extensions/color-palette-update-notifications.md

Terminal emulators that are implementing CSI 2031 are:
- Ghostty
- Contour
- Kitty

Applications that take advantage of CSI 2031 are:
- Neovim: https://github.com/neovim/neovim/pull/31350
- (probably others.)

The request here is to make sure that applications running in zellij can query the current theme (light/dark), but can also use CSI 2031 to subscribe to theme updates.

One edge case that would complicate things for zellij is a situation where a single window is attached in two different terminals, one with a light theme and one with a dark theme. In that case zellij would have to choose what to present to the applications.

Related issues on other projects:
- tmux: https://github.com/tmux/tmux/issues/4269
- wezterm: https://github.com/wez/wezterm/issues/6454

imsnif:
### Dark/Light theme auto and manual switching
The main user-facing addition in this PR is support for light/dark theme switches, both through the [CSI2031 protocol](https://github.com/contour-terminal/contour/blob/master/docs/vt-extensions/color-palette-update-notifications.md) and manually through 3 new actions: `zellij action set-dark-theme`, `zellij action set-light-theme` and `zellij action toggle-theme`.

In addition, Zellij now supports the `dark_theme` and `light_theme` optional configuration parameters. So one can define both in the same way one now defines the default `theme`:

```kdl
theme "dracula" // default
theme_dark "dracula" // will be used when the host terminal notifies the desktop now uses a dark theme or through the action
theme_light "catppuccin-latte" // will be used when the host terminal notifies the desktop now uses a light theme or through the action
```

This also includes a `HostTerminalThemeChanged` new plugin Event to be fired when any of the above changes happen.

Fixes https://github.com/zellij-org/zellij/issues/3831

## Remove STDIN parsing cache
Due to some recent refactoring of the codebase, it has become possible to remove the notorious STDIN cache. That is the HD cache file Zellij creates on first run of each version, including the various responses received from the host terminal regarding the system (eg. color codes, pixel/cell ratio, etc.) Since this change was required in order to implement the above feature, I added it to this PR.

This means that now queries regarding these values (eg. OSC10/11) will be sent directly to the host terminal, with the reply streaming back into the relevant pane. Thus, these values will never be stale (unless no client is connected). It also greatly simplifies our STDIN loop and likely fixes some issues with partial sequence interpretation.

## Acknowledgement
This feature has been prioritized due to a generous sponsorship from @colemickens - your support is very much appreciated.


colemickens:
Seems to be working great:

This shows `zellij` configured with `theme_dark` and `theme_light`. Helix is similarly configured. So the video demonstrates both `zellij` changing it's theme in response to macOS changing light/dark mode, and also shows `helix` getting the signal and changing as well. Amazing!

https://github.com/user-attachments/assets/407bf36f-7fca-49db-bb2a-d2638cb3bf86

Ehab-Ibrahim:
I think this also fixes https://github.com/zellij-org/zellij/issues/3590, right?

tautropfli:
> This means that now queries regarding these values (eg. OSC10/11) will be sent directly to the host terminal, with the reply streaming back into the relevant pane

I would strongly recommend against this approach. It means that replies may arrive out-of-order with replies to other queries. 

For example: Sending an `OSC 10` query followed by `DA1` will result in the reply to `DA1` (handled by zellij) being sent first, folllowed by the reply to `OSC 10` from the host terminal.

This will break clients which assume that order is preserved e.g. `terminal-colorsaurus` [assumes this](https://docs.rs/terminal-colorsaurus/latest/terminal_colorsaurus/feature_detection/index.html) (for transparency: I am the author of said library).

imsnif:
> I would strongly recommend against this approach. It means that replies may arrive out-of-order with replies to other queries.

Fair point. Thanks for pointing this out. I think this is technically undefined behavior, but this method (tacking on a DA query after a query the program isn't sure is supported) is so prevalent in the ecosystem I think it's fair to say we should behave accordingly. Fixed: https://github.com/zellij-org/zellij/pull/5113

PR #5112, opened 2026-04-29
feat(web): support Unix socket listener

rirze:
## Summary
- add `zellij web --socket <SOCKET>` to serve the web client over a Unix domain socket
- support `zellij web --status --socket <SOCKET>` using Isahc Unix socket dialing
- share the web-client router between TCP/TLS and Unix listeners while keeping web-server shutdown support
- add CLI coverage and a Unix-socket version endpoint test

## Motivation
This gives reverse proxies a secure local backend path without binding Zellij to a non-loopback TCP interface or requiring backend TLS certificates. Caddy, Nginx, Traefik, and similar proxies can proxy to the socket while public TLS remains at the reverse proxy.

## Notes
This PR intentionally keeps the change CLI-only. A config key and public URL/share integration can be a follow-up because the current status/share plumbing assumes an IP/port URL.

The CLI tests now run their Clap parser work on a larger test-thread stack. Clap 3 recursively validates this large command graph, and adding another web option exceeded the default test thread stack even though normal binary execution parses the flag correctly.

## Validation
- `cargo fmt`
- `git diff --check`
- `cargo test -p zellij --no-default-features --features web_server_capability,disable_automatic_asset_installation tests::cli -- --nocapture`
- `cargo test -p zellij-client --features web_server_capability test_version_endpoint_over_unix_socket -- --nocapture`
- `cargo check -p zellij-utils`
- `cargo check -p zellij-client --features web_server_capability`
- `cargo check -p zellij --no-default-features --features web_server_capability,disable_automatic_asset_installation`

I used the `disable_automatic_asset_installation` top-level check because this checkout does not include the prebuilt default-plugin WASM files needed by the default `cargo check --features web_server_capability` path.

### DIFF
--- a/src/commands.rs
+++ b/src/commands.rs
@@ -5,6 +5,9 @@ use std::{fs::File, io::prelude::*, path::PathBuf, process, time::Duration};
 #[cfg(feature = "web_server_capability")]
 use isahc::{config::RedirectPolicy, prelude::*, HttpClient, Request};
 
+#[cfg(all(feature = "web_server_capability", unix))]
+use isahc::config::Dialer;
+
 use zellij_client::{
     old_config_converter::{
         config_yaml_to_config_kdl, convert_old_yaml_files, layout_yaml_to_layout_kdl,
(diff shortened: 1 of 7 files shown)


IngmarStein:
I would love to see this merged, thanks for creating the PR!

rirze:
I would really love some contributor review this but it's been a while and I'm using a dummy socket listener to sit between my reverse proxy and zellij web since opening this PR, which is inefficient and roundabout. 

I really believe this approach will let end users self host zellij in a normal industry standard (since http hosting is specifically against zellij's security policy). It's a good compromise and hope we can move this approach into zellij itself.

PR #5119, opened 2026-05-01
fix: do not panic the screen thread when ClosePane cannot reach background_jobs

simon0191:
## Problem

When a pane closes (e.g. its child process exits), the screen thread's `ClosePane` handler in `screen.rs` propagates errors from `log_and_report_session_state()` via `?`. That function calls `send_to_background_jobs` (twice) and `send_to_plugin`. If `background_jobs_main` has already returned — cleanly via `BackgroundJob::Exit` or after a panic — the channel is closed and the send returns `SendError`. The `?` propagates it, `screen_thread_main` returns `Err`, and `.fatal()` at `lib.rs:1560` panics the screen thread. The server stays alive but its UI is dead, so `zellij attach <session>` hangs forever after "Loading Zellij... done" — the IPC handshake completes but no rendering thread exists to draw anything.

Reproduced twice in the wild on `v0.43.1` running on macOS. Reproducer sequence: a pane runs `claude --worktree X` (Claude Code, in worktree mode); user presses Ctrl+C twice → claude prompts "delete worktree?" → user confirms → claude removes its worktree and exits. The pane-close fires `ScreenInstruction::ClosePane` (sent by `pty.rs`'s `quit_cb` on natural child exit); if `background_jobs` is unreachable at that moment, the server is dead.

The same arm exists unchanged on `main` (verified by reading `screen.rs`). The wild panic from logs:

```
Panic occured:
    thread: screen
    location: At zellij-server/src/lib.rs:1560:18

Caused by:
    0: Failed to log and report session state
    1: failed to send message to background jobs
    2: Originating Thread(s)
        1. screen_thread: ClosePane
    3: failed to send message to channel
```

## Fix

Treat `log_and_report_session_state()` as best-effort in the `ClosePane` arm, matching the precedent at `screen.rs:4455` (and elsewhere in the same file). Use `non_fatal()` so the error is logged but not propagated.

## Test

Adds `close_pane_does_not_panic_when_background_jobs_unreachable` in `zellij-server/src/unit/screen_tests.rs`, plus a `MockScreen::run_without_silent_failures` helper. The existing `MockScreen::run` constructs the screen bus with `.should_silently_fail()`, which makes every `send_to_*` return `Ok(())` regardless of channel state — that flag had to be opt-out for the new test, since otherwise the bug is unreproducible from a unit test. Existing tests still default to silent-failure mode (no behaviour change). The test sends `ScreenInstruction::ClosePane` directly, mirroring the production `quit_cb` path on natural child exit.

Verified: the new test fails on `main` without the `screen.rs` change and passes with it. Full `cargo test -p zellij-server --lib` runs clean (1081 pass, 0 fail).

## Out of scope

Other arms in `screen.rs` use the same `log_and_report_session_state()?` pattern (~17 sites, including `CloseFocusedPane` and `HoldPane`). A follow-up sweep is worthwhile but kept out of this PR to keep the diff focused on the path that triggered the wild crash.

The parallel `.fatal()` calls in `pty.rs` (where `TerminalBytes::listen()` panics the runtime when the screen channel is closed) cascade after the screen-thread panic this PR fixes. The `terminal_bytes.rs:78–90` block-comment already acknowledges that wart; addressing it is a separate change.


imsnif:
Hey, `log_and_report_session_state` should never crash. Happy to accept the fix, but there's no need to add a regression text for it.

Also, please remove the commits from Claude (we don't give commercial companies ad space in our commit log unless they sponsor the project), as well as the verbose comment. Essentially, you can just change it to add `non_fatal()` in the relevant place and that should do it.

Thanks!

imsnif:
My apologies - this has already been fixed as part of a different change!

PR #5129, opened 2026-05-04
fix(sixel): clip overlaps and reap fully erased images

pacjuvenile:
Fixes stale/partial sixel image artifacts when image previews are replaced frequently, such as in yazi inside Zellij.

  This version avoids the previous broad behavior change and keeps partial overwrite behavior closer to bare WezTerm.

  Changes:

  - Fix full-cover detection in `end_image()`: `PixelRect::intersecting_rect()` returns coordinates relative to the old image, but the
previous full-cover check compared them against the old image's absolute coordinates. This meant fully covered images whose origin was not
`(0, 0)` were not detected as fully covered.
  - When a new sixel image partially overlaps an old one, clip only the overlapping region from Zellij's stored old image.
  - When text overwrites an image, keep using partial `cut_out()`.
  - If repeated text/space overwrites erase all pixels in an image, reap the now-empty image as cleanup.
  - Allocate new image IDs with `max(id) + 1` instead of `len()` to avoid reusing an ID while older images may still be queued for
reaping.

  I tested the original issue with yazi image previews in Windows + WezTerm + WSL + Zellij. With upstream/main, switching previews often
leaves the image blank or partially rendered. With this patch, previews switch normally.

  Local tests:

  ```sh
  cargo fmt --check
  cargo test -p zellij-server sixel
```

### DIFF
--- a/zellij-server/src/panes/sixel.rs
+++ b/zellij-server/src/panes/sixel.rs
@@ -160,24 +160,24 @@ impl SixelGrid {
                     image_pixel_size.1,
                 );
 
-                // here we remove images which this image covers completely to save on system
-                // resources - TODO: also do this with partial covers, eg. if several images
-                // together cover one image
+                let mut images_to_reap = vec![];
+                let mut images_to_cut_out = vec![];
                 for (image_id, pixel_rect) in &self.sixel_image_locations {
                     if let Some(intersecting_rect) =
                         pixel_rect.intersecting_rect(&image_size_and_coordinates)
                     {
-                        if intersecting_rect.x == pixel_rect.x
-                            && intersecting_rect.y == pixel_rect.y
-                            && intersecting_rect.height == pixel_rect.height
-                            && intersecting_rect.width == pixel_rect.width
-                        {
-                            self.image_ids_to_reap.push(*image_id);
+                        if rect_covers_image(&intersecting_rect, pixel_rect) {
+                            images_to_reap.push(*image_id);
+                        } else {
+                            images_to_cut_out.push((*image_id, intersecting_rect));
                         }
                     }
                 }
-                for image_id in &self.image_ids_to_reap {
-                    self.sixel_image_locations.remove(image_id);
+                for image_id in images_to_reap {
+                    self.queue_image_to_reap(image_id);
+                }
+                for (image_id, rect_in_image_to_cut_out) in images_to_cut_out {
+                    self.remove_pixels_from_image(image_id, rect_in_image_to_cut_out);
                 }
 
                 self.sixel_image_locations
@@ -262,7 +262,13 @@ impl SixelGrid {
         }
     }
     pub fn next_image_id(&self) -> usize {
-        self.sixel_image_store.borrow().sixel_images.keys().len()
+        self.sixel_image_store
+            .borrow()
+            .sixel_images
+            .keys()
+            .max()
+            .map(|id| *id + 1)
+            .unwrap_or(0)
     }
     pub fn new_sixel_image(&mut self, sixel_image_id: usize, sixel_image: SixelImage) {
         self.sixel_image_store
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by imsnif** [comment]
Hey @pacjuvenile - while I understand there are some issues here, I remember implementing this explicitly because this was the behavior of other terminal emulators supporting this protocol. Could you please test this change in other bare compliant terminals (eg. wezterm, foot, xterm with sixel) and see how they behave?

**Comment by pacjuvenile** [comment]
Thanks for the context.I tested this on Windows with WezTerm + WSL.

  I attached three recordings:

  - `screenshot.gif`: yazi running directly in bare WezTerm, without Zellij
  
  screenshots

  - `screenshots_1.gif`: yazi running inside Zellij with current upstream/main
  
screenshots_1

  - `screenshots_2.gif`: yazi running inside Zellij with this PR applied.
  
  screenshots2

  The bare WezTerm case behaves correctly: when moving through image previews in yazi, the images are rendered consistently.

  With current upstream/main inside Zellij, the same yazi preview workflow frequently leaves the preview blank or partially rendered while switching files.

  With this PR applied, the behavior inside Zellij matches the bare WezTerm behavior much more closely: image previews continue to render correctly while moving between files.

  So from this test, the issue seems specific to Zellij’s sixel image handling rather than WezTerm/yazi themselves.

**Comment by imsnif** [comment]
Thanks, but I mean regarding the behavior of sixel images and text interacting that is changed in this pr.

**Comment by azinsharaf** [comment]
> Thanks for the context.I tested this on Windows with WezTerm + WSL.
> 
> I attached three recordings:
> 
>     * `screenshot.gif`: yazi running directly in bare WezTerm, without Zellij
> 
> 
> screenshots
> 
>     * `screenshots_1.gif`: yazi running inside Zellij with current upstream/main
> 
> 
> screenshots_1
> 
>     * `screenshots_2.gif`: yazi running inside Zellij with this PR applied.
> 
> 
> screenshots2
> 
> The bare WezTerm case behaves correctly: when moving through image previews in yazi, the images are rendered consistently.
> 
> With current upstream/main inside Zellij, the same yazi preview workflow frequently leaves the preview blank or partially rendered while switching files.
> 
> With this PR applied, the behavior inside Zellij matches the bare WezTerm behavior much more closely: image previews continue to render correctly while moving between files.
> 
> So from this test, the issue seems specific to Zellij’s sixel image handling rather than WezTerm/yazi themselves.

have you tested it in Windows nativity (outside of WSL)? I can test it if you need it.

**Comment by pacjuvenile** [comment]
@azinsharaf   I have only tested under WSL so far, not native Windows. If you can test native Windows, that would be helpful.

**Comment by pacjuvenile** [comment]
@imsnif Thanks, I tested the two specific interactions in bare WezTerm under WSL.
                                                                                                                                         
Environment:
                                                                                                                                         
- Windows + WezTerm + WSL
- `chafa -f sixels`
- no Zellij involved in these two screenshots
                                                                                                                                         
### 1. Text overwriting a sixel image
                                                                                                                                         
Command:
                                                                                                                                         
```sh
clear
chafa -f sixels --size=40x20 "$IMG1"
printf '\033[5;10HHELLO_TEXT_OVER_IMAGE'
```
                                                                                                                                         
Result in bare WezTerm: the image is partially overwritten/clipped where the text intersects it. The whole image is not removed.
                                                                                                                                         
demo-202605162230
                                                                                                                                         
### 2. One sixel image overlapping another
                                                                                                                                         
Command:

```sh                                                                                                                                       
clear
printf '\033[5;10H'
chafa -f sixels --size=40x20 "$IMG1"
printf '\033[8;20H'
chafa -f sixels --size=40x20 "$IMG2"
```        
                                                                                                                             
Result in bare WezTerm: the second image partially covers the first image. The first image is not removed entirely.
                                                                                                                                         
demo-202605162231
                                                                                                                                         
So, this PR does not exactly match bare WezTerm's behavior. Bare WezTerm keeps the remaining parts of the old image, while this PR intentionally takes a more conservative approach in Zellij: if text or a new image intersects an existing sixel image, the old image is reaped.
                                                                                                                                         
My reasoning is that, in a multiplexer, keeping partially invalid sixel image state can easily leave visually broken previews or stale image fragments, especially in applications like yazi where previews are replaced frequently. Reaping the intersected image gives a cleaner and more predictable result.
                                                                                                                                         
That said, if the preferred direction is to match bare terminal behavior more closely, I can adjust the PR to implement partial clipping instead of removing the whole image.

imsnif:
Yeah - you really need to reverse engineer every minuscule source of behavior with this protocol, specified or unspecified, because you can never know which application depends on it.

If there is an issue in a specific case and you find that it can be solved by changing said minuscule behavior - even if it seems to make sense - the problem is likely elsewhere.

If you'd like to keep looking at it, I'd try looking in that sort of direction. Friendly advice: be extremely skeptical of anything an LLM confidently tells you in this area. It's a niche and underspecified protocol - they know even less than they usually do.

pacjuvenile:
@imsnif   Thanks, I agree with this direction. I reworked the PR to remove the intentional broad behavior change from the earlier version.

  I tested the two interactions in bare WezTerm:

  - text over a sixel image: partial overwrite/clipping, not whole-image removal
  - sixel image over another sixel image: partial overlap, not whole-image removal

  The updated patch now tries to preserve that behavior more closely:

  - text overwrites still use partial `cut_out()`
  - partial image/image overlap only clips the overlapping region from Zellij's stored old image
  - full image coverage reaps the old image
  - if repeated text/space overwrites erase all pixels in an image, the now-empty image is reaped as cleanup

  The concrete bug I found while reworking this is that `PixelRect::intersecting_rect()` returns coordinates relative to the old image, but the old full-cover check compared those coordinates against the old image's absolute coordinates. So fully covered images away from `(0, 0)` were not detected correctly.

  I retested the yazi case after this narrower change in Windows + WezTerm + WSL + Zellij, and the preview switching issue is still fixed.

  `cargo fmt --check` and `cargo test -p zellij-server sixel` pass locally.

PR #5135, opened 2026-05-06
feat: reload config on theme change

### LINKED ISSUE #5085: Handle SIGUSR for config-reload
Currently, the zellij config reloads and updates the themes if the config.kdl is touched. But if used with nix and home-manager that file is read-only. I have specialisations that update a theme, but that alone does not trigger the config reload as the theme is in a separate file. 

It would be amazing if either one of these options would be supported:
- also reload config if any of the theme files change
- reload config on SIGUSR1 signal (similar to how helix/waybar handles reloads)

hannahfluch:
closes #5085


imsnif:
Hey @hannahfluch - thanks for this. I fixed a super minor TOCTOU issue and added some logging. Will merge once CI passes.

PR #5139, opened 2026-05-06
fix(windows): bump windows-sys to 0.59

### LINKED ISSUE #5136: Can't update with Cargo on Windows
Trying to update zellij to the latest version on Windows using `cargo binstall zellij`, but while compiling zellij-server, I get an error:
```
error[E0432]: unresolved import `windows_sys::Win32::Storage::FileSystem::WriteFile`
  --> C:\Users\edvinas.steponaviciu\.cargo\registry\src\index.crates.io-1949cf8c6b5b557f\zellij-server-0.44.2\src\os_input_output_windows.rs:21:36
   |
21 |     CreateFileW, FlushFileBuffers, WriteFile, FILE_FLAG_OVERLAPPED, OPEN_EXISTING,
   |                                    ^^^^^^^^^ no `WriteFile` in `Win32::Storage::FileSystem`

For more information about this error, try `rustc --explain E0432`.
error: could not compile `zellij-server` (lib) due to 1 previous error
```

divens:
The workspace `windows-sys` constraint was pinned to 0.52 while the Windows source uses APIs and namespaces from 0.59 (e.g. WriteFile in Win32::Storage::FileSystem). CI built fine because Cargo.lock pinned 0.59, but downstream `cargo install`/`cargo binstall` did fresh resolution and selected 0.52, breaking the build for users on Windows.

Align the manifest with the code:
- Bump workspace dep `windows-sys` from 0.52 to 0.59.
- HANDLE is now `*mut c_void` instead of `isize`: replace `handle == 0` / `!= 0` with `is_null()`, initialise HANDLE locals with `std::ptr::null_mut()`, and pass `null_mut()` as the hTemplateFile argument to CreateFileW.
- Pass the process HANDLE through the exit-monitor thread as `usize` (raw pointers are not Send; OS handles are safe to use cross-thread).
- Note: HPCON is still `isize` in 0.59, so its zero-init is unchanged.

fixes https://github.com/zellij-org/zellij/issues/5136

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 
 ## [Unreleased]
 * feat: allow tabs to have different sizes if clients aren't focused on the same one (https://github.com/zellij-org/zellij/pull/5133)
+* fix(windows): bump windows-sys to 0.59 to align manifest with code, fixing source builds via `cargo install`/`cargo binstall` (https://github.com/zellij-org/zellij/pull/5139)
 
 ## [0.44.2] - 2026-05-05
 * fix: idle CPU + disk i/o, CommandChanged plugin Event, GetSessionList plugin command (https://github.com/zellij-org/zellij/pull/5063)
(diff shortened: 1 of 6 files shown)


ofek:
Thanks!

Issue #4564, opened 2025-12-21
Bug: Backspace inserts space instead of deleting in web client


vultuk:
## Description

When using the Zellij web client, pressing the backspace key inserts a space character instead of deleting the previous character.

## To Reproduce

1. Start `zellij web` with HTTPS enabled and bound to `0.0.0.0`
2. Connect to the web client via browser
3. Start typing in a terminal session
4. Press backspace to delete a character

**Expected behavior:** The previous character is deleted

**Actual behavior:** A space character is inserted instead

## Environment

- **Zellij version:** 0.44.0 (built from main branch, commit 1296a46a)
- **OS (server):** Amazon Linux 2023 (Linux 6.1.147)
- **Browser:** Chrome on macOS
- **Configuration:**
  - `web_server_ip "0.0.0.0"`
  - `support_kitty_keyboard_protocol true`
  - HTTPS enabled with self-signed certificates

## Additional context

- Rebinding `Ctrl h` (which can conflict with backspace sending `^H`) did not resolve the issue
- Setting `stty erase "^?"` did not resolve the issue
- This appears to be specific to the web client; terminal sessions work correctly

This may be related to how the xterm.js integration handles the backspace key event, possibly sending the wrong character code to the backend.

Horkyze:
I've had similar issue. But in iTerm2. 
I've tried opening s new zellij session, it worked fine (in iTerm2).
Switched back to my old session, the problems was there, even in new tabs/panes.

asivura:
I traced this further and my kitty-keyboard diagnosis was wrong. The actual root cause on my system is an **empty `$TERM`** in web-client sessions, i.e. (at least on my setup) this is the same underlying problem as #5144, not a keyboard-encoding issue.

Evidence that rules out the kitty/keyboard-encoding angle:

- `showkey -a` in the affected web pane shows Backspace arriving as the **correct** byte:
  ```
  ^?    127 0177 0x7f
  ```
  So the right byte *does* reach the shell — the encoding is fine.
- `bindkey '^?'` → `backward-delete-char` (correct binding).
- But `echo "[$TERM]"` prints `[]` — `$TERM` is **unset**. Without it, zsh's line editor (ZLE) can't emit the cursor-movement / redraw sequences, so `backward-delete-char` *does* run but the line can't repaint — which visually looks like a space was inserted instead of a delete.
[...]

PR #5145, opened 2026-05-07
set TERM=xterm-256color for the web server

### LINKED ISSUE #5144: Unset $TERM leads to broken text rendering in web server

## Issue description
If you don't set `$TERM`, the web app renders anything that isn't plain text weirdly. It will duplicate text, move the cursor forward when I press backspace, duplicate my shell prompt, just things that result in an unpleasant experience.

## Minimal reproduction
1. Have no `$TERM` set
2. Have zsh-syntax-highlighting or something similar
3. Spin up a web server
4. Try typing

## Other relevant information
Loom video showing bug: https://www.loom.com/share/77b171fd1a054567834ec36c2d4397cb

Quick fix: set $TERM in zshrc. But $TERM should be set by the terminal, not in your .zshrc, hence the fix

ziidonato:
This fixes #5144

Sets TERM when spinning up the web server so the terminal has full text support

### DIFF
--- a/zellij-client/src/web_client/mod.rs
+++ b/zellij-client/src/web_client/mod.rs
@@ -69,6 +69,9 @@ pub fn start_web_client(
     custom_server_key: Option<PathBuf>,
     startup_timeout: Option<u64>,
 ) {
+    // Web server may launch without a TTY; ensure spawned panes inherit a usable TERM.
+    std::env::set_var("TERM", "xterm-256color");
+
     std::panic::set_hook({
         Box::new(move |info| {
             let thread = thread::current();


Pierre-Mike:
+1 works for me thanks

asivura:
Confirming this is the right fix. Hit the exact symptom from #5144 on the 0.44.3 release build (web client, zsh): plain Backspace appeared to insert a space instead of deleting.

Verified the mechanism is `$TERM`, not keyboard encoding:
- `showkey -a` in the web pane showed Backspace arriving as the correct byte (`^? 127 0x7f`)
- `bindkey '^?'` → `backward-delete-char` (correct)
- but `echo "[$TERM]"` → `[]` (unset) — so ZLE couldn't repaint, making the delete look like an inserted space

Applying the equivalent of this PR (`TERM=xterm-256color` for the web server, via `Environment=TERM=xterm-256color` in the systemd unit) fixed it immediately: `echo helloWORLD` + Backspace×5 → `echo hello`.

Also worth noting: #4564 ("Backspace inserts space instead of deleting in web client") appears to be the same root cause — I left a correction there pointing here. This PR would close both. +1 to merge.

PR #5147, opened 2026-05-07
fix(zellij-tile): propagate host-call errors instead of panicking

dsaenztagarro:
## Summary

`current_session_last_saved_time()` and `get_session_list()` in `zellij-tile/src/shim.rs` currently call `bytes_from_stdin().unwrap().as_slice()` on the host response. When the host declines to write a response, both `unwrap()`s panic and abort the plugin with the unhelpful `<NO PAYLOAD>` panic message.

This PR converts those two helpers to the same fallible-decode pattern that `open_file()` already uses, surfacing `None` / `Err(String)` on read or decode failure rather than aborting.

## When this matters

The host **does** legitimately suppress responses — for example, when a `ReadApplicationState`-gated command is rejected because the plugin's permission grant is still pending. Built-in plugins almost never see this because their permissions are auto-granted, but plugins loaded via `file:` (the documented mechanism for forks and custom builds) hit it on every reload.

Concrete repro from a custom session-manager fork loaded via `file:` after upgrading to v0.44.2:

```
ERROR:  <NO PAYLOAD>
PanicHookInfo {
    payload: Any { .. },
    location: Location {
        file: "zellij-tile/src/shim.rs",
        line: 1917,
        column: 67,
    },
    ...
}
```

The same class of panic was reported earlier at `shim.rs:382` (the `current_session_last_saved_time()` site) in PR #5063's session-manager rework — that one is also resolved here.

## Why this fix and not something else

- The shim helpers already have proper return types for failure (`Option<u64>`, `Result<_, String>`). The `.unwrap()`s are a holdover; `open_file()` (shim.rs ~line 398) demonstrates the intended safe pattern.
- Fixing the shim is **decoupled** — every plugin, including third-party ones that load via `file:`, immediately benefits without each plugin having to defer host calls behind a `permissions_granted` flag in its own state machine.

## Scope

Intentionally limited to the two confirmed panic sites. There are ~47 other `bytes_from_stdin().unwrap().as_slice()).unwrap()` occurrences in `shim.rs` that share the same defect; happy to expand the PR to cover them in a follow-up commit if maintainers prefer a single sweep — left out here to keep the diff focused and easy to review.

## Test plan
- [x] `cargo build --release -p session-manager --target wasm32-wasip1` — compiles
- [x] `cargo check -p zellij-tile` — compiles for host target
- [x] Verified panic at `shim.rs:1917` no longer reproduces on a `file:`-loaded session-manager whose permission grant is still pending; `get_session_list()` returns `Err(...)` and the plugin keeps running.
- [ ] Maintainer check: behavior on built-in load (permissions auto-granted) is unchanged.

### DIFF
--- a/zellij-tile/src/shim.rs
+++ b/zellij-tile/src/shim.rs
@@ -375,13 +375,13 @@ pub fn save_session() -> Result<(), String> {
 /// ```
 pub fn current_session_last_saved_time() -> Option<u64> {
     let plugin_command = PluginCommand::CurrentSessionLastSavedTime;
-    let protobuf_plugin_command: ProtobufPluginCommand = plugin_command.try_into().unwrap();
+    let protobuf_plugin_command: ProtobufPluginCommand = plugin_command.try_into().ok()?;
     object_to_stdout(&protobuf_plugin_command.encode_to_vec());
     unsafe { host_run_plugin_command() };
 
+    let bytes = bytes_from_stdin().ok()?;
     let protobuf_response =
-        ProtobufCurrentSessionLastSavedTimeResponse::decode(bytes_from_stdin().unwrap().as_slice())
-            .unwrap();
+        ProtobufCurrentSessionLastSavedTimeResponse::decode(bytes.as_slice()).ok()?;
 
     protobuf_response.timestamp_millis
 }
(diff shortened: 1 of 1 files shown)


dsaenztagarro:
@imsnif @a-kenji this is ready for you to review! Happy to address any suggested feedback.

Issue #4384, opened 2025-08-29
Move built-in web server into an optional plugin (secure-by-default, reduced attack surface)


lyskov:
# Summary
Dear Developers,
Could you please consider decoupling Zellij’s built-in web server from the core binary and shipping it as an optional, officially maintained plugin?
This change would make Zellij secure-by-default, reduce the default attack surface, and better align with least-privilege and opt-in design principles.
Thanks,

# Motivation
- Security (secure-by-default): Network listeners expand the attack surface. Features that open sockets should be opt-in, not opt-out. This aligns with common guidance to reduce default privileges and capabilities.

- Least privilege & policy compliance: Many organizations restrict developer tools that bind to local or remote ports. Making the web server optional improves adoption in hardened or air-gapped environments.

- Operational clarity: If a network listener is needed, users explicitly enable it and can audit/log that choice. This is easier to reason about than a default that must be consciously disabled (ie minimize risk of accidentally providing functionality).

- Dependency minimization: Removing networking code from the default build can trim dependencies and reduce the risk/impact of CVEs for users who never use the web server.

- Binary size & startup time: A smaller default binary can improve cold-start time and packaging footprint.

imsnif:
Hey - thanks for bringing this up. Just for clarity:

1. The Zellij web server is not enabled by default and there must be an explicit user action to enable it
2. The Zellij web server cannot listen on an external interface without user setup (namely providing an SSL certificate)
3. If these measures are not sufficient, it's always possible to get a version of Zellij without this feature or its dependencies by turning off the `web-server-capability` compile-time flag. We even provide a `-no-web` variant in the releases that does this for the pre-built binaries.

To in short: I think 3 is what you want.

lyskov:
Thank you for the detailed reply and for providing the -no-web variant.

To clarify my original intent: I’m aware of the web-server-capability compile-time flag. My request is about **default posture and packaging impact** rather than availability. Most distro/OS packagers build with upstream defaults; when the web capability is compiled into the core binary (even if disabled at runtime), many organizations’ policies flag or ban the package because the binary contains network-listener code. Asking admins to install non-default artifacts or build from source introduces friction and reduces adoption.

**What would help adoption:**
- Make web-server-capability disabled by default in upstream builds; keep it easily enabled for those who need it, or
- Preferably, move the web server to an official optional plugin/subpackage, so network code and dependencies are not shipped in the core by default.

[...]

PR #5148, opened 2026-05-08
make web_server_capability opt-in

lyskov:
## Summary

This PR proposes changes to how the embedded web server / remote-attach
capability is exposed in zellij:

1. The `web` CLI subcommand is gated on the `web_server_capability` Cargo
   feature, so it appears in `--help` and is accepted by the parser if and
   only if the feature is built in.

EDIT: part-2 was removed from this PR.

~~2. `web_server_capability` is removed from the workspace's default feature
set. It remains a one-flag opt-in (`--features web_server_capability`).~~

~~`xtask` is updated internally to compensate for the new default, so its
user-facing behavior is unchanged. CI and release workflows do not require
modifications.~~

~~## Motivation~~

~~Three considerations support making this feature opt-in rather than
opt-out:~~

~~1. **Default-build dependency surface.** With the feature enabled by
default, every build links axum, axum-server, axum-extra, tokio-rustls,
isahc, time, futures, and their transitive dependencies. For users
who do not use the web server, those dependencies still occupy the
binary's audit and CVE-tracking surface. Making the feature opt-in
keeps that surface available to users who want it while reducing it
for users who do not.~~

~~2. **Downstream packaging defaults.** Distribution maintainers (Debian,
Arch, Homebrew, Nix, etc.) typically build with default features
unless they patch the feature list. The current default therefore
ships network-listening capability through every packaged binary,
including to users who may be unaware that capability is present.
An opt-in default lets each distribution decide explicitly whether
to enable it.~~

~~3. **Convention for opt-in network capabilities.** Capabilities that
open network listeners and authenticate remote callers are commonly
opt-in elsewhere in the ecosystem. Aligning with that convention
does not remove the feature — it remains one flag away — and makes
the feature's presence in a given build more transparent to users.~~

### DIFF
--- a/src/commands.rs
+++ b/src/commands.rs
@@ -1,4 +1,5 @@
 use dialoguer::Confirm;
+#[cfg(feature = "web_server_capability")]
 use std::net::IpAddr;
 use std::{fs::File, io::prelude::*, path::PathBuf, process, time::Duration};
 
@@ -213,25 +214,6 @@ pub(crate) fn start_web_server(
     );
 }
 
-#[cfg(not(feature = "web_server_capability"))]
-pub(crate) fn start_web_server(
-    _opts: CliArgs,
-    _run_daemonized: bool,
-    _ip: Option<IpAddr>,
-    _port: Option<u16>,
-    _cert: Option<PathBuf>,
-    _key: Option<PathBuf>,
-    _startup_timeout: Option<u64>,
-) {
-    log::error!(
-        "This version of Zellij was compiled without web server support, cannot run web server!"
-    );
-    eprintln!(
-        "This version of Zellij was compiled without web server support, cannot run web server!"
-    );
-    std::process::exit(2);
-}
-
 fn create_new_client() -> ClientInfo {
     ClientInfo::New(generate_unique_session_name_or_exit(), None, None)
 }
@@ -240,17 +222,6 @@ fn create_new_client() -> ClientInfo {
 pub(crate) fn stop_web_server() -> Result<(), String> {
     shutdown_all_webserver_instances().map_err(|e| e.to_string())?;
     Ok(())
-}
-
-#[cfg(not(feature = "web_server_capability"))]
-pub(crate) fn stop_web_server() -> Result<(), String> {
-    log::error!(
-        "This version of Zellij was compiled without web server support, cannot stop web server!"
-    );
-    eprintln!(
-        "This version of Zellij was compiled without web server support, cannot stop web server!"
-    );
-    std::process::exit(2);
 }
 
 #[cfg(feature = "web_server_capability")]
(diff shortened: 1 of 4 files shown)


imsnif:
Please see my reply in the original issue: https://github.com/zellij-org/zellij/issues/4384#issuecomment-3236173285

lyskov:
Thank you for the quick response and for laying out the existing mitigations — that's helpful context.

Would you be open to merging the first commit on its own? It is independent of the second and does not change the default. Its only effect is to make the existing `--no-web` build produce a coherent
CLI surface: web is hidden from `--help`, and the parser rejects `zellij web ...` rather than dispatching through the runtime error stubs in src/commands.rs. In effect, it cleans up path 3 in your
reply for users who already follow that route.

A side benefit: a built binary becomes self-describing for this feature — `zellij --help | grep web` is enough to tell whether the binary was compiled with web support, which is useful for packagers
and for shell scripts that need to branch on capability. Today this signal is not reliable because the subcommand appears either way.

If that's workable, I'll open a fresh PR with just the first commit and close this one.

imsnif:
With the disclaimer that I haven't looked at the code in this PR and am only going on your description: sure, that makes sense. I'd be happy to accept just that part.

lyskov:
Great! @imsnif I update the PR content and its description. Please let me know if anything needs to be changed before this could be merged.
Thanks,

imsnif:
Hey, I gave this a brief look. If the goal is to have a command that will instantly tell you whether a binary has the web server capability or not, can't we instead adjust the help? I think all web server commands (even status) exit with non-0 if it's compiled out.

lyskov:
Do you mean adjusting the help message that Clap generates when the `web` capability is not enabled? I suppose that could be done.

To clarify: my primary goal is to make it immediately obvious — both to humans and scripts — whether this is a web-enabled build or not, so relying on the process exit code alone is not sufficient.

Also, from an ergonomics standpoint: are you sure you want the executable to still expose the `web` option when there is nothing behind it? That feels somewhat odd to me — if web capabilities are unavailable, why still provide a `web` option at all?

imsnif:
For the same reason grayed out buttons exist in UIs when their functionality is unavailable.

Let's change the help text to something like "Web functionality unavailable in this version." Thanks!

lyskov:
Ah, I think I understand what you mean now—I must have misunderstood your earlier comment. I’ve updated the Clap help message; please let me know if anything else needs to be done. Thanks.

imsnif:
@lyskov - can we please make this change just the help message change to avoid a large diff that - as far as I understand your concerns - doesn't give us anything more than the help message change already does?

lyskov:
Well, this is your project, so I defer to your judgment, but I do not think this is a good idea. Exposing a non-functional command-line option simply because it may become functional under a different feature configuration feels confusing from a user perspective and does not seem technically justified.

That said, this is ultimately your call. If you still prefer that approach, I can update the PR accordingly. Please let me know. Thanks.

Issue #5140, opened 2026-05-06
Prompt delay after command output in 0.44.2 (regression from 0.44.1)


Chikalona:
### Description

After upgrading from 0.44.1 to 0.44.2, there is a noticeable delay of 2–3 seconds before the shell prompt returns after running commands that produce multi-line output (e.g. `lsd`, `ls`). The command output itself appears immediately, but the prompt takes several seconds to reappear.

This does not occur in 0.44.1. Downgrading to 0.44.1 resolves the issue immediately.

### Steps to reproduce

1. Run Zellij 0.44.2 with Fish shell inside Kitty terminal
2. Run a command with multi-line output, e.g. `lsd` in a directory with many files
3. Observe that the output appears instantly, but the shell prompt takes 2–3 seconds to return

### Expected behavior

Shell prompt returns immediately after command output, as it does in 0.44.1.

### Actual behavior

2–3 second delay before the shell prompt reappears after commands with multi-line output.

### Environment

- Zellij: 0.44.2 (regression not present in 0.44.1)
- Terminal: Kitty
- Shell: Fish
- OS: EndeavourOS (Arch Linux), x86_64
- No delay observed when running the same commands directly in Kitty without Zellij

### Debugging notes

- `fish --profile` confirms Fish itself renders the prompt in under 2ms — the delay is not in Fish
- All Fish event handlers (postexec hooks etc.) are fast individually
- Running `lsd; echo "DONE: $(date +%S%N)"` shows that "DONE" prints instantly — the delay occurs specifically between command completion and prompt rendering, i.e. inside Zellij
- Setting `advanced_mouse_actions false` in config.kdl does not resolve the issue
- The issue occurs regardless of directory size

SmollClover:
Can confirm myself that this is an issue.

```
Zellij 0.44.2
Shell: fish 4.7.0
Terminal: Ghostty 1.3.1-arch2.2
OS: Arch Linux x86_64
Kernel: Linux 7.0.3-1-cachyos-bore-lto
DE: KDE Plasma 6.6.4
WM: KWin (Wayland)
```

harlanwei:
I can confirm this issue on my side. I also notice that vim starts with a blank screen until I press a key or until some time has passed.

imsnif:
Could those experiencing the issue please follow the debugging instructions [here](https://github.com/zellij-org/zellij/issues/5138#issuecomment-4394958223)? Thanks.

Chikalona:
[...]

PR #5149, opened 2026-05-08
fix: occasional stall when querying host terminal (introduced in 0.44.1)

### LINKED ISSUE #5138: OSC 10/11 query replies leak into shell stdin in 0.45.0 (regression from #5105)
## Issue description

OSC 10/11 query replies emitted by an application inside a pane are
delivered into that pane's stdin instead of being consumed by the
application that asked. In a cooked-mode shell (zsh, bash) the reply
shows up at the next prompt as if it were typed.

In practice p10k's instant prompt issues an OSC 11 at shell startup,
and the leaked bytes break `gh auth login`:

    could not prompt: unexpected escape sequence from terminal: ['\x1b' ']']

This is a regression from 0.44.1.

## Minimal reproduction

Any host terminal that answers OSC 11 (WezTerm, Alacritty, kitty). In
a zsh pane:

    printf '\e]11;?\e\\'

Next prompt:

    ~  ^[]11;rgb:0000/0000/0000^[\

## Other relevant information

I built 0.44.1 and 0.45.0 from source. 0.44.1 is fine — the reply
arrives in time to be consumed on the same line, and the next prompt
is clean. 0.45.0 reproduces.

The forward path (`zellij-server/src/host_query.rs`,
`Screen::handle_forwarded_reply_from_host` writing the reply to the
pane's pty) was introduced by 227142dd (#5105). Its body says "the
reply streaming back into the relevant pane" — fine for raw-mode TUIs
that parse OSC, but cooked-mode shells just read the bytes as input.

`zellij --version`: 0.45.0
`uname -av`: Darwin 25.5.0 arm64
Host terminal: WezTerm. Shell: zsh + p10k.

imsnif:
This fixes a regression from 0.44.1 where occasionally starting applications or even printing new prompts would stall for ~500ms in certain environments and conditions.

### What was happening?
In #5105 and #5113, the behavior of OSC10/11 and some other queries was changed. Before, Zellij would query the host terminal for these values on startup, cache the reply and then use that cached reply to reply to applications running inside Zellij panes. After these PRs, in order to better support theme switching at runtime, we would instead forward each such query to the host terminal and then stream these replies back to the requesting application.

To preserve order, while these queries are in flight, Zellij would buffer application STDOUT with a 500ms timeout. This timeout was only meant to be used for a misbehaving host terminal - i.e. one that not only does not reply to the original query, but also does not reply to the following DA1 query Zellij sends after each such app query.

This worked swimmingly, except for the fact that the parser we were using to parse these incoming replies (as opposed to the one we were originally using to parse them on startup) did not recognize key sequences from the kitty keyboard protocol. Not only did it not recognize them, but it also left them inside its internal buffer - seeing them as ambiguous sequences that might later be interpreted as real keys or such query replies (due to their similar format).

So when these sequences would arrive, they would pollute the internal residual buffer of the query-reply parser (even though they were also forwarded on to the actual KKP parser), and then whenever an actual reply would arrive, it would be tacked on to these sequences and dropped as unrecognized garbage input. This would then trigger a stall (Zellij would still be buffering the application's STDOUT), which would eventually timeout with the above 500ms safety timer.

### The fix
In this PR, I made the query reply parser aware of the KKP so that it can properly strip the relevant bytes from its buffer when this happens.

A more holistic fix would be to unify all our STDIN parsers into one. Something we were until recently unable to do since we were relying on the external termwiz STDIN parser. Now that we've vendored it, we should be able to adjust it to our needs. But that's out of scope for this immediate hole that needs to be plugged.

Fixes https://github.com/zellij-org/zellij/issues/5138 and https://github.com/zellij-org/zellij/issues/5140

Thanks all for helping me troubleshoot this issue I was never able to reproduce.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 ## [Unreleased]
 * feat: allow tabs to have different sizes if clients aren't focused on the same one (https://github.com/zellij-org/zellij/pull/5133)
 * fix(windows): bump windows-sys to 0.59 to align manifest with code, fixing source builds via `cargo install`/`cargo binstall` (https://github.com/zellij-org/zellij/pull/5139)
+* fix: occasional stall before prompt/program-run (https://github.com/zellij-org/zellij/pull/5149)
 
 ## [0.44.2] - 2026-05-05
 * fix: idle CPU + disk i/o, CommandChanged plugin Event, GetSessionList plugin command (https://github.com/zellij-org/zellij/pull/5063)
--- a/zellij-client/src/stdin_ansi_parser_tests.rs
+++ b/zellij-client/src/stdin_ansi_parser_tests.rs
@@ -817,3 +817,61 @@ fn host_theme_dsr_997_unknown_param_dropped() {
         "unknown ?997;N value must not classify"
     );
 }
+
+// =====================================================================
+// Regression test: a Kitty keyboard-protocol event arriving on the
+// client's stdin must not wedge the parser. Before the fix, such an
+// event (`\x1b[<keycode>;<mods>u`) sat at the head of termwiz's
+// internal buffer because parse_csi_report rejects the unwhitelisted
+// final byte 'u' AND the keymap returned `Found::NeedData` (it's a
+// possible prefix of registered key sequences). Every host reply that
+// arrived behind it was silently swallowed, stalling host-color
+// forwards until session timeout. This was confirmed via WEDGE-CANDIDATE
+// log entries from three independent users on Kitty, Ghostty, and
+// Konsole — all triggered by the same `\x1b[<digits>;<digits>u` shape.
+//
+// Captured triggers from real logs:
+//   `\x1b[57414;129u`  (Kitty user, repro-logs/third-commit/zellij(4).log)
+//   `\x1b[27;129u`     (Ghostty user, ...zellij(5).log)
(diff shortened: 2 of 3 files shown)


bogorad:
Additional data point: this fix addresses the native-client path, but the same class of symptom still appears when using `zellij web`.

  I tested a downstream build based on `v0.44.2` with the `#5149` fix backported. In a web-client setup, Yazi still reports:

  ```text
  Terminal response timeout: The request sent by Yazi didn't receive a correct response.
  Please check your terminal environment as per: https://yazi-rs.github.io/docs/faq#trt
  ```

  The path appears different from the native-client parser issue fixed here.

  In v0.44.2 / current main, zellij-client/src/web_client/server_listener.rs handles:

```
  Some(ServerToClientMsg::ForwardQueryToHost { .. }) => {
      // Web clients do not currently participate in
      // host-terminal query forwarding ...
      // Silently drop.
  }
```

  Yazi startup sends terminal probes including CSI 16t, OSC 11;?, and CSI 0c. Zellij intercepts/forwards the host queries, but the web client drops ForwardQueryToHost. The pane is then held until the server fallback timeout, while Yazi warns after about 400ms.

  So #5149 fixes the host-reply parser wedge for native clients, but zellij web still has no immediate response path for forwarded host queries.

  A minimal fix for the web path may be to make the web client immediately send:

```
  ClientToServerMsg::ForwardedReplyFromHost {
      token,
      reply_bytes: Vec::new(),
  }
```

  for ForwardQueryToHost, allowing the server to use its existing cached fallback immediately instead of waiting for timeout.

Issue #3418, opened 2024-06-13
Resizing panes only resizes multiple lines at a time


unikitty37:
## Issue description

When resizing panes vertically, Zellij appears to resize by multiple lines at a time (5 in my case, though this may depend on overall window size). This leaves me with a choice of a pane that is either too short or too tall.

IMO, Zellij should resize by one cell at a time; if I want to resize by five lines, I can press the arrow key five times, but I cannot press it one-fifth of a time with the current behaviour :)

## Minimal reproduction

- Install the repro setup.
- Start zellij.
- Create a second pane with ⌃p n
- Enter resize mode with ⌃n
- Press the up arrow key

### config.kdl

```
plugins {
    tab-bar location="zellij:tab-bar"
    status-bar location="zellij:status-bar"
}

default_layout "demo"
```

### layouts/demo.kdl

```
layout {
    pane size=1 borderless=true {
        plugin location="tab-bar"
    }
    pane split_direction="vertical"
    pane size=2 borderless=true {
        plugin location="status-bar"
    }
}
```

## Other relevant information

macOS 14.5 (23F79)
zellij 0.40.1 (installed via homebrew)
alacritty 0.13.2 (1)

Alacritty config:

```
[font]
size = 14
normal = { family = "Hasklug Nerd Font", style = "Regular" }
bold = { family = "Hasklug Nerd Font", style = "Bold" }
italic = { family = "Hasklug Nerd Font", style = "Italic" }
bold_italic = { family = "Hasklug Nerd Font", style = "Bold Italic" }

[cursor]
style = { shape = "Block", blinking = "Off" }

[env]
TERM = "xterm-256color"

[shell]
program = "/opt/homebrew/bin/zellij"

[window]
padding.x = 4
padding.y = 6
opacity = 1
dimensions = { columns = 156, lines = 117 }
resize_increments = true
option_as_alt = "OnlyRight"
```

Issue #1063, opened 2022-02-15
Make resizes stable


har7an:
Thank you for taking the time to file this issue! Please follow the instructions and fill in the missing parts below the instructions, if it is meaningful. Try to be brief and concise.

**In Case of Graphical or Performance Issues**

1. Delete the contents of `/tmp/zellij-1000/zellij-log`, ie with `cd /tmp/zellij-1000/` and `rm -fr zellij-log/`
2. Run `zellij --debug`
3. Recreate your issue.
3. Quit Zellij immediately with ctrl-q (your bug should ideally still be visible on screen)

Please attach the files that were created in `/tmp/zellij-1000/zellij-log/` to the extent you are comfortable with.

*See end of this post*

**Basic information**

`zellij --version`: Built from main on commit 10a22c479ffb4d76627eadbee2dc4b53ae4309c3, but also happens on 0.24.0
`tput lines`: 58
`tput cols`: 237
`uname -av` or `ver`(Windows): Linux tz 5.16.8-200.fc35.x86_64 #1 SMP PREEMPT Tue Feb 8 20:58:59 UTC 2022 x86_64 x86_64 x86_64 GNU/Linux

List of programs you interact with as, `PROGRAM --version`: output cropped meaningful, for example:
`nvim --version`: NVIM v0.5.0-dev+1299-g1c2e504d5  (used the appimage release)
`alacritty --version`: alacritty 0.7.2 (5ac8060b)

**Further information**
Reproduction steps, noticeable behavior, related issues, etc

## Steps to reproduce

- Create a grid of panes. I hit `Alt+n` until I had 3 rows and 4 columns of panes
- Make one of the inner panes active
- Enter resize mode (`Ctrl+n`) and repeatedly hit `+` and `-`, alternating both buttons

## What you will see

The first two cycles the grid is restored to its original state. After this, one of the panes will permanently change its shape. The longer you repeat, the more will follow

## What I expected

The grid always restores itself and acts deterministically

## Additional observations

At some point some of the panes frame will change. I noticed that the upper-left frame ending, where it touches the pane name (i.e. the `-` ending in the name) will become a downward angle. At this point it seems that any further interaction with the pane doesn't do anything.

## The Logs

[zellij.log](https://github.com/zellij-org/zellij/files/8072136/zellij.log)
[zellij-9.log](https://github.com/zellij-org/zellij/files/8072138/zellij-9.log)
[...]

PR #5159, opened 2026-05-12
fix(resize): propagate actual resize result in stacked pane resize

konono:
## Summary

Fix two sites where `Result<bool>` return values are silently discarded,
causing stacked pane resize to skip its fallback path.

### `resize_or_stack_pane_*`: stacking fallback never triggers

These functions match `Ok(_) => true`, so `Ok(false)` (no cell change)
is treated as success. The else branch — `stack_pane_up/down/left/right` —
is never reached.

**Fix**: `Ok(_) => true` → `Ok(changed) => changed` (4 sites)

### `resize_pane_with_id`: 2× retry result discarded

When `PaneSizeUnchanged` triggers a retry with double the resize percent,
the result is discarded (`Ok(_) => {}`). `pane_size_changed` stays `false`
even when the retry succeeded.

**Fix**: `Ok(_) => {}` → `Ok(changed) => { pane_size_changed = changed; }`

### Effect on callers

`stacked_resize_pane_with_id` uses the return value of
`resize_or_stack_pane_*` to decide whether to record tombstones via
`update_tombstones_before_increase/decrease`. When a no-op resize
reports success, tombstones may record unchanged state and the stacking
fallback is skipped, making resize appear unresponsive near pane boundaries.

## Related issues

- #3418 — Resizing panes only resizes multiple lines at a time (this fix addresses the stacking fallback part of the symptom)
- #1063 — Make resizes stable

## Test plan

- [x] `cargo check -p zellij-server` passes
- [x] `cargo test -p zellij-server` — 1087 passed, 0 failed
- [ ] Manual: repeated resize keypresses respond every time
- [ ] Manual: stacked pane resize correctly falls back to stacking

### DIFF
--- a/zellij-server/src/panes/tiled_panes/mod.rs
+++ b/zellij-server/src/panes/tiled_panes/mod.rs
@@ -1365,7 +1365,7 @@ impl TiledPanes {
         strategy.invert_on_boundaries = false;
         let successfully_resized_up =
             match self.resize_pane_with_id(strategy, pane_id, Some(resize_percent)) {
-                Ok(_) => true,
+                Ok(changed) => changed,
                 Err(_) => false,
             };
         if successfully_resized_up {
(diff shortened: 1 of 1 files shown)


imsnif:
Hey, thanks for this. Can you please describe the user facing consequences this is trying to fix? Including at least one reproducible scenario?

konono:
Thanks for the feedback.
I don't have a reliable step-by-step reproduction for this, so I'll close this PR for now and revisit once I can provide a concrete scenario.

PR #5166, opened 2026-05-13
feat(osc): support and forward OSC 9 / OSC 777 desktop notifications

caojianfa:
Extends the unidirectional desktop-notification passthrough chain introduced by upstream PR #4931 (OSC 99) to also handle the simpler single-shot protocols spoken by other terminal ecosystems:

  - OSC 9   (iTerm2):  ESC ] 9 ; <body> ST
  - OSC 777 (rxvt):    ESC ] 777 ; notify ; <title> ; <body> ST

Both protocols are pushed onto the existing pending_desktop_notifications queue and forwarded byte-for-byte to the host terminal. They reuse the established BEL visual-bell pipeline (pane/tab frame flash + [!] suffix
+ folded stacked-pane indicator) by setting Grid::ring_bell on dispatch.

A new top-level configuration option allow_osc_passthrough (default true) lets users opt out of forwarding while still seeing the in-zellij visual indicator. The flag is scoped to OSC 9 / 777; OSC 99 retains its own contract from PR #4931 and is not affected.

Implementation notes:
  - OSC 9 / 777 entries are disambiguated from OSC 99 in the shared queue by a payload prefix ("9;..." or "777;..."). OSC 99 metadata is always key=value:key=value, so collisions are not possible with well-formed input.
  - The gate lives at Tab::forward_desktop_notifications instead of Grid::osc_dispatch to avoid plumbing a 13th positional param through ~135 Grid::new test sites.
  - OSC 777 sub-commands other than `notify` are silently ignored (no queue push, no visual bell). OSC 9 with empty body still rings the bell but does not forward.

Tests:
  - 6 new Grid-level tests covering payload shape, BEL/ST terminator handling, empty body, and unknown OSC 777 sub-command
  - 5 new Tab-level integration tests covering verbatim forwarding, no namespacing leakage, allow_osc_passthrough=false gate, OSC 9 / OSC 99 non-interference, and OSC 99 unaffected by the new flag

Snapshots updated for new KDL output / Options dump.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 
 ## [Unreleased]
+* feat: support and forward OSC 9 (iTerm2) / OSC 777 (rxvt) desktop notification sequences from panes to the host terminal, behind a new `allow_osc_passthrough` option (default `true`)
 * feat: allow tabs to have different sizes if clients aren't focused on the same one (https://github.com/zellij-org/zellij/pull/5133)
 * feat: PWA support for the web client (manifest + icons + iOS meta tags) so the page can be installed as a standalone app (https://github.com/zellij-org/zellij/pull/5184)
 * feat: mobile UI (https://github.com/zellij-org/zellij/pull/5241)
--- a/zellij-server/src/lib.rs
+++ b/zellij-server/src/lib.rs
@@ -437,6 +437,7 @@ impl SessionMetaData {
                     visual_bell: new_config.options.visual_bell.unwrap_or(true),
                     focus_follows_mouse: new_config.options.focus_follows_mouse.unwrap_or(false),
                     mouse_click_through: new_config.options.mouse_click_through.unwrap_or(false),
+                    allow_osc_passthrough: new_config.options.allow_osc_passthrough.unwrap_or(true),
                 })
                 .unwrap();
             self.senders
--- a/zellij-server/src/panes/grid.rs
+++ b/zellij-server/src/panes/grid.rs
@@ -630,6 +630,14 @@ pub struct Grid {
     pub is_scrolled: bool,
     pub link_handler: Rc<RefCell<LinkHandler>>,
     pub ring_bell: bool,
+    /// Visual-only bell signal: set by OSC 9 / OSC 777 dispatch to drive
+    /// the in-Zellij pane/tab/folded-stacked-pane indicators **without**
+    /// forwarding an ANSI BEL byte to the host terminal. The desktop
+    /// notification itself is the audible/visible alert; layering an
+    /// audio BEL on top would be a double-notification, and ST-terminated
+    /// OSC sequences carry no BEL in the input at all so synthesizing one
+    /// is surprising. Consumed by `Tab::check_and_handle_bell_notifications`.
+    pub pending_visual_bell: bool,
     scrollback_buffer_lines: usize,
     pub mouse_mode: MouseMode,
     pub mouse_tracking: MouseTracking,
@@ -959,6 +967,7 @@ impl Grid {
             is_scrolled: false,
             link_handler,
             ring_bell: false,
+            pending_visual_bell: false,
             scrollback_buffer_lines: 0,
             mouse_mode: MouseMode::default(),
             mouse_tracking: MouseTracking::default(),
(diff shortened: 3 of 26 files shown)


lens0021:
Is the goal of this the duplicate of #5099?

caojianfa:
> Is the goal of this the duplicate of #5099?

Hi @lens0021, thanks for the cross-reference!

Both this PR and #5099 address the same gap (OSC 9 / OSC 777 silently dropped at `Grid::osc_dispatch`'s `_ =>` arm) but the two take meaningfully different design approaches. Quick side-by-side so the maintainers can pick whichever direction fits the codebase:

**#5099 (DaveDev42)** — sanitize + canonical rebuild:
- Parses each notification into a typed `LegacyDesktopNotification` enum
- Strips C0 / DEL / C1 control characters from every field
- Caps field length at 1024 chars
- Rebuilds the sequence from a fixed template before emitting (`to_canonical_bytes`)
- Normalizes terminator to BEL on emit; original ST is discarded

**This PR (#5166)** — raw passthrough with a user opt-out + visual-bell integration:
- Forwards the original payload bytes verbatim, preserves the original terminator (BEL or ST)
- Adds a top-level `allow_osc_passthrough` config option (default `true`). When `false`, the visual indicator inside Zellij still fires but the OSC bytes are dropped at `Tab::forward_desktop_notifications` — i.e. users who don't want their notification center driven by every CI pipe can still see "something happened" via the regular pane/tab indicator
- Reuses the existing BEL visual-bell pipeline (`Grid::ring_bell = true` on dispatch) so panes that emit OSC 9 / 777 get the pane/tab frame flash + `[!]` suffix + folded stacked-pane indicator for free, matching the in-Zellij UX of a plain BEL
- 11 new unit + integration tests (Grid-level VTE-parser tests for both protocols and edge cases, Tab-level `ServerInstruction::Render` capture, the opt-out gate, OSC 9 / OSC 99 non-interference, OSC 99 unaffected by the new flag). 25 OSC tests pass in total (11 new + 11 OSC 99 from #4931 + 3 namespace helpers).

The main philosophical split is **sanitize-and-rebuild vs. trust-and-forward**: #5099 treats the notification body as untrusted input and rebuilds it from a fixed template to prevent escape-sequence injection; this PR forwards it as-is on the premise that any program already running inside a pane can write to its own stdout anyway, and adds a user-visible opt-out for those who'd rather not be pinged at all.

For what it's worth, **I've been running this patch personally for ~3 weeks across daily work** (dev / builds / CI alerts via yazi and various shell prompts) and haven't hit any issues — no crashes, no visible interaction with OSC 99 or the regular BEL flow, and the opt-out has been a real quality-of-life improvement on heads-down days.

Happy to defer to whichever direction the maintainers prefer. If there's interest in combining the two approaches (e.g. raw passthrough by default + an opt-in sanitize-strict mode for security-conscious users) I'd be glad to rebase.

yamam:
Thanks for working on this! I tested the PR locally with WezTerm, and desktop notifications are now forwarded successfully from inside Zellij.

One thing I noticed is that OSC 9/777 also triggers an audible bell in the host terminal, because the implementation sets `ring_bell = true`.

I wonder if this should be separated from OSC notification passthrough. Even when OSC 9 is terminated by BEL, that BEL is an OSC string terminator rather than a terminal bell. In WezTerm without Zellij:

  printf '\e]9;Build complete\a'

shows a desktop notification but does not ring the bell.

For ST-terminated OSC 9:

  printf '\e]9;Build complete\e\\'

there is no BEL in the input, so it feels surprising for Zellij to synthesize one.

Could we avoid sending BEL to the host terminal for OSC 9/777 notifications?

Happy to help test.

caojianfa:
> Thanks for working on this! I tested the PR locally with WezTerm, and desktop notifications are now forwarded successfully from inside Zellij.
> 
> One thing I noticed is that OSC 9/777 also triggers an audible bell in the host terminal, because the implementation sets `ring_bell = true`.
> 
> I wonder if this should be separated from OSC notification passthrough. Even when OSC 9 is terminated by BEL, that BEL is an OSC string terminator rather than a terminal bell. In WezTerm without Zellij:
> 
> printf '\e]9;Build complete\a'
> 
> shows a desktop notification but does not ring the bell.
> 
> For ST-terminated OSC 9:
> 
> printf '\e]9;Build complete\e\'
> 
> there is no BEL in the input, so it feels surprising for Zellij to synthesize one.
> 
> Could we avoid sending BEL to the host terminal for OSC 9/777 notifications?
> 
> Happy to help test.

Just pushed the fix as part of this PR's head commit. Verified locally on macOS Apple Silicon with WezTerm — both terminator forms now behave as you'd expect on bare WezTerm:

- `printf '\e]9;build done\a'` → desktop notification, no audible bell
- `printf '\e]9;build done\e\\'` → desktop notification, no audible bell (and crucially, no synthesized BEL byte from a BEL-less input)
- `printf '\e]777;notify;Title;Body\a'` → desktop notification, no audible bell
- A literal BEL byte (`printf '\a'`) still rings the bell, as you'd expect

What changed:

- New `Grid::pending_visual_bell` flag, distinct from `Grid::ring_bell`. OSC 9 / OSC 777 dispatch sets the visual-only flag; only a literal BEL byte sets `ring_bell`.
- `Tab::check_and_handle_bell_notifications` now drains both kinds. Real bells still set `tab_bell_ring` (the BEL-forwarding signal); visual-only bells populate `panes_with_pending_bell` / `tab_has_pending_bell` / `set_bell_notification(true)` but leave `tab_bell_ring` alone.
- Return type widened to `(Vec<PaneId>, bool, bool)` with a new `had_audio_bell`. `Screen::render` uses that to decide whether to emit `\u{7}` to the host, instead of inferring it from whether any pane/tab visual flash got scheduled (which incorrectly conflated the two paths before).
- 3 new regression tests lock the contract:
    - `osc9_does_not_synthesize_audio_bell_on_host` (both BEL- and ST-terminated)
    - `osc777_does_not_synthesize_audio_bell_on_host`
    - `real_bel_byte_still_triggers_audio_bell` (sanity — make sure the visual-only path didn't regress real BEL forwarding)
- Existing OSC 9 / 777 tests are tightened to assert that `pending_visual_bell` is set and `ring_bell` stays `false`.

I kept the in-Zellij visual indicator (pane frame flash, tab `[!]`, folded stacked-pane title cue) on the visual-only path, since folded panes can't show the notification text and benefit from a "something happened here" hint. Let me know if you'd rather have that suppressed as well and I'll remove it.

Thanks again for the careful test — this bug had been there since the first iteration and I hadn't noticed it until you ran the WezTerm comparison.

caojianfa:
Rebased onto latest `main` and resolved the merge conflict — this is ready for review whenever you have a chance.

The conflict came from the Mobile Layout PR (#5241): both it and this PR added new fields at the end of the `Options` message, and the protobuf tags collided at 48. Resolved by keeping the mobile fields at 48–50 and moving `allow_osc_passthrough` to tag 51. The same end-of-`Options` clash showed up in `options.rs`, `protobuf_conversion.rs`, `common_types.proto`, the generated prost file, the KDL parser, and the roundtrip test — all resolved so both features coexist.

Full local test run after the rebase, all green:
- `zellij-server`: 1189 passed
- `zellij-utils`: 396 passed (incl. KDL snapshot + protobuf roundtrip)
- OSC 9/777 + bell-pipeline subset: 35 passed

Recap of the change for convenience:
- Forwards OSC 9 (iTerm2) and OSC 777 (rxvt) desktop notifications to the host terminal, gated behind a new `allow_osc_passthrough` option (default `true`).
- Per @yamam's earlier review, the in-Zellij visual indicator and the host-forwarded BEL are decoupled (`pending_visual_bell` vs `ring_bell`), so OSC 9/777 no longer synthesize an audible bell on the host. Regression tests cover both BEL- and ST-terminated forms.

Happy to rebase again if it drifts, or to port the tests to the new `zellij-integration-tests` framework if you'd prefer they live there.

jknlsn:
Is this now superseded by this work in #5485? I've been using this on my fork and keeping an eye on it, just saw this merged into main though. Remembered this PR and wanted to flag it!

Issue #5164, opened 2026-05-12
Control characters appear in shell prompt line when new zellij session is opened


leakybits:
## Issue description

Sometimes, when I open a new zellij session, I see my shell prompt line contains some kind of control characters: `6;25;12t`. This happens maybe 1/5 times I start zellij, not every time.

## Minimal reproduction

Loop this until you see garbage on the prompt line:

```
ssh -t someuser@somehost zellij attach -c
```

This is real shell input, not a rendering glitch:

```
$ ssh -t someuser@somehost zellij attach -c

someuser@somehost ~> 6;25;12t # hit enter
fish: Unknown command: 6
fish: Unknown command: 25
fish:
6;25;12t
  ^^
fish: Unknown command: 12t
fish:
6;25;12t
     ^~^
```

## Other relevant information

```
$ zellij --version
zellij 0.45.0

$ fish --version
fish, version 4.7.1
```

imsnif:
Hey @leakybits - just to make sure I don't go down another rabbit hole here, you are certain that the session you're attaching to is also the one from `main` with the fix and not a hanger on from a previous version right? When this happens to you, if you open the `about` screen (`Ctrl o` + `a` or `Ctrl g` + `o` + `a`) does it say `0.45.0` too? (sorry for doubting, I just want to be sure because these honestly should already be fixed)

imsnif:
I have just tried unsuccessfully over 20 consecutive times with a combination of detaching/closing the remote session. Cannot reproduce.

imsnif:
Sorry for the stress - I was planning to release the patch today. I'll wait another hour in case there are insights. Worst case if we find the problem afterwards we'll roll the fix into the next version.

leakybits:
No worries! Yes, I'm absolutely certain it's a fresh 0.45.0 session, running zellij built from main with the fix you made yesterday for the other issue. I can reproduce after a full system reboot, and can also reproduce with just `zellij` (not `zellij attach -c`) so it's definitely not picking up an old session.

The fix you made for the "missing shell prompt" is already very good to have so my 2 cents would be to release that already; this input garbage issue is much less of a bother and shouldn't delay the other fix.

leakybits:
If it helps, my original repro steps might have been misleading -- I just reproduced with `ssh user@host` followed by `zellij`, so it happens if you start `zellij` from within an existing `ssh` session, not necessarily immediately on startup via `ssh -t user@host zellij`.

imsnif:
[...]

PR #5168, opened 2026-05-13
fix: Startup prompt leak

leakybits:
This is a potential fix for https://github.com/zellij-org/zellij/issues/5164. It solves the problem on my side but I'm entirely unfamiliar with this codebase so can't speak to its broader correctness.

I used gemini to make the change; I already see some ugly things (hardcoded 0 as the startup token, arbitrary 500ms timeout).


leakybits:
Closing since this served its purpose as a reference for discussion but will never be merged.

PR #5172, opened 2026-05-14
fix: use / as daemon working directory instead of cwd

### LINKED ISSUE #5171: Web UI: sessions briefly visible then disappear ~500ms after connect (Failed to send control message to client)
**zellij version:** 0.44.1

**Describe the bug**

When connecting to the zellij web UI, the session list is visible for ~500ms then most sessions disappear. Only one session (the one being initialized by the web connect) remains.

**To reproduce**

1. Have 2+ detached zellij sessions running
2. Start `zellij web --start -d`
3. Open the web UI in a browser
4. Observe the session list briefly showing all sessions, then most disappearing within ~500ms

**Expected behavior**

All running sessions should remain visible in the web session picker.

**Logs**

From `/tmp/zellij-<uid>/zellij-log/zellij.log`, every web connection attempt produces paired errors from the `server_listener` thread immediately before/during session initialization:

```
WARN  zellij_server::os_input_o  Starting a server without a controlling terminal, using the default termios configuration.
INFO  zellij_server              Starting Zellij server!
ERROR zellij_client::web_client  [server_listener] connection_manager.rs:128: Failed to send control message to client
INFO  zellij_server              FirstClientConnected: initializing session
INFO  zellij_server              Loading plugins using Wasmi interpreter
INFO  zellij_server::screen      ApplyLayout: sending QueryTerminalSize to client 1
ERROR zellij_client::web_client  [server_listener] connection_manager.rs:128: Failed to send control message to client
INFO  zellij_server::plugins      Loaded plugin 'session-manager' in ~14ms
```

The errors appear consistently in pairs on every web connection. The same pattern is present across multiple days in the log. The `:160` variant (`Failed to send close message to client`) also appears on disconnect.

**Analysis**

The sequence suggests the web client's `connection_manager` fires control messages at all existing sessions when a new web client connects. For sessions without an active web client attached, this fails. The web UI then reconciles its session list against the control channel state and removes the sessions it couldn't reach, causing the visible flicker-and-disappear.

**Environment**

- OS: Linux 6.17.0 (x86_64)
- Accessed via: proxied HTTPS (Traefik), same behaviour observed on direct HTTP
- Token type: read-write

thereisnotime:
## Problem

Spawning a new session panics with:

```
thread 'main' panicked at zellij-server/src/lib.rs:810:
could not daemonize the server process: Error { kind: ChangeDirectory(13) }
```

This happens when the caller's working directory is a path the forked daemon process can't `chdir` into — for example `/root` when running as a non-root user (EACCES, errno 13).

## Root cause

```rust
daemonize::Daemonize::new()
    .working_directory(std::env::current_dir().unwrap())  // <-- problematic
```

`current_dir()` captures the caller's cwd. The caller can be in that directory (they have an open fd for it), but after `fork()` the daemon may not have permission to `chdir` into it fresh — `/root` is typically `700`.

## Fix

Use `/` as the daemon working directory. This follows POSIX daemon conventions (Linux `daemon(7)`, Stevens APUE §13.3): a daemon should `chdir("/")` so it never holds a reference to an arbitrary caller directory and the path is always accessible.

The server uses absolute paths for all its state (sockets, logs, config), so the working directory has no functional impact.

This code path is already `#[cfg(unix)]` only — no Windows impact.

## Reproduction

```bash
cd /root  # or any directory the user lacks execute permission on after fork
zellij attach --create some-session
```

Closes #5171 (related — separate from the web client control channel race)

### DIFF
--- a/zellij-server/src/lib.rs
+++ b/zellij-server/src/lib.rs
@@ -804,7 +804,7 @@ pub fn start_server(mut os_input: Box<dyn ServerOsApi>, socket_path: PathBuf) {
         let current_umask = umask(Mode::all());
         umask(current_umask);
         daemonize::Daemonize::new()
-            .working_directory(std::env::current_dir().unwrap())
+            .working_directory("/")
             .umask(current_umask.bits() as u32)
             .start()
             .expect("could not daemonize the server process");


thereisnotime:
Hey @imsnif @a-kenji — PTAL when you get a chance. Happy to address any feedback.

Wilfred:
FWIW I hit exactly this issue today, would like to see this merged :)

PR #5173, opened 2026-05-14
fix(session-manager): show all sessions to web clients in the picker

thereisnotime:
## Problem

When connecting to Zellij via the web client, the session picker (welcome screen) showed an empty list even when sessions existed. Users had no way to discover or attach to a running session — pressing Enter created a new one instead.

## Root cause

The session-manager plugin filtered sessions where `is_web_client && !web_clients_allowed` into a `forbidden_sessions` bucket that was never rendered. Since `web_sharing` defaults to `Off` for terminal-started sessions, every session ended up in that bucket for web clients.

## Why this filter is wrong

`web_sharing` controls whether the _server_ allows a web client to _attach_ to a session. The server enforces this at attachment time — a web client that tries to attach to a session with `web_sharing off` is rejected. The session-manager plugin filtering at discovery time does not add any security: a user who knows a session name can already navigate directly to `/<session-name>` and the same server-side check applies.

Hiding sessions from the picker only breaks the UX without providing any additional protection. Authenticated users (who have already passed TLS + any reverse-proxy auth) should be able to see what sessions exist.

## Fix

Remove the `is_web_client && !web_clients_allowed` filter from `update_session_infos()`. The `forbidden_sessions` list is now always empty. Server-side `web_sharing` enforcement at attachment time is unchanged.

No defaults were changed. Users who want sessions hidden from web pickers can already achieve that with `web_sharing off` in their config; that setting continues to prevent attachment — the only change is that the session name becomes visible in the picker, consistent with how direct URL navigation already works.

## Testing

- Web client session picker lists all sessions immediately after connecting
- Direct session URLs (`/session-name`) continue to work as before
- `web_sharing off` sessions: server still rejects attachment; only discovery behavior changes


thereisnotime:
Hey @imsnif @a-kenji — PTAL when you get a chance. Happy to address any feedback.

imsnif:
Counterpoint: you can also simply [configure it](https://zellij.dev/documentation/options.html#web_sharing) to your heart's content.

Also, respectfully - if we haven't worked with each other in the past, please don't tag me unless it's an emergency. I find this to generally be a good rule of thumb in online spaces. Your PR not getting attention for 5 hours despite [an explicit explanation of minor fix statuses](https://github.com/zellij-org/zellij/blob/main/CONTRIBUTING.md#status-of-code-contributions-prs) in our CONTRIBUTING.md file is not an emergency.

Have a nice evening.

PR #5178, opened 2026-05-15
fix(plugin): prevent duplicate WASM instances for background plugins on startup

### LINKED ISSUE #5177: `load_plugins` background plugin gets instantiated twice on startup
Following up from #4982 where I switched to a single-WASM-binary architecture (controller via `load_plugins`, sidebars via layout template). This uncovered a different startup issue, that was tricky to identify: the background plugin gets two WASM instances (not always, it's a race condition).

### What happens

When a background plugin is configured in `load_plugins`, sometimes two independent WASM instances are created on Zellij startup. Both receive `load()` and `PermissionRequestResult(Granted)`, both process pipe messages, and both subscribe to events. The second instance silently overwrites the first in the plugin map, but the first keeps running as an orphan.

Here's a minimal standalone reproducer: [rhuss/zellij-duplicate-instance-repro](https://github.com/rhuss/zellij-duplicate-instance-repro)

It counts `load()` calls by writing uniquely-named marker files to `/cache/`. Two markers means the bug triggered.

The bug is timing-dependent (~10% trigger rate in my testing). A stress test runs multiple iterations:

```bash
./stress.sh 10   # must run from outside Zellij
```

Example output:

```
  PASS  run 1/10
  PASS  run 2/10
  PASS  run 3/10
  FAIL  run 4/10  (2 instances loaded)
  PASS  run 5/10
  ...
BUG CONFIRMED: 1/10 runs created duplicate instances
```

### Config

```kdl
load_plugins {
    "file:/path/to/my_plugin.wasm"
}
```

**Expected:** One `load()` call, one WASM instance.
**Actual:** Sometimes two `load()` calls, two WASM instances with the same `plugin_id`.

You might to have run it multiple times to reveal the bug.

### Attempt of a root cause analysis

I (or better claude :-) traced this through the source (0.44.1, commit 120b649c), see the details inside.

<details>
<summary>Root Cause Analysis</summary>

1. `plugin_thread_main` calls `load_background_plugin` for each entry in `background_plugins` **before** entering the event loop (`zellij-server/src/plugins/mod.rs`, around line 345).

2. `load_plugin` uses `initiating_client_id` (e.g. 1) as a dummy client because `connected_clients` is empty at this point (`wasm_bridge.rs`, around line 295). It dispatches the WASM load to the pinned executor.

3. The event loop starts. `AddClient(1)` arrives from the server (`mod.rs`, around line 513).

4. `add_client` checks `client_is_connected(1)` (`wasm_bridge.rs`, around line 723). This returns `false` because the dummy client_id was never pushed to `connected_clients` (that happens at the end of `add_client` itself, around line 820).

5. `add_client` iterates `plugin_ids()` from `plugin_map`. If the executor has already completed the background load, `plugin_id=0` is in the map. It creates a **second** WASM instance for `(plugin_id=0, client_id=1)`.

6. The `plugin_map.insert((0, 1), ...)` overwrites the first instance. The first WASM store stays in memory, still receiving cached events.

### Possible fix

The simplest guard would be checking in `add_client` whether `(plugin_id, client_id)` already exists before creating a new instance:

```rust
// In add_client, around line 731:
for plugin_id in new_plugins {
    if self.plugin_map.lock().unwrap().contains_key(&(plugin_id, client_id)) {
        continue;
    }
    // ... existing clone logic ...
}
```

This prevents the duplicate without changing the loading order or client tracking.

</details>

### Environment

- Zellij 0.44.1 (macOS, also tested on Fedora 41)
- zellij-tile 0.44

### Impact for plugin authors

Any background plugin that maintains internal state sees inconsistent behavior during the overlap window. Plugins that register keybindings via `reconfigure()` register them twice. Plugins that write to `/cache/` see concurrent writes from both instances. My workaround is a leader election protocol in the plugin itself (lowest `plugin_id` wins, loser goes dormant), but that adds significant complexity for something that should be prevented at the host level.

I can work on a PR to fix this, if this a confirmed bug.

rhuss:
## Summary

Fixes #5177

Background plugins configured via `load_plugins` can get instantiated twice on startup due to a race condition between `load_background_plugin` and `add_client`. When `AddClient` arrives while a background plugin is already loaded (or loading), `add_client()` creates a second WASM instance for the same `(plugin_id, client_id)` pair because it doesn't check for existing entries in the plugin map.

This PR adds a simple guard: before creating a new plugin instance in the `add_client()` loop, check if the `(plugin_id, client_id)` pair already exists in the plugin map and skip if so.

## Changes

- **`plugin_map.rs`**: Add `contains(plugin_id, client_id)` method to `PluginMap` (the internal `plugin_assets` HashMap is private)
- **`wasm_bridge.rs`**: Add guard in `add_client()` to skip instance creation when the plugin/client pair already exists

## Root Cause

1. `plugin_thread_main` calls `load_background_plugin` for each `load_plugins` entry *before* entering the event loop, using `initiating_client_id` as a dummy client
2. The event loop starts and `AddClient(client_id)` arrives
3. `add_client()` iterates all `plugin_ids()` from the plugin map and unconditionally creates new instances for the connecting client
4. If `initiating_client_id == client_id` (common case) and the background plugin load already completed, `plugin_map.insert` overwrites the first instance. The original WASM store stays in memory as an orphan, still receiving cached events.

## Testing

- All existing tests pass (`cargo test` in `zellij-server/`, 1088 tests)
- Validated with the [standalone reproducer](https://github.com/rhuss/zellij-duplicate-instance-repro): 20/20 runs pass (0% failure rate, previously ~10%)

Assisted-By: 🤖 Claude Code

### DIFF
--- a/zellij-server/src/plugins/plugin_map.rs
+++ b/zellij-server/src/plugins/plugin_map.rs
@@ -208,6 +208,9 @@ impl PluginMap {
         }
         cloned_plugin_assets
     }
+    pub fn contains(&self, plugin_id: PluginId, client_id: ClientId) -> bool {
+        self.plugin_assets.contains_key(&(plugin_id, client_id))
+    }
     pub fn all_plugin_ids(&self) -> Vec<(PluginId, ClientId)> {
         self.plugin_assets
             .iter()
--- a/zellij-server/src/plugins/wasm_bridge.rs
+++ b/zellij-server/src/plugins/wasm_bridge.rs
@@ -706,6 +706,9 @@ impl WasmBridge {
             new_plugins.insert(plugin_id);
         }
         for plugin_id in new_plugins {
+            if self.plugin_map.lock().unwrap().contains(plugin_id, client_id) {
+                continue;
+            }
             let Some(run_plugin) = self.run_plugin_of_plugin_id(plugin_id).map(|r| r.clone())
             else {
                 log::error!("Failed to find plugin with id: {}", plugin_id);


rhuss:
## Stress Test Results

Tested with the [standalone reproducer](https://github.com/rhuss/zellij-duplicate-instance-repro) (20 iterations each).

<details>
<summary>With fix (patched binary): 20/20 PASS</summary>

```
$ PATH=~/Development/misc/zellij/zellij/target/release:$PATH ./stress.sh 20

Running 20 iterations (each starts Zellij, waits 2s, auto-quits)

  PASS  run 1/20
  PASS  run 2/20
  PASS  run 3/20
  PASS  run 4/20
  PASS  run 5/20
  PASS  run 6/20
  PASS  run 7/20
  PASS  run 8/20
  PASS  run 9/20
  PASS  run 10/20
  PASS  run 11/20
  PASS  run 12/20
  PASS  run 13/20
  PASS  run 14/20
  PASS  run 15/20
  PASS  run 16/20
  PASS  run 17/20
  PASS  run 18/20
  PASS  run 19/20
  PASS  run 20/20

ALL PASSED: 20/20 runs had exactly 1 instance
```

</details>

<details>
<summary>Without fix (stock 0.44.1): 5/20 FAIL</summary>

```
$ ./stress.sh 20

Running 20 iterations (each starts Zellij, waits 2s, auto-quits)

  PASS  run 1/20
  PASS  run 2/20
  FAIL  run 3/20  (2 instances loaded)
  PASS  run 4/20
  PASS  run 5/20
  PASS  run 6/20
  PASS  run 7/20
  PASS  run 8/20
  FAIL  run 9/20  (2 instances loaded)
  PASS  run 10/20
  PASS  run 11/20
  PASS  run 12/20
  PASS  run 13/20
  FAIL  run 14/20  (2 instances loaded)
  FAIL  run 15/20  (2 instances loaded)
  FAIL  run 16/20  (2 instances loaded)
  PASS  run 17/20
  PASS  run 18/20
  PASS  run 19/20
  PASS  run 20/20

BUG CONFIRMED: 5/20 runs created duplicate instances
```

</details>

imsnif:
Thank you!

PR #5180, opened 2026-05-16
Fix ECH artifacts when erasing wide characters

### LINKED ISSUE #5101: Neovim rendering artifacts in empty buffer area inside Zellij
#### Summary

Neovim shows small white rectangular artifacts in the empty buffer area when running inside Zellij.

This does **not** reproduce when running Neovim directly in the terminal emulator outside Zellij.

I reproduced this in both:

- GNOME Terminal 3.58.0 using VTE 0.84.0 (+BIDI +GNUTLS +ICU +SYSTEMD)
- WezTerm 20260331-040028-577474d8

#### Screenshot
Image
The attached screenshot shows two vertical Neovim splits, with a row of small white rectangular artifacts visible in otherwise empty buffer area near the lower part of both panes.

#### Basic information

- OS: Ubuntu 26.04
- Zellij version: built from `main`, commit 5ab8b687
- Neovim: NVIM v0.12.3-dev-30+g0f6f5cd0a7
- Terminal (primary reproduction): GNOME Terminal 3.58.0 using VTE 0.84.0
- Also reproduced in: WezTerm 20260331-040028-577474d8

#### Minimal reproduction

```bash
mkdir -p /tmp/zellij-empty
zellij --debug --config-dir /tmp/zellij-empty
```
Then inside Zellij:
```
nvim --clean
```
Then inside Neovim:
```
:vs
```
At that point, the rendering artifacts appear in the empty area.
#### Important observations

- Reproduces inside Zellij with:
  - TERM=wezterm
  - TERM=vte-256color
  - TERM=xterm-256color
- Does **not** reproduce inside Zellij with:
  - TERM=tmux-256color

Additionally, I compared terminfo entries and found differences in bce and ech between wezterm and tmux-256color.
As a further test, I created custom terminfo entries based on wezterm:

disabling only bce (bce@) -> issue does not reproduce
disabling only ech (ech@) -> issue does not reproduce

This suggests the issue may be related to how Neovim's erase/background optimization paths interact with Zellij rendering.

#### Debug log
I am attaching the zellij.log captured with zellij --debug.

[zellij.log](https://github.com/user-attachments/files/27118318/zellij.log)

yamam:
## Description

This fixes rendering artifacts that can appear when Neovim erases wide characters with `CSI X` / ECH inside Zellij.

When a wide character is replaced with a single-width blank, `Row::replace_character_at` inserts padding cells for the remaining width. Previously these padding cells used `EMPTY_TERMINAL_CHARACTER`, which has reset/default styling.

If the replacement character had a non-default background, the first erased cell kept the expected style, but the padding cell lost it. This could leave a small rectangular artifact in otherwise empty buffer areas.

This change makes the padding cell inherit the replacement character's styles.

## Reproduction

This was observed with Neovim inside Zellij:

```sh
zellij --debug --config-dir /tmp/zellij-empty
nvim --clean
````

Then inside Neovim:

```vim
:vs
```

In my case, the Neovim start screen contained wide characters, and opening a vertical split caused small rectangular artifacts in the empty buffer area.

## Fix

When replacing a wide character with a narrower character, use a padding cell that inherits the replacement character's styles instead of using `EMPTY_TERMINAL_CHARACTER` directly.

## Tests

Added a regression test that:

1.  Writes a wide character with a background color
2.  Moves the cursor back to it
3.  Erases it with ECH
4.  Verifies that both cells keep the expected background style

Tested with:

```sh
cargo test -p zellij-server ech_over_wide_character_preserves_background_on_padding_cell
cargo test -p zellij-server --lib
```

Fixes #5101

### DIFF
--- a/zellij-server/src/panes/grid.rs
+++ b/zellij-server/src/panes/grid.rs
@@ -5541,11 +5541,14 @@ impl Row {
             let overwrite_end = x + character.width().max(terminal_character_width);
             self.osc133_markers
                 .retain(|marker| marker.column < x || marker.column >= overwrite_end);
+
+            let mut padding_character = EMPTY_TERMINAL_CHARACTER;
+            padding_character.styles = terminal_character.styles.clone();
             let character = std::mem::replace(character, terminal_character);
             let excess_width = character.width().saturating_sub(terminal_character_width);
             for _ in 0..excess_width {
                 self.columns
-                    .insert(absolute_x_index, EMPTY_TERMINAL_CHARACTER);
+                    .insert(absolute_x_index, padding_character.clone());
             }
         }
         self.width = None;
--- a/zellij-server/src/panes/unit/grid_tests.rs
+++ b/zellij-server/src/panes/unit/grid_tests.rs
@@ -8772,3 +8772,31 @@ fn sixel_dcs_is_not_confused_with_xtgettcap() {
         "a sixel image must not produce an XTGETTCAP reply"
     );
 }
+
+#[test]
+fn ech_over_wide_character_preserves_background_on_padding_cell() {
+    use crate::panes::terminal_character::NamedColor;
+    // Use a wide character to exercise replacement of a wide cell.
+    assert_eq!(crate::panes::TerminalCharacter::new('Ａ').width(), 2);
+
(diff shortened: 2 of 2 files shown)


imsnif:
Thank you!

Issue #5128, opened 2026-05-04
Statusbar does not update mode section after CloseFocus keybinding.


shans10:
# 1. Graphical issue inside a terminal pane (eg. something does not look as it should)

**Basic information**

`zellij --version`: zellij 0.44.1

`stty size`: 43 16

`uname -av`: Linux prestige 7.0.3-1-cachyos, Sat, 02 May 2026 02:50:09 +0000 x86_64 GNU/Linux

## Issue description
I mostly use zellij in Tmux mode with `default layout` set to `compact` and I have the following keybinding for it:
```kdl
// Close focused tab/pane
bind "x" { CloseFocus; SwitchToMode "Normal"; }
```
which works with the prefix Ctrl+B. Now whenever I use this keybinding to kill a pane or a tab it works fine and zellij returns back to the Normal mode which is evident because I am able to type and use the terminal however this change is not reflected in the statusbar. The mode section of the compact bar still shows Tmux however all the normal mode keybindings work however for the statusbar to reflect the change I have to press Ctrl+B twice to first change into Tmux mode and then to Normal mode for the statusbar to update.

## Minimal reproduction
- Open zellij.
- Create a new tab.
- Presss `Ctrl+B x` to close the newly created tab/pane.
- The statusbar show show  `Normal` in mode however it still shows `Tmux`.

## Other relevant information
I have attached the zellij log below:
**log**:  [zellij.log](https://github.com/user-attachments/files/27351001/zellij.log)

Ape:
I think I can reproduce this issue with:
1. ctrl+t n (open a new tab)
2. close the new tab, returing to the original (e.g. ctrl+d to exit the new shell)
3. the original tab shows "TAB" mode while it's actually in "NORMAL"

Ape:
Probably duplidate of #5114.

PR #5181, opened 2026-05-17
Fix stale plugin mode updates after closing tabs

### LINKED ISSUE #5114: A tab can display stale state when it becomes visible again after another tab is closed because its ModeUpdate state is stale

## Summary

I think there are two closely related issues involving tab-local `tab-bar` / `status-bar` plugin panes when a tab is closed immediately after it becomes active.

Both issues happen with the following layout:

```kdl
layout {
  default_tab_template {
    pane size=1 borderless=true {
      plugin location="tab-bar"
    }
    pane
    pane size=1 borderless=true {
      plugin location="status-bar"
    }
  }
  tab
  tab
}
```

### Steps to reproduce

1. Start Zellij with zellij setup --clean
1. Use the layout above
1. Immediately press Ctrl-D in one of the initial tabs to close it

### Expected behavior
The remaining tab should show a complete tab-bar / status-bar, including session-related UI and normal powerline rendering.

### Actual behavior
The remaining tab sometimes shows an incomplete tab-bar.
For example, session-related UI and/or powerline-style fragments are missing.

### Log and screenshot
[zellij-bug1.txt](https://github.com/user-attachments/files/27266500/zellij-bug1.txt)
Image

### Root cause hypothesis
This looks like the surviving tab’s tab-bar / status-bar instance renders before it has a valid ModeInfo.
That would explain why:

- mode-dependent display decisions are still at default/empty values
- session-name-dependent fragments are missing
- powerline/feature decisions that rely on style or capability metadata behave as if disabled

### Steps to reproduce

1. Start Zellij with zellij setup --clean
1. Use the layout above
1. Create a new tab via Ctrl-t n
1. Immediately press Ctrl-D in the new tab to exit the shell
1. Return to the previous tab automatically

### Expected behavior
The previous tab's tab-bar should no longer show TAB.
### Actual behavior
The previous tab's tab-bar can still display TAB.

### Log and screenshot
[zelli-bug2.txt](https://github.com/user-attachments/files/27266503/zelli-bug2.txt)
Image

### Root cause hypothesis
This looks like the previous tab’s plugin instance becomes visible again without receiving a fresh ModeUpdate, so it continues to render an outdated mode string.

## Possibly related issues
These may be related, but I could not find one that matches this exact ModeInfo / stale plugin-state behavior:

#667 Broken rendering when closing a tab
#2234 Crashes after closing a recently created tab
#2248 Panic on closing a tab
#628 [Plugin] Status bar plugin issue when the status bar is at the top

yamam:
## Summary

Fixes #5114.

When a tab is closed, clients are moved to another tab through `move_clients_from_closed_tab()`. This path did not refresh the destination tab's plugin mode state, unlike the regular tab-switch path, which calls `update_input_modes()` for the destination tab.

As a result, tab-local plugins such as `tab-bar` / `status-bar` could render stale mode information after the previous tab became visible again.

Also addresses #5128.

## Changes

- Track the tabs that receive clients after a tab is closed.
- Call `update_input_modes()` for those destination tabs.
- Add a regression test to ensure a tab-local plugin receives a fresh `ModeUpdate` after the active tab is closed and a previous tab becomes visible again.

## Testing

```bash
cargo test -p zellij-server closing_active_tab_sends_fresh_mode_update_to_revealed_tab_plugins
````

Manually verified that the stale tab-bar/status-bar mode display no longer reproduces after closing a tab.

### DIFF
--- a/zellij-server/src/screen.rs
+++ b/zellij-server/src/screen.rs
@@ -1686,6 +1686,9 @@ impl Screen {
             .next()
             .context("screen contained no tabs")
             .with_context(err_context)?;
+
+        let mut tabs_to_update = HashSet::new();
+
         for (client_id, client_mode_info) in client_ids_and_mode_infos {
             let client_tab_history = self.tab_history.entry(client_id).or_insert_with(Vec::new);
             if let Some(client_previous_tab) = client_tab_history.pop() {
@@ -1694,6 +1697,7 @@ impl Screen {
                     client_active_tab
                         .add_client(client_id, Some(client_mode_info))
                         .with_context(err_context)?;
+                    tabs_to_update.insert(client_previous_tab);
                     continue;
                 }
             }
@@ -1703,7 +1707,18 @@ impl Screen {
                 .with_context(err_context)?
                 .add_client(client_id, Some(client_mode_info))
                 .with_context(err_context)?;
+            tabs_to_update.insert(first_tab_index);
         }
+
+        // Closing a tab moves its clients without going through the normal tab-switch path.
+        // Refresh ModeUpdate for the destination tabs so tab-local plugins such as
+        // tab-bar/status-bar do not render stale mode information when they become visible again.
+        for tab_id in tabs_to_update {
+            if let Some(tab) = self.tabs.get_mut(&tab_id) {
+                tab.update_input_modes().with_context(err_context)?;
+            }
+        }
+
         Ok(())
     }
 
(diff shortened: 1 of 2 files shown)


yamam:
Closing this PR as superseded by #5535, which has been merged.

#5535 fixes the same stale input-mode updates after closing or switching tabs, including the tab-bar/status-bar behavior described here. It also covers additional tab-switch paths beyond this PR.

The changes from this PR are therefore no longer needed separately.

PR #5184, opened 2026-05-17
feat: PWA support for the web client

### LINKED ISSUE #5094: Adding PWA support for zellij web
Hi,
Thought it would be nice if PWA support was added for the zellij web client, so you can launch it in its own window and have it look more native.
At minimum would need:
- A logo/icon served at /<icon-subpath>
- A manifest served at /manifest.webmanifest containing at minimum https://developer.mozilla.org/en-US/docs/Web/Progressive_web_apps/Guides/Making_PWAs_installable#required_manifest_members, for example:
```
{
    "name": "Zellij Web Client",
    "start_url": "/<subpath>/", <- these and below are based on the base_url in zellij.kdl
    "display": "standalone",
    "display_override": ["window-controls-overlay", "standalone", "minimal-ui"],
    "icons": [
        {
            "src": "/<subpath>/icon.svg",
            "sizes": "any",
            "type": "image/svg+xml",
            "purpose": "any maskable"
        }
    ]
}
```

I might give this a shot myself soon

jzyrobert:
## Summary

Closes #5094.

Adds Progressive Web App support to the Zellij web client so the page can be installed as a standalone app via the browser's "Install" / "Add to Home Screen" affordance. Implementation is intentionally minimal — manifest + a single 192×192 icon + iOS `apple-touch-icon` link, no service worker, no offline mode (a terminal client requires a live WebSocket).

The whole change is additive: existing behaviour is unchanged for users who don't install the PWA.

On opening a localhost/HTTPS web client:
image

Then opening the tab in standalone mode:
image

What a standard session looks like in standalone mode:
image

Working over a reverse-proxy on `/zellij/`:
image

### What's in the diff

- **Manifest** (`zellij-client/assets/manifest.webmanifest`): required W3C members plus `short_name`, `scope`, `display_override`, `theme_color`, `background_color`. All path-bearing fields (`start_url`, `scope`, `icons[*].src`) are relative to the manifest URL, so the same static file works under any reverse-proxy `base_url` prefix (which must end with a trailing slash — `base_url "/zellij/"`, not `"/zellij"`) without per-deployment templating. `display_override` is `["standalone", "minimal-ui"]`.
- **Icon** (`zellij-client/assets/icon-192.png`): a single 192×192 square PNG rendered from `assets/logo.svg` and composited onto a transparent square canvas. Chromium's installability gate requires a square icon ≥144px; the existing 2269×2620 `assets/logo.png` fails that gate, so a pre-rasterized square variant is committed alongside. Re-render with `rsvg-convert` + ImageMagick `convert` if the logo changes.
- **MIME mapping** (`zellij-client/src/web_client/utils.rs`): adds `.webmanifest` → `application/manifest+json`. Safari rejects manifests served as `application/json`.
- **HTML head** (`zellij-client/assets/index.html`): adds `<link rel="manifest">`, `<link rel="apple-touch-icon" href="assets/icon-192.png">`, and `<meta name="theme-color" content="#000000">`. Tags placed after the existing favicon link and before any `<script>` so the manifest hits the browser's preload scanner early. The three `apple-mobile-web-app-*` meta tags are intentionally omitted — they're backward-compat scaffolding for iOS <16.4, and iOS 16.4+ reads the equivalents (`display: standalone`, `short_name`) from the manifest. iOS still ignores manifest `icons`, so `apple-touch-icon` is the one Apple-specific tag genuinely needed on every iOS version.
- **Tests** (`zellij-client/src/web_client/unit/web_client_tests.rs`): 5 new tests for Content-Type, required JSON members, the relative-URL contract, index.html references, icon reachability, and `base_url` invariance. They follow the existing `serve_web_client` + `wait_for_server` + `isahc::get` pattern used elsewhere in the file.

## Test plan

Automated coverage is in place; the items below require human/device-specific verification that CI can't run.

- [x] `cargo build -p zellij-client --features web_server_capability` clean
- [x] `cargo test -p zellij-client --features web_server_capability --lib web_client::web_client_tests` — 28 passed (existing 23 + new 5)
- [x] `cargo clippy -p zellij-client --features web_server_capability --lib --tests` — no new warnings in changed files
- [x] **Manual Chromium desktop testing**: install the PWA via "Install Zellij Web Client", launch from OS, log in, confirm terminal works in standalone window
- [x] **Sub-path deployment**: configure `web_client.base_url "/zellij/"` (with trailing slash) in `zellij.kdl`, run behind a reverse proxy that strips the prefix, confirm install works and `start_url`/icon paths resolve correctly.


jzyrobert:
The changes were driven by Claude Code but it's been scoped down to the minimum changes needed and the code looks fine to me, and I have tested it locally using the root path and a base_url behind nginx.

The additional icon is required because the base one is not square and this is required by Chromium browsers.
Not tested on iOS devices as I don't own any.

Additional features could be:
- Add support for setting the manifest name, e.g. to distinguish the PWA app shortcuts on different hosts. Would require serving the manifest over code though rather than static.

imsnif:
Beautiful, thank you!

PR #5189, opened 2026-05-19
fix: preserve shifted character when encoding kitty keys to non-kitty panes

lyskov:
## Summary

When a host terminal speaking the kitty keyboard protocol sends a shifted punctuation key (e.g. Alt+Shift+? as `\x1b[47;4u` — code point 47 = `/` + modifiers Alt|Shift) to a pane that does **not** speak kitty, zellij re-serializes the key via `serialize_non_kitty()` and the inner terminal receives `\x1b/` (ESC + `/`) instead of `\x1b?` (ESC + `?`).

The kitty protocol reports the *unshifted* code point alongside the SHIFT modifier. The downstream termwiz encoder, however, assumes shift has already been baked into the character (xterm convention) and strips SHIFT for punctuation — so `/` + SHIFT becomes plain `/`, dropping the user-intended `?`.

This patch applies a US-keyboard shift mapping (`/` → `?`, `1` → `!`, etc.) in `serialize_non_kitty()` before handing the key to termwiz, so the inner pane receives the shifted character the user actually pressed. Letters were already handled correctly via termwiz's `normalize_shift_to_upper_case` (a → A); only ASCII punctuation/digits needed bridging.

## How to reproduce

Inside a terminal pane in zellij (running under a kitty-protocol-capable host terminal such as Kitty, Ghostty, WezTerm, foot, or Alacritty 0.13+): `sed -n l`

sed -n l echoes every keystroke as its literal byte sequence. Press Alt+Shift+? and observe the output.

- Before this patch (broken): `^[/`
- After this patch (fixed):   `^[?`

Running the same sed -n l outside zellij in the host terminal directly will show `^[?` — confirming the host produced the right keystroke and zellij was mistranslating it.

Other affected combinations include `Alt+Shift+1`, `Alt+Shift+;`, `Alt+Shift+.` , and so on.


lyskov:
Just to add some context: this bug causes a number of keyboard shortcuts to stop working in Midnight Commander (mc), making it difficult to use effectively inside Zellij. Happy to provide more details if helpful!

lyskov:
@imsnif Gentle ping on this one — this bug still affects some shortcuts in Midnight Commander inside Zellij. Happy to provide more details or adjust the patch if helpful. Thanks!

imsnif:
Hi @lyskov - thanks for the reminders and good find on the issue.

I'm hesitant to hard-code things to a US keyboard layout. Do you think we can make this more agnostic?

lyskov:
Good call @imsnif  — made it layout-agnostic.

The client now also requests kitty's "report alternate keys" (CSI >5u). For modified keys the terminal reports the layout-correct shifted codepoint (unicode-key:shifted:base), which we use when down-converting to non-kitty panes — so Alt+Shift+ß on a German layout correctly yields ?. The US table stays only as a fallback for terminals that don't report alternates.

imsnif:
Hey @lyskov - thanks for taking another look. I however don't at this time want to go any deeper than we already are into the KKP. Ever since we added even minimal support for this protocol, it has created significant issues of compatibility with other pieces of software that have degraded the experience for Zellij users (much more so than not having support for it at all). I generally view this protocol as problematic (just a small example being that it does not specify any sort of mandatory progressive enhancement level negotiation, hoping instead that software+terminals just play nice with each other).

It sounds like for your use case it might be easier for you to [disable it entirely](https://zellij.dev/documentation/options.html#support_kitty_keyboard_protocol)?

lyskov:
Makes sense — thanks for the context on the compatibility tradeoffs.

  I'm honestly not sure what else could address this without diving deeper into KKP, since the shifted character just isn't in the data the terminal sends at the minimal level. Do you have anything in mind?

  As for disabling it entirely — that would work, but I was hoping to keep it on so I can use the Meta modifier in some of my shortcuts (which is partly why I'm on KKP in the first place). Any recommendation for a workaround that keeps that working?

  Thanks!

imsnif:
The direction I'd look at is trying to hold on to the original bytes before they are converted to KKP and possibly sending them directly (or modified as needed) to the pane in question. Might not work or need some extra attention, but I'd start there.

PR #5208, opened 2026-05-24
fix: convert paste body \n to \r for panes without bracketed paste

### LINKED ISSUE #5207: Multi-line paste loses newlines when target program doesn't speak bracketed paste (e.g. UW pico on macOS)
Spent a few hours running this one down, posting in case it helps someone
or a maintainer wants to take a look.

The setup: macOS, alacritty + zellij, pasting into `/usr/bin/nano`. On
macOS that's UW PICO 5.09, not GNU nano — caught me by surprise too.
Every newline in the paste gets eaten and lines collapse into wrapped
paragraphs separated by spaces. Doesn't reproduce in alacritty alone or
in Terminal.app + zellij, only the combination.

*(A short screen recording demonstrating the bug — pasting into pico
shows the mangling, pasting the same content into GNU nano works
correctly — will be attached as a comment below this body.)*

## Reproducing

1. Open alacritty, start zellij.
2. Inside zellij: `/usr/bin/nano /tmp/test.txt` — header bar will say
   "UW PICO 5.09".
3. Copy any multi-line text to your clipboard, e.g.
   ```
   # comment
   function() {
     a = 1
     b = 2
   }
   ```
4. Cmd+V to paste.
5. Expected: five lines in the buffer matching the source.
6. Actual: two lines, with words from the source separated by spaces
   where newlines should be.

## What's actually happening

Pico isn't broken. Its `^J` (= `\n`, 0x0a) is bound to "Justify
Paragraph", and pico doesn't enable bracketed-paste mode so it has no
way to know it's receiving a paste vs typed input. Every `\n` in the
body fires Justify.

The reason this works in every other configuration is that
single-terminal paste paths translate `\n` → `\r` when the inner program
doesn't have bracketed paste on. Alacritty does this explicitly in its
non-bracketed paste branch, macOS Terminal.app does it, xterm does it.
So in those setups pico sees `\r` (= Enter) and inserts a proper line
break.

Zellij in the middle doesn't do that translation. When the inner pane
hasn't enabled BP, zellij strips the `\x1b[200~`/`\x1b[201~` markers
(correctly) but leaves the body bytes unchanged. So `\n` from the host
clipboard reaches pico verbatim, and Justify fires once per line.

Confirmed with `cat | xxd` inside zellij + alacritty — body arrives
with `0a` bytes intact, no markers, no `\r` conversion. With
`printf '\e[?2004h'; cat | xxd; printf '\e[?2004l'` the markers come
through but body is still `0a`. So it's specifically the
marker-stripping path that's missing the `\n` → `\r` translation the
host terminal would have applied.

## Proposed fix

When zellij strips BP markers for a pane that hasn't enabled BP mode,
also translate `\n` → `\r` in the body bytes until the END marker.
Matches what every other terminal does in the same scenario. Panes
that *have* enabled BP (vim, GNU nano, shells at the prompt, etc.) are
unaffected since the body flows through verbatim with markers intact,
same as today.

Platform-independent change — no `cfg(target_os)`, no Mac-specific
paths. Anyone running pico or another `^J`-binding program on Linux/BSD
via zellij gets the same fix.

I have a small patch ready (~15 lines + two tests covering both
directions: translation happens during a paste, doesn't happen outside
one). Happy to put up a PR if there's interest.

## Versions

- macOS 26.4.1
- zellij 0.44.3 (also reproduced on latest main built locally)
- alacritty 0.17.0
- `/usr/bin/nano`: UW PICO 5.09

edvorg:
Closes #5207.

Diagnostic and reproducer are in the issue. Short version: when
forwarding a bracketed paste to a pane whose inner program hasn't
enabled BP, zellij strips the markers but leaves `\n` bytes in the body
unchanged. Programs like UW pico that bind `^J` to a command mangle the
paste. Other terminal emulators do `\n` → `\r` translation in this case;
this PR brings zellij in line.

The fix is a per-pane `pending_paste_passthrough` flag on `TerminalPane`,
set when the `\x1b[200~` BEGIN marker is stripped, cleared by `\x1b[201~`.
While set, `\n` bytes in the body are rewritten to `\r` before
forwarding. Outside the paste window the flag is off and bytes pass
through untouched — so ordinary input is never affected.

*(A short screen recording verifying the fix end-to-end — pasting the
same multi-line snippet from the issue into UW pico through alacritty +
patched zellij — will be attached as a comment below.)*

## Tests

Two new tests next to the existing BP regression test:

- **positive**: body `line one\nline two\nline three` arrives at the
  inner program as `line one\rline two\rline three`
- **negative**: ordinary `\n` arriving outside a paste window is left
  alone (no false positives on normal input)

All 1090 existing `zellij-server` unit tests still pass.

## End-to-end verification

Verified on macOS by pasting a multi-line snippet into UW pico through
alacritty + patched zellij — the file saved by pico now contains real
newlines instead of justified paragraphs. Behavior unchanged for panes
with bracketed-paste mode enabled (vim, GNU nano, shells at prompt,
etc.) since those receive the markers and full body verbatim, same as
today.

## Platform safety

Platform-independent — no `cfg(target_os)` or OS-specific paths anywhere
in the diff. Linux/BSD users running UW pico (or any other program with
`^J` bound to a command) through zellij benefit identically. The bug
just happens to be most visible on macOS because that's where
`/usr/bin/nano` ships as UW pico.

### DIFF
--- a/zellij-server/src/panes/terminal_pane.rs
+++ b/zellij-server/src/panes/terminal_pane.rs
@@ -166,6 +166,15 @@ pub struct TerminalPane {
     /// remaining bytes in the queue to be drained after the host reply
     /// has been written.
     pending_pty_input: VecDeque<u8>,
+    /// `true` while a host paste is being forwarded into a pane whose inner program
+    /// has not enabled bracketed paste mode. Set when the `\x1b[200~` begin marker is
+    /// stripped, cleared by the matching `\x1b[201~` end marker. While set, `\n`
+    /// bytes in the body are translated to `\r` before reaching the inner program —
+    /// matching the convention used by Alacritty's own non-bracketed paste path and
+    /// macOS Terminal.app. Without this, inner programs that don't speak bracketed
+    /// paste (e.g. UW pico, where `^J` is bound to Justify Paragraph) interpret the
+    /// `\n` bytes as commands and mangle the paste.
+    pending_paste_passthrough: bool,
 }
 
 impl Pane for TerminalPane {
@@ -251,7 +260,7 @@ impl Pane for TerminalPane {
     fn adjust_input_to_terminal(
         &mut self,
         key_with_modifier: &Option<KeyWithModifier>,
-        raw_input_bytes: Vec<u8>,
+        mut raw_input_bytes: Vec<u8>,
         raw_input_bytes_are_kitty: bool,
         client_id: Option<ClientId>,
     ) -> Option<AdjustedInput> {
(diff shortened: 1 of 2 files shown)


edvorg:
Video that is demostrating the fix 

https://github.com/user-attachments/assets/194f71e9-8ff2-4103-ad50-1ade304c8044

There is also a video demonstrating the problem in the issue #5207

edvorg:
Closing it to clean up my pull requests view. Feel free to reopen if this raises any interest

PR #5225, opened 2026-06-01
fix: Preserve z-order of floating panes when using MoveFocusOrTab

mjkpolo:
fix: preserve z-order of floating panes when using MoveFocusOrTab

MoveFocusOrTab is often more convenient than the modal Tab mode. When switching tabs normally, z-order is preserved. However when using MoveFocusOrTab, an "edge" floating window is selected. It is far more convenient to not re-focus on an "edge" floating window. That way we can quickly move between tabs with the relevant floating panes remaining visible.

Before (always go to pane with "bar" even if "foo" was on top):
wrong_zellij

After (always go to pane that was on top last time):
correct_zellij

### DIFF
--- a/zellij-server/src/panes/floating_panes/floating_pane_grid.rs
+++ b/zellij-server/src/panes/floating_panes/floating_pane_grid.rs
@@ -625,7 +625,8 @@ impl<'a> FloatingPaneGrid<'a> {
             .copied();
         next_index
     }
-    pub fn pane_id_on_edge(&self, direction: Direction) -> Option<PaneId> {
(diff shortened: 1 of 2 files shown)


mjkpolo:
I realized just removing the functionality to find edges of floating panes probably isn't good and may break someone's workflow. Instead, I change the tiebreaker to use z order instead of the opposite dimension (if direction is down and y pos equal, take highest zorder instead of x pos)

Still not sure if this is the right approach, and what the original behavior was intended to be... IMO switching tabs should not trigger a new focus, but perhaps others depend on this.

PR #5236, opened 2026-06-07
fix: query terminal palette in a single command

jatwin:
During `zellij attach`, the terminal color registers are queried, and previously each register was queried separately with a OSC 4 sequence. This creates a delay when attaching to a session from within a session, on my machine of about ~10s. This PR batches the queries into a single one


imsnif:
Hey @jatwin, thanks for this! I was planning on removing the startup query entirely since we now also do this on demand and I think it's not needed anymore. Want to give that a shot instead? It'll essentially involve removing it, then testing the various query types from within a terminal pane to make sure they still get a proper response.

jatwin:
> Hey @jatwin, thanks for this! I was planning on removing the startup query entirely since we now also do this on demand and I think it's not needed anymore. Want to give that a shot instead? It'll essentially involve removing it, then testing the various query types from within a terminal pane to make sure they still get a proper response.

Hey, just saw the reply, unfortunately don't have the bandwidth but in that case will close this one given that a better fix is coming

PR #5241, opened 2026-06-08
Mobile layout

imsnif:
mobile-web-3
mobile-web-2
mobile-web-1

This adds a new "mobile" plugin, set out to solve the bad UX we get when attaching to a Zellij session through a mobile browser.

The mobile plugin acts as its own "hidden" tab, internally employing a new "shadow focus" mechanism in order to both report which pane the user is actually focused on and to allow the user to focus on different panes.

This allows the plugin to do its own secondary render, rather than relying on the actual pane render, which allows us to support panning (so that the size of the pane won't be locked to the size of the mobile phone if for example we'd like to keep working on the session through a desktop simultaneously), as well as pinch-to-resize and other similar conveniences.

The mobile plugin also includes its own built-in version of the welcome screen / session manager, allowing users to switch between sessions and create new ones.

It's of course also possible to switch-to-desktop and return to the normal experience.


imsnif:
@tlinford - if you can test this on an iphone it'll make my life easier :)

Zykino:
Wow, that sounds like a huge change. Congrat on that! :tada: 

I know quite a few peoples want this kind of feature but for embedded terminal in IDE. Do you think it could be used there as well?
Especially those are real terminal emulator (not browser's tab) and are most of the time in thin landscape instead of mobile's portrait.
Or maybe this add the architecture and another plugin would be more tailored for that purpose?

I'm asking not primary for me as a user of the feature but more because it is a recurring question/need to integrate zellij in IDE's terminal. So I prepare myself for when the question drop next time :)

imsnif:
Good point! This also works as is with a non-browser terminal attach assuming it starts small enough. Both the behavior and the rows/cols "breakpoints" are configurable. I hope this will do the trick?

jjmachan:
this is great!!!! wanted this so badly
thanks for shipping this ❤️ 🙌🏽

Issue #3409, opened 2024-06-09
bug: `LaunchOrFocusPlugin` creates new plugin instances when using plugin aliases


cristiand391:
# 2. Issues with the Zellij UI / behavior / crash

## Issue description

both the CLI action `launch-or-focus-plugin` and keybind action `LaunchOrFocusPlugin`  create a new plugin instance instead of focusing on the existing one when using plugin aliases.

**Workaround**: use file schema URL

## Minimal reproduction

1. download [zj-quit](https://github.com/cristiand391/zj-quit) and set up an alias for it:
```
plugins {
  zj-quit location="file:/path/to/zj-quit.wasm"
}
```
2. start zellij and launch it via with `launch-or-focus-plugin` via the CLI using the alias:
```
zellij action launch-or-focus-plugin --floating zj-quit
```

press `Esc` to make it hide itself:
https://github.com/cristiand391/zj-quit/blob/1ab13ff06af29215bba33ed03b116ef407acc8c1/src/main.rs#L40

repeat that 3 times and then open the session-manager and check the panes in the current tab, you should see this:
![image](https://github.com/zellij-org/zellij/assets/6853656/d05093f0-0e62-478e-8cb1-c1d34fc2c935)

note that there's 3 `zj-quit` panes instead of just 1.

now repeat step 2 but use the file schema URL:
```
zellij action launch-or-focus-plugin --floating "file:/path/to/zj-quit.wasm"
```

check the session-manager again, only 1 `zj-quit` instance is open.

## Other relevant information
I noticed this first when using the keybind action, doesn't seem to affect the `zellij:` schema.

`zellij --version`: `zellij 0.40.1`

pnx:
This is even present with the default config as it uses an alias for the session manager:

```kdl
keybinds {
    ...
    session {
        bind "Ctrl o" { SwitchToMode "Normal"; }
        bind "Ctrl s" { SwitchToMode "Scroll"; }
        bind "d" { Detach; }
        bind "w" {
            LaunchOrFocusPlugin "session-manager" {
                floating true
                move_to_focused_tab true
            };
            SwitchToMode "Normal"
        }
    }
}

plugins {
    ...
    session-manager location="zellij:session-manager"
}
```

you can just do `ctr+o w` then `<esc>` 3 times to produce the same result as the screenshot below: 

![image](https://github.com/user-attachments/assets/138ab032-3be1-4405-9fe0-2e732c527275)

imsnif:
[...]

PR #5248, opened 2026-06-10
fix(plugins): prevent spawning duplicated aliased plugins

acr4:
Addresses #3409

When using `LaunchOrFocusPlugin` with a plugin alias (e.g., "session-manager") and custom configuration (like `floating true`), Zellij would incorrectly spawn a new plugin instance instead of focusing the existing one. This caused severe resource leakage (CPU burn/zombie plugins) on rapid keypresses.

The issue occurred because Zellij dynamically injects a `caller_cwd` into an invoked alias's configuration. The equivalence check failed to properly ignore this dynamic `caller_cwd` when matching against the background plugin. Furthermore, `is_equivalent_to_run` lacked the logic to compare an `Alias` directly against a canonical `RunPlugin` instance.

This commit fixes the issue by:
1. Implementing custom `PartialEq` and `Hash` for `PluginUserConfiguration` to explicitly ignore the `caller_cwd` key during comparisons.

2. Enhancing `is_equivalent_to_run` to properly match an `Alias` with a `RunPlugin` instance (e.g. matching alias "session-manager" to location "zellij:session-manager").

### DIFF
--- a/zellij-utils/src/input/layout.rs
+++ b/zellij-utils/src/input/layout.rs
@@ -193,30 +193,29 @@ impl RunPluginOrAlias {
             (
                 RunPluginOrAlias::Alias(self_alias),
                 Some(Run::Plugin(RunPluginOrAlias::Alias(run_alias))),
-            ) => {
-                self_alias.name == run_alias.name
-                    && self_alias
-                        .configuration
-                        .as_ref()
-                        // we do the is_empty() checks because an empty configuration is the same as no
-                        // configuration (i.e. None)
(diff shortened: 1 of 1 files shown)


acr4:
Hi @imsnif,
When you have time, will you please review this PR (and associated issue #3409)? I was able to reproduce and resolve the original issue. For my particular use-case, I see slowly-increasing CPU usage as the session-manager plugin instances accumulate (I am a heavy user of multiple sessions, constantly switching between them), resulting in Zellij session bankruptcy about once a week. This patch seems to have completely resolved the issue for me.

PR #5249, opened 2026-06-11
fix(plugins): close session-manager instead of hiding it when switchi…

Kittease:
# What this fixes

When switching to another session (or resurrecting one) from the session-manager, the plugin hid itself instead of closing. The hidden pane stayed alive in the session being left, attached to the tab it was launched from. The next `LaunchOrFocusPlugin` in that session would then focus that stale tab instead of the user's current one — even with `move_to_focused_tab true`.

This is the remaining part of #3834. #5055 fixed the Esc/Ctrl-c paths by replacing `hide_self()` with `close_self()`, but the switch-session paths still called `hide_self()`, so the bug kept reproducing whenever a session switch was performed through the manager (see [this comment](https://github.com/zellij-org/zellij/issues/3834#issuecomment-4254399953) from April reporting the issue still occurs).

# Reproduction

1. Create a session with two tabs, focus tab 1
2. Open the session-manager and press Enter to switch to another session
3. Come back to the first session, focus tab 2
4. Open the session-manager again → focus jumps to tab 1, while the manager pane is moved to tab 2

# Changes

Replace `hide_self()` with `close_self()` on all switch/resurrect paths, mirroring what #5055 did for the Esc/Ctrl-c handlers:

- `main.rs`: the `AttachToSession` and `ResurrectSession` Enter handlers, and the three `SingleScreen` flows (search-and-select, typed name match, layout selection)
- `new_session_info.rs`: the `NewSession` screen's Enter handler (both the with-layout and without-layout branches, plus the trailing dismissal)

The welcome-screen `quit_zellij()` branches are untouched, and the comments guarding the rename-validation early `return`s were updated to match ("don't hide self" → "don't close self").

# Testing

- `cargo fmt --all -- --check` — clean
- `cargo xtask test` — all tests pass
- `cargo xtask ci e2e --build` + `cargo xtask ci e2e --test` — 42/42 e2e cases and 146/146 plugin system tests pass

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 * feat: allow tabs to have different sizes if clients aren't focused on the same one (https://github.com/zellij-org/zellij/pull/5133)
 * feat: PWA support for the web client (manifest + icons + iOS meta tags) so the page can be installed as a standalone app (https://github.com/zellij-org/zellij/pull/5184)
 * feat: mobile UI (https://github.com/zellij-org/zellij/pull/5241)
+* fix: close session-manager instead of hiding it when switching sessions (https://github.com/zellij-org/zellij/pull/5249)
 
 ## [0.44.3] - 2026-05-13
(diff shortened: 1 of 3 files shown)


myriadbytes:
Just want to say that I am running into this issue all the time, so I really wish this gets merged eventually !

Issue #5261, opened 2026-06-14
Misleading ERROR logs are emitted during normal pane and CLI operations


yamam:
# 2. Issues with the Zellij UI / behavior / crash

## Description

Some normal pane / CLI operations emit `ERROR` logs even though the operations do not appear to indicate actual failures.

I found three cases so far:

1. Exiting a pane shell logs `Input/output error`
2. `zellij pipe` can be treated as repeated unknown messages
3. Action completion sender drop is reported as a timeout

These logs are misleading because they make normal behavior look like an internal failure.

### Steps to reproduce

1. Open a normal terminal pane.
2. Exit the shell in that pane, for example with:

```bash
exit
````

or `Ctrl-d`.

### Actual behavior

The following ERROR log is emitted:

```text
ERROR |zellij_server::terminal_b| ... zellij-server/src/terminal_bytes.rs:58: Input/output error (os error 5)
```

### Expected behavior

Exiting a pane shell should not emit an ERROR log.

This appears to be a normal PTY closure case rather than an actual application error.

### Steps to reproduce

Run a CLI pipe command, for example:

```bash
zellij pipe ...
```

### Actual behavior

The following ERROR log is emitted:

```text
ERROR  |zellij_server::route     | ... zellij-server/src/route.rs:2697: Client sent over 1000 consecutive unknown messages, this is probably an infinite loop, logging client out
```

### Expected behavior

A CLI pipe client disconnect should not be treated as repeated unknown or malformed messages.

A disconnected client and a malformed / unknown message should be handled separately.

### Steps to reproduce

This can be reproduced with `zellij pipe`:

```bash
zellij pipe ...
````

It can also be reproduced by switching to the current session:

```bash
zellij action switch-session <current-session-name>
```

### Actual behavior

The following ERROR log is emitted:

```text
ERROR |zellij_server::route| ... zellij-server/src/route.rs:75: Action CliPipe did not complete within 1s timeout
```

However, this does not appear to be a real 1 second timeout. The action completion sender is dropped before sending a result.

### Expected behavior

A dropped action completion sender should not be reported as a timeout.

Actual timeouts should still be logged as errors.

## Notes

These cases likely have different causes and can be fixed independently.

Possible directions:

* Treat normal PTY closure / EOF-like read errors as non-error terminal termination.
* Distinguish client disconnects from malformed / unknown IPC messages.
* Distinguish actual action completion timeouts from dropped completion senders.

PR #5263, opened 2026-06-14
fix(route): distinguish client IO errors from malformed messages

yamam:
## Summary

Part of #5261.

This PR fixes a misleading route error that can happen with `zellij pipe`.

When the server fails to read a client message, the route thread previously
treated that failure the same way as a malformed or unknown message. This could
cause the route loop to keep counting consecutive unknown messages until it hit
the 1000-message guard.

This could produce an error like:

```text
Client sent over 1000 consecutive unknown messages, this is probably an infinite loop, logging client out
````

even though the client did not actually send 1000 unknown messages.

## Problem

`recv_client_msg()` previously returned `None` for multiple different cases:

* the client message was read successfully but could not be converted into
  `ClientToServerMsg`
* reading the client protobuf message failed
* the client disconnected or the client stream could no longer be read

`route.rs` treated `None` as an unknown message.

As a result, IO/read failures could be counted as malformed or unknown messages.
In the `zellij pipe` case, this could cause the route loop to keep reading and
eventually hit the 1000-message guard.

## Change

Introduce an explicit result type for receiving client messages:

* `Message(ClientToServerMsg, ErrorContext)`
* `IoError`
* `Malformed`

`IoError` is returned when reading the client protobuf message fails with an IO
error. The route loop exits for that client in this case.

`Malformed` is returned when a protobuf message cannot be converted into a valid
`ClientToServerMsg`.
Malformed messages still use the existing unknown-message counter and
1000-message guard.

## Why this is correct

An IO/read failure is different from receiving a malformed or unknown client
message.

If the client stream cannot be read, continuing to read from the same route loop
is not useful and can lead to misleading repeated unknown-message logs.

Malformed messages are still counted by the existing guard, because after a
malformed message the stream may still be usable and a later message may be
valid.

### DIFF
--- a/zellij-server/src/route.rs
+++ b/zellij-server/src/route.rs
@@ -28,7 +28,10 @@ use zellij_utils::{
         actions::{Action, SearchDirection, SearchOption},
         command::TerminalAction,
     },
-    ipc::{ClientToServerMsg, ExitReason, IpcReceiverWithContext, ResizeCause, ServerToClientMsg},
+    ipc::{
+        ClientToServerMsg, ExitReason, IpcReceiverWithContext, RecvClientMsg, ResizeCause,
+        ServerToClientMsg,
+    },
 };
 
 use crate::ClientId;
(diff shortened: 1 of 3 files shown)


yamam:
Closing this as superseded by #5492.

The equivalent route/IPC error handling was included in that PR and is now implemented on `main` with `IpcReceiveError` and `try_recv_client_msg()`.

Issue #5270, opened 2026-06-16
`There is no active session!` can be reported incorrectly under concurrent `zellij pipe`


yamam:
# 2. Issues with the Zellij UI / behavior / crash
## Description

After repeatedly running `zellij pipe` concurrently, a Zellij CLI command can incorrectly fail to detect the active session and report:

```text
There is no active session!
````

even though the Zellij session is still running.

## Steps to reproduce

Start a Zellij session, then run the following script from a shell:

```bash
#!/usr/bin/env bash

for ((i = 0; i < 10000; i++)); do
  echo | zellij pipe --name=some-pipe >/dev/null &
  echo | zellij pipe --name=some-pipe >/dev/null &
  wait
done
```

After a while, one of the `zellij pipe` commands can print:

```text
There is no active session!
```

even though the session is still active.

## Expected behavior

`zellij pipe` should continue to detect and connect to the active session correctly.

Concurrent short-lived CLI clients should not cause active session detection to incorrectly fail.

## Actual behavior

Under concurrent `zellij pipe` usage, session detection can intermittently fail with:

```text
There is no active session!
```

## Investigation

I found two related races that can lead to this symptom.

### 1. Unassociated CLI pipe responses can be broadcast to unrelated clients

This appears to be caused by a race involving CLI pipe clients and session socket probing.

The sequence I observed is:

1. A `zellij pipe` client is run with stdin that is not a terminal.
2. The pipe client reaches EOF on stdin.
3. The client sends the final empty pipe message but exits before waiting for the corresponding server response.
4. The server removes the pipe association for that client.
5. Later, when `UnblockCliPipeInput` or `CliPipeOutput` is produced for that pipe, the server no longer has an associated client.
6. The server broadcasts the pipe response to all clients.
7. A temporary client created by `assert_socket()` while probing active sessions can receive this unrelated pipe response instead of the expected `Connected` response.
8. `assert_socket()` treats the non-`Connected` message as a failed socket probe.
9. As a result, session detection can incorrectly conclude that there is no active session.

### 2. Stale client-id based cleanup can affect a newer client after `client_id` reuse

There is also a separate race involving short-lived `ConnStatus` probe clients and quick `client_id` reuse.

`assert_socket()` creates a temporary client and sends `ClientToServerMsg::ConnStatus` in order to check whether a session socket belongs to a live Zellij server.

[...]

Issue #5252, opened 2026-06-12
Server leaks ~100-400 MB/min with zero clients attached after a skewed-client flood; wasm plugin host threads pegged ~99% CPU, plugin reload does not recover (0.44.3)


phsb5321:
## Summary

After a client desync event, the zellij **server** process leaks memory at ~100–400 MB/min **with provably zero clients attached**. The server's wasm plugin host threads (`plugin-*`) stay pegged at ~99% CPU indefinitely. Reloading plugins does not recover; only killing the server does. This looks distinct from (but possibly triggered alongside) #4797 — the allocation continues with no clients connected and no render targets, and it survives plugin reloads.

## Environment

- zellij **0.44.3** (nixpkgs build), Linux 6.12.92 (NixOS), x86_64
- Long-lived detached session under a systemd user unit (server daemonized), ~31 tabs, most panes running busy TUI programs (continuous title/output churn)
- Plugins: built-in `tab-bar`/`status-bar`/`session-manager` + one custom wasm plugin (subscribes to `TabUpdate`/`ModeUpdate`/`Mouse`/`Timer` only). Disclosed for completeness — the pegged plugin threads survived reloading **both** the custom plugin and the built-ins.
- `zellij web` server running on localhost (this is how the trigger happened, see below)

## Trigger (preceding event)

The session server and the `zellij web` client ended up being **two different builds of the same 0.44.3** (NixOS generations drifted). The skewed web client produced this loop in the server log every ~20s for ~2.6 days:

```
ERROR  |zellij_server::route     | 2026-06-12 09:33:57.106 [server_router] zellij-server/src/route.rs:2645: Client sent over 1000 consecutive unknown messages, this is probably an infinite loop
```

pattern: `unknown message from client → Broken pipe → 1000 consecutive → logging client out`, leaving a ghost client on a broken pipe. Same log signature as seen in #5007 (different platform). Over those days the server grew to:

```
VmRSS: 42650440 kB   VmSwap: 2236964 kB   Threads: 68
```

(~17 GB/day with 30 busy tabs — that part is plausibly #4797's unbounded buffering for a dead/never-draining client.)

## The part that looks like a separate bug

After stopping the web server and disconnecting **every** client (verified: zero zellij client processes on the host, nothing attached):

- 6× `plugin-*` threads pegged at 98.5–99.9% CPU (`top -H`), continuously
- clean 3-minute measurement window, zero clients, zero probes: **+1238 MB**
[...]

PR #5272, opened 2026-06-17
fix(server): avoid immediate client id reuse

yamam:
## Summary

Part of #5270.

This PR avoids immediate reuse of `client_id` values for short-lived clients.

Previously, new clients were allocated by starting from the lowest available
client id. Under heavy concurrent CLI activity, this meant the same `client_id`
could be reused almost immediately after being removed.

This is fragile because some server instructions are queued asynchronously and
refer to clients only by `client_id`. If an old instruction is processed after
that `client_id` has already been reused, it can affect a newer, unrelated
client.

## Problem

The issue was observed while investigating intermittent failures where
concurrent `zellij pipe` commands could report:

```text
There is no active session!
```

even though the session was still active.

One observed stale instruction path involved `ConnStatus`, which is used by
`assert_socket()` as a short-lived session socket probe.

That path can issue cleanup for the probe client more than once: the server
handles `ServerInstruction::ConnStatus(client_id)` by sending `Connected` and
removing the temporary probe client, while the route thread can also send its
generic cleanup for the same `client_id` when the route loop exits.

This PR does not try to audit or fix every duplicate cleanup path. Instead, it
mitigates the broader issue by avoiding immediate `client_id` reuse, so stale
client-id based instructions are much less likely to affect newly created
clients.

## Change

Track the next client id to try in `SessionState`.

Instead of always starting allocation from the lowest available id, client id
allocation now continues from the previously issued id and wraps around when
needed.

This makes immediate `client_id` reuse much less likely.

## Limitations

This is a mitigation rather than a complete correctness guarantee.

Client ids can still wrap around eventually, so a stale instruction could
theoretically still affect a later client after wraparound.

A more complete long-term fix would be to use generation-aware client handles,
so stale client-id based instructions can be detected and ignored explicitly.

### DIFF
--- a/zellij-server/src/lib.rs
+++ b/zellij-server/src/lib.rs
@@ -616,6 +616,16 @@ pub(crate) struct SessionState {
     /// empty synthetic reply so `Screen`'s `forward_in_flight` slot
     /// releases and the queued forwards keep moving.
     forwards_in_flight: HashMap<u32, ClientId>,
+    next_client_id: ClientId,
+}
+
+fn next_non_zero_client_id(client_id: ClientId) -> ClientId {
+    let next_client_id = client_id.wrapping_add(1);
+    if next_client_id == 0 {
+        1
+    } else {
+        next_client_id
+    }
 }
 
 impl SessionState {
@@ -626,6 +636,7 @@ impl SessionState {
             watchers: HashMap::new(),
             last_active_client: None,
             forwards_in_flight: HashMap::new(),
+            next_client_id: 1,
         }
     }
     pub fn new_client(&mut self) -> ClientId {
@@ -636,15 +647,20 @@ impl SessionState {
             .chain(self.watchers.keys().copied())
             .collect();
 
-        let mut next_client_id = 1;
+        let start_client_id = self.next_client_id;
+        let mut next_client_id = self.next_client_id;
         loop {
             if all_ids.contains(&next_client_id) {
-                next_client_id += 1;
+                next_client_id = next_non_zero_client_id(next_client_id);
+                if next_client_id == start_client_id {
+                    panic!("no available client ids");
+                }
             } else {
                 break;
             }
         }
         self.clients.insert(next_client_id, None);
+        self.next_client_id = next_non_zero_client_id(next_client_id);
         next_client_id
     }
     pub fn associate_pipe_with_client(&mut self, pipe_id: String, client_id: ClientId) {
--- a/zellij-server/src/plugins/mod.rs
+++ b/zellij-server/src/plugins/mod.rs
@@ -88,6 +88,10 @@ pub enum PluginInstruction {
     Resize(PluginId, usize, usize), // plugin_id, columns, rows
     AddClient(ClientId),
     RemoveClient(ClientId),
+    PruneDisconnectedPluginInstances {
+        plugin_id: PluginId,
+        keep_client_id: ClientId,
+    },
     NewTab(
         Option<PathBuf>,
         Option<TerminalAction>,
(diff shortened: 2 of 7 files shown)


yamam:
I found an additional plugin lifecycle issue while testing this change and
added a follow-up commit to address it.

Avoiding immediate client ID reuse means that a reattached client normally
receives a different ID. As a result, its plugin instances are registered under
new `(plugin_id, client_id)` keys and no longer replace the instances associated
with the disconnected client. Without explicit cleanup, those old instances
remain active and accumulate across detach and reattach.

The new commit cleans up disconnected-client plugin instances after the
corresponding instance for the newly attached client has started successfully.
It preserves instances belonging to currently connected clients and cleans up
the removed instance's workers, key interception, data directory, and pending
pipe state.

I also confirmed another consequence of immediate client ID reuse: a timer
scheduled by a previous plugin instance can be delivered to its replacement
when both instances use the same `(plugin_id, client_id)` key. In testing, the
replacement instance received two independent timer sequences even though the
previous `RunningPlugin` had already been dropped.

The client ID allocation change prevents the old timer from being routed to
the replacement instance.

yamam:
While testing this PR, I found a related plugin event delivery race.

When a background plugin is loaded during a detach/attach cycle, the initial
`TabUpdate`/`PaneUpdate` events can be generated before the plugin subscription
is registered. If no state report is generated after the subscription, the new
plugin instance does not receive the initial state.

On `upstream/main`, the client ID may be reused after detach/attach. In that
case, the previous subscription entry for the same `(plugin_id, client_id)`
pair remains effective. The plugin ID itself is unchanged; the relevant
difference is the reused client ID. This causes events generated before the
new plugin instance loads to be cached and delivered, masking the race.

Therefore, avoiding immediate client ID reuse in this PR may expose this
latent plugin event race. This PR should also ensure that the current
`TabUpdate`/`PaneUpdate` state is re-delivered after a background plugin
subscription is registered, so that the new plugin instance does not depend on
a subsequent state change.

Avicennasis:
I reproduced one cleanup gap that remains at the current head of this PR (`7648fa44`). It matches a live multi-client failure in which a status plugin kept changing session-wide pane frames after the client that created one instance had disconnected.

The reproducer is a small plugin that starts a 100 ms `set_timeout` chain. Each instance appends its client ID and tick count to its private `/data/timers.log`. The harness starts two real PTY clients, detaches the second through the normal `DetachSession` IPC message while leaving the first connected, then samples both logs for one second.

On this PR's current head:

```text
connected client 1:  +10 ticks
detached client 4:   +10 ticks
client-4 data dir:   still present
```

This happens because `PruneDisconnectedPluginInstances` is sent only after a new client-specific instance finishes loading. In the two-client case, an instance for the remaining client already exists and no later load occurs, so the disconnected instance continues running until some future attach.

I tested a small amendment on top of this PR. During `RemoveClient`, it removes a disconnected instance only if the same plugin ID already has an instance belonging to another currently connected client. It retains the last instance when no replacement exists, preserving this PR's template for a future attach. The later post-load prune remains in place for that final-template case.

With the amendment:

```text
connected client 1:  +10 ticks
detached client 4:     0 ticks
client-4 data dir:   removed
```

The cleanup also removes the instance's background-subscription entry and cached worker messages, along with the workers, key interception, data directory, and pending pipe state already handled by this PR.

Validation:

- real PTY A/B reproduction: 10 stale ticks before, 0 after
- new real-WASM two/three-client lifecycle regression: pass
- `cargo fmt --all -- --check`: pass
- `zellij-server` library suite: 1,407 passed, 0 failed, 151 ignored

The local amendment commit is `157eac82`, based directly on `7648fa44`. I have a one-commit format patch and can publish a branch or open a focused PR if that is useful.

This also gives a deterministic reproduction for #5414. It may explain part of #5252 as well, although this amendment intentionally retains the final instance until a replacement loads, so it does not eliminate all possible plugin activity while a session has zero clients.

yamam:
Closing this PR as its fixes are now covered by upstream changes in #5630.

- Client IDs are allocated monotonically, preventing immediate reuse after a client disconnects.
- Client-scoped plugin instances are cleaned up on detach, and fresh instances are loaded on attach.
- Newly registered background plugin subscriptions receive targeted initial pane/tab state updates.

Issue #5048, opened 2026-04-15
AltGr/Alt and special characters in non-US keyboard layout


wwwald:
# 2. Issues with the Zellij UI / behavior / crash

## Current setup: 

* Zellij 0.44.1 on Ubuntu 24.04, running in WSL.
* using a _Belgian_ AZERTY keyboard layout with an AltGr key for some special characters.  
Examples are `|` (AltGr + &) and `\` (AltGr + <).

## Issue description

In plain zsh or bash, AltGr-based special characters work fine.
In Zellij, they don't. Examples:
* `|` (AltGr + &) or `]` (AltGr + $) does not appear 
* `\` (AltGr + <) triggers the typing of a `chsh` command
* `#` (Alt+") triggers the typing of `''` 
* and more

Narrowing down, the problem seems to be that AltGr is interpreted as Alt, and the resulting key combo triggers one of the built-in Zellij keybindings.

I confirmed using `cat -v` that the keyboard signals differ between regular bash and Zellij: in bash, I get the regular `|` while Zellij shows me `^[&`.

I could unbind all the `Alt` keybindings - but I'd like to keep these shortcuts, and properly distinguishing between AltGr and Alt would be an improvement. 

## Other relevant information

I'm running WSL/Ubuntu in Windows Terminal, with the Kitty Keyboard protocol enabled. That enables distinguishing between Alt and AltGr, as the zsh/bash result shows.

Zykino:
Zellij 0.44.1, WSL, Debian 12, Windows Terminal, AZERTY (FRA FR)

Fish:
<img width="122" height="52" alt="Image" src="https://github.com/user-attachments/assets/841be04a-6b78-437e-8b66-4fd76991a346" />

(Same for bash, no issue.)

Only I don't see an issue :s

wwwald:
@Zykino What version of Windows Terminal? There may be some changes in the Kitty keyboard protocol between versions...

Zykino:
WT: 1.23.20211.0

`zellij` only support the first revision of the Kitty Keyboard Protocol and software treating support as either "all" or "none" are known to break things. Usually they are inside `zellij` not the terminal itself (in the report I saw).

wwwald:
OK - I'm on 1.25.622.0. It seems that this is the first WT version that natively supports the Kitty protocol, according to [the release notes](https://github.com/microsoft/terminal/releases/tag/v1.25.622.0). So wouldn't be unlogical to encounter some issues.

[...]

PR #5281, opened 2026-06-18
fix(kkp): handle 3-parameter Kitty CSI-u sequences

divens:
Terminals supporting REPORT_ASSOCIATED_TEXT (Kitty flag 16) send the OS-produced character alongside the keycode in a third CSI-u parameter: `\e[<keycode>;<modifier>;<text>u`. The previous parser only knew the 2-parameter form, so the second `;` and the text codepoints were pushed into modifier_bytes. KeyModifier::from_bytes then failed to parse "3;124" as an integer, returning an empty modifier set and emitting the base-layer keycode (e.g. `-` for AltGr+6 instead of `|`).

Three changes:

* Bump ENTER_KITTY_KEYBOARD_MODE from `\e[>1u` to `\e[>17u` so terminals that support REPORT_ASSOCIATED_TEXT will include the produced character. Terminals that don't support the bit silently ignore it per the Kitty spec.

* Extend KittyKeyboardParser with a ParsingAssociatedText state and a third byte buffer. The second `;` transitions into the new state; both `u` and `~` terminate from it.

* KeyWithModifier::from_bytes_with_u / _tilde now take the associated text bytes and, when the produced character differs from the keycode (AltGr glyphs, Shift+letter→uppercase, dead-key composition, IME), return that character with no modifiers. When text matches the keycode (real Alt+f reporting `f` as text), modifiers are preserved so Alt-bindings still fire.

Together these fix the AltGr regression reported for Windows Terminal 1.25 in #5048 (verified on WSL + WT 1.25): AltGr+6 now produces `|`, AltGr+8 produces `\`, etc. The parser change is also load-bearing for the flag bump itself — terminals that honor REPORT_ASSOCIATED_TEXT switch all modifier+key reports to the 3-parameter form, so without the parser update, Alt+f / Ctrl+f bindings would silently break on those terminals (WezTerm, Alacritty 0.13+) instead of continuing to work.

Known limitation: on Windows native PowerShell, ConPTY strips the third parameter before bytes reach zellij, so the fix only takes effect in environments where the 3-parameter form survives the input pipeline (Linux/macOS native, WSL bash).

### DIFF
--- a/zellij-client/src/keyboard_parser.rs
+++ b/zellij-client/src/keyboard_parser.rs
@@ -7,6 +7,11 @@ enum KittyKeysParsingState {
     ReceivedEscapeCharacter,
     ParsingNumber,
     ParsingModifiers,
+    /// Third CSI-u parameter (after the second `;`), present when the
+    /// terminal has `REPORT_ASSOCIATED_TEXT` enabled. Holds the produced
+    /// character codepoint(s) as decimal digits, optionally separated by
+    /// `:` for multi-codepoint text.
+    ParsingAssociatedText,
     DoneParsingWithU,
     DoneParsingWithTilde,
 }
@@ -32,6 +37,7 @@ pub struct KittyKeyboardParser {
     state: KittyKeysParsingState,
     number_bytes: Vec<u8>,
     modifier_bytes: Vec<u8>,
+    associated_text_bytes: Vec<u8>,
 }
 
 /// CSI final-byte range (0x40..=0x7E), minus `u` and `~` which trigger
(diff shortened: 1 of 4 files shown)


imsnif:
Hey, I'm sorry to intervene, but let's please not support any more progressive enhancements of the kitty keyboard protocol. It has caused nothing but trouble and I don't want to weigh more deeply into it. Specifically, the protocol doesn't include a mandatory negotiation phase. Meaning that client/server implementing it are essentially supposed to hope they each play nicely with each other.

The most I'm willing to consider is to do a blanket implementation of the whole thing, but that's a pretty big undertaking.

PR #5283, opened 2026-06-19
feat(config): add mouse_scroll_resize option

### LINKED ISSUE #5086: Allow disabling Ctrl+ScrollWheel pane resize independently of other `advanced_mouse_actions` behaviors
## Summary

`advanced_mouse_actions` is currently an all-or-nothing toggle. I'd like to disable
just the Ctrl+ScrollWheel pane resize while keeping the other behaviors it
controls (drag-to-resize borders, hover effects, Alt+click pane grouping).

## Motivation

When using a trackpad, accidentally triggering a pane resize while scrolling
through terminal output (or scrolling in a TUI) is disruptive — the focused
pane jumps in size and the cursor/content shifts, breaking the flow of reading.
The other features bundled into `advanced_mouse_actions` are useful and I'd
like to keep them.

There doesn't seem to be a workaround:
- `mouse_mode false` is too aggressive (loses click-to-focus, mouse selection,
  scroll-history, clickable tabs).
- `advanced_mouse_actions false` removes the wanted features (drag-resize,
  hover, pane grouping) along with the unwanted scroll-resize.
- Scroll events are not bindable as keybinds, so they can't be intercepted /
  no-op'd via configuration.

## Current behavior

Per https://zellij.dev/documentation/options, `advanced_mouse_actions`
"Enables mouse hover effects, multiple select functionality (pane grouping),
and mouse-based pane resizing" — where "mouse-based pane resizing" covers both
border-drag and Ctrl+ScrollWheel.

## Proposed behavior

Either of the following would work:

**Option A — sub-option (table/block form):**

```kdl
advanced_mouse_actions {
    scroll_resize false
}
```

**Option B — separate top-level option (default `true` for back-compat):**

```kdl
mouse_scroll_resize false
```

I'd lean toward Option B for ergonomics, but I have no strong preference.

## Environment

- Zellij version: `0.44.1`
- Terminal: kitty (macOS)
- OS: macOS (Darwin 25.3.0)

WeZZard:
Adds a new `mouse_scroll_resize` config option that independently disables Ctrl+ScrollWheel pane resizing while preserving other mouse actions.

Fixes #5086.

Behavior:
- When `mouse_scroll_resize true` (default), Ctrl+ScrollWheel resizes panes as before.
- When `mouse_scroll_resize false`, Ctrl+ScrollWheel is ignored for resizing and the pane-frame help text changes from `Ctrl <MouseScroll> or <drag borders> to resize` to `<drag borders> to resize`.

Verification:
- `cargo xtask test` passes.
- `cargo fmt --check` passes.
- `cargo xtask clippy` still reports pre-existing errors in untouched files (`zellij-server/src/output/mod.rs` and `zellij-server/src/unit/screen_tests.rs`).

### DIFF
--- a/zellij-server/src/tab/mouse_handler.rs
+++ b/zellij-server/src/tab/mouse_handler.rs
@@ -230,6 +230,7 @@ struct MouseEventContext {
     pinned_unselectable: Option<PaneId>,
     focus_follows_mouse: bool,
     mouse_click_through: bool,
+    mouse_scroll_resize: bool,
 }
 
 fn edge_and_delta_to_strategies(
@@ -405,6 +406,7 @@ impl MouseHandler {
             pinned_unselectable,
             focus_follows_mouse: tab.focus_follows_mouse,
             mouse_click_through: tab.mouse_click_through,
+            mouse_scroll_resize: tab.mouse_scroll_resize,
         })
     }
 
@@ -1282,6 +1284,9 @@ impl MouseHandler {
         }
 
         if event.wheel_up || event.wheel_down {
+            if event.ctrl && !ctx.mouse_scroll_resize {
+                return Ok(MouseAction::NoAction);
+            }
             if let Some(pane_id) = ctx.pane_id_at_position {
                 if event.ctrl {
                     if event.wheel_up {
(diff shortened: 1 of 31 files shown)


zellij-server/src/tab/mouse_handler.rs
```diff
@@ -1228,7 +1230,7 @@ impl MouseHandler {
 
         if event.wheel_up || event.wheel_down {
             if let Some(pane_id) = ctx.pane_id_at_position {
-                if event.ctrl {
+                if event.ctrl && ctx.mouse_scroll_resize {
```
imsnif: I'd rather not have this fall down the rest of the if ladder. Right now it's benign, but we might miss stuff if we add more functionality in the future.

CHANGELOG.md
```diff
@@ -5,6 +5,7 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 
 ## [Unreleased]
+* feat: add `mouse_scroll_resize` option to disable Ctrl+ScrollWheel pane resizing
```
imsnif: Please remove this, I'd rather manage the changelog on my own.

imsnif (changes requested):
Hi, thanks for this. I added two minor notes. Otherwise if you'd like to rebase from main there might be some more changes.

zellij-server/src/tab/mouse_handler.rs
WeZZard: Fixed in 4f8bf9b. When `mouse_scroll_resize` is disabled, Ctrl+wheel now returns `MouseAction::NoAction` before the regular scrolling branches. I also added a direct regression test to ensure it cannot fall through.

CHANGELOG.md
WeZZard: Removed the changelog entry in 4f8bf9b.

imsnif:
Thank you!

Nautman:
Thank you!! So nice to see this fixed 🎉

PR #5299, opened 2026-06-23
fix(scroll): keep scroll position when leaving scroll mode

valentynkit:
Leaving Scroll mode (to Normal or Locked) snapped the active pane back to the bottom. Switching panes means leaving Scroll mode, so that reset is what made it impossible to keep one pane scrolled up to read from while working in another.

Follows the discussion in https://github.com/zellij-org/zellij/issues/638.

## Fix

Remove the implicit scroll reset in `Screen::change_mode`. Scroll position is already per-pane and survives focus changes, so with the reset gone the position is simply kept. It was also redundant: the scroll keymap already binds an explicit `ScrollToBottom` on exit (`Ctrl c`), so following live output stays one keypress away, while a plain exit (`Ctrl s`) now preserves the position.

This does not reintroduce the scroll-buffer overwrite that buffering guards against: output is buffered while a pane `is_scrolled`, independent of input mode, and every path that clears scroll already replays the buffered events, so a pane catches up to live output the moment it is scrolled back to the bottom.

Added a `screen_tests` snapshot test asserting the pane stays scrolled after a Scroll to Normal transition.

## Open question

#638 literally asks to be "placed in scroll mode again" when returning to the pane. This implements the underlying goal (the position is no longer lost) without re-entering scroll mode, which fits the current per-client mode model better than making mode per-pane. If you would rather keep the old default and put this behind a config option, that is a small change and I can do that instead.


imsnif:
Thanks, but I think it's a better UX to have the mode returned to scroll if focusing on a scrolled pane. Want to give it a shot?

valentynkit:
> Thanks, but I think it's a better UX to have the mode returned to scroll if focusing on a scrolled pane. Want to give it a shot?

Thanks for feedback, sound interesting and this approach could provide better UX, i will try give it a shot.

valentynkit:
Done. Focusing a scrolled pane now drops you into Scroll mode, and focusing an unscrolled one takes you back to Normal. Only the Normal/Scroll pair is synced, so Pane, Tab, Search and Locked are left alone.

Two calls I'd like your take on:

- I made it symmetric (leaving a scrolled pane returns to Normal). Could keep Scroll sticky instead, closer to the #696 spec. Happy either way.
- Wired into keyboard focus moves only for now. Want mouse-click and tab switches covered too?

Also as it is touching UX, please check if it is how you expect it to behave, or any adjustments.

imsnif:
Hey, thanks for taking a gander at this. I haven't yet checked this out, but from your comments a few directions to continue:
1. I like the symmetric direction, good call. Let's make it more precise: this should affect switching between scroll mode and the default mode which we keep on screen under the default_mode_info struct. This is so that both the default and unlock-first keybindings will work as intended. Makes sense?
2. This definitely needs to affect the mouse as well. Anything mouse-related that changes focus.

valentynkit:
> Hey, thanks for taking a gander at this. I haven't yet checked this out, but from your comments a few directions to continue:
> 
>     1. I like the symmetric direction, good call. Let's make it more precise: this should affect switching between scroll mode and the default mode which we keep on screen under the default_mode_info struct. This is so that both the default and unlock-first keybindings will work as intended. Makes sense?
> 
>     2. This definitely needs to affect the mouse as well. Anything mouse-related that changes focus.

Done, both points.

1. It now syncs between Scroll and the default mode read from `default_mode_info`, not a hardcoded Normal, so unlock-first sessions return to Locked instead of Normal. Only the default/Scroll pair is touched, so Pane, Tab, Search and Locked (when it isn't the default) are left alone.

2. Mouse focus changes are covered. I hooked the spot in `handle_mouse_event` where the active pane changes, so clicks, click-through and focus-follows-mouse all sync the same way a keyboard focus move does.

Added a Locked-default test and a mouse-click test next to the keyboard one.

imsnif:
Hey @valentynkit - looks great!

I added some behaviors (eg. make sure to also do this with showing/hiding floating panes, closing panes, switching tabs, etc.) as well as automatically go into scroll mode when scrolling with the mouse on the focused pane.

I'll merge once CI passes. Thanks for this, it has been annoying me personally for a while and I'm sure I'm not the only one. :)

PR #5302, opened 2026-06-24
fix(config): allow binding theme actions in config.kdl

### LINKED ISSUE #5297: Can't create a key binding for `ToggleTheme`
## Issue description
I'd like to create a key binding for the `ToggleTheme` action
## Minimal reproduction
Add this line to your `config.kdl`
``` kdl
bind "Ctrl t" { ToggleTheme; }
```

## Other relevant information
This is the error message I get when I start a new session:
```
    ╭─[/home/swimmer/.config/zellij/config.kdl:31:1]
 31 │         bind "Alt 9" { GoToTab 9; SwitchToMode "Normal"; }
 32 │           bind "Ctrl t" { ToggleTheme; }
    ·                           ─────┬─────
    ·                                ╰── Unsupported action: ToggleTheme
 33 │     }
    ╰────
```

Running the action in the command line with `zellij action toggle-theme` just works.

truffle-dev:
## Problem

`ToggleTheme`, `SetDarkTheme` and `SetLightTheme` work from the CLI (e.g. `zellij action toggle-theme`) but cannot be bound in `config.kdl`. Binding one fails at startup:

```kdl
bind "Ctrl t" { ToggleTheme; }
```

```
   ╰── Unsupported action: ToggleTheme
```

Fixes #5297.

## Cause

The inner `parse_kdl_action_arguments!` macro already handles these three actions, but the main `match action_name` arm in `Action::try_from` (the keybinding path) had no entry that dispatches to the macro for them, so they fell through to the `_ => "Unsupported action"` arm. The `to_kdl` serialization side was likewise missing them.

## Fix

- Add the missing parse arms for `SetDarkTheme` / `SetLightTheme` / `ToggleTheme`.
- Add the matching `to_kdl` arms so the bindings round-trip.

## Test

Added `can_bind_theme_actions`, which binds all three actions, asserts they parse to the expected `Action`s, and checks a serialize → deserialize round-trip. It fails on `main` with the exact `Unsupported action: ToggleTheme` error and passes with the fix.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 * fix(windows): backslashes dropped when quoting arguments passed to spawned processes (https://github.com/zellij-org/zellij/pull/5336)
 * fix: occasional high resource utilization and sluggishness on machines with tens of thousands of concurent processes (https://github.com/zellij-org/zellij/pull/5324)
 * feat: allow opening panes/tabs from the CLI without changing focus (https://github.com/zellij-org/zellij/pull/5346)
+* feat: allow binding `SetDarkTheme`, `SetLightTheme` and `ToggleTheme` as direct keybinding actions (https://github.com/zellij-org/zellij/pull/5302)
 
 ## [0.44.3] - 2026-05-13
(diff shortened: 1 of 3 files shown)


imsnif:
Hey, thanks for this! I added some round trip tests for the serialization to match other actions. Otherwise it's solid.

Issue #4069, opened 2025-03-18
Unselected frame border color


oredaze:
I was so excited when I saw zellij 42. Finally I can change that awful white unselected frame border color. Nope. I cannot. Or at least [this](https://zellij.dev/documentation/themes.html) says so.

It is literally the only reason I still use tmux... I know it sounds silly, but to me it is a large enough friction point.

Why? Because it is unintuitive for me to parse in a fraction of a second what is the selected frame. Is it the bright white or the more tame colored one? I have to think every time. The white draws too much attention. This stuff needs to be subconscious. I need to change the unfocused frame color to dark gray and since that is my text color I can't do it.

Please consider adding that very tiny change (`frame_unselected` in the theme spec).
And thank you for making such an amazing tool.

GrzegorzKozub:
I reported a similar concern in #4074

alexgorbatchev:
https://zellij.dev/documentation/themes.html#frame_unselected

need to restart for the change to pickup, unlike some other theme options

Issue #1175, opened 2022-03-05
Change pane_frames colour?


Dhamma-valence:
This is quite low priority, just an aesthetic change. Would it be possible to change the `pane_frames` variable colours, ideally using the existing theme colours? It would be lovely to change the active frame, inactive, and the whatever the orange one signifies in this way.

a-kenji:
Thanks for reaching out! You can change the `pane_frame` colors indirectly through the theme at the moment. But not directly yet.

Igonato:
I've just set the theme to the:
```
        fg 0 0 0
        bg 0 0 0
        black 0 0 0
        red 0 0 0
        green 0 0 0
        yellow 0 0 0
        blue 0 0 0
        magenta 0 0 0
        cyan 0 0 0
        white 0 0 0
        orange 0 0 0
```
Inactive pane frames are still blindingly white (I understand it has to do with my terminal's color scheme, but in other apps I quite like it). Overriding colors for individual UI elements would be a great feature!

PR #5303, opened 2026-06-24
feat(action): add `--frame` to set-pane-color for runtime frame/border color

dchersey:
## What

Adds a `--frame` option to `zellij action set-pane-color`, letting a running process set its pane's frame/border color at runtime:

```sh
zellij action set-pane-color --pane-id $ZELLIJ_PANE_ID --frame "#8a2be2"
zellij action set-pane-color --reset    # clears fg / bg / frame
```

`--frame` accepts the same color formats as `--fg`/`--bg` (parsed via `xparse_color`: hex, `rgb:r/g/b`, X11 names). `--pane-id` defaults to `$ZELLIJ_PANE_ID`, so a program can color its own pane.

## Why

Zellij already overrides a pane's frame color internally — red on a failed command, and the highlight color via the plugin `highlight_and_unhighlight_panes` API — but both use fixed theme colors and neither is reachable from the CLI with an arbitrary color. Users have asked for runtime/finer frame coloring (#1175, #4069, #3373). This exposes it for scripting: flag panes by CI status, a REPL/editor mode, an agent's activity, etc. It mirrors the per-pane fg/bg support added in #4737.

## How

- Adds a `frame` field to the `SetPaneColor` action, its CLI option, and the `SetPaneColorAction` proto message (regenerated via `cargo xtask build`). The plugin `set_pane_color` path passes `frame: None`, so plugin behavior is unchanged.
- Server-side, `Tab::set_pane_color` parses the frame color with the same `xparse_color` used for fg/bg and stores it via a new `Pane::set_frame_color_override` that takes an arbitrary `PaletteColor` (the existing red/highlight setters use fixed theme colors). The user-set color is kept in its **own** `frame_color_user_override` field, separate from the transient `pane_frame_color_override` used by the red-on-error and highlight flashes (see *Persistence across bell/highlight flashes* below); `frame_color_override()` returns the transient override when present and otherwise falls back to the user color. `--reset` clears the user override. The renderer is unchanged.
- `frame` semantics: omitted = leave the override untouched (so plugin and `--fg`-only calls don't disturb it); a parseable color = set; the empty string (emitted by `--reset`) = clear.

## Persistence across bell/highlight flashes

A user-set `--frame` color and zellij's *transient* frame indicators — the red exit-code flash, the `highlight_and_unhighlight_panes` highlight, and the visual-bell flash — originally shared a single `pane_frame_color_override` slot. So a terminal bell or highlight would overwrite a user-set color, and — worse — clearing that transient (`clear_bell_notification_for_pane` / `ClearPaneFrameColorOverride`) wiped the slot entirely, dropping the pane back to its focus/unselected color (e.g. the theme's green focused-pane frame) instead of the color the user set.

This PR keeps the two separate: the persistent user color lives in `frame_color_user_override`, the transient indicators keep using `pane_frame_color_override`, and `frame_color_override()` prefers the transient while it's active and falls back to the user color when it clears. Net effect: a bell still flashes the highlight color, and afterward the frame returns to the user's `--frame` color rather than vanishing. Panes with no user-set color are unaffected — behavior is identical to before.

## Testing

- `cargo check` clean; unit tests for the CLI→action conversion (`--frame` passthrough and `--reset` clearing).
- Exercised end-to-end against a running session — the target pane's frame recolors and resets as expected, and a `--frame` color now **persists across terminal-bell flashes** (flash the highlight → restore to the set color) instead of reverting to the focus color.

## Notes

- Plugin-API symmetry (a `frame` arg on the plugin `set_pane_color` command) is a straightforward follow-up; left out here to avoid a breaking plugin-API change.


imsnif:
Looks great! Happy to merge, just two requests:
1. Can you please remove the LLM authored commits? As a policy we do not allow commercial companies free ad-space in our commit log unless they sponsor the project.
2. Can you please remove the changelog entry? I'd rather manage these things on my own (eg. ordering, wording and such)

Issue #3090, opened 2024-01-26
Get Tab name (via env or cli query)


tennox:
I'd like to know the current tab title - I found these:
```
❯ export | rg -i zellij
ZELLIJ 0
ZELLIJ_PANE_ID 10
ZELLIJ_SESSION_NAME unflappable-ukulele #lol
❯ zellij action query-tab-names
Tab #1
Test Tab Name
```
But no way to know which tab I'm in, so I don't see a way currently.

### Ideas
```
zellij action query-tab-name
```
```
ZELLIJ_TAB_ID 1
ZELLIJ_TAB_NAME Test Tab Name
```

Can PR, if I get a go & guidance which way to go :tipping_hand_man:

Jedsek:
bash -c 'zellij action query-tab-names | sed -n "$(($ZELLIJ_PANE_ID + 1))p"'

bschucha:
Would love to have this feature.

tzhouhc:
For anyone looking for a workaround for this, I have a dirty little script that sort of does this by changing then reverting the name of the current tab and watching for the changed row. It seems to work?

```bash
#!/usr/bin/env bash
set -e

original_tabs=$(zellij action query-tab-names)
zellij action rename-tab TEMP
# seems like nerdfont icons don't necessarily work for tab names

new_tabs=$(zellij action query-tab-names)
readarray -t original_array <<< "$original_tabs"
readarray -t new_array <<< "$new_tabs"

temp_index=-1
for i in "${!new_array[@]}"; do
    if [[ "${new_array[i]}" == "TEMP" ]]; then
        temp_index=$i
        break
    fi
done

if [[ $temp_index -eq -1 ]]; then
    exit 1
fi

original_name="${original_array[temp_index]}"

zellij action rename-tab "$original_name"

echo $original_name
```

Limitations:
- your tab status bar -- if any -- will blink once as it gets changed then reverted.
- might not work if you sometimes explicitly name your tabs `TEMP`.
- makes assumptions like "zellij keep tab ordering / outputs them deterministically" and "no race tab changes"

I use this in combination with `neovim-remote` to have a per-tab local neovim server and send all files opened within the tab to the single neovim instance. You can see why being able to identify the tab programmatically would be important for this, which worked nicely in tmux but isn't quite easily doable in zellij as of writing this comment.

psomani16k:
#5307 might do what you need @tennox

PR #5307, opened 2026-06-25
feat(cli): add --name and --id flags to zellij action current-tab-info

psomani16k:
Adds a `--name` and `--id` flag to `zellij action current-tab-info`

- Get current tab name:
```fish
❯ zellij action current-tab-info --name
test-tab
```
- Get current tab id
```fish
❯ zellij action current-tab-info --id
1
```

Addresses the needs of #3090


imsnif:
Hey, thanks for this. However all of this info is already there in the base command. It's also possible to get its output as json with `--json` and then pipe it to `jq` to get the exact string you like for both cases.

psomani16k:
Hey @imsnif, I would be using this to display the tab (and session) name in my prompt, I am using the jq workaround right now. I was quite surprised to find that here is not command to directly get the tab name so added this (seemed like the natural place to add this).

Also didn't want to invoke anything that wasn't absolutely required in my prompt!

Edit: Also there should be a straight forward way to get the name of the current tab's name, I'd imagine this would be used alot!

PR #5323, opened 2026-07-01
fix: input spam when moving mouse while typing by splitting the input bytes into their respective events

### LINKED ISSUE #4894: Typing while moving mouse or scrolling causes input spam
## Issue description
Typing something while moving the mouse or scrolling causes input spam similar to https://github.com/zellij-org/zellij/issues/4553
That issue is fixed in v0.44.0 but typing while there is mouse input still causes similar behaviour.

## Minimal reproduction
I keep typing "-" and move the mouse or scroll while doing that.
The terminal outputs: `-----------35;62;16m35;61;16m35;60;16m35;60;15m-----35;38;16m35;39;16m------`

## Other relevant information

[zellij-0.log](https://github.com/user-attachments/files/26181603/zellij-0.log)
[zellij.log](https://github.com/user-attachments/files/26181604/zellij.log)

`zellij --version`:
zellij 0.44.0

`stty size`:
26 118

`uname -av` or `ver`(Windows):
Linux server1 5.14.0-570.17.1.el9_6.x86_64 #1 SMP PREEMPT_DYNAMIC Tue May 13 06:41:18 EDT 2025 x86_64 x86_64 x86_64 GNU/Linux

lhns:
Fixes #4894. Typing (or scrolling) while moving the mouse leaks raw mouse-motion reports into the focused pane as garbage like `a35;84;9M35;83;9Ma…`.

**Notice:** This PR is created with AI assistance (claude code). After fighting with this bug for several months now, I threw claude at the problem. It identified and fixed a few rare input corruption bugs, but only in the fourth attempt, it actually got the one that was causing this issue. I reviewed the code and it looks very reasonable to me.

## Root cause

When a single stdin read decodes into multiple input events, the client gave *all* of the read's bytes to the first event. A read like `a\x1b[<35;52;16M\x1b[<35;49;16M` parses correctly into `[Key('a'), Mouse, Mouse]`, but `Key('a')` was handed the whole buffer, including the two raw SGR mouse reports. Since a keystroke's raw bytes are forwarded to the focused pane, the mouse sequences leaked in. The mouse always parsed fine — this was purely a raw-byte attribution bug.

## Fix

Surface the bytes consumed per event from the parser (`parse_with_consumed` on the vendored termwiz `InputParser`) and drain exactly that many per event in the stdin loop, so each event keeps only its own bytes. `parse` is kept as a thin wrapper, so all other callers are unchanged.

Scope: 2 files (vendored `termwiz/input.rs` + `stdin_handler.rs`).

## Tests

Added regression tests: a keystroke before and after mouse reports in one read, consecutive keystrokes before a report, and a single multi-byte key keeping all its bytes.

## Further investigation

During the investigation I identified a few other, rarer input-corruption bugs. I wanted to keep this PR small and focused only on a solution for the issue above, so these aren't addressed here, I'm listing them for completeness. Also while I was creating the PR, I saw, that you already fixed one error scenario.

| Scenario | When it can occur | What causes the corruption | Status |
|---|---|---|---|
| Keystroke + whole mouse report(s) in one read | Any terminal, whenever you type or scroll while the mouse moves and the report(s) arrive intact | All of the read's bytes are attributed to the first event, so a keystroke carries the trailing `\x1b[<…M` reports and forwards them to the pane | **Fixed in this PR** |
| Report split at the `\x1b[<` boundary (or any partial CSI/OSC) | Fragmentation over laggy / multi-hop links (e.g. SSH) | An incomplete `\x1b[<…` prefix could be committed on the idle finalize, dispatching `\x1b[` as `Alt+[` and leaking the tail | **Handled by #5320** (verified: the partial is held up to 1000 ms until it completes, with no added input latency since a partial CSI is never a keypress) |
| Report split exactly at the lone `\x1b` | Same links, when a read boundary lands right on the report's leading `\x1b` and a >50 ms gap follows | The bare `\x1b` is ambiguous (the Esc key vs a sequence introducer), so it is flushed after 50 ms to keep the Esc key responsive; it commits as `Esc` and the continuation `[<…M` leaks as literal chars | **Still present on `main`** (verified). Follow-up fix: an opt-in `escape_sequence_timeout` option that holds the lone Esc for a configured grace, delaying the Esc key by the same amount (https://github.com/lhns/zellij/commit/65e1dd662f807d98eb3343c7ee070c65a2d4f36d). The proper no-delay fix is the Kitty keyboard protocol, where Esc arrives as `\x1b[27u` and the ambiguity disappears. |
| Pure mouse motion, no typing | - | No leak: a lone mouse event's raw bytes are discarded server-side | No fix needed |

### DIFF
--- a/src/tests/e2e/cases.rs
+++ b/src/tests/e2e/cases.rs
@@ -23,7 +23,7 @@ use super::remote_runner::{RemoteRunner, RemoteTerminal, Step};
 
 pub const QUIT: [u8; 1] = [17]; // ctrl-q
 pub const ESC: [u8; 1] = [27];
-pub const ENTER: [u8; 2] = [10, 13]; // '\n\r'
+pub const ENTER: [u8; 1] = [13]; // '\r'
 pub const SPACE: [u8; 1] = [32];
 pub const LOCK_MODE: [u8; 1] = [7]; // ctrl-g
 
--- a/src/tests/e2e/snapshots/zellij__tests__e2e__cases__send_blocking_command_through_the_cli.snap
+++ b/src/tests/e2e/snapshots/zellij__tests__e2e__cases__send_blocking_command_through_the_cli.snap
@@ -4,10 +4,8 @@ expression: last_snapshot
 ---
  Zellij (e2e-test)  Tab #1  + 
 $ /usr/src/zellij/zellij run --blocking --floating --close-on-exit -- bash -c 'sleep 2 && exit 42'                                                    
-$                                                                                                                                                     
 $ echo $?                                                                                                                                             
 42                                                                                                                                                    
-$                                                                                                                                                     
 $ █                                                                                                                                                   
                                                                                                                                                       
                                                                                                                                                       
@@ -25,4 +23,6 @@ $ █
                                                                                                                                                       
                                                                                                                                                       
                                                                                                                                                       
- Ctrl + <g> LOCK  <p> PANE  <t> TAB  <n> RESIZE  <h> MOVE  <s> SEARCH  <o> SESSION  <q> QUIT  Alt +  <n> New Pane  <f> Floating 
+                                                                                                                                                      
+                                                                                                                                                      
+ Ctrl + <g> LOCK  <p> PANE  <t> TAB  <n> RESIZE  <h> MOVE  <s> SEARCH  <o> SESSION  <q> QUIT  Alt +  <n> New Pane  <f> Floating  
--- a/zellij-client/src/stdin_handler.rs
+++ b/zellij-client/src/stdin_handler.rs
@@ -177,11 +177,11 @@ pub(crate) fn stdin_loop(
                         // Ambiguous events (if any) will be finalized later only if 50ms
                         // passes with no new input
                         let maybe_more = true;
-                        let mut events = vec![];
-                        input_parser.parse(
+                        let mut events: Vec<(InputEvent, usize)> = vec![];
+                        input_parser.parse_with_consumed(
                             &residue,
-                            |input_event: InputEvent| {
-                                events.push(input_event);
+                            |input_event: InputEvent, consumed: usize| {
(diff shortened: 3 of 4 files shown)


imsnif:
Hey @lhns - thanks for this. This looks generally good, a few things:
1. Could you please remove the LLM generated comments? They're written in the context of your session and don't make a lot of sense to a person reading the code without knowing which problem you were trying to solve or what the state was before hand.
2. In that same direction, could you please remove the LLM generated commits? I don't want to give a commercial company free ad space in our commit log unless they sponsor the project.
3. Could you please add an explicit test to what this PR fixes? Specifically: emit one lone esc in one batch, then a mouse report in a second batch, then make sure you get both events with their consumed bytes.
4. Please rebase from main, I suspect that will fix the e2e failure

lhns:
@imsnif, thank you for the review! I made the requested changes and hope that the comments are a bit more understandable now.

imsnif:
Thanks. Seems like CI is consistently failing on this one e2e test. It's probably a timing issue introduced with this fragmentation guard. Would you like to take a look?

lhns:
So the tests actually caught a real change caused by this PR. Before, it seems like multiple newline characters like `\n\r` as used in the e2e cases were emitted as a single event. Now that these events are split "properly" they are emitted as multiple events. If we accept this new behaviour we could fix the tests by either changing the snapshot or by sending only `\r` on ENTER. The latter one sounds more correct to me.

imsnif:
I think the right fix would be changing the e2e harness to send only `\r` - should probably have been that way from the beginning.

lhns:
Done! I ended up needing to update the snapshot anyway but the tests should run now.

imsnif:
Great, thank you!

tgross35:
Would it be possible to do a patch release with this fix? This issue has kept me kind of stuck on 0.43.1.

PR #5324, opened 2026-07-01
perf: foreground pane-command discovery without scanning the whole process table

### LINKED ISSUE #5304: New-pane/tab creation scans the entire /proc table (execs "ps"), causing multi-second delays on hosts with many processes
## Issue description

Creating a pane/tab makes the Zellij server exec `ps` and read `/proc` for every process on the host to discover the command running in each pane. This is O(total host processes). On hosts with many processes (shared servers, build boxes, etc.) the scan takes **seconds and blocks pane creation** (the server's pty thread waits (`wait4`) on the `ps` child), so `Alt+n` / new-tab hangs for several seconds. It is instant on machines with few processes.

## Minimal reproduction

1. On a host with a very large process count, create a new pane/tab.
2. It takes ~1.5–5+ s to appear (instant on a low-process-count host).
3. During the delay, a `ps` child of the Zellij server is present and reading `/proc` for all
   PIDs. To confirm:
   ```
   strace -f -e trace=execve -p "$(pgrep -f 'zellij --server')"
   ```
   then create a pane — you'll see `execve(".../ps", ["ps", "-ao", "ppid,args"], …)` on each
   creation, followed by a storm of `/proc/<pid>` reads.

## Other relevant information

**`zellij --version`:** reproduced on released 0.44.3; the responsible code is also present on
`main` @ ff0925f (0.45.0-dev).

**`uname -av`:** `Linux 6.18.36-1-lts x86_64 GNU/Linux` (host has ~59,000 processes/threads).

**Root cause:** `ServerOsInputOutput::get_all_cmds_by_ppid`
(`zellij-server/src/os_input_output.rs`) runs:
```rust
Command::new("ps").args(vec!["-ao", "ppid,args"]) // whole-host ppid -> command map
```
to find the foreground command of each pane (for pane titles / session serialization). `ps` reads every `/proc/<pid>` entry to enumerate the process table, so the cost scales with the host's process count, not the number of panes.

**Evidence** (`strace -f` of the server during one new-pane creation): 66,340 syscalls, 99.9% `read`; 65,216 from a single `ps` child reading 16,558+ distinct processes from pid 1 upward. No syscall took >6 ms — a large fan-out of cheap reads, not a wait.

**Ruled out:** not config-fixable (`session_serialization false` and `pane_frames false` don't stop it); not NFS/network/shell — purely a local `/proc` walk (plain `pgrep` is just as slow on this host).

**Proposed fix (PR incoming):** on Linux, read each pane's controlling-terminal foreground process group (`tpgid` from `/proc/<pid>/stat`) and that leader's `/proc/<tpgid>/cmdline` — O(panes), two small reads per pane, no subprocess; keep `ps` as a fallback for other unix targets. Verified on a ~59k-process host: creating panes execs **0** `ps` (vs **5** per discovery on 0.44.3) while command discovery still works (records `command="sleep" args "314159"`).

**Notes:** supersedes #5301 (which I filed with an incorrect early diagnosis of pty-thread accumulation; please disregard it). A separate ~1 s action-completion timeout on the 2nd+ NewPane/NewTab is filed separately.

### LINKED ISSUE #5305: 2nd and subsequent NewPane/NewTab always block ~1s: action-completion signal not delivered, route.rs hits ACTION_COMPLETION_TIMEOUT
## Issue description

Creating panes/tabs has a consistent ~1 second latency on the 2nd and every subsequent creation in a session. The first NewPane/NewTab completes relatively quickly (well, ~0.6s, but that's yet another issue); every one after that blocks for almost exactly 1 second (just over), then succeeds.

The 1 s is `route.rs`'s `ACTION_COMPLETION_TIMEOUT`: `route_action` waits for the action's completion signal (a `NotificationEnd` whose `Drop` sends a oneshot result), never receives it
for these later creations, times out, and returns a default result. It is not keybinding-specific (the `zellij action …` CLI path hits it too) and it is independent of host load (it reproduces with the unrelated #5304 `ps`-scan fix applied). Net effect: every pane/tab after the first costs ~1 extra second.

## Minimal reproduction

1. Start a session.
2. Create several tabs in a row. For clean timing, via the CLI:
   ```
   for i in 1 2 3 4 5; do { time zellij --session <name> action new-tab >/dev/null; } 2>&1 | grep real; done
   ```
3. The first completes relatively quickly; each subsequent one takes ~1.0–1.1 s.
4. The server log (`/tmp/zellij-<uid>/zellij-log/zellij.log`) shows, for the 2nd onward:
   ```
   ERROR zellij_server::route  route.rs:75: Action NewTab did not complete within 1s timeout
   ```

Same behavior with `new-pane` and when triggered by a keybinding rather than the CLI.

## Other relevant information

**`zellij --version`:** reproduced on `main` @ ff0925f (0.45.0-dev); the same `route.rs:75`
timeout is also logged on released 0.44.3.

**`uname -av`:** `Linux 6.18.36-1-lts x86_64 GNU/Linux`

**Measured timings** (`action new-tab` ×5):
```
real 0m0.675s   <- 1st (no timeout)
real 0m1.098s   <- 2nd  } each ≈ 1s ACTION_COMPLETION_TIMEOUT + ~0.1s
real 0m1.097s
real 0m1.096s
real 0m1.097s
```

**Internal timeline for the fast 1st new-tab** (completes without timing out):
```
pty.rs:500   PtyInstruction::NewTab: spawning terminals for tab 1
pty.rs:1304  spawn_terminals_for_layout: ... sending ApplyLayout
screen.rs    ApplyLayout: applying layout for tab 1
screen.rs    ApplyLayout: sending QueryTerminalSize to client 1
→ action returns ~0.675s
```
The 2nd+ run the same steps but `route.rs:75` logs the 1 s timeout — the completion signal is
not delivered.

**Where it seems to live (localized, not fully root-caused):** for a normal NewTab
(`block_on_first_terminal == false`) the completion travels as the regular `completion_tx`
into `ScreenInstruction::ApplyLayout` and drops at the end of that arm — which works for the
first tab. The first `ApplyLayout` also sends `QueryTerminalSize` (the "recover from a startup
size race" block), whose response is handled as a resize. Only the 2nd+ creations stop
delivering the completion, which points at state touched after the first one
(`pending_tab_ids` / `pending_events_waiting_for_client` / `…_for_tab`, the
QueryTerminalSize→resize path, and/or the blocking-pane completion plumbing). I localized it
to this area but did not pin down the exact 1st-vs-Nth state difference.

**Note:** distinct from #5304 (whole-host `ps` scan); found while investigating that, and
reproduces with that fix applied — confirming it's a separate bug.

ccfrost:
## What

Foreground command discovery was reading information about every process on the host, on each pane/tab creation and on every periodic poll. On hosts with many processes (eg, 50k) this stalled pane creation for tens of seconds and overran the action-completion timeout.

This replaces the `O(host processes)` lookup with an `O(panes)` one:

- `tcgetpgrp()` on each pane's pty master yields the foreground process-group
  leader's pid.
- argv comes from a single targeted `sysinfo` refresh of just those pids
  (scan-free on Linux thanks to sysinfo 0.37's `ProcessesToUpdate::Some`).
- When the foreground group is the shell itself (idle prompt, or a command
  sharing the shell's group without job control) the pane is omitted and the
  caller falls back to the shell command — matching prior behavior. With job
  control on, it now reports the true foreground job instead of possibly a
  background one.
- Linux, macOS, and BSD take the new path; Windows keeps the existing
  ppid-based path.

## Commits

- `build(deps): bump sysinfo to 0.37`: enables the scan-free
  `ProcessesToUpdate::Some` targeted refresh.
- `perf: discover foreground pane commands w/o scanning all processes`: the
  core change.
- `refactor(os_input): extract apply_post_command_hook helper`: de-duplicates
  the post-command-hook application.

## Fixes

Fixes #5304
Fixes #5305

## Testing

- `cargo fmt` clean; `cargo test -p zellij-server` passing, including new tests
  `tcgetpgrp_returns_foreground_group` and the `pty_tests` foreground-discovery
  cases (`foreground_command_emitted_with_is_foreground_true`,
  `empty_foreground_falls_back_to_shell_command`, …).
- **Manually confirmed on a test host with ~50k processes**: new-pane/tab
  creation no longer stalls (no multi-second delay, no `ps`/`/proc` scan),
  resolving both #5304 and #5305.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 * feat: new title-frames UI (https://github.com/zellij-org/zellij/pull/5318)
 * chore: upgrade wasmi to 1.1.0 (https://github.com/zellij-org/zellij/pull/5319)
 * fix: fragmented STDIN causing "garbage" input on slow SSH connections (https://github.com/zellij-org/zellij/pull/5320)
+* build(deps): bump sysinfo to 0.37 for scan-free targeted process refresh (https://github.com/zellij-org/zellij/pull/5324)
 
 ## [0.44.3] - 2026-05-13
 * fix(windows): bump windows-sys to 0.59 to align manifest with code, fixing source builds via `cargo install`/`cargo binstall` (https://github.com/zellij-org/zellij/pull/5139)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 * chore: upgrade wasmi to 1.1.0 (https://github.com/zellij-org/zellij/pull/5319)
 * fix: fragmented STDIN causing "garbage" input on slow SSH connections (https://github.com/zellij-org/zellij/pull/5320)
 * build(deps): bump sysinfo to 0.37 for scan-free targeted process refresh (https://github.com/zellij-org/zellij/pull/5324)
+* perf: discover foreground pane commands without scanning the whole process table (https://github.com/zellij-org/zellij/pull/5324)
 
 ## [0.44.3] - 2026-05-13
 * fix(windows): bump windows-sys to 0.59 to align manifest with code, fixing source builds via `cargo install`/`cargo binstall` (https://github.com/zellij-org/zellij/pull/5139)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,8 +14,6 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 * feat: new title-frames UI (https://github.com/zellij-org/zellij/pull/5318)
 * chore: upgrade wasmi to 1.1.0 (https://github.com/zellij-org/zellij/pull/5319)
 * fix: fragmented STDIN causing "garbage" input on slow SSH connections (https://github.com/zellij-org/zellij/pull/5320)
-* build(deps): bump sysinfo to 0.37 for scan-free targeted process refresh (https://github.com/zellij-org/zellij/pull/5324)
-* perf: discover foreground pane commands without scanning the whole process table (https://github.com/zellij-org/zellij/pull/5324)
 
 ## [0.44.3] - 2026-05-13
 * fix(windows): bump windows-sys to 0.59 to align manifest with code, fixing source builds via `cargo install`/`cargo binstall` (https://github.com/zellij-org/zellij/pull/5139)
--- a/zellij-server/src/os_input_output.rs
+++ b/zellij-server/src/os_input_output.rs
@@ -574,27 +574,7 @@ impl ServerOsApi for ServerOsInputOutput {
                 if command.is_empty() {
                     continue;
                 }
-                match post_hook {
-                    Some(post_hook) => {
-                        let stringified = command.join(" ");
-                        let cmd = match run_command_hook(&stringified, post_hook) {
-                            Ok(command) => command,
-                            Err(e) => {
-                                log::error!("Post command hook failed to run: {}", e);
-                                stringified.to_owned()
-                            },
-                        };
-                        let line_parts: Vec<String> = cmd
-                            .trim()
-                            .split_ascii_whitespace()
-                            .map(|p| p.to_owned())
-                            .collect();
-                        cmds.insert(ppid_str, line_parts);
-                    },
-                    None => {
-                        cmds.insert(ppid_str, command);
-                    },
-                }
+                cmds.insert(ppid_str, apply_post_command_hook(command, post_hook));
             }
         }
         cmds
@@ -649,23 +629,7 @@ impl ServerOsApi for ServerOsInputOutput {
             if command.is_empty() {
                 continue;
             }
-            let command = match post_hook {
-                Some(post_hook) => {
-                    let stringified = command.join(" ");
-                    let cmd = match run_command_hook(&stringified, post_hook) {
-                        Ok(command) => command,
-                        Err(e) => {
-                            log::error!("Post command hook failed to run: {}", e);
-                            stringified
-                        },
-                    };
-                    cmd.trim()
-                        .split_ascii_whitespace()
-                        .map(|p| p.to_owned())
-                        .collect()
-                },
-                None => command,
-            };
+            let command = apply_post_command_hook(command, post_hook);
             cmds.insert(terminal_id, command);
         }
         cmds
@@ -777,6 +741,26 @@ impl Drop for ResizeCache {
                 log::error!("Failed to apply cached resizes: {}", e);
             });
     }
+}
+
+/// Apply the post-command-discovery `post_hook` to a discovered command, returning the
+/// (possibly rewritten) command in argv form. With no hook the command is returned unchanged.
+fn apply_post_command_hook(command: Vec<String>, post_hook: &Option<String>) -> Vec<String> {
+    let Some(post_hook) = post_hook else {
+        return command;
+    };
+    let stringified = command.join(" ");
+    let cmd = match run_command_hook(&stringified, post_hook) {
+        Ok(command) => command,
+        Err(e) => {
+            Err::<(), _>(anyhow!("post command discovery hook failed to run: {e}")).non_fatal();
+            stringified
+        },
+    };
+    cmd.trim()
+        .split_ascii_whitespace()
+        .map(|p| p.to_owned())
+        .collect()
 }
 
 #[cfg(not(windows))]
--- a/zellij-server/src/os_input_output.rs
+++ b/zellij-server/src/os_input_output.rs
@@ -348,6 +348,18 @@ pub trait ServerOsApi: Send + Sync {
     fn get_all_cmds_by_ppid(&self, _post_hook: &Option<String>) -> HashMap<String, Vec<String>> {
         HashMap::new()
     }
+    /// For each `(terminal_id, shell_pid)` pane, return the foreground command running in its
+    /// controlling terminal, keyed by `terminal_id`. This is `O(panes)` and performs no full
+    /// process-table scan. Panes whose foreground process group is led by the shell itself
+    /// (idle prompt, or a command sharing the shell's group without job control) are omitted,
+    /// so the caller falls back to reporting the shell.
+    fn get_foreground_cmds(
+        &self,
+        _panes: &[(u32, u32)],
+        _post_hook: &Option<String>,
+    ) -> HashMap<u32, Vec<String>> {
+        HashMap::new()
+    }
     /// Writes the given buffer to a string
     fn write_to_file(&mut self, buf: String, file: Option<String>) -> Result<()>;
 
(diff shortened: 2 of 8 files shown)


victorhooi:
I've been having a *horrible* experience with zellij and multiple Claude Code sessions - zellij would often lock up for 5-10 seconds, or in some cases, even over a minute, whenever I created a new tab or pane =(.

I managed to compile a zellij binary from this PR, and did some A/B testing on my machine (macOS 26.5.1, ~1,300 processes on the host, config with `serialize_pane_viewport true` + `scrollback_lines_to_serialize 10000`)

_Disclaimer: Investigation was AI-assisted, but I ran all the commands by hand._

I ran two isolated A/B TMPDIR sandboxes, with identical synthetic sessions in each (19 tabs; 6 panes with 10k-line
scrollback running a long-lived command)

A is Zellij 0.44.3 release (nixpkgs), and B is zellij from this branch.

| | 0.44.3 | this PR |
|---|---|---|
| `action dump-layout` | ~1.03 s, **empty** output (route.rs 1 s timeout; 31/32 timed out) | 44 ms steady, full output |
| `action new-tab` ×12 | 11/12 ≈ 1.03 s | 37–43 ms |
| pty thread, 10 s `sample` | 8,675/8,677 samples in `get_all_cmds_by_ppid` | 0; `tcgetpgrp` ≈ 89 samples |
| `GetPaneCwd timed out` + action-timeout log lines | 56 | 0 |

Side observation on 0.44.3: responses from timed-out DumpLayouts were later delivered to unrelated CLI clients - a full stale layout printed into a subsequent `new-tab` invocation's stdout - presumably freed client-id reuse. Not reproducible under this PR's timings.

I can attach the raw commands I ran, and also the output from sample, etc. if it would help - but figured the above is enough.

@imsnif Do you know if there's any chance you could take a look, and approve this? Hoping this would improve my situation with having lots of things that redraw (e.g. Claude Code), and zellij, and lots of panes/tabs.

zellij-server/src/os_input_output.rs
```diff
@@ -348,6 +348,18 @@ pub trait ServerOsApi: Send + Sync {
     fn get_all_cmds_by_ppid(&self, _post_hook: &Option<String>) -> HashMap<String, Vec<String>> {
         HashMap::new()
     }
+    /// For each `(terminal_id, shell_pid)` pane, return the foreground command running in its
+    /// controlling terminal, keyed by `terminal_id`. This is `O(panes)` and performs no full
```
imsnif: Could you please remove everything after "keyed by `terminal_id`"? I don't think this context is relevant to the reader of the comment.

zellij-server/src/os_input_output.rs
```diff
@@ -610,29 +574,82 @@ impl ServerOsApi for ServerOsInputOutput {
...
+    fn get_foreground_cmds(
+        &self,
+        panes: &[(u32, u32)],
+        post_hook: &Option<String>,
+    ) -> HashMap<u32, Vec<String>> {
+        // Resolve each pane's foreground command via `tcgetpgrp` on the pty master: the
+        // foreground process group leader has `pid == pgid`, so the returned pgid is the pid we
+        // want. When it equals the shell's pid the shell itself is in the foreground (idle, or a
+        // command sharing the shell's group without job control) and we omit the pane so the
+        // caller falls back to the shell command. No full process-table scan.
```
imsnif: Please remove this comment, it's restating the obvious :)

zellij-server/src/os_input_output.rs
```diff
@@ -610,29 +574,82 @@ impl ServerOsApi for ServerOsInputOutput {
...
+                if fpgid > 0 && fpgid as u32 != shell_pid {
+                    terminal_to_fg_pid.insert(terminal_id, fpgid as u32);
                 }
             }
         }
+        if terminal_to_fg_pid.is_empty() {
+            return HashMap::new();
+        }
+
+        // Refresh the foreground pids.
```
imsnif: Also this one. Generally I prefer no LLM written comments in this repo as much as possible.

zellij-server/src/os_input_output.rs
```diff
@@ -726,6 +743,26 @@ impl Drop for ResizeCache {
     }
 }
 
+/// Apply the post-command-discovery `post_hook` to a discovered command, returning the
+/// (possibly rewritten) command in argv form. With no hook the command is returned unchanged.
```
imsnif: This one too.

zellij-server/src/os_input_output_unix.rs
```diff
@@ -414,6 +414,25 @@ impl UnixPtyBackend {
         }
     }
 
+    /// Return the foreground process group of the controlling terminal behind
+    /// `terminal_id` (i.e. `tcgetpgrp` on the pty master). A process group's leader
+    /// has `pid == pgid`, so the returned value is also the pid of the foreground
+    /// group leader. Returns `None` if the fd is unknown or the syscall fails.
```
imsnif: This one too

zellij-server/src/os_input_output_unix.rs
```diff
@@ -414,6 +414,25 @@ impl UnixPtyBackend {
...
+        let guard = self.terminal_id_to_raw_fd.lock().ok()?;
+        let fd = match guard.get(&terminal_id) {
+            Some(Some(fd)) => *fd,
+            _ => return None,
+        };
+        // Release the lock before the syscall (as `write_to_tty_stdin` does) rather than hold it
+        // across `tcgetpgrp`. This opens a small TOCTOU window: if the pane is torn down here the
+        // fd could be closed (or reused) before the syscall runs, giving a stale/wrong result.
+        // The syscall is non-blocking and the worst case is a momentarily wrong or missing
+        // foreground command, so we accept the race instead of serializing pty I/O on this lookup.
```
imsnif: I don't understand why?

CHANGELOG.md
```diff
@@ -14,6 +14,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 * feat: new title-frames UI (https://github.com/zellij-org/zellij/pull/5318)
 * chore: upgrade wasmi to 1.1.0 (https://github.com/zellij-org/zellij/pull/5319)
 * fix: fragmented STDIN causing "garbage" input on slow SSH connections (https://github.com/zellij-org/zellij/pull/5320)
+* build(deps): bump sysinfo to 0.37 for scan-free targeted process refresh (https://github.com/zellij-org/zellij/pull/5324)
+* perf: discover foreground pane commands without scanning the whole process table (https://github.com/zellij-org/zellij/pull/5324)
```
imsnif: Please remove the changelog entries, I prefer managing this on my own

imsnif:
Hey @ccfrost - thanks for this. I left some inline comments (mostly about removing LLM generated comments).

ccfrost:
@imsnif: I don't see any comments on the contents of the PR, only these messages. Just in case, did github publish the comments? If it did, perhaps share a link to one or a screenshot? Thanks.

imsnif (changes requested):
My bad @ccfrost ! Here they are.

zellij-server/src/os_input_output_unix.rs
ccfrost: I originally did this because write_to_tty_stdin() did similar. But I also see that other approach in set_terminal_size() and tcdrain(). That avoids the TOCTOU caveat, so I switched the code to hold the lock. Does this look good to you?

ccfrost:
@imsnif: thanks for the review. I've resolved all your asks. I have three questions:
1) Does the change to tcgetpgrp() look good to you?
2) Do you prefer the changes in response to your comments to be a new commit vs. merged into the logical commits? I started as a new commit, and am happy to switch it.
3) Benchmarking question: I wrote a short program that creates and destroys N processes, so that I could benchmark the changes even when the host doesn't otherwise have many processes. Would this be helpful to share? Asking just in case. (It's a ~100 line C program.)

imsnif:
Looks good, thank you! No need for anything else.

PR #5335, opened 2026-07-06
fix(windows): route Windows Terminal through the VT input path

### LINKED ISSUE #4885: # [Windows] Multi-line paste broken -- newlines treated as Enter
## Issue description

When running zellij natively on Windows (e.g. in Windows Terminal), pasting multi-line text causes each line to be submitted separately — as if I'm hitting Enter after every line. This completely breaks pasting code blocks, multi-line commands, or anything with newlines into apps running inside zellij (shells, editors, TUI apps like opencode, etc.).

The same paste works perfectly when running the application directly in Windows Terminal without zellij.

**Basic information**

`zellij --version`: 0.44.0 (built from source with Windows support)

`ver` (Windows): Windows 11 (Build 26220)

## Minimal reproduction

1. Open Windows Terminal (or any terminal where the `TERM` environment variable is **not** set — this is the default on Windows)
2. Run `zellij`
3. Inside zellij, open a shell (or any app that accepts multi-line input, e.g. `cat`)
4. Copy some multi-line text to clipboard:
   ```
   line one
   line two
   line three
   ```
5. Paste (Ctrl+V) into the app

**Expected**: All three lines appear as a single paste operation (the inner app sees them wrapped in bracketed paste sequences)

**Actual**: Each newline triggers a real Enter keypress, causing each line to be submitted/executed separately

### What's going on

Zellij on Windows has two stdin input paths, chosen at `stdin_handler.rs:94`:

```rust
let use_vt_reader = std::env::var("TERM").is_ok() && enable_vt_input();
```

| Condition      | Input Path         | Read Method                      | Paste Support                       |
| -------------- | ------------------ | -------------------------------- | ----------------------------------- |
| `TERM` not set | **Native Console** | `ReadConsoleInput` via crossterm | No `Event::Paste`                   |
| `TERM` is set  | VT Reader          | `ReadFile` raw bytes via termwiz | Parses `\x1b[200~...\x1b[201~`     |

In most Windows environments (Windows Terminal, cmd.exe, PowerShell), `TERM` is **not set**, so zellij uses the native console path. On this path:

1. Zellij enables bracketed paste mode by sending `\x1b[?2004h` to the outer terminal
2. Windows Terminal receives this and wraps pasted content with `\x1b[200~...\x1b[201~`
3. But zellij reads input via `ReadConsoleInput` (native console API), not raw VT bytes
4. `ReadConsoleInput` delivers each character as an individual `KEY_EVENT` — including newlines
5. Crossterm converts each `KEY_EVENT` to `Event::Key` — `Event::Paste` is **never produced** on this path (crossterm only recognizes bracketed paste sequences on the Unix parse path)
6. Every `\n` in the paste becomes `Event::Key(Enter)` → forwarded to the PTY as a real Enter

The paste handling code in `input_handler.rs` (lines 176-227) is correct — it just never gets triggered because crossterm never produces the paste event on the native console path.

### Affected code locations

| File                                               | Relevance                                       |
| -------------------------------------------------- | ----------------------------------------------- |
| `zellij-client/src/stdin_handler.rs:94`            | `TERM` check decides input path                 |
| `zellij-client/src/stdin_handler_windows.rs:17-63` | Native console event loop (no paste detection)  |
| `zellij-client/src/input_handler.rs:151-227`       | Paste event handling (correct but unreachable)   |
| `zellij-server/src/panes/terminal_pane.rs:237-248` | Bracket stripping for non-bracketed-paste panes |

### Workaround

Set `TERM` as a persistent environment variable so zellij always uses the VT reader path:

```powershell
[Environment]::SetEnvironmentVariable('TERM', 'xterm-256color', 'User')
```

This makes zellij use `ENABLE_VIRTUAL_TERMINAL_INPUT` + termwiz's `InputParser`, which correctly parses bracketed paste sequences from the raw VT byte stream. Note: this changes ALT key behavior from modifier flags to ESC-prefix encoding, which may affect some keybindings.

### Suggested fix

The simplest fix would be to detect Windows Terminal (via `WT_SESSION` env var) and automatically use the VT reader path, even when `TERM` isn't set:

```rust
#[cfg(windows)]
let use_vt_reader = (std::env::var("TERM").is_ok()
    || std::env::var("WT_SESSION").is_ok())  // Windows Terminal
    && enable_vt_input();
```

Related: #3865 (similar symptom reported from the SSH client side)

divens:
Fixes #4885 and, more broadly, moves Windows Terminal onto the VT byte input path.

### The multi-line paste issue

The native-console input path (crossterm `INPUT_RECORD`s via `ReadConsoleInput`) can't observe bracketed-paste sequences (`\e[200~ ... \e[201~`). They're filtered before the app ever sees them. Multi-line paste therefore ran each line as a separate command because every `\r\n` inside the paste hit the shell as a real Enter.

An alternative fix would be to synthesize paste events on the native path (watching for input surges, sampling the clipboard, emitting a synthetic paste event). That approach is fragile and, more importantly, doesn't address the other limitations of the native path (KKP, emoji, richer mouse).

### Moving WT to the VT path

WT sets `WT_SESSION` per-process. Adding it as a second trigger for the VT reader path (`use_vt_path()` helper) means default WT installs get raw VT bytes without users having to add `TERM=xterm-256color` to their profile.

The stdin handler and the mouse setup each had their own `TERM`-only check. Bumping just the stdin gate made mouse setup fall through to `crossterm::EnableMouseCapture`, which calls `SetConsoleMode` and clobbered the `ENABLE_VIRTUAL_TERMINAL_INPUT` flag the VT reader had just set, breaking all WT input.

To make that class of drift impossible, this PR introduces `use_vt_path()` in `os_input_output_windows.rs` as the single source of truth. All VT-path-conditional code (stdin handler + mouse setup) calls it, so future additions can't misalign.

### What was tested

- **Standard WT usage (`WT → pwsh → zellij`), `TERM` unset, `WT_SESSION` set** → the common scenario: a default WT profile spawns pwsh, from which the user launches zellij. Confirmed the fix is what unlocks the new behavior (`main` is broken, this branch works).
- **WT profile launching zellij directly (`WT → zellij`, no pwsh in the chain), `TERM` unset** → a custom WT profile whose commandline is `zellij.exe`, no parent shell process. Same working outcome; confirms the fix doesn't depend on a wrapper shell.
- **Alacritty / WezTerm on Windows** → already worked, verified no regression.

Applications verified to receive multi-line paste correctly under the fix:

- WSL bash (readline)
- vim
- Claude CLI (Ink-based TUI)

**Note**: PowerShell, cmd, and Nushell still don't handle multi-line paste, even without zellij in the chain. All three are upstream bracketed-paste-support gaps (see [PSReadLine#1471](https://github.com/PowerShell/PSReadLine/issues/1471), blocked on [dotnet/runtime#60107](https://github.com/dotnet/runtime/issues/60107)). Nothing zellij can do about those from this side.

### Why this is the right long-term direction

The native-console input path on Windows exists as a bridge for the legacy console host. Everything zellij cares about is either already better on the VT path or is arriving there:

- **Bracketed paste** : only visible on the VT byte stream (fixed here).
- **Kitty keyboard protocol** : WT 1.25+ ships full KKP support. Zellij's kitty parser expects raw VT bytes; the native path can't feed it.
- **Emoji / supplementary-plane input** : the native path suffers from the crossterm surrogate-pair bug (key-up events corrupting the pair buffer). The VT path delivers UTF-8 bytes directly and sidesteps it entirely.
- **SGR mouse encoding** : richer than the legacy console mouse events.
- **Path parity with Unix** : same parser, same code paths, fewer platform-specific quirks.

Native console will still be selected when neither `TERM` nor `WT_SESSION` is set (legacy conhost or any environment where `cmd`/`pwsh` are spawned without WT wrapping)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 * fix: fragmented STDIN causing "garbage" input on slow SSH connections (https://github.com/zellij-org/zellij/pull/5320)
 * feat: stacked lists - new stack UI (https://github.com/zellij-org/zellij/pull/5331)
 * fix: more verbose startup errors (https://github.com/zellij-org/zellij/pull/5334)
+* fix(windows): route Windows Terminal through the VT input path (https://github.com/zellij-org/zellij/pull/5335)
 
 ## [0.44.3] - 2026-05-13
 * fix(windows): bump windows-sys to 0.59 to align manifest with code, fixing source builds via `cargo install`/`cargo binstall` (https://github.com/zellij-org/zellij/pull/5139)
--- a/zellij-client/src/os_input_output_windows.rs
+++ b/zellij-client/src/os_input_output_windows.rs
@@ -9,6 +9,30 @@ use std::io::Write;
 use std::path::Path;
 use zellij_utils::ipc::{IpcReceiverWithContext, IpcSenderWithContext};
 
+/// Whether zellij should use the VT byte path on Windows (raw stdin via
+/// `ReadFile` + termwiz/kitty parsing) instead of the native-console path
+/// (crossterm `INPUT_RECORD`s).
+///
+/// True when either of these environment variables is set:
+///
+/// * `TERM` — set automatically by terminal emulators that expose a Unix-style
+///   environment (Alacritty, WezTerm, WSL sessions inheriting bash's env), or
+///   set manually by users who want zellij on the VT path (e.g. adding
+///   `TERM=xterm-256color` to a Windows Terminal / PowerShell profile).
+/// * `WT_SESSION` — set per-session by Windows Terminal itself. Detecting it
+///   means default WT installs land on the VT path without the user having to
+///   set `TERM`, so they get bracketed paste, kitty keyboard, SGR mouse, and
+///   the other features that depend on raw VT input.
+///
+/// Every VT-path-conditional code site (stdin handler, mouse mode setup,
+/// future input-related code) should call this so the gates can't drift —
+/// when they did drift, mouse setup using `crossterm::EnableMouseCapture`
+/// clobbered the `ENABLE_VIRTUAL_TERMINAL_INPUT` flag the stdin path had
+/// just set and broke all input from WT.
+pub(crate) fn use_vt_path() -> bool {
+    std::env::var("TERM").is_ok() || std::env::var("WT_SESSION").is_ok()
+}
+
 /// Windows async signal listener.
 ///
 /// Polls `crossterm::terminal::size()` at 100ms intervals for resize events,
@@ -206,7 +230,7 @@ fn enable_vt_processing_on_stdout() {
 /// Windows Terminal) and use crossterm's Console API approach.
 pub(crate) fn enable_mouse_support(stdout: &mut dyn Write) -> Result<()> {
     let err_context = "failed to enable mouse mode";
-    if std::env::var("TERM").is_ok() {
+    if use_vt_path() {
         enable_vt_processing_on_stdout();
         stdout
             .write_all(super::os_input_output::ENABLE_MOUSE_SUPPORT.as_bytes())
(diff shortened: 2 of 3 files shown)


MitchelliJ:
Thanks for this!

PR #5353, opened 2026-07-10
fix(plugins): deliver right-clicks to plugin panes

### LINKED ISSUE #5350: Plugins never receive Mouse::RightClick, the forwarding path is dead code
Zellij 0.44.3, macOS 15. The relevant code looks unchanged on current main.

## Summary

The plugin API advertises right-click support: `Event::Mouse(Mouse::RightClick(line, col))` exists in zellij-tile (zellij-utils/src/data.rs:900), plugins can subscribe to `EventType::Mouse`, and `PluginPane` even implements a forwarder for it. But no right-click ever reaches a plugin, because the forwarder is never called. I shipped a right-click interaction in a plugin and found out in production that the event is unproducible.

## Where it breaks

`PluginPane::handle_right_click` (zellij-server/src/panes/plugin_pane.rs:719-727) sends `Event::Mouse(Mouse::RightClick(..))` to the plugin, which is presumably the intended delivery path. A repo-wide search for `handle_right_click` finds only its definitions: that impl, and the empty default trait method at zellij-server/src/tab/mod.rs:529-530, which carries its own confession:

```rust
// TODO: this should probably be merged with the mouse_right_click
fn handle_right_click(&mut self, _to: &Position, _client_id: ClientId) {}
```

Nothing calls it. The live mouse pipeline routes right-clicks through `mouse_right_click(position, is_held) -> Option<String>` and `mouse_right_click_release` (tab/mod.rs:540-543), which exist for terminal-pane mouse-reporting passthrough; plugin panes don't turn them into event delivery.

The net effect is that `LeftClick`, scroll, and hold events reach plugins while `RightClick` never does, despite the enum variant and the ready-made forwarder. `MiddleClick` looks affected the same way, though I haven't depended on it.

## Repro

Subscribe a plugin to `EventType::Mouse` and log every mouse event it receives, then click around its pane: left clicks and scrolls arrive, right clicks never do. My evidence is the source trace above plus production behavior of my plugin (its right-click handler, covered by unit tests that inject the event, never fires against a real server, while the terminal emulator side does forward the click into the mouse-reporting stream). A minimal logging plugin reproducing this is straightforward to construct from the above.

## Ask

Either wire the existing forwarder into the mouse pipeline, routing pane-targeted right-clicks to `handle_right_click` with the same coordinate translation left clicks get, or document right-click as unsupported for plugins and remove the dead forwarder so the API stops implying otherwise.

gretzke:
Fixes #5350.

Plugins subscribe to EventType::Mouse and the API exposes Mouse::RightClick, but no right-click ever reached a plugin: PluginPane::handle_right_click existed with no callers, and the live mouse pipeline only routed right-clicks through the terminal-pane mouse-reporting passthrough. I shipped a right-click interaction in a plugin and found out in production that the event is unproducible.

This routes pane-targeted right-click presses through the same path and coordinate translation left-clicks get, into the existing forwarder. Terminal-pane right-click passthrough behavior is unchanged, with its focused test still passing. Middle-click is deliberately not included: it has no plugin-pane forwarder today, so wiring it is a bigger decision than this fix should carry.

The integration test mirrors the existing left-click delivery coverage for the right-click path.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 
 ## [Unreleased]
+* fix: deliver right-clicks to plugin panes (https://github.com/zellij-org/zellij/issues/5350)
 * feat: allow tabs to have different sizes if clients aren't focused on the same one (https://github.com/zellij-org/zellij/pull/5133)
 * feat: PWA support for the web client (manifest + icons + iOS meta tags) so the page can be installed as a standalone app (https://github.com/zellij-org/zellij/pull/5184)
 * feat: mobile UI (https://github.com/zellij-org/zellij/pull/5241)
--- a/zellij-server/src/tab/mouse_handler.rs
+++ b/zellij-server/src/tab/mouse_handler.rs
@@ -176,6 +176,10 @@ enum MouseAction {
         pane_id: PaneId,
         position: Position,
     },
+    RightClick {
+        pane_id: PaneId,
+        position: Position,
+    },
     SendToTerminal {
         pane_id: PaneId,
         event: MouseEvent,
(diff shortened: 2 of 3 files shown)


jknlsn:
Yep, hit this exact bug on my fork and landed on basically the same fix, so +1. The forwarding path is dead code, right-clicks never reach plugin panes, and it really hurts the zjstatus-style bar panes that never get focused. Couple of small things from my version that might help: `RightClick` doesn't need to carry `position`, since `execute_mouse_action` can read it off the event, which let me fold the press-only cases into the existing `event.right` branch and drop ~20 lines. And it's worth having the test assert the pane *isn't* focused, since that unfocused delivery is really the main behaviour change. I've got my version on a branch in case it's useful for comparison: https://github.com/jknlsn/zellij/tree/fix/mouse-passthrough-to-plugin

PR #5354, opened 2026-07-10
fix(input): support Ctrl with non-letter keys like backslash

### LINKED ISSUE #4509: Unable to use C-\ as keybind

Socialpranker:
With `support_kitty_keyboard_protocol false`, `Ctrl \` (and a few of its neighbors) can't be used as a keybind. `Ctrl \` sends the control byte 0x1c, but termwiz's `build_basic_key_map()` only maps Ctrl+A through Ctrl+Z, which covers bytes 0x01-0x1a. Byte 0x1c falls outside that range, so termwiz hands it back as a plain `KeyCode::Char('\x1c')` with no Ctrl modifier attached, and the keybind never matches.

As diagnosed in the thread by @m-morino, the same gap applies to Ctrl+] (0x1d), Ctrl+^ (0x1e) and Ctrl+_ (0x1f). `cast_termwiz_key()` in `zellij-utils/src/input/mod.rs` already has a special case for raw byte 0x08 (Ctrl+H, which collides with Backspace), so this PR extends that same pattern to the four missing bytes, mapping each one straight to the matching Ctrl-modified key instead of trying to fix this further upstream in the vendored termwiz key map.

Added unit tests for all four bytes, plus a regression test that runs the existing Ctrl+A..Z path through the real parser to make sure this change doesn't touch that behavior. `cargo test -p zellij-utils` passes (414 tests) and `cargo clippy -p zellij-utils --lib --tests` reports no new warnings.

Fixes #4509

---
Disclosure: this PR (diagnosis review, implementation, and tests) was written with AI assistance (Claude).

### DIFF
--- a/zellij-utils/src/input/mod.rs
+++ b/zellij-utils/src/input/mod.rs
@@ -92,6 +92,23 @@ mod not_wasm {
             return KeyWithModifier::new(BareKey::Char('h')).with_ctrl_modifier();
         };
 
+        // termwiz's basic_key_map only maps Ctrl-A through Ctrl-Z (bytes
+        // 0x01-0x1a). Ctrl combined with a few non-letter keys is left
+        // unmapped, so those bytes come back as a plain, unmodified
+        // KeyCode::Char. Map them here the same way Ctrl-A..Z are mapped.
+        if raw_bytes == [0x1c] {
+            return KeyWithModifier::new(BareKey::Char('\\')).with_ctrl_modifier();
+        }
+        if raw_bytes == [0x1d] {
+            return KeyWithModifier::new(BareKey::Char(']')).with_ctrl_modifier();
+        }
+        if raw_bytes == [0x1e] {
+            return KeyWithModifier::new(BareKey::Char('^')).with_ctrl_modifier();
+        }
+        if raw_bytes == [0x1f] {
+            return KeyWithModifier::new(BareKey::Char('_')).with_ctrl_modifier();
+        }
+
         if raw_bytes == [10] {
             if let Some((keybinds, mode)) = keybinds_mode {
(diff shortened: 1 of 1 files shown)


Socialpranker:
Friendly ping — open since 10 July, no conflicts.

It makes `Ctrl \` and its neighbours usable as keybinds when `support_kitty_keyboard_protocol` is false: termwiz's `build_basic_key_map()` only covers Ctrl+A..Ctrl+Z (bytes 0x01-0x1a), so 0x1c comes back as a plain char with no Ctrl modifier and the keybind never matches.

No urgency — happy to rebase or rework it if you'd rather handle this upstream in termwiz than in zellij.

Socialpranker:
One more nudge, sorry for the repeat — this has been quiet a while and I don't want it to just get buried.

PR #5364, opened 2026-07-13
fix(windows): prefer PATHEXT variants over bare filename in resolver

### LINKED ISSUE #5361: Crash: %1 is not a valid Win32 application
# Issue description

Crash when creating a pane in a template to run composer commands.

# Minimal reproduction

Run `zellij --layout layout.kdl` with the following content on a windows machine with [composer](https://getcomposer.org/) installed.

```kdl
layout {
    pane {
        command "composer"
    }
}
```

# Other relevant information

```
~: zellij --layout layout.kdl
Error occurred in server:

  × Thread 'pty' panicked.
  ├─▶ Originating Thread(s)
  │     1. ipc_server: NewClient
  │     2. screen_thread: NewTab
  │     3. plugin_thread: NewTab
  │     4. pty_thread: NewTab
  │
  ├─▶ At zellij-server\src\lib.rs:2030:58
  ╰─▶ Program terminates: a fatal error occured

      Caused by:
          0: failed to open new tab
          1: failed to apply run instruction
          2: failed to spawn terminal
          3: failed to spawn terminal for 'C:\ProgramData\ComposerSetup\bin\composer'
          4: %1 is not a valid Win32 application. (os error 193)

      Stack backtrace:
         0: <unknown>
         1: <unknown>
         2: <unknown>
         3: <unknown>
         4: <unknown>
         5: <unknown>
         6: <unknown>
         7: <unknown>
         8: <unknown>
         9: <unknown>
        10: BaseThreadInitThunk
        11: RtlUserThreadStart

      Panic backtrace:
         0: <unknown>
         1: <unknown>
         2: <unknown>
         3: <unknown>
         4: <unknown>
         5: <unknown>
         6: <unknown>
         7: <unknown>
         8: <unknown>
         9: aws_lc_0_39_0_jent_entropy_switch_notime_impl
        10: aws_lc_0_39_0_jent_entropy_switch_notime_impl
        11: <unknown>
        12: <unknown>
        13: <unknown>
        14: BaseThreadInitThunk
        15: RtlUserThreadStart

        -> To get additional information, check the log at: C:\Users\owner\AppData\Local\Temp\zellij\zellij-log\zellij.log
        -> To see a backtrace next time, reproduce the error with: RUST_BACKTRACE=1 zellij [...]
        -> To help us fix this, please open an issue: https://github.com/zellij-org/zellij/issues
```

Log file

```
INFO   |zellij_client            | 2026-07-13 00:32:24.960 [main      ] zellij-client\src\lib.rs:735: Starting Zellij client! 
INFO   |zellij_server            | 2026-07-13 00:32:25.019 [main      ] zellij-server\src\lib.rs:826: Starting Zellij server! 
INFO   |zellij_server            | 2026-07-13 00:32:25.114 [main      ] zellij-server\src\lib.rs:982: FirstClientConnected: initializing session 
INFO   |zellij_server            | 2026-07-13 00:32:25.115 [main      ] zellij-server\src\lib.rs:2423: Loading plugins using Wasmi interpreter 
INFO   |zellij_server::plugins   | 2026-07-13 00:32:25.115 [wasm      ] zellij-server\src\plugins\mod.rs:313: Wasm main thread starts 
INFO   |zellij_server            | 2026-07-13 00:32:25.116 [main      ] zellij-server\src\lib.rs:994: FirstClientConnected: session initialized, spawning tabs 
INFO   |zellij_server::plugins::p| 2026-07-13 00:32:25.138 [plugin-exec-0] zellij-server\src\plugins\plugin_loader.rs:151: Loaded plugin 'link' in 9.6644ms 
INFO   |zellij_server::pty       | 2026-07-13 00:32:25.147 [pty       ] zellij-server\src\pty.rs:500: PtyInstruction::NewTab: spawning terminals for tab 0 
ERROR  |zellij_utils::errors::not| 2026-07-13 00:32:25.184 [pty       ] zellij-utils\src\errors.rs:819: Panic occured:
             thread: pty
             location: At zellij-server\src\lib.rs:2030:58

Caused by:
    0: failed to open new tab
    1: failed to apply run instruction
    2: failed to spawn terminal
    3: failed to spawn terminal for 'C:\ProgramData\ComposerSetup\bin\composer'
    4: %1 is not a valid Win32 application. (os error 193)

Stack backtrace:
   0: <unknown>
   1: <unknown>
   2: <unknown>
   3: <unknown>
   4: <unknown>
   5: <unknown>
   6: <unknown>
   7: <unknown>
   8: <unknown>
   9: <unknown>
  10: BaseThreadInitThunk
  11: RtlUserThreadStart 
INFO   |zellij_server::plugins   | 2026-07-13 00:32:25.185 [wasm      ] zellij-server\src\plugins\mod.rs:1273: wasm main thread exits 
ERROR  |zellij_server::os_input_o| 2026-07-13 00:32:25.185 [screen    ] zellij-server\src\os_input_output.rs:724: Failed to apply cached resizes: failed to send message to pty writer 
INFO   |zellij_server::plugins::w| 2026-07-13 00:32:25.185 [wasm      ] zellij-server\src\plugins\wasm_bridge.rs:494: Bye from plugin 0 
```

# Working workaround
Using spawning a shell to run the command is a working method.
```kdl
layout {
    pane {
        command "nu"
        args "-c" "composer"
    }
}
```

divens:
Fixes https://github.com/zellij-org/zellij/issues/5361

`find_executable` used to short-circuit on the first exists()-and-is_file match and only consult PATHEXT if no bare file was present. That returned the wrong file for tools whose Windows installer ships both a Unix-style launcher and a `.bat` alongside it — the reported case is Composer, whose setup places bare `composer` (a Unix launcher, invalid PE on Windows) next to `composer.bat`. The bare file was resolved and handed to CreateProcessW, which rejected it with ERROR_BAD_EXE_FORMAT.

On Windows, when the candidate has no user-supplied extension, walk PATHEXT first and only fall back to the bare match if nothing matches. This mirrors cmd.exe's resolution.


imsnif:
Hum. I'm a little bit worried about this as it's the sort of thing that's easily exploitable. How is the behavior in windows shells? If I just execute "composer" and there's both a bare composer file and a composer.bat file, is the latter the one to be executed?

divens:
> How is the behavior in windows shells? If I just execute "composer" and there's both a bare composer file and a composer.bat file, is the latter the one to be executed?

Both cmd and powershell resolve to `composer.bat`. I verified this locally: 
- for cmd, I created a trivial test script (printing a message) both with and without extension, changing the message, ran it with `cmd /c test_script`. The message printed was the one from the `.bat` test script.
- for powershell, I did the same empirical verification, same result. On top of that powershell has `Get-Command -All` which is [documented](https://learn.microsoft.com/en-us/powershell/module/microsoft.powershell.core/get-command?view=powershell-7.6#example-12-get-all-instances-of-the-notepad-command) to return matching commands in _execution precedence order_. The `.bat` script comes before the bare name. 

A note on why it works this way : on Windows, the `PATHEXT` environment variable contains the ordered list of file extensions that Windows considers executable. This is documented [here](https://learn.microsoft.com/en-us/powershell/module/microsoft.powershell.core/about/about_environment_variables?view=powershell-7.6#other-environment-variables-used-by-powershell). Powershell refers to this when resolving a command and respects the order for precedence.

When we're resolving commands ourselves in zellij, we must first honor extensions present in `PATHEXT` before falling back to a bare name option. `CreateProcessW` which we use in `spawn_child_process` doesn't resolve the command that we pass to it. 

> I'm a little bit worried about this as it's the sort of thing that's easily exploitable. 

The specific new behavior in this patch is that when a bare file and a same-name `.bat` (or other PATHEXT extension) sit in the same directory, we now pick the `.bat`. For that to be exploitable, an attacker needs write access to a directory holding a legitimate extensionless PE binary. At that point they can just replace that binary directly.

If the attacker doesnt have access to the directory containing the legitimate binary (or script) we're trying to run, another vector would be PATH-hijacking via an earlier-PATH directory (planting a similarly name malicious binary or script in a PATH directory that appears before the one we want to run from). This is not specific to this patch.

One thing to note : the pre-patch situation wasn't really a safe fallback, it just happened to hand a non-Windows script to CreateProcessW which failed with ERROR_BAD_EXE_FORMAT and caused the crash. It's also worth noting that extension-less executables are uncommon on native Windows; CLIs, scripts and tools rely on extension resolution to give users the shorter form (`composer` instead of `composer.bat`) while conforming to the platform's file-naming conventions. That's what PATHEXT exists for, and this patch just brings zellij's resolver in line with it.

imsnif:
If this is also the behavior of the windows shells, it's fine for me (as long of course as this is just a windows change, which it seems it is).

PR #5369, opened 2026-07-15
fix(panes): clear custom name when renaming

### LINKED ISSUE #3309: pane rename doesn't respect `PaneNameInput 0` after a first rename
2. Issues with the Zellij UI / behavior / crash

When switching to renamepane mode via keybindings zellij stops clearing the current pane name after a first rename, in this video I create 3 new panes, on the first rename zellij clears the name and puts the "Enter name..." placeholder, on subsequent renames the name isn't cleared and I have to manually delete it.

[Screencast from 04-26-2024 06:13:17 PM.webm](https://github.com/zellij-org/zellij/assets/6853656/42eaab31-91ff-4536-ad4a-b99aa849f643)

The keybind that triggers the rename calls `PaneNameInput 0` like the default config so I'm inclined to think this is a bug, don't see anything in the debug log.

**Basic information**

`zellij --version`: `zellij 0.40.0`

`stty size`: `16 157`

## How to repro

Install zellij >= 0.39.2 (didn't try lower than that), starts with the default config/layouts and press `Ctrl + p` then `c` to do a pane rename. Observe how after a first rename it never clears the name again.

### LINKED ISSUE #4600: [Bug] `PaneNameInput 0` fails to clear existing pane name on second rename (inconsistent with Tab behavior)
When renaming a pane that already has a custom name, entering the rename mode does not clear the existing name. This forces the user to manually backspace through the old name.

  This behavior is inconsistent with Tab renaming, where entering rename mode (default Ctrl + t -> r) correctly clears the existing tab name allows for immediate typing of the new name.

  Reproduction Steps
   1. Open a new Zellij session.
   2. Rename the current pane to "Test" (Ctrl + p -> c -> type "Test" -> Enter).
   3. Attempt to rename the pane again (Ctrl + p -> c).
   4. Actual Result: The input buffer still contains "Test". Typing "New" results in "TestNew".
   5. Expected Result: The input buffer should be empty (cleared), allowing the user to type "New" directly.

  Technical Analysis
  I have investigated the codebase and identified the cause.

   1. The Trigger: The default configuration correctly sends a PaneNameInput 0 (ASCII \0 null character) when entering rename mode to signal a clear operation:

   1    // zellij-utils/assets/config/default.kdl
   2    bind "c" { SwitchToMode "RenamePane"; PaneNameInput 0; }

   2. The Bug: In zellij-server/src/tab/mod.rs, the function update_active_pane_name passes the input string to clean_string_from_control_and_linebreak before checking specific control
      codes.

     Since \0 is considered a control character, it is filtered out by clean_string... and the logic never receives the signal to clear the string.

     Current Code (`zellij-server/src/tab/mod.rs`):

    1    // ...
    2    .map(|active_pane| {
    3        let to_update = match s {
    4            "\u{007F}" | "\u{0008}" => {
    5                // delete and backspace keys
    6                s
    7            },
    8            // BUG: The \0 character is filtered out here because it is a control char
    9            _ => &clean_string_from_control_and_linebreak(s),
   10        };
   11        active_pane.update_name(&to_update);
   12    })?;

   3. Comparison with Working Logic (Tabs):
     The Tab renaming logic in zellij-server/src/screen.rs (update_active_tab_name) correctly handles this by matching \0 explicitly:

   1    // zellij-server/src/screen.rs
   2    match s {
   3        "\0" => {
   4            active_tab.name = String::new(); // Correctly clears name
   5        },
   6        // ...
   7    }

MiroFurtado:
## Summary

- preserve the NUL pane-name terminator when entering rename mode
- add regression coverage for clearing and replacing an existing custom name
- verify other control characters remain sanitized

## Implementation

`Tab::update_active_pane_name` now forwards NUL alongside its existing backspace and delete exceptions instead of stripping it with `clean_string_from_control_and_linebreak`. This is the shared input path for terminal and plugin panes, whose `update_name` implementations already clear the custom pane name when they receive NUL.

## Test plan

- `cargo xtask format --check`
- `cargo test -p zellij-server interactive_rename_ -- --nocapture`
- `cargo test -p zellij-server --lib`

Fixes #4600.
Fixes #3309.

### DIFF
--- a/zellij-server/src/tab/mod.rs
+++ b/zellij-server/src/tab/mod.rs
@@ -6134,6 +6134,7 @@ impl Tab {
             .with_context(|| format!("no active pane found for client {client_id}"))
             .map(|active_pane| {
                 let to_update = match s {
+                    "\0" => s, // pane name terminator
                     "\u{007F}" | "\u{0008}" => {
                         // delete and backspace keys
                         s
--- a/zellij-server/src/tab/unit/tab_tests.rs
+++ b/zellij-server/src/tab/unit/tab_tests.rs
@@ -16187,6 +16187,40 @@ pub fn interactive_rename_appends_to_existing_name() {
 }
 
 #[test]
+pub fn interactive_rename_nul_clears_existing_name() {
+    let size = Size {
+        cols: 121,
+        rows: 20,
+    };
+    let mut tab = create_new_tab(size, true);
+    let client_id = 1;
+    let pane_id = PaneId::Terminal(1);
+    let _ = tab.rename_pane_by_pane_id(pane_id, "flame".as_bytes().to_vec());
+    let _ = tab.update_active_pane_name(vec![0], client_id);
+    let pane = tab.get_pane_with_id(pane_id).unwrap();
+    assert_eq!(pane.custom_title(), None);
+
+    let _ = tab.update_active_pane_name(b"spark".to_vec(), client_id);
+    let pane = tab.get_pane_with_id(pane_id).unwrap();
+    assert_eq!(pane.custom_title(), Some("spark".to_owned()));
+}
+
+#[test]
+pub fn interactive_rename_sanitizes_other_control_characters() {
+    let size = Size {
+        cols: 121,
+        rows: 20,
+    };
+    let mut tab = create_new_tab(size, true);
+    let client_id = 1;
+    let pane_id = PaneId::Terminal(1);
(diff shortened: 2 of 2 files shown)


MiroFurtado:
Have been running this patch on top of master `HEAD` and it is matching the desired behavior/existing-tab-rename-behavior without any issues that I've noticed. This is particularly nice with agentic development like `copilot` which has a tendency to rename panes with very verbose names that I would previously have to spend a fair bit of time backspacing through.

imsnif:
Great, thank you!

PR #5388, opened 2026-07-19
fix(action): normalize CRLF line endings for paste

### LINKED ISSUE #5387: zellij action paste does not normalize CRLF line endings
## Issue description
`zellij action paste` does not normalize CRLF line endings before sending
the text through bracketed paste.
When the pasted text contains Windows-style `\r\n` line endings, some
terminal applications interpret both `\r` and `\n` as separate line
breaks. This results in an extra blank line for every CRLF sequence.

## Steps to reproduce
Create a text file containing CRLF line endings:
```sh
printf 'first\r\nsecond\r\nthird\r\n' > /tmp/crlf.txt
```
Paste the contents through the Zellij CLI:
```sh
zellij action paste "$(cat /tmp/crlf.txt)"; cat >/dev/null
```

## Actual behavior
After the `zellij action paste` command finishes, the pasted text is
displayed with an extra blank line for each CRLF sequence:
```
first

second

third
```
The trailing `cat >/dev/null` prevents the pasted lines from being
executed as shell commands.
## Expected behavior
Each CRLF sequence should be treated as one logical line ending:
```
first
second
third
```

yamam:
Closes #5387
## Summary
Normalize CRLF line endings before sending text through the `Paste`
action.
Without normalization, applications that interpret both carriage return
and line feed as input line breaks receive two line breaks for each CRLF
sequence.
## Cause
The `Action::Paste` handler converted its input directly to bytes:
```rust
let bytes = chars.into_bytes();
```
This preserved Windows-style CRLF line endings as two separate control
characters.
The resulting bytes were then passed to `ScreenInstruction::Paste` and
wrapped in bracketed paste markers by `paste_to_pane_id()`.
When CRLF is sent as terminal input, CR and LF can each be processed as
a line break, resulting in an extra blank line for every CRLF sequence.
## Changes
Normalize CRLF to LF in the `Action::Paste` path before converting the
text to bytes:
```rust
let bytes = chars.replace("\r\n", "\n").into_bytes();
```
The normalization is intentionally limited to CRLF sequences:
```text
CRLF -> LF
LF   -> LF
CR   -> CR
```
Standalone carriage returns remain unchanged.
`WriteChars` and `WriteCharsToPaneId` are also left unchanged. These are
lower-level input actions and should preserve the exact characters
provided by the caller.

### DIFF
--- a/zellij-server/src/route.rs
+++ b/zellij-server/src/route.rs
@@ -322,7 +322,7 @@ pub(crate) fn route_action(
             senders
                 .send_to_screen(ScreenInstruction::ClearScroll(client_id))
                 .with_context(err_context)?;
-            let bytes = chars.into_bytes();
+            let bytes = chars.replace("\r\n", "\n").into_bytes();
             senders
                 .send_to_screen(ScreenInstruction::Paste(
                     bytes,


yamam:
Based on the investigation summarized in #5387, this change is no longer
necessary. I will handle the newline normalization before invoking
`zellij action paste`, so I am closing this PR.

PR #5393, opened 2026-07-20
Add client-local routing for keybind and plugin pipes

### LINKED ISSUE #5392: Allow keybind and plugin pipes to target the originating client only
## Summary
When multiple clients are connected to the same session, keybind pipes created
with `MessagePlugin` and plugin-to-plugin pipes can be delivered to multiple
client-specific instances of the destination plugin.
I would like to propose an option that restricts delivery to the destination
plugin instance associated with the client from which the pipe originated.
The proposed option name is `client_local`.
To preserve the existing behavior, the option would default to `false`.
Client-local routing would be enabled only when `client_local` is explicitly
set to `true`.
## Background
In a multi-client session, the same plugin can have a separate instance for
each client.
I encountered this while developing a plugin named `vim-tab-conditional`. It is
a custom plugin that changes the behavior of keybindings depending on whether
the active tab is named `vim`.
For example, the `right` message behaves as follows:
- On a tab named `vim`, the plugin sends `Alt-l` to the focused terminal.
- On any other tab, the plugin moves to the tab on the right.
The keybinding is configured with `MessagePlugin`:
```kdl
bind "Alt l" {
    MessagePlugin "vim-tab-conditional" {
        name "right"
    }
}
```
If client A and client B are connected to the same session and each has an
instance of `vim-tab-conditional`, pressing `Alt-l` on client A can deliver the
message to both instances:
```text
keybind on client A
    -> vim-tab-conditional instance for client A
    -> vim-tab-conditional instance for client B
```
For this type of action, only the plugin instance associated with the client
that generated the keybinding should receive the message.
The same problem can occur with plugin-to-plugin pipes.
## Backward compatibility
The existing broadcast behavior might be used by existing keybindings or
plugins.
For this reason, the proposed `client_local` option would preserve the current
behavior by default:
```text
client_local = false
    -> preserve the existing broadcast behavior
client_local = true
    -> deliver only to the matching instance associated with the
       originating client
```
The default value would be `false`, so existing keybindings and plugins would
continue to behave as before.
## Keybind pipes
For `MessagePlugin`, the option could be specified in KDL as follows:
```kdl
bind "Alt l" {
    MessagePlugin "vim-tab-conditional" {
        name "right"
        client_local true
    }
}
```
When `client_local` is `true`, the message would be delivered only to the
destination plugin instance associated with the client that executed the
keybinding.
The destination instance associated with another client would not receive the
message.
If the option is omitted, the existing broadcast behavior would be preserved:
```kdl
bind "Alt l" {
    MessagePlugin "vim-tab-conditional" {
        name "right"
    }
}
````
## Plugin-to-plugin pipes
The same option could be added to `MessageToPlugin`.
For example:
```rust
let message = MessageToPlugin::new("message-name")
    .with_plugin_url("destination-plugin")
    .client_local();
pipe_message_to_plugin(message);
```
When `client_local` is `true`, the pipe would be delivered only to the
destination plugin instance associated with the same client as the source
plugin instance.
When `client_local` is `false`, the existing broadcast behavior would be
preserved.
## CLI pipes

A CLI pipe is sent through a short-lived CLI client rather than an attached
interactive client. Therefore, its client ID should not be interpreted as the
client ID of a particular attached client whose plugin instance should receive
the pipe.

For this reason, I do not propose applying `client_local` to CLI pipes.

CLI pipes would be delivered to matching plugin instances associated with
currently connected clients, while instances associated with disconnected
clients would be excluded.
## Alternative default behavior
I expect client-local delivery to be the desired behavior for most keybind
pipes and many plugin-to-plugin pipes. Broadcasting to every client-specific
instance seems likely to be a less common use case.
An alternative design would therefore make client-local delivery the default
and provide an explicit option for delivery to all client-specific instances:
```text
all_clients = false
    -> deliver only to the matching instance associated with the
       originating client
all_clients = true
    -> deliver to all matching client-specific instances
```
This would arguably provide a more natural default behavior, but it would
change the behavior of existing keybindings and plugins that do not specify an
option.
For compatibility, my primary proposal is currently the `client_local` option
with a default value of `false`. However, I would also appreciate feedback on
whether client-local delivery should instead become the default behavior.
## Relationship to #4094
PR #4094 proposed exposing the source `client_id` and `pane_id` in
`PipeMessage`, allowing receiving plugins to inspect the origin of a pipe.
The `client_local` proposal in this issue is related to #4094, but it addresses
a different part of pipe handling: selecting which plugin instances receive
the pipe.
With the approach proposed in #4094, the pipe can be delivered to all matching
instances, and each receiving plugin can inspect the source client ID and
decide whether to process it.
With `client_local`, the server would instead deliver the pipe only to the
matching plugin instance associated with the originating client.
This avoids requiring every receiving plugin to implement its own filtering
based on the source client ID.

yamam:
## Summary
This PR adds optional client-local routing for keybind and
plugin-to-plugin pipes.
Closes #5392.
## Changes
- Add a `client_local` option to `MessagePlugin` keybinds.
  ```kdl
  bind "Alt l" {
      MessagePlugin "vim-tab-conditional" {
          name "right"
          client_local true
      }
  }
  ```
- Add `client_local` to `MessageToPlugin`.
  ```rust
  MessageToPlugin::new("message-name")
      .with_plugin_url("destination-plugin")
      .client_local()
  ```
- When enabled, deliver the pipe only to the matching plugin instance
  associated with the originating client.
- Preserve the existing broadcast behavior when `client_local` is omitted or
  set to `false`.
- Exclude plugin instances associated with disconnected clients from CLI pipe
  delivery.
- Treat `client_local` as a routing option rather than part of the destination
  plugin configuration.
## Compatibility
`client_local` defaults to `false`, so existing keybindings and plugins retain
their current broadcast behavior.
CLI pipes do not support `client_local` because they are sent through a
short-lived CLI client rather than a particular attached client.


yamam:
While testing the disconnected-client filtering, I found an additional issue affecting CLI pipes.

When a CLI pipe targets only plugin instances associated with detached clients, the filtering introduced by this PR removes all entries from `pipe_messages`. In that case, `WasmBridge::pipe_messages` has nothing to process and no `UnblockCliPipeInput` is emitted, so the short-lived `zellij pipe` client waits indefinitely.

This is fixed by releasing the CLI pipe directly when the filtered message list is empty:

- keep excluding disconnected plugin instances from CLI pipe delivery
- send `UnblockCliPipeInput` when there are no connected targets
- call `WasmBridge::pipe_messages` only when there are messages to process
- add a regression test covering a CLI pipe with no connected plugin targets

This prevents CLI pipes from hanging when the target plugin instances belong only to detached clients.

PR #5401, opened 2026-07-23
fix: preserve background transparency upon receiving OSC 11 sequence

adillari:
This solves for the `pane_default_bg/pane_default_fg` values being used for two distinct purposes, being both pane colors and capturing/storing OSC 10/11 sequences. To keep transparency intact we should catch OSC 10/11 in a separate variable.

This is an issue when running zellij in a transparent terminal emulator, with any program inside of it that sends an OSC 10/11 sequence (Starship prompt in my case).

main currently
Screenshot From 2026-07-23 10-20-48

main with patch
Screenshot From 2026-07-23 10-17-33

Issue created here:
https://github.com/zellij-org/zellij/issues/5402

See also:
https://github.com/zellij-org/zellij/pull/4737
https://github.com/zellij-org/zellij/commit/f77cdec298576164f34baf0456e4b986bde269e6

### DIFF
--- a/zellij-server/src/panes/grid.rs
+++ b/zellij-server/src/panes/grid.rs
@@ -749,6 +749,8 @@ pub struct Grid {
     /// Pane-scoped override for the default background colour. Same
     /// invariant as `pane_default_fg`.
     pub pane_default_bg: Option<(u8, u8, u8)>,
+    pub osc_default_fg: Option<(u8, u8, u8)>,
+    pub osc_default_bg: Option<(u8, u8, u8)>,
     pub plugin_highlights: HashMap<u32, Vec<(String, CompiledHighlight)>>,
     // key: plugin_id (u32), inner vec: (pattern, compiled) pairs
     pub hover_position: Option<Position>, // pane-relative cursor cell; None when outside pane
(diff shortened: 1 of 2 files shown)


adillari:
@imsnif hello, I must request workflow approval for this PR- may I have that? This is a fairly localized bugfix. any review would also be much appreciated. I know the queue is long- thanks for any attention here!

imsnif:
Hey, first thing: as a policy we do not accept PRs that have bots from commercial companies that do not sponsor the project in the commit log.

Otherwise: I'm not sure I understand this change. Why do we need an additional attribute? What's the differentiator?

adillari:
@imsnif
the pane_default_fg/bg vars currently are used to store colors from 2 separate origins which is the problem:

1: pane colors originating from Zellij itself
2: OSC sequence from programs running inside zellij (this is presumably to answer OSC queries later)

The problem is that just because a program, like in my case Starship prompt, sends an OSC sequence broadcasting what background color, that doesn't mean Zellij should change the pane color (destructively in a way that breaks effects like transparency).

It only needs to be stored to answer OSC queries later. Seperating out these 2 things into distinct variables solves this. It's the reason the transparency is broken in new versions of zellij for me.

As far as the bot policy, very sorry! I did not know about that. If you are understanding the above explanation and are interested in having this change should I close this PR and reopen a new one done manually? I noticed this issue only after https://github.com/zellij-org/zellij/commit/f77cdec298576164f34baf0456e4b986bde269e6 merged.

Thank you for your attention!

imsnif:
Hey, I'm sorry - but I'm afraid this is not true. OSC11 is exactly there to set the background of a terminal pane. You can try it in+out of Zellij and the result will be the same: `printf '\033]11;#1e1e2e\007'`. I'm not saying there is no bug introduced by that specific commit, but it might be more particular. Would you like to investigate further?

adillari:
@imsnif Sure thing, I'll investigate further. The issue for me is the broken transparency either way. zellij is only setting opaque-colored blocks, from OSC or otherwise.

I'll update here after some further research/investigation, thank you very much!

PR #5403, opened 2026-07-24
fix(stack): scrollback truncation when adding a stacked pane in list mode

divens:
When `stacked_pane_list` is on and the user adds a stacked pane, the classic in-grid stack path first collapses the pre-existing pane to `Fixed(1)` and then `focus_pane` fires `reapply_pane_frames`, which sends `resize_pty!` to the collapsed pane at 1 row. The stack-list groupify pass (invoked later during `render`) then extracts the pane to `suppressed_panes` and resizes it back to the visible size, issuing a second `resize_pty!`. The two rapid WINCH events cause the shell (nushell, pwsh, bash, …) inside the pane to redraw its prompt twice, overwriting previously-visible rows of output. The grid data is preserved in `lines_above`, but the visible viewport ends up mostly blank — the user sees only the pre-existing rows that had already scrolled off, giving the impression of a truncated buffer.

Sync stacked-pane-list mode right after the classic mutation, before `focus_pane`. `groupify_all` moves the collapsed member into `suppressed_panes` first, so the subsequent `reapply_pane_frames` (which only iterates `tiled_panes`) never sends a 1-row `resize_pty!` to that shell. The pane is then resized once to the visible size via `resize_stack_list_hidden_members`.

Regression test asserts that after adding a stacked pane in this mode, the pre-existing pane is in `suppressed_panes` and the new visible member has non-fixed row dimensions — proving groupify happened before any classic `reapply_pane_frames` step.


imsnif:
Good find, thanks! (I removed the entry from the changelog since this has to do with a yet-unreleased feature)

PR #5406, opened 2026-07-24
feat: expose per-pane bell state and add list-panes --bell filter

nerdrew:
Add a `has_bell_notification` field to `PaneInfo` so the pending-bell state of each pane is visible through the plugin API (GetPaneInfo, PaneUpdate/PaneManifest), the `list-panes` JSON/table output, and persisted session info.

Also add a `-b/--bell` flag to `zellij action list-panes` to filter to only panes with a pending bell, and a `BELL` column to the `--state` table. KDL session decoding tolerates the field's absence for backwards compatibility with older dumps.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 
 ## [Unreleased]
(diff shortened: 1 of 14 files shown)


nerdrew:
Follow up feature to https://github.com/zellij-org/zellij/pull/4781

nerdrew:
@imsnif Do you have any thoughts on this feature?

Issue #5256, opened 2026-06-13
Extra empty line is inserted when using ANSI scrollback output


yamam:
# 2. Issues with the Zellij UI / behavior / crash

## Description

When using ANSI-enabled scrollback output, an extra empty line is inserted before the viewport content.

This affects both:

```bash
zellij action edit-scrollback --ansi
````

and:

```bash
zellij action dump-screen --full --ansi
```

## Reproduction condition

This reproduces when the viewport is not yet fully filled.

In this state, the first line of the viewport is empty. Once the viewport becomes fully filled, that empty line disappears and the issue no longer reproduces.

## Steps to reproduce with dump-screen

1. Start a new Zellij session or pane.

2. Make sure the viewport is not fully filled yet.

3. Run:

   ```bash
   zellij action dump-screen --full --ansi --path /tmp/zellij-dump.txt
   ```

4. Inspect the output:

   ```bash
   cat -v /tmp/zellij-dump.txt
   ```

5. An extra empty line appears before the viewport content.

   When inspected with `cat -v`, the first line contains only:

   ```text
   ^[[m
   ```

   This corresponds to the ANSI reset sequence `ESC[m`.

## Steps to reproduce with edit-scrollback

1. Start a new Zellij session or pane.

2. Make sure the viewport is not fully filled yet.

3. Run:

   ```bash
   zellij action edit-scrollback --ansi
   ```

4. The editor opens the scrollback content.

5. An extra empty line appears before the viewport content.

   The first line contains only the ANSI reset sequence:

   ```text
   ESC[m
   ```

   Depending on the editor, this may be displayed as:

   ```text
   ^[[m
   ```

## Expected behavior

No extra empty line should be inserted before the viewport content.

When there is no scrollback above the viewport, ANSI-enabled scrollback output should start directly with the viewport content.

## Actual behavior

An extra empty line is inserted before the viewport content when the viewport is not fully filled.

This happens with both:

```bash
zellij action edit-scrollback --ansi
```

and:

```bash
zellij action dump-screen --full --ansi
```

## Cause

The ANSI dump path appends the ANSI reset sequence `ESC[m` even when no cells were emitted for the scrollback area.

As a result, an otherwise empty scrollback dump becomes non-empty because it contains only `ESC[m`. When the full output is assembled, this causes an extra separator newline to be inserted before the viewport content.

PR #5409, opened 2026-07-24
fix(grid): keep spaces before a wrapped line continuation when dumping

### LINKED ISSUE #5259: Spaces disappear in scrollback output when a line wraps
## Description

Spaces can disappear from scrollback output when a line wraps.

For example, if a logical line contains spaces near the wrap boundary, opening the scrollback in an editor or dumping the screen can drop some of those spaces.

This affects scrollback output such as:

```bash
zellij action edit-scrollback
````

and:

```bash
zellij action dump-screen
```

It also affects the ANSI variants:

```bash
zellij action edit-scrollback --ansi
```

and:

```bash
zellij action dump-screen --ansi
```

## Example

With a narrow pane, a single logical line can be visually wrapped across two
physical terminal rows.

For example, the original logical line is:

```text
a   b     c
````

If the pane is 10 columns wide, it can be displayed like this:

```text
a   b     
c
```

In this case, the spaces at the end of the first physical row are not just
terminal padding. They are part of the original logical line before the wrapped
continuation.

Expected scrollback output:

```text
a   b     c
```

Actual scrollback output:

```text
a   bc
```
The spaces before the wrapped continuation are dropped.

Rohitjangra7370:
Fixes #5259

## Problem

`dump_screen` and `dump_screen_with_ansi` trim trailing spaces from **every physical
row**. That is right for a row that ends a logical line (the row is padded out to the
pane width), but wrong for a row that wraps: the continuation row is appended with no
newline, so those trailing spaces sit *between* content and are part of the logical
line.

With a 10 column pane, the logical line `a   b     c` wraps as:

```text
a   b␣␣␣␣␣
c
```

and dumps as `a   bc` instead of `a   b     c`. This affects `zellij action dump-screen`
and `zellij action edit-scrollback`, both with and without `--ansi` (the ANSI path has
the same per-row truncation via `last_non_space`).

## Changes

Both macros now look ahead one row and only trim the padding when the current row
actually ends a logical line — i.e. when the next row is canonical, or there is no next
row. A row followed by a non-canonical (continuation) row is emitted in full.

The fix is applied to both `dump_screen!` and `dump_screen_with_ansi!` so the `--ansi`
variants behave the same.

## Tests

Added two tests in `grid_tests.rs`:

- `dump_screen_keeps_spaces_before_a_wrapped_continuation` — the reproduction from the
  issue. **Fails without the change** with exactly the reported output:
  ```
  assertion `left == right` failed
    left: "a   bc"
   right: "a   b     c"
  ```
- `dump_screen_still_trims_padding_at_end_of_a_logical_line` — guards the existing
  behaviour, so the padding trim isn't simply removed and every dumped line padded out
  to the pane width.

## Verification

- `cargo test -p zellij-server --lib` — 1200 passed, 0 failed (run twice, deterministic)
- baseline on `main` for comparison — 1198 passed, 0 failed
- `cargo fmt --check -p zellij-server` — clean
- `cargo clippy -p zellij-server --lib` — no new warnings in the changed macros (the
  remaining warnings are pre-existing elsewhere in the file)

## Note

I deliberately kept this scoped to the per-row trimming. `dump_screen` joins
`lines_above` and the viewport with an unconditional `\n`, so a logical line spanning
that boundary is still split — that looked like separate behaviour (closer to #5256), so
I left it alone rather than widening the change.

### DIFF
--- a/zellij-server/src/panes/grid.rs
+++ b/zellij-server/src/panes/grid.rs
@@ -392,15 +392,23 @@ macro_rules! dump_screen {
     ($lines:expr) => {{
         let mut is_first = true;
         let mut buf = String::with_capacity($lines.iter().map(|l| l.len()).sum());
+        let mut lines = $lines.iter().peekable();
 
-        for line in &$lines {
+        while let Some(line) = lines.next() {
             if line.is_canonical && !is_first {
                 buf.push_str("\n");
             }
             let s: String = (&line.columns).into_iter().map(|x| x.character).collect();
-            // Replace the spaces at the end of the line. Sometimes, the lines are
-            // collected with spaces until the end of the panel.
-            buf.push_str(&s.trim_end_matches(' '));
+            // A row followed by a wrapped continuation row ends mid-logical-line, so its
+            // trailing spaces are content rather than padding and must be kept. Only the
+            // last row of a logical line gets the padding trimmed. Sometimes, the lines
+            // are collected with spaces until the end of the panel.
+            let wraps_onto_next_row = lines.peek().map_or(false, |next| !next.is_canonical);
+            if wraps_onto_next_row {
+                buf.push_str(&s);
+            } else {
+                buf.push_str(&s.trim_end_matches(' '));
+            }
             is_first = false;
         }
         buf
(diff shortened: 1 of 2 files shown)


rhuss:
thanks! It would be great if this PR (or https://github.com/zellij-org/zellij/pull/5260, which tackles the very same issue) could be integrated to fix this annoying scrollback buffer issue.

PR #5412, opened 2026-07-24
Add missings sugar functions `NestedListItem::*color*` in plugin API

### LINKED ISSUE #5411: `NestedListItem::color_substring` does not exist

2. I need a new API

Zykino:
Fix #5411


imsnif:
Thank you! If you'd like to also find a way to keep them synced automatically as you mentioned in the issue, by all means. Just merging this now to unstuck you.

Issue #1906, opened 2022-11-04
Feature request: allow floating panes to go full screen


xJonathanLEI:
Currently it's only possible for embedded panes to go fullscreen mode. It would be nice if floating panes can do this too.

Sometimes I need extra screen space for the floating pane, but for now I have to:

- embed it
- go full screen
- exit full screen
- float it

Not to mention that I lose the size and position setting for the floating pane after doing this.

imsnif:
I'd love to see this too. Are you interested in working on this by any chance?

xJonathanLEI:
I might, but it's not exactly my priority right now :/

naosense:
Hey @imsnif maybe I can help, if @xJonathanLEI is not working on it. I need some hints, perhaps

xJonathanLEI:
@naosense Thanks for asking. Please feel free to. I'm not working on this at the moment.

imsnif:
Hey @naosense - sorry for the delayed response.

I think the best way to approach this would be to start by mapping the behavior from the user's perspective. Namely: what should happen when a floating pane becomes full screen? If in such a case the user wants to focus other panes, what should happen? Should it be possible? Should it work the same way as focusing other floating panes when one of them is not full screen (eg. move the focus in the relevant direction)? When adding  new floating panes and one of them is in full screen, should the fullscreen be unset like with tiled panes?

I'd gather answers to all of these questions, then do some investigation to see how it works with tiled panes and fullscreen in various situations (no need to look at the code, just play around with it and see what happens). Then write everything down and start making an implementation plan (at that stage I'd be happy to offer more help if you want it).

malikbenkirane:
I'm excited to share my user perspective on this feature and contribute to the discussion. I hope my input brings positive energy to move the conversation forward!

Hey @imsnif - let's follow up on this 😇 

When the pane goes full screen, I intend to give more space to that embedded pane. So, as a user, I would like to have this "zoom out" on the pane I'm currently focusing on (getting more space for a moment) and then "zoom back" to get back to my previous view (how the Zellij's space organization was before my "zoom-out"). It is always a better workflow, in my opinion, when the number of available views stored in the mind map of my workspace is limited to a few additional thoughts (as, in general, two is already enough when I have more to focus on, I am also considerate regarding neurodivergent users and I plaid in favor of more inclusive software).

[...]

PR #5417, opened 2026-07-27
Nested session handling

imsnif:
nested-sessions-demo

This adds built-in handling of nested Zellij sessions. When Zellij detects another Zellij session running inside a pane, it will ask the user whether they would like to "fullscreen" this session - meaning that the session will take up the entire screen and provide a "breadcrumbs" navigation on the top tab-bar - or "descend" into it - meaning that any keybinding pressed will be sent into the nested session. Users can ascend back from the session into the host session with `Ctrl g` + `o` + `[` or `]` (hints appear on screen).

When choosing option 2 (descending into the session), other than using the session navigation (`[` and `]` mentioned above in the session mode), when navigating into the session, it will be focused. Navigating away from it will focus the nearest pane depending on direction.

As a by-product of this change, floating panes can now also be directly fullscreened (as normal panes, with `Ctrl p` + `f`).


thiblahute:
Looks like this fixed https://github.com/zellij-org/zellij/issues/1906 "Feature request: allow floating panes to go full screen" -- right?

imsnif:
> Looks like this fixed #1906 "Feature request: allow floating panes to go full screen" -- right?

Yep, thanks for pointing this out!

NovaViper:
Hey does this address #387 ?

PR #2611, opened 2023-07-06
Use CARGO_TARGET_DIR when getting plugins

poliorcetics:
Using `CARGO_MANIFEST_DIR` is brittle when accessing the cargo target directory since the latter's path could be overriden by a user trying to compile `zellij`.

PR #5426, opened 2026-07-30
fix(build): remove manpage, honor CARGO_TARGET_DIR, add cargo argument passthrough, remove clippy

imsnif:
This performs some fixes and cleanups to our build system:

1. `cargo x install` will now honor CARGO_TARGET_DIR (rather than hardcoding `target`)
2. Removed the extremely outdated manpage and its related generation infrastructure
3. Cleaned up stale paths in the deb section of Cargo.toml (I think this section should be removed completely, but it is conceivable that something external depends on it)
4. Removed clippy (we're not using it anyway)
5. Added cargo argument passthrough to `build` and `install`


Zykino:
> `cargo x install` will now honor CARGO_TARGET_DIR (rather than hardcoding target)

In ".cargo/config.toml", `CARGO_TARGET_DIR` is overrided. I think the issue was with `cargo x build` that it looked for its wasm files in the wrong directory.

If you really want to honor it, ".cargo/config.toml" should be reverted

imsnif:
Hey @Zykino - thanks for pointing this out! I did not notice this. However, I tested and I think `CARGO_TARGET_DIR` takes precedence... or did I misunderstand the concern?

Zykino:
At home I have it configured (I think in the "~/.config/cargo.toml" file or something). I needed the change in "zellij/.cargo/config.toml" to override my default setting.

Anyway, removing the config in "zellij/.cargo/config.toml" should be OK now. I can try this evening if you want and report back if I have compilation issue with/without it. (If I no longer need it, I think we should remove it anyway.)

imsnif:
That would be great, thank you!

Zykino:
# So before this PR on main
And after removing the 2 lines to get the old error message:
```toml
[build]
target-dir = "target"
```

Extract of the last error:
```log
error: couldn't read `/mnt/Data/Code/zellij/zellij-utils/../target/wasm32-wasip1/debug/mobile.wasm`: No such file or directory (os error 2)
   --> zellij-utils/src/consts.rs:148:17
    |
148 | /                 include_bytes!(concat!(
149 | |                     env!("CARGO_MANIFEST_DIR"),
150 | |                     "/../target/wasm32-wasip1/debug/",
151 | |                     $plugin
152 | |                 ))
    | |__________________^
...
174 |               add_plugin!(assets, "mobile.wasm");
    |               ---------------------------------- in this macro invocation
    |
    = note: this error originates in the macro `include_bytes` which comes from the expansion of the macro `add_plugin` (in Nightly builds, run with -Z macro-backtrace for more info)

error: could not compile `zellij-utils` (lib) due to 13 previous errors
Error: failed to build '.'

Caused by:
    command exited with non-zero code `/home/zykino/.rustup/toolchains/stable-x86_64-unknown-linux-gnu/bin/cargo build --profile dev-opt`: 101
```

The zellij/target local folder exist but is empty (I suspect xtask create it around the error time)
```sh
$ ls target/

```

# Now with this PR but the `target-dir = "target"` removed
We still have the same issue as before:
```log
error: couldn't read `/mnt/Data/Code/zellij/zellij-utils/../target/wasm32-wasip1/debug/mobile.wasm`: No such file or directory (os error 2)
   --> zellij-utils/src/consts.rs:148:17
    |
148 | /                 include_bytes!(concat!(
149 | |                     env!("CARGO_MANIFEST_DIR"),
150 | |                     "/../target/wasm32-wasip1/debug/",
151 | |                     $plugin
152 | |                 ))
    | |__________________^
...
174 |               add_plugin!(assets, "mobile.wasm");
    |               ---------------------------------- in this macro invocation
    |
    = note: this error originates in the macro `include_bytes` which comes from the expansion of the macro `add_plugin` (in Nightly builds, run with -Z macro-backtrace for more info)
```

## #2611 Fix does not work
```log
error: environment variable `CARGO_TARGET_DIR` not defined at compile time
   --> zellij-utils/src/consts.rs:149:21
    |
149 |                     env!("CARGO_TARGET_DIR"),
    |                     ^^^^^^^^^^^^^^^^^^^^^^^^
...
174 |             add_plugin!(assets, "mobile.wasm");
    |             ---------------------------------- in this macro invocation
    |
    = help: there is a similar Cargo environment variable: `CARGO_TARGET_TMPDIR`
    = note: this error originates in the macro `env` which comes from the expansion of the macro `add_plugin` (in Nightly builds, run with -Z macro-backtrace for more info)
```

I tried more, but couldn't fix it… But I don't understand at which stage the `add_plugin` is called…
Maybe it require a build.rs? But build.rs and xtask are things I don't really understand (since I did not use them myself) so I'm sure there is a fix, not sure how to do it myself.

# This PR as-is
Compile but there are things in the local "target" folder:
```sh
$ ls target/
CACHEDIR.TAG  debug/  dev-opt/  wasm32-wasip1/
```

Note: For all tests, to build I set the toolchain's channel to "stable. It should not change anything for this test. I do it to avoid downloading a specific toolchain (trying to preserve my disk usage).

imsnif:
Thanks for checking @Zykino - I think I'm missing a bit of context though. So first off: you say with this PR as is without any changes it compiles for you and places some stuff in the target folder... this sounds correct to me, no? Otherwise - what problem are we trying to solve?

WeepingClown13:
fyi, I also ran into this issue, with a plain `cargo build --release --bin zellij --offline`. I thought it was maybe because I didn't run `cargo x build`, but looks like that is not the case.

imsnif:
@WeepingClown13 - what issue are you running into exactly? Could you walk me through?

Zykino:
Yesterday was busy at work and I slacked a bit to come back to you today.

> this PR as is without any changes it compiles for you and places some stuff in the target folder... this sounds correct to me, no? Otherwise - what problem are we trying to solve?

`CARGO_TARGET_DIR` or the same as global user config file in "~/.cargo/config.toml":
```toml
[build]
target-dir = "/path/to/global/Cargo/folder"
```
tell cargo "The user want all its compilation to be done in '/path/to/global/Cargo/folder'". As a user I like having all my project compile on the same folder:
If I have multiple plugins that all use `zellij-tile` on the same version, cargo can reuse the compilation of this dependency when I switch between them. I save on compilation time, disk space (since I don't have multiple "target" folder with the same crate compiled in each).

Now zellij-tile is quite small and specific, but centralizing my compile folder I get to mutualise all the crates compilation. `serde`, thiserror` or `tokio` are much more often used and each project using any crate in common is a win.


Maybe I did not understood at all what you meant by this and got carried away?
> * cargo x install will now honor CARGO_TARGET_DIR (rather than hardcoding target)

Last note on my section "#2611 Fix does not work" that I forgot:
Sometimes rustc tell me to use `std::env::vars()` instead of `env!()`, but it does not work in the macro… and I don't really know how rust macro work. Especially I saw that `add_plugin!` is called in `lazy_static!` which I suspect makes my understanding even harder.

imsnif:
Hey @Zykino - thanks for the explanation. I think in your case if you symlink the target folder in the repo to the global target folder, it should do the trick and you can keep working in this way.

As for @WeepingClown13 - if I understood your issue correctly, I think it is fixed by #5433. This should allow you to move forward without the bundled plugins (or any internal attempt to access them, and thus the error should be preventes). Let me know if it doesn't work or if I misunderstood your issue.

WeepingClown13:
@imsnif sorry, I've been quite busy and couldn't followup. The issue I was getting was the same "No such file or directory" from above comments. I'll try the new fixes later when I have a bit more breathing room and update.

PR #5428, opened 2026-07-31
Support the Kitty Image Protocol

imsnif:
kitty-pr-images

This implements the kitty graphics protocol in Zellij - providing image support for terminals that do not support Sixel.

While this has been a long requested feature, I have put it off because I am not a believer in this protocol and I was hoping the ecosystem will converge on Sixel. A protocol which I am also not a big fan of, but that I view as the lesser of two evils.

The last straw for me was that [recent events](https://github.com/alacritty/alacritty/issues/910#issuecomment-4905562551) have shown me that Alacritty (my favorite terminal) - the one I hold in the highest regard as the gateway to the ecosystem - will not be implementing either one. I respect the decision of the maintainers even though I disagree with it. While personally I don't use images in the terminal, I believe in this day and age not providing users with image support is selling them short.

Practical implications: this is a full protocol implementation*, meaning images should "just work" in any terminal that supports this protocol while inside Zellij. No need for passthroughs or any special handling. Some applications hard-code image behavior to Zellij environment variables (eg. Yazi) and those will have to be adjusted upstream, but otherwise this puts this issue to bed.

This also provides an escape-hatch that disables support for this protocol even in supported terminals: `support_kitty_image_protocol false` in the config or through the CLI flag.

I thank the community for your patience.

*minus the heavier animations and unicode placeholders which might follow later


sxyazi:
👋 Hi, Yazi maintainer here. 

What great news! Happy to see Zellij adding Kitty graphics protocol support and this will be a huge QoL improvement for people like me who use images on a daily basis but have terminals without Sixel. Thank you so much for implementing this!

I'm looking into adapting Yazi to it and had a question: if the underlying terminal doesn't support KGP, will Zellij still report KGP as supported, like it currently does with Sixel? In other words, does Zellij report its own capabilities, or the capabilities of both Zellij and the underlying terminal?

Previously, Yazi always used Sixel in Zellij, since that was the only protocol Zellij supported. Now that Zellij supports both Sixel and KGP, how can I tell which one the terminal actually supports and choose the right one?

In tmux, I use its passthrough sequences to probe the outer terminal's capabilities to figure that out, but I'm not sure if Zellij has something similar. Any ideas would be greatly appreciated!

dayioh:
Thank you so much for that ♥️

imsnif:
Hey @sxyazi - thanks for pointing this out! I issued a fix for the Sixel reporting issue: https://github.com/zellij-org/zellij/pull/5432 and we should now only advertise (and indeed, render) Sixel if the underlying terminal itself also supports it. This is already the case with the kitty graphics protocol in this PR. So once released, you should be able to query both and get a definitive answer (at the time of query, this might change if for example the user detached from an active session and attached from a different terminal).

I don't know about you, but I personally don't always manage to keep up with issues/PRs in this repo. For this reason, I might miss integration issues such as these. Please see this as a personal invitation to tag me whenever you have an integration issue with Zellij. You maintain a popular tool and I would like to give you and your users close attention.

sxyazi:
Thanks for the quick fix on the Sixel report, it's a much friendlier and more reliable way than passthrough!

I've done the adaptation work on the Yazi side in https://github.com/sxyazi/yazi/pull/4216, really appreciate your help and your love for the open source community. You've done an awesome job with Zellij, it's one of my favorite tools and one I recommend to people the most :)

luccahuguet:
♥️🤠🙏🤩

amakeenk:
Could you please tell me when the new release, which includes these changes, is expected?

mati865:
There is watch button on GH which lets you configure custom notifications
obraz

amakeenk:
> There is watch button on GH which lets you configure custom notifications

Thank you very much for the tip; I’ve had this feature enabled for a long time, but the question was about something else.

PR #5429, opened 2026-08-01
fix(mouse): forward scroll modifier state to applications

### LINKED ISSUE #5306: Mouse modifier state for scroll events is not properly forwarded to applications running inside panes
**Basic information**

`zellij --version`: 0.44.3

`stty size`: 14 96

`uname -av` or `ver`(Windows): Linux terminal 6.14.0-37-generic #37-Ubuntu SMP PREEMPT_DYNAMIC Fri Nov 14 22:10:32 UTC 2025 x86_64 x86_64 x86_64 GNU/Linux

## Issue description
When creating a crossterm/ratatui based app and running in zellij, I cant make use of scroll + ctrl/alt/shift modifier events. 
May be the same issue as #2014.

Thanks for a great tool!

## Minimal reproduction
Provided crossterm-based script shows mouse modifiers when invoked directly in alacritty and konsole. However, when running the test in a zellij pane inside those terminal emulators, I do not see the modifiers when scrolling. I get no events when holding ctrl/alt and when holding shift I see a scroll event but it doesnt carry the shift modifier. 

## Other relevant information
```rust
#!/usr/bin/env cargo

//! ```cargo
//! [dependencies]
//! crossterm = "0.29.0"
//! ```
use std::io;

use crossterm::{
    event::{
        self, DisableMouseCapture, EnableMouseCapture, Event, KeyCode, KeyEvent, KeyModifiers,
        MouseEventKind,
    },
    execute,
    terminal::{disable_raw_mode, enable_raw_mode},
};

fn main() -> io::Result<()> {
    enable_raw_mode()?;
    execute!(io::stdout(), EnableMouseCapture)?;

    println!("scroll with and without modifiers; press Ctrl+c quit");

    loop {
        let event = event::read()?;
        match event {
            Event::Key(KeyEvent {
                code: KeyCode::Char('c'),
                modifiers,
                ..
            }) if modifiers.contains(KeyModifiers::CONTROL) => break,

            Event::Mouse(mouse) => match mouse.kind {
                MouseEventKind::ScrollUp
                | MouseEventKind::ScrollDown
                | MouseEventKind::ScrollLeft
                | MouseEventKind::ScrollRight => {
                    dbg!(mouse.kind, mouse.modifiers);
                    println!("\r\n");
                }
                _ => {}
            },
            _ => {}
        }
    }

    execute!(io::stdout(), DisableMouseCapture)?;
    disable_raw_mode()?;
    println!();

    Ok(())
}

```

Rohitjangra7370:
Fixes #5306

## Problem

Wheel events lose their modifier state on the way to the application running in a
pane. `mouse_scroll_up_signal` / `mouse_scroll_down_signal` took only a `Position`,
so they emitted a bare wheel button value — `64`/`65` for SGR, `` ` ``/`a` for
x10/utf8 — with no modifier bits. Every other mouse signal encodes them through
`mouse_buttons_value_sgr` / `mouse_buttons_value_x10` (shift `0x04`, alt `0x08`,
ctrl `0x10`).

The result is what the issue reports: a crossterm/ratatui app asking for
shift+scroll sees a plain scroll with no modifier.

## Changes

- `mouse_scroll_up_signal` / `mouse_scroll_down_signal` now take the `&MouseEvent`
  and OR the modifier bits onto the wheel value. The bare values are unchanged when
  no modifier is held, and the x10 byte and SGR button number are otherwise
  untouched.
- The event is threaded through `handle_scrollwheel_up` / `_down`, which already had
  it in scope at the dispatch site. Position is still reported relative to the pane,
  mirroring the existing `event_for_pane` pattern used for clicks.
- The public `Tab::handle_scrollwheel_up` / `_down` keep their `&Position` signature
  and synthesize a modifier-free event, so `screen.rs` and existing tests are
  unaffected.

## Scope

Ctrl+scroll is deliberately left alone. It is claimed by zellij for pane resizing,
and dropped entirely when `mouse_scroll_resize` is off, so it never reaches this
code — that part of the issue ("no events when holding ctrl") is a separate
behaviour decision I did not want to make unilaterally. This PR fixes the shift/alt
case the reporter demonstrated.

I also noticed while here that `mouse_buttons_value_x10` returns `68`/`69` for
wheel up/down, while the scroll path uses `` ` ``/`a` (96/97, i.e. `32 + 64`). Those
disagree, so I deliberately did **not** route scroll through the shared helper —
that would have silently changed the emitted bytes. Flagging it as possibly worth a
separate look.

## Tests

Two tests in `grid_tests.rs`:

- `scroll_signals_forward_modifier_state_to_the_application` — shift+wheel-up must
  encode `68`, alt+wheel-down must encode `73`. **Fails without the change:**
  ```
  left: Some("\u{1b}[<64;1;1M")
  right: Some("\u{1b}[<68;1;1M")
  ```
- `scroll_signals_are_unchanged_without_modifiers` — guards the existing encoding so
  a plain scroll still reports the bare `64`/`65`.

## Verification

- `cargo test -p zellij-server --lib` — 1445 passed, 0 failed
- `cargo fmt --check -p zellij-server` — clean
- `cargo clippy -p zellij-server --lib` — no new warnings from these changes

### DIFF
--- a/zellij-server/src/panes/grid.rs
+++ b/zellij-server/src/panes/grid.rs
@@ -3732,11 +3732,27 @@ impl Grid {
             },
         }
     }
-    pub fn mouse_scroll_up_signal(&self, position: &Position) -> Option<String> {
+    /// The modifier bits shared by the x10/utf8 and SGR mouse encodings.
+    fn mouse_modifiers_value(event: &MouseEvent) -> u8 {
+        let mut value = 0;
+        if event.shift {
+            value |= 0x04;
+        }
+        if event.alt {
+            value |= 0x08;
+        }
+        if event.ctrl {
+            value |= 0x10;
+        }
+        value
+    }
+    pub fn mouse_scroll_up_signal(&self, event: &MouseEvent) -> Option<String> {
+        let position = &event.position;
+        let modifiers = Self::mouse_modifiers_value(event);
         match (&self.mouse_mode, &self.mouse_tracking) {
             (_, MouseTracking::Off) => None,
             (MouseMode::NoEncoding | MouseMode::Utf8, _) => {
-                let mut msg: Vec<u8> = vec![27, b'[', b'M', b'`'];
+                let mut msg: Vec<u8> = vec![27, b'[', b'M', b'`' | modifiers];
                 msg.append(&mut utf8_mouse_coordinates(
                     position.column() + 1,
                     position.line() + 1,
@@ -3745,7 +3761,8 @@ impl Grid {
             },
             (MouseMode::Sgr, _) => {
                 let mouse_event = format!(
-                    "\u{1b}[<64;{:?};{:?}M",
+                    "\u{1b}[<{:?};{:?};{:?}M",
+                    64 | modifiers,
                     position.column.0 + 1,
                     position.line.0 + 1
                 );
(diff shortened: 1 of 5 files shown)


cengiz-io:
@Rohitjangra7370 hey, ran into the bug this fixes. pi coding agent loses alt+scroll inside zellij, see #5306. 

read through the diff, OR-ing the modifier bits onto the wheel value matches how clicks already do it and the bit math checks out. 

however it's conflicting with main now so it'll need a rebase. 

also the new tests only cover the SGR encoding path, the x10/utf8 branch got the same change but nothing checks it. 

happy to help with the rebase or the missing test if you want.

PR #5441, opened 2026-08-05
Native web mobile UI

imsnif:
This swaps the yet-unreleased built-in mobile plugin - replacing it with a dedicated web mobile UI. I felt this is much nicer and performs better.

The UI includes native web session switching, pane switching, a single-pane mode (where the UI only shows the focused pane to save space, allowing to switch the focus through the mobile menu), a fit to screen toggle and a light/dark theme switch toggle.

This is detected according to the browser client and can be disabled with the "Switch to desktop" option in the hamburger menu.


Zentrik:
wow cool thanks, I'm going to try this out. I had hacked a bunch of improvements upon the old version to make it work better on mobile, but this looks much nicer.

Zentrik:
If I may provide some feedback, some initial thoughts
1. I find the default font size far too large, even zoomed out as far as it lets me this feels larger than I would like. I believe I had the same issue with the previous UI though.
2. It would be nice if desktop mode didn't have a persistent mobile UI button. On android I use a keyboard that provides the control, escape and arrow keys, so I want to hide the keyboard bar. I also found the session top bar useless, so I want to hide that as well given the limited screen estate I have on my phone. Thus I use desktop mode, but then I have this floating mobile UI button.
3. Whenever I swipe on my screen to scroll, the keyboard pops up. I'd prefer if it didn't

jurriaan:
@Zentrik agree!

It would indeed be nice if the font size was configurable, by default it is too large. 

Also it seems to reset the font size whenever you in/out of the PWA app. So a combination of configuration + remembering of the "zoom level" would be nice.

PR #5446, opened 2026-08-06
fix(input): realign raw-byte buffer with the keyboard parser to stop stale bytes eating keystrokes

ZrjaK:
## Problem

After a paste (e.g. via a clipboard manager like maccy, or plain bracketed paste), zellij's next keystrokes are lost or garbled: keys go missing, produce stray characters, or drop into vi command mode. Reproduced on current `main`, works on 0.44.3.

Reproduced in both **Ghostty** and the **macOS built-in Terminal.app**, both running `zsh`, on macOS (Apple Silicon), zellij built from `main`.

## Regression

This is caused by #5323, which changed `current_buffer.drain(..)` (drain the whole buffer per event, as in 0.44.x) to `current_buffer.drain(..take)` (drain only the reported `consumed` bytes). Between `drain(..take)` and `InputParser`'s separately-accumulated `self.buf`, the two can drift (see Root cause below), and stale bytes get forwarded with unrelated key events.

Reverting to the 0.44.x `drain(..)` also resolves the paste issue. We kept the more conservative `realign` fix below because `drain(..)` was replaced in #5323 for a reason (splitting input bytes across per-event payloads); `realign` keeps `#5323`'s per-event byte attribution and only trims bytes the parser has already consumed. The original `#5323` mouse-spam scenario (scrolling with an Apple Magic Trackpad on macOS) was not observable in our environment, so we did not validate that path.

## Root cause

`zellij-client/src/stdin_handler.rs` keeps two byte sources:
- `current_buffer` – a raw byte ring that `stdin_loop` and `drain_partial_to_keyboard` accumulate from `StdinAnsiParser` residue, drained by each event's reported `consumed` bytes.
- `InputParser::self.buf` – termwiz's own internal buffer.

`parse_with_consumed` reports `consumed` relative to termwiz's `self.buf` (its length drops), while `stdin_handler` drains the separately-accumulated `current_buffer`. When a read leaves a partial sequence that termwiz holds internally under `maybe_more`, the two buffers can drift: `current_buffer` keeps a small number of bytes termwiz already decoded into events. Those stale bytes then get attached to the raw payload of the *next* `KeyEvent`, desyncing each keystroke from the bytes that produced it.

## How the drift was confirmed

The drift was reproduced in **Ghostty** and the **macOS built-in Terminal.app** (both running `zsh`), pasting via maccy or plain `Cmd/Ctrl+V`. To replicate it:

1. Temporarily add a log line after the termwiz parse in `stdin_loop` (`zellij-client/src/stdin_handler.rs`):

```rust
for (input_event, consumed) in events.into_iter() {
    let take = consumed.min(current_buffer.len());
    log::info!(
        "DRIFT-CHECK pre-drain cb={} consumed={} take={} parser_buf={}",
        current_buffer.len(),
        consumed,
        take,
        input_parser.buffered_len()
    );
    let raw_bytes: Vec<u8> = current_buffer.drain(..take).collect();
    log::info!(
        "DRIFT-CHECK post-drain cb={} parser_buf={}",
        current_buffer.len(),
        input_parser.buffered_len()
    );
    // ... forward KeyEvent ...
}
```

2. Build and run zellij with the `--debug` flag:

```sh
cargo build --release
zellij --debug
```

The client logs to `$TMPDIR/zellij-<UID>/zellij-log/zellij.log` (e.g. `/var/folders/.../T/zellij-501/zellij-log/zellij.log` on macOS).

3. In the zellij session, paste a chunk of text (plain `Cmd/Ctrl+V`, or via a clipboard manager such as maccy).

4. The log shows the stale bytes that would have been forwarded to unrelated key events, e.g.:

```
DRIFT-CHECK pre-drain cb=50 consumed=44 take=44 parser_buf=0
DRIFT-CHECK post-drain cb=6 parser_buf=0
```

A single read delivered 50 bytes; termwiz decoded 44 of them into one event but its `self.buf` is already empty (`parser_buf=0`), while `current_buffer` still holds the 6 bytes that termwiz consumed. Those 6 stale bytes get attached to the raw payload of the next `KeyEvent` (subsequent `cb=1` reads), which is what eats/garbles the following keystrokes.

## Fix
After every parse, realign `current_buffer` to termwiz's internal buffered length so no already-decoded byte is ever forwarded again:

- add `InputParser::buffered_len()` to the vendored termwiz copy
- call `realign_current_buffer` after the termwiz parse in both `stdin_loop` and `drain_partial_to_keyboard`

```rust
fn realign_current_buffer(current_buffer: &mut Vec<u8>, input_parser: &InputParser) {
    let buffered = input_parser.buffered_len();
    if current_buffer.len() > buffered {
        current_buffer.truncate(buffered);
    }
}
```

## Tests

- `cargo test -p zellij-utils --lib` – 426 passed, 2 ignored.
- `cargo test -p zellij-client --lib` – 84 passed; the single failure (`stdin_ansi_parser::tests::open_forward_debug_asserts_on_reentry`) also fails on a clean checkout of `main` and is unrelated to this change.
- `cargo build --release` – builds.
- Manually verified: bracketed paste and maccy-style paste no longer eat subsequent keystrokes.

### DIFF
--- a/zellij-client/src/stdin_handler.rs
+++ b/zellij-client/src/stdin_handler.rs
@@ -224,6 +224,7 @@ pub(crate) fn stdin_loop(
                                 break 'stdin;
                             }
                         }
+                        realign_current_buffer(&mut current_buffer, &input_parser);
 
                         schedule_finalization(
                             &stdin_ansi_parser,
@@ -329,6 +330,18 @@ fn drain_partial_to_keyboard(
         send_input_instructions
             .send(InputInstruction::KeyEvent(input_event, raw_bytes))
             .unwrap();
+    }
+    realign_current_buffer(current_buffer, input_parser);
+}
+
+/// Trim `current_buffer` to the parser's own buffered length so it can
+/// never drift from the parser's internal state: a trailing incomplete
+/// sequence is held by the parser (and mirrored here) until the next
+/// read completes it, while bytes already decoded into events are dropped.
+fn realign_current_buffer(current_buffer: &mut Vec<u8>, input_parser: &InputParser) {
+    let buffered = input_parser.buffered_len();
+    if current_buffer.len() > buffered {
+        current_buffer.truncate(buffered);
     }
 }
 
--- a/zellij-client/src/stdin_handler.rs
+++ b/zellij-client/src/stdin_handler.rs
@@ -340,8 +340,9 @@ fn drain_partial_to_keyboard(
 /// read completes it, while bytes already decoded into events are dropped.
 fn realign_current_buffer(current_buffer: &mut Vec<u8>, input_parser: &InputParser) {
     let buffered = input_parser.buffered_len();
-    if current_buffer.len() > buffered {
-        current_buffer.truncate(buffered);
+    let excess = current_buffer.len().saturating_sub(buffered);
+    if excess > 0 {
+        current_buffer.drain(..excess);
     }
 }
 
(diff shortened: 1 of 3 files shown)


imsnif:
Hey @ZrjaK - good find, thanks for this!

I made a small adjustment: we should realign the buffers by truncating from the *start* rather than the finish. I think in your tests this happened to pass because the buffer we were aligning to (the parser_buf) was always 0.

Will merge later today after CI passes.

PR #5448, opened 2026-08-07
NestedListItem: forward the remaining Text methods

### LINKED ISSUE #5411: `NestedListItem::color_substring` does not exist

karpovantonme:
Fixes #5411

`NestedListItem` wraps `Text` and forwards its builders one at a time, so a method that was never forwarded does not exist on it. `color_substring` is one of seventeen in that state, which is why its `success_` and `error_` variants are there and it is not.

Diffing the public API of the two files: `Text` has 37 public methods, `NestedListItem` had 21, and four of the difference are the free print/serialize functions that have their own nested-list counterparts. The rest:

```
color_all              dim_all           unbold_all        content
color_substring        dim_indices       unbold_indices    len
color_nth_substring    dim_range         unbold_range
color_last_substring   dim_substring     unbold_substring
success_color_last_substring   error_color_last_substring   disabled
```

All seventeen forwarded here, in the same one-line style as the existing ones. Added `is_empty` next to `len` since clippy asks for the pair.

Nothing else in the type changes: `indent` stays its only own behaviour, and `Ribbon` needs no equivalent work -- `print_ribbon` takes a `Text` directly, so the asymmetry was only here.

`cargo build`, `cargo test -p zellij-tile` and `cargo fmt --check` pass. `cargo clippy` reports the same single pre-existing warning on this file before and after.

### DIFF
--- a/zellij-tile/src/ui_components/nested_list.rs
+++ b/zellij-tile/src/ui_components/nested_list.rs
@@ -90,6 +90,82 @@ impl NestedListItem {
         self.content = self.content.success_color_all();
         self
     }
+    pub fn success_color_last_substring<S: AsRef<str>>(mut self, substr: S) -> Self {
+        self.content = self.content.success_color_last_substring(substr);
+        self
+    }
+    pub fn error_color_last_substring<S: AsRef<str>>(mut self, substr: S) -> Self {
+        self.content = self.content.error_color_last_substring(substr);
+        self
+    }
+    pub fn color_all(mut self, index_level: usize) -> Self {
+        self.content = self.content.color_all(index_level);
+        self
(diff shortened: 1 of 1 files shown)


Zykino:
Your PR made me realize what is really happening here. I've done #5611 which should prevent the API drifting appart.
Do you want to give feed-backs on it?

karpovantonme:
Left the feedback on #5611. Short version: taking a `Text` is the right fix and supersedes this, my one reservation is that narrowing `Text::new` to `String` rides along with it and turns 42 call sites into 432.

Closing this in favour of yours once #5611 lands, or say the word and I close it now.

PR #5452, opened 2026-08-07
fix(windows): load OpenConsole's conpty.dll when available

divens:
The pseudoconsole shipped in kernel32 silently discards VT queries the child writes (OSC 10/11 for fg/bg colors being the most visible symptom), so apps inside a pane never receive a reply. OpenConsole's conpty.dll (from microsoft/terminal) forwards those queries and carries years of terminal-emulation improvements over the system default. It is not distributed standalone by Microsoft but ships bundled inside WezTerm, Visual Studio's terminal component, and similar installations.

Match Alacritty/WezTerm: LoadLibraryW("conpty.dll") on first use, resolve Create/Resize/Close via GetProcAddress, fall back to kernel32 on failure. Log which backend was picked, and — when conpty.dll was loaded — the resolved full path of the DLL, so users can diagnose which one won on their machine.


mattcargile:
I think [this](https://www.nuget.org/packages/Microsoft.Windows.Console.ConPTY/1.24.260710001) is the modern conpty.

PR #5456, opened 2026-08-08
Forward terminal focus events to active panes

### LINKED ISSUE #5438: Focus events from the terminal aren't forwarded to the focused pane
Zellij forwards focus events between its own panes, but it doesn't pass the terminal's focus-in/focus-out through to the program running in the focused pane. So an app inside a pane never finds out when you switch away from the terminal and come back.

The practical version: I copy something in my browser on one monitor, click into the Neovim pane on the other monitor, and Neovim never gets a `FocusGained`. It has no idea it just came back into focus.

## Steps to reproduce

1. Start zellij with a single pane running Neovim.
2. Register a focus handler:

   ```lua
   :lua vim.api.nvim_create_autocmd("FocusGained", { callback = function() vim.notify("focus!") end })
   ```

3. Click into another application (a browser on a second monitor works well), then click back into the zellij pane running Neovim.

Nothing fires.

Now split a second zellij pane. Click into that one, then back into the Neovim pane. The notification fires.

## What works and what doesn't

| Transition | Focus event delivered |
| --- | --- |
| zellij pane to zellij pane | yes |
| zellij tab to zellij tab | yes |
| other application to zellij pane | no |

So the internal pane-switch path works. The path from the terminal down into the focused pane doesn't.

## Expected

When the terminal emulator reports focus-in or focus-out, zellij should forward it to the currently focused pane, the same way it already does for pane-to-pane switches.

## Environment

- zellij: `0.44.3`
- OS: macOS `26.5.2`
- Terminals tested: Ghostty `1.3.1` and iTerm2 `3.6.11`. Same behavior in both. Doesn't appear to be terminal-specific.
- Neovim: `v0.12.4`

## Why it matters

Focus events are how a lot of terminal programs stay in sync with the outside world. Two things this breaks in Neovim, and I doubt it's only Neovim:

- **Clipboard sync.** Plugins that track the system clipboard (yanky.nvim, for one) refresh their state on `FocusGained`. Without it, they keep serving stale internal history and never notice you copied something in another app. You end up reaching for `"+p` every time, which defeats the point of having the sync.

- **Reloading files changed on disk.** Neovim's `autoread` runs `checktime` on `FocusGained`. Without it, you can edit a file in another editor, or pull in git, switch back, and the buffer still shows the old contents. Save from there and you quietly clobber the change. That one's a real hazard, not just an annoyance.

Both are working around a missing event rather than a missing feature, so forwarding it would fix them without any change on the client side.

yamam:
## Summary

Fixes #5438

Forward terminal emulator focus-in and focus-out events to the active terminal pane when the pane has enabled focus event tracking.

## Changes

- Enable terminal focus reporting (`CSI ?1004h`) during client terminal setup.
- Disable it during terminal teardown.
- Parse `CSI I` and `CSI O` in the stateful CSI parser.
- Preserve focus-like sequences inside bracketed paste payloads.
- Forward focus events through the client/server IPC layer.
- Send events only to the active terminal pane that has requested focus tracking.
- Preserve existing pane-switch focus event behavior.

## Testing

- `cargo test -p zellij-client --lib`

### DIFF
--- a/zellij-client/src/input_handler.rs
+++ b/zellij-client/src/input_handler.rs
@@ -264,6 +264,14 @@ impl InputHandler {
                             raw_bytes,
                         });
                 },
+                Ok((InputInstruction::TerminalFocusGained, _error_context)) => {
+                    self.os_input
+                        .send_to_server(ClientToServerMsg::TerminalFocusGained);
+                },
+                Ok((InputInstruction::TerminalFocusLost, _error_context)) => {
+                    self.os_input
+                        .send_to_server(ClientToServerMsg::TerminalFocusLost);
+                },
                 Ok((
                     InputInstruction::ForwardedReplyFromHostComplete { token, reply_bytes },
                     _error_context,
--- a/zellij-client/src/lib.rs
+++ b/zellij-client/src/lib.rs
@@ -53,6 +53,8 @@ const RESET_STYLE: &str = "\u{1b}[m";
 const SHOW_CURSOR: &str = "\u{1b}[?25h";
 const ENTER_KITTY_KEYBOARD_MODE: &str = "\u{1b}[>1u";
 const EXIT_KITTY_KEYBOARD_MODE: &str = "\u{1b}[<1u";
+const ENABLE_FOCUS_REPORTING: &str = "\u{1b}[?1004h";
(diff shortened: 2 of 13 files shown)


yamam:
Closing this as superseded by #5485, which implements host terminal focus events with additional multi-client and reconnect handling.

PR #5458, opened 2026-08-08
fix(clipboard): don't kill copy_command processes that own the selection

### LINKED ISSUE #5422: Regression: copy_command stopped working after v0.42.2
v0.42.2: with `copy_command: "xsel -b"` mouse selection copies the selection remotely and can be pasted locally.

Starting from v0.43 - stopped working. Used no-web flavour of zellij.

teddytennant:
Fixes #5422

## The bug

`CopyCommand::set` spawns the configured `copy_command`, writes the copied text to its stdin, and then — since 19c1591 (#4298, released in 0.43.0) — polls the process for one second and `kill()`s it if it is still running:

```rust
if start.elapsed() > timeout {
    let _ = process.kill();
    log::error!("Copy operation times out after 1 second");
    return;
}
```

X11 and Wayland clipboard helpers are supposed to still be running at that point. On X11 the clipboard has no storage of its own: whoever set the selection stays alive and hands the data over when something pastes it. That is why `xsel`, `xclip` and `wl-copy` keep running after they have consumed stdin. Killing them one second later throws the clipboard away, so pasting either yields nothing or yields whatever was on the clipboard before.

This affects the commands zellij itself suggests in the generated config (`xclip -selection clipboard`, `wl-copy`, `xsel -b`). Whether a given helper survives depends on how quickly it manages to detach (several of them fork a background process and let the foreground one exit), which is why this shows up as an intermittent "copying just stopped working" rather than a consistent failure — and why it is more likely on a remote/forwarded X11 connection, as in #5422, where every X round trip is slow.

### Reproducer

I extracted the `set()` body into a small binary so both versions could be run against a real X server (`Xvfb`) with real clipboard tools, since running zellij itself needs a terminal:

```
########## 19c1591 (0.43.0) — tool='xsel --nodetach -b' ##########
[before] KILLED child after 1s timeout
readback: []
########## 19c1591 (0.43.0) — tool='xclip -selection clipboard' ##########
[before] KILLED child after 1s timeout
readback: []
########## this patch — tool='xsel --nodetach -b' ##########
readback: [CLIPBOARD-PAYLOAD]
########## this patch — tool='xclip -selection clipboard' ##########
[after] child exited on its own: ExitStatus(unix_wait_status(0))
readback: [CLIPBOARD-PAYLOAD]
```

(`readback` is `xsel -b -o` run afterwards. xsel 1.2.1, xclip 0.13.)

## The fix

Move the reaping into a small `reap()` helper that just blocks on `process.wait()` instead of polling with a deadline, and drop the `kill()`.

**#4298's zombie fix is preserved**: the child is still `wait()`ed for on a background thread, so a `copy_command` that exits by itself (`pbcopy`, `wl-copy --foreground`, a user's shell script, …) is still reaped the moment it exits — in fact sooner than before, since there is no 50ms polling granularity any more. The only thing removed is the killing, which is what makes a long-lived selection owner survive.

There is no timeout left at all, deliberately: any timeout is wrong here, because a helper that owns the selection is *meant* to outlive it, with no upper bound. The reaper thread lives as long as the process it is waiting for. In practice that is at most one thread per live helper, because setting the clipboard again makes the previous owner lose the selection and exit, which ends its reaper thread.

## Tests

Two unit tests in `zellij-server/src/tab/copy_command.rs`, `#[cfg(all(test, unix))]` since they need a shell:

* `copy_command_is_not_killed_when_it_outlives_the_copy` — runs a stand-in helper that consumes stdin and then keeps running past the old one-second deadline, and asserts it both received the text and was still alive to finish. Fails on `main`:

  ```
  ---- tab::copy_command::tests::copy_command_is_not_killed_when_it_outlives_the_copy stdout ----
  thread '...' panicked at zellij-server/src/tab/copy_command.rs:124:9:
  the copy command should not be killed for outliving the copy operation

  test result: FAILED. 1 passed; 1 failed; 0 ignored; 0 measured; 1556 filtered out
  ```

* `copy_command_is_reaped_once_it_exits` — guards #4298: after `reap()` has finished with a process that exits immediately, `waitpid(pid, WNOHANG)` must fail with `ECHILD`, ie. the process is gone from the process table rather than left as a zombie. Replacing the `wait()` with a plain `drop()` makes it fail:

  ```
  thread '...' panicked at zellij-server/src/tab/copy_command.rs:122:9:
  assertion `left == right` failed: the copy command's process should have been reaped
    left: None
   right: Some(ECHILD)
  ```

`cargo test -p zellij-server --lib` with the patch: `1408 passed; 0 failed; 150 ignored` (1406 before, plus these two). `cargo fmt --check` is clean.

One flake to flag, unrelated to this change: `screen::screen_tests::send_cli_scroll_up_action` fails intermittently under load on a clean checkout of the base commit too (it passed on run 1 and failed on run 2 of the unmodified suite here). I have not touched it.

## What I could not verify

I do not have a display or a real clipboard on the machine I tested on, so the end-to-end runs above are against `Xvfb` with the `set()` logic lifted out of zellij, not against zellij driving a real desktop session. I also could not reproduce the reporter's exact command: on my setup `xsel -b` detaches fast enough that it was never the one killed — it is `xclip -selection clipboard` and `xsel --nodetach -b` that I could make fail on demand. The mechanism is the same for all of them, but someone with a desktop confirming #5422 against this branch would be worth having.

## Not changed

The `copy_command` string is still split on plain whitespace in `CopyCommand::new`, so a command or path containing spaces still cannot be configured. That is a separate pre-existing issue and I left it alone.


imsnif:
Good catch, thank you!

Issue #899, opened 2021-11-25
 Feature: Support for OSC 133 ANSI escape sequence


a-kenji:
Related discussion: #892, #642

https://wezfurlong.org/wezterm/shell-integration.html :

> OSC 133 Escape sequence to define Input, Output and Prompt zones

How could this be implemented?

leelavg:
- Hi, glad that I checked issues before raising a new one 😅 
- May I know the progress of this or whom to ask?
- I'm on `foot` terminal and when I add below to my bash prompt `foot` terminal is able to recognise the prompts and support jumping to them
```
prompt_marker() {
    printf '\e]133;A\e\\'
}

set_ps1() {
    # my current prompt function
}

PS1="$(set_ps1)$(prompt_marker)"
```
- Even with locked interface, it seems zellij isn't sending this sequence to the emulator (or zellij is handling it incorrectly 🤔)

tobias-haenel:
There should at least be some action like `scroll-next-prompt` or `scroll-previous-prompt` similar to `tmux` (https://man7.org/linux/man-pages/man1/tmux.1.html). I imagine that `zellij` needs to keep track of the various semantic parts prompt, input/command, output in the scrollback buffer based on OSC 133 sequences that are emitted by applications like `nushell`. Maybe some `tmux` contributors or people who understand their codebase might give more details on a potential implementation.

tummetott:
Is someone working on this? OSC 133 is so usefull, it would be a killer feature in zellij.

beojan:
Another request for this. It's really useful to be able to scroll to the start of a build command's output when you're dealing with C++ template error messages.

fculpo:
I opened #5460 with an initial OSC 133 integration focused on selection: triple-clicking marked output selects the command and its output while excluding the prompt.

It handles scrollback, reflow, and incomplete commands while preserving existing selection behavior elsewhere. Testing and feedback across shells and terminals are welcome.

fculpo:
This can be closed, merged via #5460

beojan:
Scrolling to the previous prompt would still be useful I think.

fculpo:
Indeed, maybe better to open a specific issue now that the base is implemented ?

schrieveslaach:
@fculpo, thanks for providing #5460. I would be also nice to edit output of last command instead of the whole scrollback. Is that something you implemented to or has it to be done, still?

fculpo:
[...]

PR #5460, opened 2026-08-09
feat: select OSC 133 command output on triple click

fculpo:
Disclaimer: I am not a Rust developer, this was analyzed and implemented by LLMs for my own usage (Fable and Sol)

## Summary

- parse OSC 133 `A`/`P`, `B`/`I`, `C`, and `D` into row-local semantic boundaries
- make triple-click on marked output select the command and output while excluding the prompt, with the existing logical-line fallback elsewhere
- preserve boundaries through scrollback, wrapping, reflow, clearing, truncation, and overwriting, and render click selections without mouse movement

Related to #899.

## Testing

- `cargo fmt --all -- --check`
- `cargo test --no-default-features -p zellij-server --lib` (1423 passed, 150 ignored)
- `cargo test` (10 passed, 5 ignored)
- `cargo xtask build`
- manual Ghostty and Powerlevel10k verification for command/output selection, prompt exclusion, and stationary triple-click highlighting

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,6 +34,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 * feat: implement support for the kitty graphics protocol to display images in the terminal (https://github.com/zellij-org/zellij/pull/5428)
 * fix: only advertise Sixel support if the attached terminal also supports it (https://github.com/zellij-org/zellij/pull/5432)
 * fix: properly unbundle plugin wasm executables when stripping them (to assist distro packagers) (https://github.com/zellij-org/zellij/pull/5433)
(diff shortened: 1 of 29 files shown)


imsnif:
Hey @fculpo - thanks for this, it's a lovely feature (especially now that many shells support OSC133 out of the box). I made some adjustments:
1. I changed the algorithm so that we only look for the beginning/end indication around the clicked position until we find one rather than scanning the whole buffer every time (and this we only do if the buffer has ever seen an OSC133 to begin with, so as to short circuit panes that do not support it)
2. I made this behavior configurable (opt-out) with `osc133_command_selection`
3. A bit out of scope, but while I was there I also added `word_separators` as a configurable option to allow users to configure which characters separate a word for the double-clicking word selection. The default is (as it is now) `[]{}<>()` but can be attached to (for example: `[]{}<>():` for users who want to include colons in this separation (by default they are not included so as to copy urls whole in this way).

PR #5461, opened 2026-08-09
fix: don't log error when resizing unspawned or cleared terminal

luchev:
## The error

Users see recurring ERROR lines in the zellij log, sometimes dozens per day, from `pty_writer`:

```
ERROR [pty_writer] zellij-server/src/os_input_output_unix.rs:377: a non-fatal error occured
Caused by: failed to set terminal id 0 to size (49, 171)
Caused by: failed to find terminal fd for id 0
```

The terminal id varies (0, 1, 2...), and the size alternates between a full-size and a 1-row state in ~700ms bursts.

## The reason

`set_terminal_size` looks up the terminal id in `terminal_id_to_raw_fd` (a map of terminal id → optional raw fd). The lookup has three states:

- `Some(Some(fd))` — live terminal, resize proceeds
- `Some(None)` — id **reserved but not yet spawned** (`reserve_terminal_id` inserts this placeholder before the pty spawn completes)
- entry missing — terminal **already cleared** (`clear_terminal_id` removes it when its pane closes)

`ResizePty` instructions can legitimately arrive for both non-live states:

1. During the reserve→spawn window (a resize fired between `reserve_terminal_id` and `spawn_terminal` — pane creation paths re-fire the resize after spawn, so the intermediate one is redundant).
2. After a terminal is cleared: a stale pane/layout pass re-fires `resize_pty!` for the dead id (also reachable through `apply_cached_resizes`, which drains cached resizes for terminals that may have been cleared in the meantime).

Both are normal transient/stale conditions, not failures. The old `_` arm turned them into an `anyhow!` error logged at **ERROR** level via `.non_fatal()`, producing pure log noise with no behavioral effect (the error was already swallowed by `non_fatal` and callers never observed it).

## The fix

Treat a missing/unspawned terminal id as a benign skip: log at **debug** level and return `Ok(())`:

```rust
_ => {
    log::debug!("tried to resize terminal id {} which is not yet spawned or already closed", terminal_id);
}
```

`write_to_tty_stdin` and `tcdrain` keep their `_ => Err(...)` behavior — writing to a genuinely dead fd is still a real failure that the pty_writer handles.

## Test

Added `resize_for_unspawned_or_cleared_terminal_is_not_an_error` in `os_input_output_unix.rs`: with a real `UnixPtyBackend`, reserves a terminal id, resizes it before spawn, then clears it and resizes again — asserting no ERROR-level "failed to find terminal fd" record is logged (via a test-scoped `log` capture). Fails on main, passes with the fix. Full `zellij-server` suite: 1407 passed, 0 failed.

### DIFF
--- a/zellij-server/src/os_input_output_unix.rs
+++ b/zellij-server/src/os_input_output_unix.rs
@@ -374,9 +374,10 @@ impl UnixPtyBackend {
                 }
             },
             _ => {
-                Err::<(), _>(anyhow!("failed to find terminal fd for id {terminal_id}"))
-                    .with_context(err_context)
-                    .non_fatal();
+                log::debug!(
+                    "tried to resize terminal id {} which is not yet spawned or already closed",
+                    terminal_id
+                );
             },
         }
         Ok(())
@@ -470,6 +471,7 @@ mod tests {
     use nix::fcntl::{fcntl, FcntlArg, OFlag};
     use nix::sys::termios;
     use std::io::Read;
+    use std::sync::OnceLock;
 
     /// Verify that `try_write_to_fd` writes as many bytes as the kernel will
     /// accept in one pass and returns a partial count (not an error) when the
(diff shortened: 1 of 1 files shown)


imsnif:
Hey, thanks for this. Happy to merge, but I think this is far too much code for such small a change. We really don't need to test this. It's even fine not to log this but to turn this arm into a no-op.

PR #5473, opened 2026-08-13
fix(themes): skip invalid files while loading theme directories

### LINKED ISSUE #2598: Difference in error handling of .kdl files from 0.36.0 to 0.37.2 - improve theme file error handling
**Basic information**

`zellij --version`: 0.37.2
`uname -av` or `ver`(Windows): Linux MSI 5.15.90.1-microsoft-standard-WSL2

**Further information**
I updated `zellij` from version 0.36.0 to 0.37.2 and did not change any configuration files. Upon running `zellij` it immediately quit and print `De serialization error: Expected a valid node entry.` to the terminal. Nothing was added to the log file.

After reinstalling zellij version 0.36.0 it run successfully run, while still logging the following:
```
ERROR  |zellij_utils::setup      | 2023-07-01 14:41:24.688 [main      ]
 [/home/user/.cargo/registry/src/index.crates.io-6f17d22bba15001f/zellij-utils-0.36.0/src/setup.rs:325]: error loading theme file: 
KdlDeserializationError(KdlError { input: "themes {\n  catppuccin-latte{\n    bg: \"#acb0be\" # Surface2\n    fg: \"#acb0be\" # Surface2\n  
  red: \"#d20f39\"\n    green: \"#40a02b\"\n    blue: \"#1e66f5\"\n    yellow: \"#df8e1d\"\n    magenta: \"#ea76cb\" # Pink\n    orange: 
\"#fe640b\" # Peach\n    cyan: \"#04a5e5\" # Sky\n    black: \"#4c4f69\" # Text\n    white: \"#dce0e8\" # Crust\n    }\n  catppuccin-frappe 
{\n    bg: \"#626880\" # Surface2\n    fg: \"#c6d0f5\"\n    red: \"#e78284\"\n    green: \"#a6d189\"\n    blue: \"#8caaee\"\n    yellow: 
\"#e5c890\"\n    magenta: \"#f4b8e4\" # Pink\n    orange: \"#ef9f76\" # Peach\n    cyan: \"#99d1db\" # Sky\n    black: \"#292c3c\" # 
Mantle\n    white: \"#c6d0f5\"\n    }\n\n  catppuccin-macchiato {\n    bg: \"#5b6078\" # Surface2\n    fg: \"#cad3f5\"\n    red: 
\"#ed8796\"\n    green: \"#a6da95\"\n    blue: \"#8aadf4\"\n    yellow: \"#eed49f\"\n    magenta: \"#f5bde6\" # Pink\n    orange: 
\"#f5a97f\" # Peach\n    cyan: \"#91d7e3\" # Sky\n    black: \"#1e2030\" # Mantle\n    white: \"#cad3f5\"\n    }\n\n  catppuccin-mocha {\n
    bg: \"#585b70\" # Surface2\n    fg: \"#cdd6f4\"\n    red: \"#f38ba8\"\n    green: \"#a6e3a1\"\n    blue: \"#89b4fa\"\n    yellow: 
\"#f9e2af\"\n    magenta: \"#f5c2e7\" # Pink\n    orange: \"#fab387\" # Peach\n    cyan: \"#89dceb\" # Sky\n    black: \"#181825\" # 
Mantle\n    white: \"#cdd6f4\"\n    }\n}\n", span: SourceSpan { offset: SourceOffset(47), length: 1 }, label: Some("plain identifiers can't be
 used here"), help: Some("If this was supposed to be a string, wrap it in quotes.\nIf this was supposed to be a new node, terminate the 
previous node with `;` or a newline."), kind: Context("a valid node entry") }) 
```

There was a syntax error in a theme definition file. This theme file was in the theme directory and did _not_ have a theme that was currently being used. 

The difference of handling the error between the 2 version is what I consider the bug:
- Version 0.37.2 quits the program without logging the error and prints ```De serialization error: Expected a valid node entry.``` 
- Version 0.36.2 will run the program and log the same error 

I would believe the intended behavior in Version 0.37.2 would be the same behavior as Version 0.36.2 . 

This was not an easy error to diagnose as there was little information given. I had no change in configuration and not much of a direction to head towards in order to solve it. Zellij so nicely handles config.kdl syntax errors, and such an implementation for handling theme configuration files would be helpful for users. 

**Incorrectly formatted file for reference** 
```
themes {
  catppuccin-latte{
    bg: "#acb0be" # Surface2
    fg: "#acb0be" # Surface2
    red: "#d20f39"
    green: "#40a02b"
    blue: "#1e66f5"
    yellow: "#df8e1d"
    magenta: "#ea76cb" # Pink
    orange: "#fe640b" # Peach
    cyan: "#04a5e5" # Sky
    black: "#4c4f69" # Text
    white: "#dce0e8" # Crust
    }
  catppuccin-frappe {
    bg: "#626880" # Surface2
    fg: "#c6d0f5"
    red: "#e78284"
    green: "#a6d189"
    blue: "#8caaee"
    yellow: "#e5c890"
    magenta: "#f4b8e4" # Pink
    orange: "#ef9f76" # Peach
    cyan: "#99d1db" # Sky
    black: "#292c3c" # Mantle
    white: "#c6d0f5"
    }

  catppuccin-macchiato {
    bg: "#5b6078" # Surface2
    fg: "#cad3f5"
    red: "#ed8796"
    green: "#a6da95"
    blue: "#8aadf4"
    yellow: "#eed49f"
    magenta: "#f5bde6" # Pink
    orange: "#f5a97f" # Peach
    cyan: "#91d7e3" # Sky
    black: "#1e2030" # Mantle
    white: "#cad3f5"
    }

  catppuccin-mocha {
    bg: "#585b70" # Surface2
    fg: "#cdd6f4"
    red: "#f38ba8"
    green: "#a6e3a1"
    blue: "#89b4fa"
    yellow: "#f9e2af"
    magenta: "#f5c2e7" # Pink
    orange: "#fab387" # Peach
    cyan: "#89dceb" # Sky
    black: "#181825" # Mantle
    white: "#cdd6f4"
    }
}
```

fzlzjerry:
Fixes #2598.

An invalid file in the theme directory currently makes `Themes::from_dir` return immediately, so startup and config reload fail even when the malformed theme is not selected.

This restores the earlier per-file behavior: theme-file read and parse errors are logged as non-fatal, valid files continue to be merged, and errors opening or iterating the theme directory itself still propagate.

The regression test loads a directory containing one valid theme and one malformed theme, then verifies that the valid theme remains available.

Tests:

- `cargo test -p zellij-utils`
- `cargo xtask format --check`

### DIFF
--- a/zellij-utils/src/kdl/mod.rs
+++ b/zellij-utils/src/kdl/mod.rs
@@ -6,6 +6,7 @@ use crate::data::{
     DEFAULT_STYLES,
 };
 use crate::envs::EnvironmentVariables;
+use crate::errors::FatalError;
 use crate::home::{find_default_config_dir, get_layout_dir};
 use crate::input::config::{Config, ConfigError, KdlError};
 use crate::input::keybinds::Keybinds;
@@ -19,6 +20,7 @@ use crate::input::permission::{GrantedPermission, PermissionCache};
 use crate::input::plugins::PluginAliases;
 use crate::input::theme::{FrameConfig, Theme, Themes, UiConfig};
 use crate::input::web_client::WebClientConfig;
+use anyhow::Context;
 use kdl_layout_parser::KdlLayoutParser;
 use std::collections::{BTreeMap, BTreeSet, HashMap, HashSet};
 use std::net::{IpAddr, Ipv4Addr};
@@ -5756,7 +5758,14 @@ impl Themes {
             let path = entry.path();
             if let Some(extension) = path.extension() {
                 if extension == "kdl" {
-                    themes = themes.merge(Themes::from_path(path)?);
+                    match Themes::from_path(path.clone()) {
+                        Ok(themes_from_file) => themes = themes.merge(themes_from_file),
+                        Err(error) => Err::<(), anyhow::Error>(anyhow::anyhow!("{error:?}"))
+                            .with_context(|| {
+                                format!("Failed to load theme file {}", path.display())
+                            })
+                            .non_fatal(),
+                    }
                 }
             }
         }
@@ -7208,6 +7217,42 @@ fn themes_to_string() {
         "Deserialized serialized config equals original config",
     );
     insta::assert_snapshot!(serialized.to_string());
+}
+
+#[test]
+fn invalid_theme_file_does_not_prevent_loading_valid_themes() {
+    let theme_dir = tempfile::tempdir().unwrap();
+    std::fs::write(
+        theme_dir.path().join("valid.kdl"),
+        r##"
+        themes {
+            valid {
+                fg "#D8DEE9"
+                bg "#2E3440"
+                black "#3B4252"
+                red "#BF616A"
+                green "#A3BE8C"
+                yellow "#EBCB8B"
+                blue "#81A1C1"
+                magenta "#B48EAD"
+                cyan "#88C0D0"
+                white "#E5E9F0"
+                orange "#D08770"
+            }
+        }
+        "##,
+    )
+    .unwrap();
+    std::fs::write(
+        theme_dir.path().join("invalid.kdl"),
+        "themes { invalid { fg }",
+    )
+    .unwrap();
+
+    let themes = Themes::from_dir(theme_dir.path().to_path_buf()).unwrap();
+
+    assert!(themes.get_theme("valid").is_some());
+    assert!(themes.get_theme("invalid").is_none());
 }
 
 #[test]


xhon-pelushi (approved):
Built this and checked it properly rather than eyeballing. The change is correct and the test is a real regression guard — I reverted just the `from_dir` hunk while keeping the test, and it fails exactly as it should:

```
test kdl::invalid_theme_file_does_not_prevent_loading_valid_themes ... FAILED
called `Result::unwrap()` on an `Err` value: KdlDeserializationError(KdlError { ... })
```

and passes with the hunk restored. Using `non_fatal()` is also the right call here — it routes through `to_log()`, so the broken file is still recorded rather than silently ignored, and it matches how the rest of the codebase handles recoverable problems.

One small suggestion, and one thing I checked so nobody else has to.

### `{error:?}` vs `{error}`

`ConfigError` is a `thiserror` enum with useful `Display` impls, so the `Debug` formatting is noisier than it needs to be. Same error, both ways:

```
Display = Deserialization error: Expected valid node terminator.

Debug   = KdlDeserializationError(KdlError { input: "themes { invalid { fg }",
          span: SourceSpan { offset: SourceOffset(19), length: SourceOffset(3) },
          label: Some("parsed node"), help: Some("Nodes can only be terminated by
          `;` or a valid line ending."), kind: Context("valid node terminator") })
```

So `anyhow::anyhow!("{error}")` would give a log line a user can act on, alongside the `Failed to load theme file <path>` context you already add.

Worth being even-handed about the trade-off, since it is not purely one-sided: `Debug` is the only one that carries kdl's `help` text ("Nodes can only be terminated by `;` ..."), which `Display` drops. It also embeds the file's contents in the log via `input:`, which for a theme file is harmless but is not a habit worth forming. If the help text is the part you wanted, pulling it out explicitly would beat dumping the whole struct.

### On the `anyhow!("{error:?}")` wrapper generally

I was going to suggest `Err::<(), ConfigError>(error).with_context(...)` to preserve the error chain instead of stringifying, then checked and it does not compile: `anyhow::Context` requires `E: std::error::Error + Send + Sync + 'static`, and `ConfigError` is not `Send + Sync` because of

```rust
#[error("Config error: {0}")]
Std(#[from] Box<dyn std::error::Error>),
```

`Box<dyn Error>` without `+ Send + Sync` poisons the whole enum. So the stringify is a necessary workaround, not an oversight — leaving this here so it does not get flagged as one in a later review. (Tightening that variant to `Box<dyn std::error::Error + Send + Sync>` would let the whole codebase use `anyhow`'s context directly, but that is well outside this PR.)

### Also checked

Both `from_dir` callers are consistent with making this non-fatal: `setup.rs:338` propagates with `?` into config loading, and `cli_assets.rs:84` already discards the error with `.ok()`. So no caller was relying on one bad theme file aborting the whole load.

LGTM as-is; the `{error}` change is a nit, not a blocker.

Tested on `22c2e60`, rustc 1.97.1, linux x86_64, `cargo test -p zellij-utils`.

fzlzjerry:
Thanks, @xhon-pelushi. Addressed the `{error}` nit in `329ec43f`. The theme loader now uses `ConfigError`’s Display output, with the existing file-path context and `.non_fatal()` handling retained. No change to the `ConfigError` enum or error-chain workaround.

Validation on macOS arm64, Rust/Cargo 1.97.0:
- `cargo +stable test --locked -p zellij-utils --features disable_automatic_asset_installation`: 418 unit tests and 3 doc tests passed; 2 ignored.
- A temporary executable called `Themes::from_dir` with one valid and one malformed theme: the valid theme loaded, the invalid one was skipped, and the captured log included the file path and `Deserialization error: Expected valid node terminator.` without the raw theme contents. The temporary executable was removed before committing.
- Package formatting checked. Compiler warnings remain in unchanged code; this commit only changes the formatting expression.

PR #5478, opened 2026-08-14
fix: writing to a pane id no longer clears the active pane's scroll

### LINKED ISSUE #5476: writing to a targeted pane-id clears the active pane

When using `zellij action write-chars --pane-id <pane_id>`, it will always clear the active (viewed) pane. 

Affected paths:
1. `write --pane-id`
2. `write-chars --pane-id`
3. `send-keys --pane-id`
4. `paste --pane-id`

Basically, if I have something in the background using write-chars to a different tab/pane, it will reset the pane I'm currently viewing as part of that. If I've scrolled up in my pane for example, it will reset my scrollback to the bottom.

Desired behavior is resetting the pane that was written to.

**Basic information**

`zellij --version`: zellij 0.44.3

`stty size`: 44 211

`uname -av` or `ver`(Windows): 6.18.33.2-microsoft-standard-WSL2 #1 SMP PREEMPT_DYNAMIC Thu Jun 18 21:54:43 UTC 2026 x86_64 x86_64 x86_64 GNU/Linux

## Minimal reproduction

1. send `write-chars` to a pane that is not your active pane
2. See the active pane reset

VXNCXNX:
Fixes #5476.

## The problem

`WriteToPaneId`, `WriteCharsToPaneId` and a pane-targeted `Paste` each sent `ScreenInstruction::ClearScroll(client_id)` before performing the write:

```rust
// route.rs:296, 308, 321
senders.send_to_screen(ScreenInstruction::ClearScroll(client_id))?;
senders.send_to_screen(ScreenInstruction::WriteToPaneId(bytes, pane_id))?;
```

But `ClearScroll` isn't pane-targeted. It goes `screen.rs:9045` → `active_tab_and_connected_client_id!` → `tab.clear_active_terminal_scroll(client_id)` → `get_active_pane_or_floating_pane_mut(client_id)` (`tab/mod.rs:6297`). It takes no pane id at all.

So the bytes land in pane B while the scroll reset lands in whatever pane A the client happens to be focused on. Scripting a write to a background pane yanks you out of your scrollback in the pane you're actually reading.

## The fix

The `ClearScroll` is there for a good reason — typing into a pane you've scrolled back in should jump you to the bottom — so this keeps the semantics and just aims them at the right pane.

zellij already has the pane-targeted equivalent: `ScreenInstruction::ScrollToBottomWithPaneId(PaneId, Option<NotificationEnd>)` → `tab.scroll_to_bottom_by_pane_id` (`tab/mod.rs:7805`). Its body is the same as `clear_active_terminal_scroll` but keyed on a pane id, and it already calls `sync_scroll_mode_for_pane_id` so scroll-mode state stays consistent. No new instruction needed.

`Action::Paste` now branches: targeted → `ScrollToBottomWithPaneId`, untargeted → the unchanged `ClearScroll`.

Nothing in the plugin/wasm boundary is touched; `zellij_exports.rs` never sent `ClearScroll` for these paths.

## Tests

Two in `route.rs`'s existing test module, driving a mock screen channel and comparing `ScreenContext` names:

- the three targeted actions emit `ScrollToBottomWithPaneId` and never `ClearScroll`
- the non-targeted paths still emit `ClearScroll`, so the fix can't over-correct

I checked the first is load-bearing by temporarily restoring the old `ClearScroll` line — it fails. No snapshots regenerated.

```
baseline  cargo test -p zellij-server:  1467 passed; 0 failed; 150 ignored
final     cargo test -p zellij-server:  1469 passed; 0 failed; 150 ignored
cargo fmt -p zellij-server -- --check:  clean
```

Two environment notes, for transparency rather than as excuses: a full-workspace build needs prebuilt wasm plugin artifacts that aren't in a fresh checkout, so I built and tested `zellij-server` only. And `screen::screen_tests::send_cli_scroll_up_action` is flaky here — it failed 1 run in 4 with my change, and also 1 in 5 on an unmodified baseline, so it's pre-existing and I left it alone.


VXNCXNX:
Rebased both onto current `main`, 35 commits of drift each, no conflicts, and each touches exactly one file still.

Being straight about what I did and did not verify: `cargo xtask build` succeeds on both rebased branches and `cargo fmt --check` is clean, but I could not run `cargo xtask test` here because the test build needs OpenSSL headers this machine does not have. So the build is verified locally and the test suite is not.

The push also moved CI into `action_required`, so there is no signal on the PR page until someone approves the run.

- #5478, `+102/-5` in `zellij-server/src/route.rs`: writing to a pane by id clears the active pane's scroll.
- #5480, `+43/-13` in `zellij-client/src/lib.rs`: the client aborts instead of exiting cleanly when the tty is gone at exit.

Independent of each other, so either can go alone.

PR #5485, opened 2026-08-17
Implement host focus gained/lost and missing notifications

imsnif:
This PR implements 2 things:

1. The missing host terminal's focus gained/lost events are now parsed and forwarded to panes who subscribe to them. Previously we would only do this with internal focused changes (eg. moving to a different pane/tab) - now we also do this for example when the user changes focus from the terminal pane.

2. This PR also implements the OSC9 and OSC777 desktop notifications protocols (in addition to the already implemented OSC99). These are implemented in a way that if a program sends anyone of the three to Zellij, we will endeavor to translate it to a notification protocol supported by the host terminal and then forward it.

It's also possible to explicitly choose which notification protocol Zellij will used through the config with `host_notification_protocol`.


MBA380:
I get spammed with progress notifications whenever I run a command such as 'flatpak update' or save any file or do an action in neovim.

Kitty terminal discards OSC 9 notifications that start with 4. Is it possible to do the same for zellij?
kovidgoyal/kitty#8011

imsnif:
Already fixed in main.

Issue #3453, opened 2024-06-26
Doesn't render emoji with zero-width-joiners (like the trans flag emoji)


VivianWilde:
# 1. Graphical issue inside a terminal pane (eg. something does not look as it should)

**Basic information**
Multi-character emoji, like the [trans flag emoji](https://emojipedia.org/transgender-flag) or the rainbow flag are rendered as their component emoji rather than combined.
Some multi-character emoji without zero-width-joiners are rendered correctly, like the [UK flag](https://emojipedia.org/flag-united-kingdom)

`zellij --version`:
zellij 0.40.1

`stty size`:
19 90

`uname -av` or `ver`(Windows):
Linux AdvancedResearchTransport 6.9.5-arch1-1 #1 SMP PREEMPT_DYNAMIC Sun, 16 Jun 2024 19:06:37 +0000 x86_64 GNU/Linux

## Further information

I'm using [wezterm](https://github.com/wez/wezterm) and fish. 
Outside of zellij the trans flag emoji renders correctly (both in the prompt and when pasted into the command).
Inside zellij it renders as "white flag + trans symbol" in both cases (screenshots are attached)

Inside Zellij:
![Inside Zellij](https://github.com/zellij-org/zellij/assets/45487680/93a0d253-5835-4e13-8514-899303f596bc)

Outside Zellij
![Outside Zellij](https://github.com/zellij-org/zellij/assets/45487680/0b5ab949-4c01-4038-a4af-3fd781e708c3)

XMoon:
I can reproduce this with a cursor-position probe as well, and the difference is quite large for ZWJ sequences.

Directly in Windows Terminal / Ghostty:

👨‍👩‍👧‍👦  terminal=2  expected=2

Inside Zellij:

👨‍👩‍👧‍👦  terminal=8  expected=2

The same Zellij result is reproducible with different outer terminals.

This suggests the ZWJ sequence is effectively being treated as separate wide emoji cells instead of one grapheme cluster. In TUIs this also causes cursor/layout drift because the application advances by 2 columns while Zellij's grid advances by 8.

#4800 looks directly relevant here since it stores full EGCs and adds grapheme segmentation support. Happy to test this exact case against that branch if useful.

PR #5500, opened 2026-08-20
fix(grid): attach combining marks instead of dropping them

### LINKED ISSUE #1538: Not drawing strikethrough text as strikedthrough

### LINKED ISSUE #3667: Unicode decomposed characters are not displayed correctly

[zellij.log](https://github.com/user-attachments/files/17348864/zellij.log)
[zellij-12.log](https://github.com/user-attachments/files/17348866/zellij-12.log)

**Basic information**

`zellij --version`: zellij 0.40.1

`stty size`: 41 168

`uname -av` or `ver`(Windows): Linux DESKTOP-J73IICD 5.10.102.1-microsoft-standard-WSL2 #1 SMP Wed Mar 2 00:30:59 UTC 2022 x86_64 x86_64 x86_64 GNU/Linux

## Further information

Input decomposed character (e.g. `echo é`).
`é` is displayed as `e`.
The code points of `é` are `\u0065 \u0301`.
The composed character of `é`, the code point of which is `\u00e9`, is displayed correctly.

nonpawite:
Fixes #1538
Fixes #3667

## The problem

`Grid::add_character` dropped every zero width codepoint outright. The comment above it acknowledged that this breaks grapheme segmentation and pointed at #1538.

For scripts that write their vowels and tone marks as combining marks, this is not lost decoration, it is lost text. Thai:

```
$ printf 'สวัสดีครับ ที่ไหน ผู้ใหญ่\n'
```

| | |
|---|---|
| outside zellij | `สวัสดีครับ ที่ไหน ผู้ใหญ่` |
| inside zellij | `สวสดครบ ทไหน ผใหญ` |

`ผู้ใหญ่` means "adult". `ผใหญ` is not a word. The same applies to Lao, Khmer, Myanmar, Devanagari, Hebrew points and Arabic harakat, and to decomposed Latin, where `e` + U+0301 lost its acute (#3667).

## The change

Combining marks (general category Mn, Mc, Me) are attached to the character they follow and rendered as part of it. Zero width codepoints that are *not* marks — controls, format characters such as the soft hyphen and the zero width joiner, fillers — have nothing to combine with and are still dropped, exactly as before.

Column count and cursor position are unchanged. The existing `precomposed_and_decomposed_forms_occupy_the_same_number_of_columns` test already pinned that invariant and still passes untouched.

Variation selectors are marks, so they survive too: U+26A0 U+FE0F now keeps its emoji presentation (⚠️) instead of degrading to the text one (⚠). That is the case #1538 was originally filed about.

## Memory

I assumed this was deliberate and load bearing:

```rust
// This size has significant memory and CPU implications for long lines,
// be careful about allowing it to grow
const _: [(); 16] = [(); std::mem::size_of::<TerminalCharacter>()];
```

So marks are not stored in the cell. They live in an interner keyed by cluster content, and the cell holds a 24 bit id into it. That id occupies the three padding bytes that already sat between `width` and `styles`:

```rust
pub struct TerminalCharacter {
    pub character: char,           // 4
    width: u8,                     // 1
    combining_marks: [u8; 3],      // 3, was padding
    pub styles: RcCharacterStyles, // 8
}
```

`size_of::<TerminalCharacter>()` is still 16. Carrying combining marks costs no additional memory per cell, and a cell without marks does no work at all beyond a three byte comparison. I extended the assertion to cover `aarch64` as well as `x86_64`, so it is actually checked on Apple Silicon.

The interner is thread local rather than shared, which is sound because `TerminalCharacter` holds an `Rc` and is therefore `!Send`: a cell can never be read from a thread other than the one that built it. No locking.

Keying by content means repeated text reuses ids, so a session rendering Thai holds a handful of entries no matter how much scrolls past.

## Things I would rather flag than hide

- **The interner is never evicted.** It is bounded by two caps: 8 marks per character, and 2^24 distinct clusters. Those exist to contain pathological input rather than because real sessions approach them. Doing this properly means refcounting cluster ids against live cells, which seemed like a lot of machinery for a table that holds a few hundred entries. Happy to add it if you disagree.
- **ZWJ emoji (#3453) is not fixed.** The zero width joiner is a format character, not a mark, and joining emoji into one cluster also requires recomputing the width of that cluster. That is a separate change and I did not want to smuggle it in here.
- **Search still matches on base characters only.** `search.rs` reads `.character`, so searching text with marks behaves as it did before this PR. Not a regression, but not fixed either.
- **New dependency: `unicode-properties`.** Needed to tell a mark from any other zero width codepoint. It is from unicode-rs, data only, no transitive dependencies, and `default-features = false` with just `general-category`. If you would rather not take it, the alternative is a hand maintained range table, which I did not think was a good trade.

## Testing

`cargo test -p zellij-server --lib`: 1534 passed, 0 failed. `cargo fmt --all --check` clean.

Existing tests updated, both because the behavior they pinned is the behavior being fixed:

- `combining_marks_are_dropped_and_do_not_advance_the_cursor` → `combining_marks_attach_to_the_preceding_character_and_do_not_advance_the_cursor`
- `variation_selector_and_zero_width_joiner_do_not_advance_the_cursor` now expects the variation selector to survive; the joiner and the zero width space are still dropped

Added:

- `thai_vowels_and_tone_marks_are_preserved` — `สวัสดีครับ`, ten codepoints, seven columns
- `combining_marks_with_no_preceding_character_are_dropped`
- `a_combining_mark_attaches_to_a_wide_character_rather_than_its_second_column`

Unchanged and still passing, which is the part that matters: the soft hyphen, hangul filler, halfwidth katakana sound mark, ZWJ emoji and vttest snapshot tests.

Also checked a release build by hand in a real session: Thai renders correctly, and the status bar, pane frames and tab names are unaffected.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,7 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 
 ## [Unreleased]
-* fix: combining marks (Thai and Lao vowels and tone marks, Devanagari matras, Hebrew points, Arabic harakat, decomposed Latin accents, emoji variation selectors) were dropped instead of being attached to the character they modify
+* fix: combining marks (Thai and Lao vowels and tone marks, Devanagari matras, Hebrew points, Arabic harakat, decomposed Latin accents, emoji variation selectors) were dropped instead of being attached to the character they modify (https://github.com/zellij-org/zellij/pull/5500)
 
 ## [0.45.0] - 2026-08-20
 * feat: allow tabs to have different sizes if clients aren't focused on the same one (https://github.com/zellij-org/zellij/pull/5133)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 
 ## [Unreleased]
+* fix: combining marks (Thai and Lao vowels and tone marks, Devanagari matras, Hebrew points, Arabic harakat, decomposed Latin accents, emoji variation selectors) were dropped instead of being attached to the character they modify
 
 ## [0.45.0] - 2026-08-20
 * feat: allow tabs to have different sizes if clients aren't focused on the same one (https://github.com/zellij-org/zellij/pull/5133)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,7 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 
 ## [Unreleased]
-* fix: combining marks (Thai and Lao vowels and tone marks, Devanagari matras, Hebrew points, Arabic harakat, decomposed Latin accents, emoji variation selectors) were dropped instead of being attached to the character they modify (https://github.com/zellij-org/zellij/pull/5500)
+* fix: combining marks (Thai and Lao vowels and tone marks, Devanagari matras, Hebrew points, Arabic harakat, decomposed Latin accents, emoji variation selectors) and the conjoining jamo of decomposed Hangul syllables were dropped instead of being attached to the character they modify (https://github.com/zellij-org/zellij/pull/5500)
 
 ## [0.45.0] - 2026-08-20
 * feat: allow tabs to have different sizes if clients aren't focused on the same one (https://github.com/zellij-org/zellij/pull/5133)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,7 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 
 ## [Unreleased]
-* fix: combining marks (Thai and Lao vowels and tone marks, Devanagari matras, Hebrew points, Arabic harakat, decomposed Latin accents, emoji variation selectors) and the conjoining jamo of decomposed Hangul syllables were dropped instead of being attached to the character they modify (https://github.com/zellij-org/zellij/pull/5500)
+* fix: combining marks (Thai and Lao vowels and tone marks, Devanagari matras, Hebrew points, Arabic harakat, decomposed Latin accents, emoji variation selectors) and the conjoining jamo of decomposed Hangul syllables were dropped instead of being attached to the character they modify, and were lost again when a screen was dumped, a session serialized or a selection copied (https://github.com/zellij-org/zellij/pull/5500)
 
 ## [0.45.0] - 2026-08-20
 * feat: allow tabs to have different sizes if clients aren't focused on the same one (https://github.com/zellij-org/zellij/pull/5133)
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4057,6 +4057,12 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "3b09c83c3c29d37506a3e260c08c03743a6bb66a9cd432c6934ab501a190571f"
 
 [[package]]
+name = "unicode-properties"
+version = "0.1.4"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "7df058c713841ad818f1dc5d3fd88063241cc61f49f5fbea4b951e8cf5a8d71d"
+
+[[package]]
 name = "unicode-width"
 version = "0.1.14"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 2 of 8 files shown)


shk95:
Tested this on macOS (aarch64) against a build of 8e02033 on top of v0.45.0: Thai marks now attach, but a decomposed Hangul syllable still loses its vowel and final consonant. `is_combining_mark` accepts general category Mark only, and Hangul jungseong/jongseong (U+1160–U+11FF, U+D7B0–U+D7FB) are `Lo` — letters that `unicode-width` gives zero width because they conjoin with the choseong before them — so `add_character` still drops them on the zero-width path. Since #3667 is the Korean report, that case seems worth covering here.

I've opened nonpawite/zellij#1 against this branch with the two ranges added beside category Mark and a grid test (`U+1112 U+1161 U+11AB` → one wide cell); `cargo test -p zellij-server` passes with it, and measured from the bytes an attached client receives the syllable now arrives whole (`e1 84 92 e1 85 a1 e1 86 ab`, equal to a bare pty). Happy to rebase or fold it however you prefer.

One thing noticed while measuring, not addressed there: `dump_screen!`, selection and `serialize` still push only `tc.character`, so `zellij action dump-screen` and copy lose the marks the screen now keeps.

nonpawite:
Thank you for building this and testing it on hardware I don't have — both of your findings were right, and both are now on the branch.

**Hangul jamo.** Folded as [`f9805a1`](https://github.com/nonpawite/zellij/commit/f9805a15), with you as the author. The reasoning holds: jungseong and jongseong are category `Lo` rather than `Mark`, so `is_combining_mark` did not accept them, and `unicode-width` gives them zero width, so they went on being dropped. Your range split around the unassigned U+D7C7–U+D7CA gap is right, and leaving choseong out is right too, since it is the base the others conjoin onto.

One correction, which is why the wording changed slightly when I folded it: #3667 is not the Korean report. It is "Unicode decomposed characters are not displayed correctly", filed against decomposed Latin `é` as U+0065 U+0301 on WSL — the case category `Mark` already covered. I could not find a Korean issue in the tracker at all, so I dropped the reference from both the commit message and the code comment rather than cite the wrong one. Your point about macOS handing terminals NFD file names is the real motivation and I kept it.

**The readback paths.** This one was the more valuable catch, because it was a hole in my own change rather than a missing case. Fixed in [`c4d1657`](https://github.com/nonpawite/zellij/commit/c4d16579): `dump_screen!`, `dump_screen_with_ansi!`, `pane_contents`, `pane_contents_with_ansi` and the selection copy all collected `character` alone and dropped the marks the grid had just been taught to keep. So the screen was right and everything read back off it was wrong — copying สวัสดี out of a pane pasted สวสด. They now go through `TerminalCharacter::push_cluster_to`, or `Row::to_text` where a whole row is wanted.

While fixing it I hit an adjacent case worth naming: trailing blanks are trimmed off a dumped or serialized line, and a space that a combining mark attached to is not a blank. `tc.character == ' '` now also requires `!tc.has_combining_marks()`, otherwise the trim takes the mark with it.

Search is still deliberately out of scope, as before — matching and highlighting index by byte offset and by column, so making it cluster-aware is its own change rather than a line in this one.

Four tests added for the readback paths (dump, copy, serialize, marked blank) alongside your jamo test. `cargo test -p zellij-server` is green at 1539 passing, and `cargo fmt --check` is clean.

shk95:
Thanks for folding both in so quickly, and for the correction on #3667 — I'd taken it for the Korean report without reading it properly, so I'm glad you caught that before it spread any further. The readback fix is a nice catch on top; copying text out of a pane losing its marks would have been a confusing one to debug later.

I'll re-pin our overlay to the branch head and drop the local addition. Closing nonpawite/zellij#1 since it's on the branch now.

nonpawite:
Merged current `main` (474ea0ce) into the branch to clear the conflict flag. The only conflict was the `[Unreleased]` entry in `CHANGELOG.md`; `grid.rs` and `grid_tests.rs` auto-merged cleanly despite the churn there since 0.45.1, and the OSC 7 work doesn't touch the `add_character` path this changes.

`cargo test -p zellij-server --lib` on the merge result: 1603 passed, 0 failed, no new warnings. That includes the nine tests this PR adds, covering Thai marks, Hangul conjoining jamo, a mark on a wide character, and the dump/serialize/selection readback paths.

GitHub reports no check runs on the branch, so I don't think CI has been triggered for this fork — glad to do whatever is needed on that side.

No hurry at all, just noting it's conflict-free and green whenever someone has a moment to look. Scope is unchanged from the discussion above: search matching and highlighting still index by byte offset and column, so making those cluster-aware stays a separate change.

PR #5513, opened 2026-08-22
Add sessions registry

divens:
## TL;DR

Sockets/pipes are now named by a stable **id**, not the session name. A KDL
registry (`sessions.kdl`) maps id → display name. Renaming a session just updates
the registry — nothing on disk is renamed — which is what fixes
[**reattach-after-rename on Windows** (#4998)](https://github.com/zellij-org/zellij/issues/4998) (named pipes can't be renamed).

## What changes

- Socket/pipe **and** the session-info cache folder are keyed by id; the registry
  is the single source of truth for name ↔ id.
- Rename = a registry update (no socket rename, no cache-folder move).
- Listing / kill / delete / resurrection resolve name ↔ id via the registry.
- Reconciliation prunes dead entries so `sessions.kdl` stays bounded after
  crashes; writes are atomic (temp + rename, with a `.bak` fallback).
- Legacy sessions migrate in with `id == name`; no protocol/plugin-API changes.

## Notes

- Ids are short (12 hex chars, collision-guarded) so the socket path fits the
  platform limit (104 bytes on macOS) — a full UUID overflows it.
- The server's id is threaded per-session (not a global env), which matters for
  the in-process integration-test harness.

## vs #5103

Supersedes it — a rework, not an update: registry is now the single source of
truth, rebuilt on current `main`, and drops the earlier exit-time re-serialization
that broke multi-tab resurrection.

## Testing

Unit tests for the registry (round-trip, reconcile, atomic/corrupt-file
handling) and a new integration test asserting the socket path is unchanged
across a rename and reattach-by-new-name works. Verified on Windows + Linux; CI
green.

supersedes #5103 and #5006

### DIFF
--- a/src/commands.rs
+++ b/src/commands.rs
@@ -776,7 +776,10 @@ pub(crate) fn start_client(opts: CliArgs) {
                             }
                             ClientInfo::Resurrect(
                                 session_name.clone(),
-                                session_layout_cache_file_name(session_name.as_ref()),
+                                session_layout_cache_file_name(
+                                    &zellij_utils::sessions::resolve_session_id(session_name)
+                                        .unwrap_or_else(|| session_name.to_string()),
+                                ),
                                 force_run_commands,
                                 new_session_cwd.clone(),
                             )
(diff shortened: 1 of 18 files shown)


divens:
@imsnif I'd like your feedback on this if you can please

imsnif:
Hey! This is on my TODO, however I'm going on vacation soon so will only get to it mid-September.

divens:
@imsnif Hey, no worries. Enjoy your well deserved vacation!

PR #5531, opened 2026-08-25
feat(kitty-graphics): support Unicode-placeholder (U=1) image placements

### LINKED ISSUE #5530: kitty graphics: Unicode-placeholder (U=1) placements are rejected with ENOTSUPPORTED
## Description

Zellij rejects the kitty graphics protocol's Unicode-placeholder (`U=1`) placement mode with `ENOTSUPPORTED`. Apps that place images this way — previews in file pickers, `fzf --preview`, chat clients with inline attachments — render nothing inside Zellij, even though they work in kitty, Ghostty, and WezTerm.

## Expected behavior

Printing `U+10EEEE` placeholder cells after a `U=1` transmit places the image at those cells and tracks them through scroll, erase, and overwrite.

## Actual behavior

Any `U=1` command fails here:

https://github.com/zellij-org/zellij/blob/main/zellij-server/src/panes/kitty_graphics/parser.rs#L489-L496

```rust
b"U" => match parse_u32(value) {
    Some(0) => {},
    Some(_) => {
        return Err(echo.error(
            KittyErrorCode::Enotsupported,
            "unicode placeholders are not supported",
        ));
    },
    ...
```

Nothing renders. No error reaches the user.

## Steps to reproduce

1. Save https://gist.github.com/drn1996/e6b243286d359c74fe7a906674f0df97
2. Run `python3 kitty_placeholder_repro.py` in kitty, Ghostty, or WezTerm: the image renders.
3. Run it inside a Zellij session in the same terminal: nothing renders. `--verbose` sets `q=0` and surfaces the `ENOTSUPPORTED` reply.

## Notes

Fixed in #5531.

drn1996:
## Summary

Fixes #5530.

Implements the kitty graphics protocol's Unicode-placeholder (`U=1`) placement mode. Zellij currently rejects it with `ENOTSUPPORTED`.

## Root cause

`parser.rs` only accepts `U=0`:

https://github.com/zellij-org/zellij/blob/main/zellij-server/src/panes/kitty_graphics/parser.rs#L489-L496

No code path registered or rendered a placement whose position comes from where the app prints placeholder cells, rather than `x=`/`y=` coordinates.

## What this does

- **Parser**: accepts and flags `U=1` (`parser.rs`).
- **Placeholder encoding** (`placeholders.rs`, new): a per-pane interner folds a placeholder cell's (image row, column, id most-significant byte) — read from its diacritics, or inherited from the left neighbor per the spec's omitted-diacritic rule — into a single Plane-16 PUA codepoint. The grid drops zero-width codepoints from stored cells, so the geometry has to live in the visible character to survive reflow and scrollback.
- **Grid integration** (`grid.rs`): builds the interned entry as the placeholder and its diacritics print, and scans the viewport at render time (`placeholder_kitty_chunks`, split into scan and per-run resolve) to turn visible placeholder cells into image chunks. Merges adjacent same-row cells into one chunk, and sizes aspect-ratio-preserving output when `c=`/`r=` gives only one axis, mirroring the existing positioned-placement logic.
- **Placement bookkeeping** (`grid_state.rs`): virtual placements key by image id with a `Vec` distinguished by placement id (carried in the cell's underline color), so multiple placements of one image coexist. Handles both `a=T,U=1` and the two-step `a=t` + `a=p,U=1` flow. Adds `d=r`/`d=R` range-delete for virtual placements, previously positioned-only.
- **Output/copy** (`output/mod.rs`, `grid.rs`): blanks interned placeholder codepoints to a space on serialize and copy, so they never leak as an undefined glyph.

## Edge cases

- A diacritic sequence interrupted by anything other than the next diacritic — cursor reposition, alternate-screen round trip, resize — ends the sequence, so a stray combining-mark character after a coincidental cursor jump can't corrupt an earlier placement.
- The interner has a capacity cap; once exhausted, new placeholder cells degrade to a space instead of leaking the raw undecodable character.
- Retransmitting under a reused image id drops the old virtual placement's reference. Re-placing under the same placement id reuses that placement's identity.

## Testing

- `cargo fmt` clean.
- `cargo test -p zellij-server`: 1568/1569 pass. The one failure, `temp_file_medium_outside_temp_dir_is_not_deleted`, is pre-existing: it assumes `CARGO_MANIFEST_DIR` sits outside the OS temp dir, false in my dev environment.
- New tests cover the interner, diacritic table, placement registration/re-registration/placement-id selection, id/range deletes, letterboxing, and the two edge cases above.
- Verified end-to-end against Ghostty running a Slack-client-style TUI: images that previously rendered nothing now display and track scroll and resize.

## Not verified

- Only tested against Ghostty. kitty and WezTerm should behave the same, since the output side is unchanged.
- `a=q,U=1` falls through to the existing query path unchanged; queries don't register a placement, so no placeholder-specific handling was needed.
- The scaled-variant cache namespaces placeholder variants from positioned ones via a high bit on the cell-count key, rather than a separate cache — a deliberate choice worth a second look.

Open to splitting this up if you'd rather review it in pieces.

### DIFF
--- a/zellij-server/src/output/mod.rs
+++ b/zellij-server/src/output/mod.rs
@@ -190,7 +190,15 @@ fn serialize_chunks_with_newlines(
             )
             .with_context(err_context)?;
             chunk_width += t_character.width();
-            vte_output.push(t_character.character);
+            // Kitty Unicode-placeholder cells are stored as encoded PUA
+            // chars (see panes::kitty_graphics::placeholders); the image
+            // is drawn over them via kitty chunks, so the host terminal
+            // gets a plain space instead of an undefined-glyph box.
+            if crate::panes::kitty_graphics::is_in_encoded_range(t_character.character) {
+                vte_output.push(' ');
+            } else {
+                vte_output.push(t_character.character);
+            }
         }
     }
     Ok(vte_output)
@@ -256,7 +264,15 @@ fn serialize_chunks(
             )
             .with_context(err_context)?;
             chunk_width += t_character.width();
-            vte_output.push(t_character.character);
+            // Kitty Unicode-placeholder cells are stored as encoded PUA
+            // chars (see panes::kitty_graphics::placeholders); the image
+            // is drawn over them via kitty chunks, so the host terminal
+            // gets a plain space instead of an undefined-glyph box.
+            if crate::panes::kitty_graphics::is_in_encoded_range(t_character.character) {
+                vte_output.push(' ');
+            } else {
+                vte_output.push(t_character.character);
+            }
         }
     }
     if let Some(sixel_image_store) = sixel_image_store {
--- a/zellij-server/src/panes/grid.rs
+++ b/zellij-server/src/panes/grid.rs
@@ -801,6 +801,12 @@ pub struct Grid {
     kitty_parser: KittyCommandParser,
     kitty_host_support: KittyHostSupport,
     sixel_host_support: bool,
+    /// (x, y, diacritics_seen) of a just-printed kitty image
+    /// placeholder cell, so the zero-width row/column/id diacritics
+    /// that follow it can be folded into the stored cell before the
+    /// generic zero-width drop discards them.
+    pending_kitty_placeholder: Option<(usize, usize, u8)>,
+    kitty_placeholder_interner: crate::panes::kitty_graphics::PlaceholderInterner,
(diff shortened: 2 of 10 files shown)


gary-host-laptop:
i hope this gets accepted because i'm not being able to display images while using zellij otherwise. :(

edit: i tested this fix and at least it fixes image previews in superfile.

PR #5537, opened 2026-08-25
fix(plugins): release a CLI pipe when the plugin handling it crashes

TAJD:
@imsnif — you were right, thanks for the pointer. I've dropped the `catch_unwind` commits entirely and force-pushed a different fix based on your suggestion. Rebased onto current `main`.

## What's actually going on

`apply_pipe_message_to_plugin` (`zellij-server/src/plugins/pipes.rs`) sends `UnblockCliPipes` on **every success path** — including the `Err(_e)` arm for old plugins that don't export `pipe` — but **not** when the call into the plugin fails.

When a plugin traps while handling a pipe message, both call sites (`wasm_bridge.rs`, in `pipe_messages` and in `apply_cached_events_and_resizes_for_plugin`) log the error and call `handle_plugin_crash`, which only sends `ScreenInstruction::UpdatePluginLoadingStage` with a "Panic!" loading indication. It does not unload the plugin, so `PendingPipes` keeps the `(plugin_id, client_id)` entry and the pipe is never released. The `zellij pipe` client stays blocked until the plugin is unloaded or the session ends.

I confirmed this by adding a `panic_while_handling_pipe` branch to the fixture plugin and logging the instruction stream. The crash is visible, and the unblock arrives only after `KillSession`:

```
server instructions: [ KillSession, UnblockCliPipeInput("input_pipe_id") ]
screen instructions: ... UpdatePluginLoadingStage(0, error: "intentional panic ...") ... Exit
```

## The fix

`apply_pipe_message_to_plugin` becomes a thin wrapper that releases the pipe on any `Err` and re-returns it; the old body moves to `apply_pipe_message_to_plugin_inner`. Doing it at that single choke point means neither call site can forget, and it covers the pre-call failures (protobuf conversion) too.

The regression test asserts `UnblockCliPipeInput` is emitted *at crash time* rather than at teardown — the earlier version of the test, which checked after `teardown()`, passed against the unfixed code, which is what made the teardown-only release visible in the first place.

## Verification

- New test fails before the change, passes after.
- `cargo test -p zellij-server --lib`: 1566 passed, 0 failed.
- Ignored plugin suite: 141 passed. 10 failures are pre-existing Windows-only path-snapshot mismatches (`/foo/bar` vs `C:/foo/bar`); they fail identically on an unmodified tree here.
- `cargo fmt --all --check` clean.

The fixture plugin `.wasm` was built locally and staged at `target/e2e-data/plugins/` to run the ignored tests.

## Separate observation about the `CliPipe` log line

Worth flagging since it's what sent me down the wrong path originally: the `Action CliPipe did not complete within 1s timeout` spam is **not** evidence of a blocked pipe, and it isn't a timeout.

`route.rs` deliberately does `drop(completion_tx)` in the `Action::CliPipe` arm, but the call to `wait_for_action_completion` at the end of `route_action` is unconditional. A dropped sender resolves the oneshot as `Err(RecvError)` immediately, which lands in the `Err(_) | Ok(Err(_))` arm and logs the timeout message. So every single `zellij pipe` invocation logs an error instantly, regardless of whether anything blocked — in my case ~21,000 lines over 9 days of normal use.

I've deliberately left that alone rather than bundling it here. Happy to open a separate PR if you'd like it silenced.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 * feat: allow setting explicit_theme_hue (light/dark) (https://github.com/zellij-org/zellij/pull/5536)
 * fix: input mode updates when switching tabs (https://github.com/zellij-org/zellij/pull/5535)
 * feat: allow starting a session with an initial command, eg. `zellij attach -a my-session -- htop` (https://github.com/zellij-org/zellij/pull/5543)
+* fix: release a CLI pipe when the plugin handling it crashes, instead of blocking the `zellij pipe` client until the plugin is unloaded (https://github.com/zellij-org/zellij/pull/5537)
 
 ## [0.45.0] - 2026-08-20
 * feat: allow tabs to have different sizes if clients aren't focused on the same one (https://github.com/zellij-org/zellij/pull/5133)
--- a/default-plugins/fixture-plugin-for-tests/src/main.rs
+++ b/default-plugins/fixture-plugin-for-tests/src/main.rs
@@ -993,6 +993,8 @@ impl ZellijPlugin for State {
             );
         } else if name == "message_to_plugin" {
             self.message_to_plugin_payload = payload.clone();
+        } else if name == "panic_while_handling_pipe" {
+            panic!("intentional panic for the pipe-release-on-crash test");
         }
         let should_render = true;
         should_render
--- a/zellij-server/src/plugins/pipes.rs
+++ b/zellij-server/src/plugins/pipes.rs
@@ -145,6 +145,47 @@ pub fn apply_pipe_message_to_plugin(
     plugin_render_assets: &mut Vec<PluginRenderAsset>,
     senders: &ThreadSenders,
 ) -> Result<()> {
+    let result = apply_pipe_message_to_plugin_inner(
+        plugin_id,
+        client_id,
+        running_plugin,
+        pipe_message,
+        plugin_render_assets,
+        senders,
+    );
+    if result.is_err() {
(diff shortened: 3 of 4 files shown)


imsnif:
Hi, thanks for this. I looked into this a bit though and I think this isn't exactly what's going on.

The quoted panic is from no_threads.rs (std's mutex shim for wasm32-wasip1). It carries the plugin's stderr log prefix, so it occurred inside the wasm sandbox, not on the server. Panics inside plugins surface as wasmi traps, returned as `Err` from `.call(...)` and handled at every call site via `handle_plugin_crash`. They cannot unwind `plugin-exec-N`, skip the counter decrement, or poison host mutexes. The `catch_unwind` therefore guards a path plugin panics never take, and for genuine host panics, the global panic hook still tears down the session before unwinding, so I'm afraid patch would not change the outcome either way.

From the symptoms you describe though, it could be that what happened is that the plugin crashed while processing a pipe message, so the pipe it was handling might not have gotten released, and thus left the CLI client blocked (would explain the `CliPipe did not complete within 1s timeout` spam. If you want to take a look and see if that's your issue, it'd be great.

TAJD:
Hey @imsnif I will do, thanks for the suggestion!

imsnif:
Hey @TAJD - could you please remove the LLM authored commits from the commit log? I don't mind you using them, but I'd rather not have bots in the commit log as much as possible.

TAJD:
Absolutely yes, let me do that

TAJD:
@imsnif and apologies if I'm wasting your time - I am trying to fix a genuine bug I'm seeing and I'm appreciating your quick feedback!

imsnif:
Thanks for the fix!

PR #5539, opened 2026-08-25
fix unlock-first keybinding preset's nested session keybinds

### LINKED ISSUE #5529: First Run Setup Wizard does not set up new nested key bindings properly for the non-colliding preset

## Issue description
When running the first run setup wizard, picking the non-colliding preset generates key bindings for nested session interactions that return to the "normal" mode rather than the "locked" mode. The affected keybindings are `[`, `]`, and `f` under the `session` section.

I think this is incorrect behaviour because the other mode keybindings all return to the `"locked"` mode except of course the key bindings that are used to enter the sub mode (e.g. `p` enters pane mode and would also return to normal mode when pressed again).

## Minimal reproduction
- Run zellij without an existing user configuration
- Select non-colliding preset
- View generated configuration file
- Observe affected keybindings ending with a `SwitchToMode "normal"` instead of `SwitchToMode "locked"`

## Other relevant information
The new keymap updater in the release notes properly adds the key bindings.

habeebweeb:
The first run wizard generated incorrect unlock first preset keybind action set for the nested session handling keybinds that return to "Normal" mode instead of "Locked" mode.

Fixes #5529

### DIFF
--- a/default-plugins/configuration/src/presets.rs
+++ b/default-plugins/configuration/src/presets.rs
@@ -129,9 +129,9 @@ keybinds clear-defaults=true {{
     session {{
         bind "o" {{ SwitchToMode "Normal"; }}
         bind "d" {{ Detach; }}
-        bind "]" {{ FocusHostSession; SwitchToMode "Normal"; }}
-        bind "[" {{ FocusGuestSession; SwitchToMode "Normal"; }}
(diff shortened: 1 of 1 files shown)


imsnif:
Thank you!

Issue #3147, opened 2024-02-17
C-q pop up to confirm


HubertBos:
Apologies for opening the issue, I know there are a lot of things in plan and to fix and I do not want to add more unfortunately this start to drive me crazy.

I'm using neovim and one of the commands is C-q for Visual vertical
I know I should block with C-g but when you are a sysadmin and you are swapping tabs and panels at 2 AM, it is not an easy task to have that good discipline to keep this in mind or muscle memory. 

Is it possible to introduce a popup to confirm close zellij when C-q, please? 
This can be enabled in config to not change current behaviour that others will think is good but it will make a diff for DevOps/Sysadmins out there. 

Thank you very much for understanding and looking forward to hearing from you.

cristiand391:
Hey, this seems a dup of https://github.com/zellij-org/zellij/issues/467 and https://github.com/zellij-org/zellij/issues/1229

I don't think there's anyone working on this right now, but you can see people workaround this by changing/disabling the Ctrl-q keybind.

Issue #1229, opened 2022-03-17
Feature request regarding Ctrl-Q


fansari:
Since I have now accidently used "Ctrl-Q" many times I would like to ask whether there can be done something preventing this.

Maybe a more complicated action or some question whether you really want to do it.

Closing everything is nothing I normally do intend to do so it would be good to change this.

a-kenji:
Hey! I am working on including this feature. And some preliminary work has been done to support it.

Related issues: #933, #467

fansari:
Great. I found this comment: "Also on other "terminating" operations, like "tab close" and "pane close".

I can confirm this. It is not a normal workflow that you want to close a tab. Normally such thing only happens by accident. So it would be good to have also this protection for tabs and panes.

imsnif:
> Great. I found this comment in #467: "Also on other "terminating" operations, like "tab close" and "pane close".
> 
> I can confirm this. It is not a normal workflow that you want to close a tab. Normally such thing only happens by accident. So it would be good to have also this protection for tabs and panes.

I do sometimes want to close many panes quickly. Maybe we add a confirmation for closing a tab/pane, and also add a "force close" shortcut (eg. `ctrl-p` + `x` => close with confirmation, `ctrl-p` + `X` => close without confirmation)

a-kenji:
Yes, since that isn't an action.

imsnif:
> I usually close zellij by closing the shell with Ctrl + D, I'm guessing that will remain without confirmation?

There's also not a lot we can do about this, tbh. We only intercept keys to the shell if they correspond to one of our UI actions. This is an internal shell thing - for one shell it can be ctrl + d, for another it can be something else.

raphCode:
> Maybe we add a confirmation for closing a tab/pane, and also add a "force close" shortcut

I feel like using a "double key" could solve the problem: The first `x` leading to a confirmation question with `x` as 'yes' and any other key as 'no'.

If you really want to close something, it is easy to get the double keypress into muscle memory, while inadvertently pressing the close combination twice seems unlikely.

This should also play nice with key remapping.

SeryiBaran:
+

If you open micro editor in zellij and try to exit it, zellij will close.

raphCode:
Since it wasn't mentioned already, it is totally possible to remap the offending actions to keys that are more difficult to hit by accident or remove the mapping altogether:
https://zellij.dev/documentation/keybindings.html

dwvisser:
[...]

Issue #467, opened 2021-05-06
Feature: Confirm on quit


lilyball:
If I have multiple panes/tabs open, quitting loses my layout. Similarly, if I have running processes, quitting terminates them. This is also true for closing tabs or panes.

By contrast, tmux prompts for all of these (when invoked via default keybinds). It also doesn't even have a default bind for `kill-session` (instead offering a bind for "detach"). This makes it very hard to accidentally lose work in tmux. And the easiest way to kill panes in tmux is to exit the process running in it (and killing windows is typically done by killing all panes).

I think zellij has the opportunity to strike a middle ground here. Here's my suggestions:

1. Like macOS Terminal.app, Zellij could have a notion of "processes that are okay to close without confirmation". Terminal.app always includes your login shell in that list, and by default includes "screen" and "tmux" (presumably because those can detach and resume later). Zellij would preferably have a config entry for processes to use.

   Until such time as Zellij has a way to override preferences on a per-preference basis, it should probably omit screen/tmux from the list (and the login shell would always be included even with an empty list).
2. When using a default keybind to close a pane, it should prompt if the pane includes processes other than what's in step 1.
3. When using a default keybind to close a tab, it should prompt if any pane in the tab would need a prompt.
4. When using a default keybind to quit zellij, it should prompt if any tab would need a prompt.

The close prompts would ideally identify any processes that are the reason for the prompt (again, this is something that Terminal.app does), though just saying it will "close running processes" is acceptable.

Zellij should support also actions to bypass the confirm prompt, for use in custom keybinds for people who want it. A few suggestions, ordered by my aesthetic preference:

```yaml
- action: [SkipConfirm: Quit,] # or [SkipConfirm: [Quit]]
- action: [QuitWithoutConfirm]
- action: [Quit: NoConfirm]
```

There's also the inverse, e.g. `ConfirmBeforeQuit` as the default, but this means anyone who's already customized their keybinds won't get the confirmations.

#### Additional thoughts

Experimentally, I believe macOS Terminal.app's notion of "login shell" is "any entry in /etc/shells that is the direct child of the root `login` instance for the tab" (meaning running `exec login -f $USER` from the shell requires a prompt before closing as the process list is now login > login > shell).

[...]

PR #5547, opened 2026-08-26
feat(server): interactive quit confirmation prompt

### LINKED ISSUE #933: Confirmation on quit?
Is it possible to configure a Confirmation on ctrl+q ?
Something like "Are you sure you want to quit? y/n"

jpds:
Adds a "Quit?" prompt when Ctrl-Q is hit, all key strokes in the session are then ignored minus Y/N/Esc.

This prompt can be disabled by setting `confirm_quit = false`.

Closes: #933, #467, #1229, #3147.

### DIFF
--- a/default-plugins/compact-bar/src/keybind_utils.rs
+++ b/default-plugins/compact-bar/src/keybind_utils.rs
@@ -729,6 +729,12 @@ impl KeybindProcessor {
                 ];
                 Self::find_predetermined_actions(mode_info, mode, ordered_predicates)
             },
+            InputMode::ConfirmQuit => {
+                vec![
+                    ("y".to_owned(), "Yes".to_owned()),
+                    ("n".to_owned(), "No".to_owned()),
+                ]
+            },
             InputMode::EnterSearch
             | InputMode::RenameTab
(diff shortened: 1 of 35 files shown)


Zykino:
I also don't like the binding in general but there are already multiples way to prevent accidental quit:
* I rebind-ed this option to execute only when I am in session mode. I'm on the "locked first" keybindings options with so I unlock, then go to session mode, then quit if I need to: `Alt g, o, Ctrl q`. -> I made the binding impossible to hit randomly
* Another option is like you did, change the binding to open a popup. Since zellij [support plugins](https://github.com/zellij-org/awesome-zellij), I think it is a better way to introduce this sort of functionality. And at least one already exist [`zj-quit`](https://github.com/cristiand391/zj-quit) (but is archived).

I would encourage you to use one of this methods (rebind, use an archived plugin, spin your own plugin) for this feature :)

PS: with your own plugin you could even merge the "detach" and "quit" bindings to only have one to "quit" using your plugin. And the plugin could let you choose which action to take. If you are fancy you could even merge the session manager plugin in there if you which to have the same bindings for all "quit this session" options ;)

PR #5582, opened 2026-09-04
Make search auto-jump on input configurable

### LINKED ISSUE #5551: Incremental search auto-jumps backward before "n" is ever pressed
**Basic information**

`zellij --version`: `zellij 0.45.0`

`stty size`: `54 243`

`uname -av` or `ver`(Windows): `Linux work 6.12.101+deb13-amd64 #1 SMP PREEMPT_DYNAMIC Debian 6.12.101-1 (2026-08-05) x86_64 GNU/Linux`

## Description

When searching the scrollback (`Ctrl s` -> scroll to a point -> `s` to enter search -> type a query), the view jumps *upward* through the scrollback before pressing `n`/`p` at all, and it does this again on every subsequent keystroke while typing the query, not just once.

There is a relevant comment in zellij-server/src/panes/search.rs:
```text
        // If the current viewport does not contain any hits,
        // we jump around until we find something. Starting
        // going backwards.
```

Because this runs per-keystroke, typing a multi-character query where intermediate substrings aren't visible on screen can cause the viewport to jump backward multiple times while still typing, before ever navigating with `n` (search down) or `p` (search up).

The practical effect: there is no way to scroll to a point in the scrollback and then search only *forward* from there, because the act of typing the query itself can relocate you above your starting point.

## Steps to reproduce

1. `zellij` with default config
2. Produce enough scrollback that a search term appears both above and below your current view
3. `Ctrl s` to enter scroll mode, scroll up to roughly the middle of the buffer
4. Press `s` to enter search mode
5. Type a query whose first few characters aren't visible in the current viewport, but which occurs earlier (above) in the buffer
6. Observe the viewport jump upward as you type, before pressing `n`/`p`

## Expected behavior

Some way to search only forward (or only backward) from the current scroll position without the search-input step itself scrolling to nearest match/wrapping backward — e.g., an option to disable the "search up first, then down" fallback in `set_search_string`, or an explicit "search forward only" toggle.

## Actual behavior

Typing the search query can silently scroll the viewport upward (once per keystroke, if no match is visible) before any navigation key is pressed, making it impossible to reliably confine a search to content below a chosen scroll position.

Huang-404-Q:
## Problem

While incrementally typing a search query in a pane, the viewport jumps around the scrollback on every keystroke — even before `n`/`p` is ever pressed (issue #5551).

## Fix

Per the discussion in #5582, jumping to the nearest match on input is the desired default (this is what Firefox and helix do). Instead of removing the auto-jump, this makes it configurable:

```kdl
config {
    // Whether to automatically jump to the nearest search result
    // while typing the search term. (Requires restart)
    search_auto_jump_on_input true
}
```

The default stays `true` (current behavior preserved); setting `search_auto_jump_on_input false` keeps the viewport in place while typing. Explicit navigation with `n`/`p` (`search_down`/`search_up`) is unaffected in both modes.

## Plumbing

`search_auto_jump_on_input` follows the existing `styled_underlines` pipeline: `Options` (new `Option<bool>` field, default `true`) → `Screen` → `Tab` → `LayoutApplier` → `TerminalPane::new` → `Grid` (via the existing `update_search_auto_jump_on_input` setter, same pattern as `update_arrow_fonts`). It is honored in `Grid::set_search_string`, where the two auto-jump blocks are now wrapped in `if self.search_auto_jump_on_input { ... }`.

The IPC contract gains `optional bool search_auto_jump_on_input = 70;` in `common_types.proto` and the matching checked-in prost code.

## Tests

- `updating_the_search_term_auto_jumps_by_default` — locks in the default: with the option on, typing `"odio"` (a term only present in the scrollback above the current viewport) jumps the viewport to the nearest hit.
- `updating_the_search_term_does_not_auto_jump_when_disabled` — with the option off, typing `"odio"` does not move the viewport and pre-selects nothing; a following `search_up` still jumps to the hit (guards the `n`/`p` path).

Regression: `cargo test -p zellij-server --lib` — baseline `1577 passed; 0 failed; 151 ignored`, after the change `1578 passed; 0 failed; 151 ignored` (only delta: the two renamed/added search tests). `cargo test -p zellij-utils` — `450 + 3 passed; 0 failed` (11 insta snapshots updated with the new option's default).

Fixes #5551


Zykino:
Is it really a sensible default to not jump on search?

Firefox jump to "adopted" when i type it to search on this page, `helix` jump to the word immediately on write…  
On the other hand `less` wait for me to press `Enter`.

I usually prefer the immediate jump since that way I can stop early if I already found what I was looking for. And more importantly, I se immediately that I made a typo (or that nothing match) before typing the full search string.  
So I can fix the typo while editing, or look for "warn" when I don't see any "err" in the search.

Huang-404-Q:
Thanks — and you're right, which is why the PR no longer removes the auto-jump. I reworked it after your comment (hence the rename): auto-jump on input stays the **default**, and the new option is an opt-out for the `less`-style behaviour.

Your Firefox/helix reasoning is the argument *for* the default as it is. The option exists only for the case in #5551 — searching forward from a specific point in the scrollback without the viewport relocating while you type. `n`/`p` are unaffected in both modes; there's a test pinning that (typing with the option off doesn't move the viewport, and a following `search_up` still jumps to the hit).

PR #5589, opened 2026-09-06
feat(nested-sessions): report a guest's input mode and key bindings to its host

### LINKED ISSUE #5561: Plugin API: Expose a nested session's input mode and key bindings to its host session's plugins
2. I need a new API

A guest advertises a `HintReporting` capability in the existing
handshake, and a host that sees it can ask:

- `GuestModeUpdate { mode, base_mode }`, guest to host
- `RequestGuestKeybinds`, host to guest
- `GuestKeybindsUpdate { keybinds }`, guest to host

Surfacing as `Event::NestedSessionModeUpdate`, `Event::NestedSessionKeybinds` and
`PluginCommand::RequestNestedSessionKeybinds(PaneId)`, under the existing
`ReadApplicationState` permission and `EventType` subscription.

Two changes from the proposal below. The mode event also carries `base_mode`, since a
host cannot render hints without knowing the guest's resting mode. And the guest reports
once unprompted when the handshake completes, not only on change, because otherwise a
host never learns a guest exists and has nothing to query about.

I'm working on a PR: https://github.com/myah-mitchell/zellij/tree/feat/nested-session-hint-events

Separately, out of the same work but standing on its own: #5588 asks for a way for a bar
plugin to give its layout row back while it has nothing to draw.

EDIT:
See the latest release of [zjhints](https://github.com/myah-mitchell/zjhints) for an example of this in use in the README's pic and video.

<details>
<summary>Original Issue Text</summary>
Original Title: Expose a nested session's current input mode to its host session's plugins

Original Body:

I'm really enjoying the 0.45 nested session support and the new possibilities that this brings. I'm looking to see if there would be any interest in supporting either an extension of the existing `ModeUpdate` event or adding a new `NestedSessionModeUpate` or something similar.

I maintaining a fork of a prior plugin called [zjstatus-hints](https://github.com/myah-mitchell/zjstatus-hints) (my fork) which is its self an extension of [zjstatus](https://github.com/dj95/zjstatus). The main feature of my plugin is providing a theme able way to show the currently valid key binds. This includes native binds, and ones the user may have added via a discovery process. Having access to input mode is critical for a plugin like this to function. 

I'm looking to update my plugin, or add a patch to zjstatus, so that it only is displayed in the host session. Then assuming the user has some system to maintain keybinds in sync the host session's zjstatus-hints can be configured to show the hints for the current pane's (including nested session panes) valid keybinds.

Honestly a way to receive the whole keybind list would be wonderful, but I felt that was to specific of a feature that it would never be added to Zellij. My hope is just extending the current process of passing session and nested data between the Zellij instances to also support passing mode would be more likely accepted.

So when a pane holds a nested session and that pane is the one currently receiving input (the same state that already triggers chrome dimming), send its mode to the host as a new event on change:

 enum Event {
      // ...
      NestedSessionModeUpdate {
          pane_id: PaneId,
          session_name: Option<String>,
          mode: InputMode,
      },
 }

Tagging pane_id and session_name is critical to support this across multiple nested Zellij sessions.

If this is interesting, I'm more than happy to start working on the issue and opening a PR once I have a working solution.
</details>

myah-mitchell:
Closes #5561.

When focus descends into a nested session, the host's `ModeInfo` still describes the
host, and Zellij exposes nothing else about the guest. A host status bar therefore
cannot say anything about the session the user is actually typing into. This adds the
missing reports. No default plugin is modified, so the diff is API surface plus the
plumbing to fill it.

### Wire protocol

Three arms on `NestedSessionMessage`, in the self-contained `nested_session_contract`
protobuf, plus a `HintReporting` capability advertised in the existing handshake:

- `GuestModeUpdate { mode, base_mode }`, guest to host
- `RequestGuestKeybinds`, host to guest
- `GuestKeybindsUpdate { keybinds }`, guest to host

Keybindings cross as an opaque `bytes keybinds_payload` holding a plugin-API
`InitialKeybindsPayload`, so the contract does not restate the `Action` schema. Modes
cross under the name `InputMode::from_str` reads, so a new mode needs no contract
change.

Everything is appended and no existing arm changed, so an older build decodes an
unknown arm as `Err` and drops the frame. An unknown capability is dropped without
losing the known ones.

### Plugin API

- `Event::NestedSessionModeUpdate { pane_id, session_name, mode, base_mode }`
- `Event::NestedSessionKeybinds { pane_id, session_name, keybinds }`
- `PluginCommand::RequestNestedSessionKeybinds(PaneId)`, with a `zellij-tile` shim

All under `PermissionType::ReadApplicationState`, matching `ModeUpdate` and
`GetPaneInfo`, delivered through the existing `EventType` subscription.

### Server and client

A guest reports its mode when the handshake completes, on every mode change, when
answering a keybinding request, and when a reconfigure rewrites its bindings.
`report_mode_to_host` returns immediately unless the session really is nested, so a
standalone session stays silent.

The unprompted report on handshake matters: without it a host does not know a guest
exists until the user happens to change mode inside it, so a host plugin has nothing
to ask key bindings about.

### Tests

- `nested_session.rs`: 21 unit tests. Every arm and every `InputMode` round-trips,
  unknown mode names reject the frame, tables keep their modes/keys/actions, empty
  tables survive, undecodable payloads are rejected, unknown capabilities are dropped
  without losing known ones, control-character bindings are dropped without losing the
  rest.
- `plugin_api/event.rs`: protobuf round trips for both events, plus one covering that
  an unrepresentable key costs only its own binding.
- `zellij-integration-tests/tests/nested_hint_reporting.rs`: 4 tests on the existing
  `NestedHarness` pattern, covering the handshake, the unprompted report, the request
  and answer, and mode changes flowing up while descended.

```
cargo test -p zellij-utils --lib                                    462 pass
cargo test -p zellij-server --lib                                  1576 pass
cargo test -p zellij-integration-tests --test nested_hint_reporting   4 pass
cargo fmt --all --check                                            clean
cargo xtask build                                                  succeeds
```

### Actual Use

I've already have a working update to my zjstatus-hints to use this API.
https://github.com/myah-mitchell/zjstatus-hints/tree/feat/nested-session-hint-events

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 
 ## [Unreleased]
+* feat: report a nested session's input mode and keybindings to its host, so host plugins can show hints for the session the user is actually typing into (https://github.com/zellij-org/zellij/issues/5561)
 
 ## [0.45.1] - 2026-08-28
 * fix: nested-session detection over SSH (https://github.com/zellij-org/zellij/pull/5522)
--- a/zellij-client/src/lib.rs
+++ b/zellij-client/src/lib.rs
@@ -672,7 +672,10 @@ pub async fn run_remote_client_terminal_loop(
                     if reannounce_scheduler.on_tick(std::time::Instant::now()) {
                         let announce = nested_session::NestedSessionMessage::Announce {
                             session_name: session_name.clone(),
-                            capabilities: vec![nested_session::NestedSessionCapability::NestedControl],
+                            capabilities: vec![
+                                nested_session::NestedSessionCapability::NestedControl,
+                                nested_session::NestedSessionCapability::HintReporting,
+                            ],
                         };
                         let mut stdout = os_input.get_stdout_writer();
                         if stdout
@@ -869,7 +872,10 @@ pub fn start_remote_client(
     stdout.write_all(ENABLE_FOCUS_REPORTING.as_bytes()).unwrap();
     let announce = nested_session::NestedSessionMessage::Announce {
         session_name: remote_session_name.clone(),
-        capabilities: vec![nested_session::NestedSessionCapability::NestedControl],
+        capabilities: vec![
+            nested_session::NestedSessionCapability::NestedControl,
+            nested_session::NestedSessionCapability::HintReporting,
+        ],
     };
     stdout
         .write_all(&nested_session::encode_frame(&announce))
@@ -1187,7 +1193,10 @@ pub fn start_client(
     stdout.write_all(ENABLE_FOCUS_REPORTING.as_bytes()).unwrap();
     let announce = nested_session::NestedSessionMessage::Announce {
         session_name: own_session_name.clone(),
-        capabilities: vec![nested_session::NestedSessionCapability::NestedControl],
+        capabilities: vec![
+            nested_session::NestedSessionCapability::NestedControl,
+            nested_session::NestedSessionCapability::HintReporting,
+        ],
     };
     stdout
         .write_all(&nested_session::encode_frame(&announce))
--- a/zellij-client/src/nested_reannounce.rs
+++ b/zellij-client/src/nested_reannounce.rs
@@ -48,7 +48,10 @@ impl NestedReannounce {
                     }
                     let announce = NestedSessionMessage::Announce {
                         session_name: session_name.clone(),
-                        capabilities: vec![NestedSessionCapability::NestedControl],
+                        capabilities: vec![
+                            NestedSessionCapability::NestedControl,
+                            NestedSessionCapability::HintReporting,
+                        ],
                     };
                     let mut stdout = os_input.get_stdout_writer();
                     if stdout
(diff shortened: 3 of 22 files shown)


imsnif:
Hey @myah-mitchell - a thousand pardons for the delay, thank you for your patience!

I took the liberty to make some changes:
1. I rebased from main
2. I adjusted the protocol so that it will also do the right thing with more levels of nesting (i.e. send the mode/keybindings from the most deeply nested session the user is focused on), as well as fix some potential races.
3. I adjusted the plugin command so that it gives feedback back to the plugin. We use this same pattern with other plugin commands. Basically, it blocks with a timeout of 1s while waiting for a reply, providing the requested info in the success arm or a relevant error (eg. timed out, too large, etc.) in the error arm so the plugin can decide how to handle it.

Example:
```rust
match get_nested_session_keybinds(PaneId::Terminal(pane_id)) {
    Ok(nested) => self.hints = Some((nested.session_path, nested.mode, nested.keybinds)),
    Err(NestedSessionKeybindsError::Timeout | NestedSessionKeybindsError::GuestUnresponsive) => set_timeout(1.0),
    Err(e) => self.hints_error = Some(e.to_string()),
}
```

I'll take a look at your other PR next. Thanks for this, great addition!

PR #5590, opened 2026-09-06
feat(plugin-api): let a plugin give its layout space back while empty

### LINKED ISSUE #5561: Plugin API: Expose a nested session's input mode and key bindings to its host session's plugins

### LINKED ISSUE #5588: Plugin API: let a plugin give its layout space back while it has nothing to draw
2. I need a new API

A plugin mounted as a bar (`size=1 borderless`) holds its row whether or not it draws
anything, so an empty bar shows as a blank line. I would like it to be able to hand the
row to its neighbors and later take it back, with the layout ending up as it started.

Nothing today does that:

- `hide_self()` does not survive the round trip. `Tab::unsuppress_pane` re-inserts
  through `add_tiled_pane`, which splits a neighbor rather than restoring the layout
  node, so a one-row borderless bar does not come back as one.
- `close_self()` does not come back at all.
- `Fixed(0)` cannot exist: `discretize_spans` errors on any span below 1.

What I am proposing is one command on the caller's own pane, alongside `hide_self` and
`close_self` and ungated for the same reason:

```rust
set_self_collapsed(collapsed: bool)
```

I'm working on a PR: https://github.com/myah-mitchell/zellij/tree/feat/plugin-collapse-self

It takes the pane out of the constraint solve rather than out of the layout, so
`constrain_spans` grows the neighbors over the row on its own and the pane keeps its
size constraint while it is out. 

Context, though the command mentions none of it: I hit this adding nested-session hint
reporting to zjstatus-hints (#5561), where a nested session's hidden bar leaves a blank
row. Any bar-shaped plugin with a quiet state wants the same thing.

EDIT:
See the latest release of [zjhints](https://github.com/myah-mitchell/zjhints) for an example of this in use in the README's pic and video.

myah-mitchell:
Closes #5588.

A plugin mounted as a bar holds its layout row whether or not it draws anything, so an
empty bar shows as a blank line. Nothing in the API covers giving that row up and
getting it back: `hide_self()` returns through `add_tiled_pane`, which splits a
neighbor rather than restoring the layout node, so a `size=1 borderless` bar does not
come back as one; `close_self()` does not come back at all; and `Fixed(0)` cannot exist
because `discretize_spans` errors on any span below 1.

This adds `PluginCommand::SetSelfCollapsed(bool)` through the usual chain, with a
`zellij-tile` shim `set_self_collapsed`. Ungated, matching `HideSelf`, `CloseSelf` and
`SetSelfMouseSelectionSupport`: it names no pane but the caller's own.

### How it works

The pane leaves the constraint solve, not the layout. `TiledPaneGrid::new` already
filters `panes_to_hide` and `constrain_spans` already forces the surviving spans to
fill the space, so the neighbors absorb the row on their own. The pane keeps its size
constraint while it is out, so a `Fixed(1)` bar is still `Fixed(1)` when it rejoins and
expanding restores the layout exactly.

Three things that are not obvious from the diff:

- Collapsed state lives in its own set rather than in `panes_to_hide`, because
  `set_fullscreen` overwrites that wholesale and `unset_fullscreen` clears it.
  `refresh_panes_to_hide` composes the two.
- `non_selectable_pane_geoms_inside_viewport` skips collapsed panes, otherwise
  `offset_viewport` still shrinks the viewport by a row nobody holds.
- Nothing writes geometry to a pane the solver cannot see, so `resize_whole_tab` and
  `relayout_tiled_panes` let collapsed panes take part and collapse them again over the
  result. `resize_whole_tab` already does this for fullscreen a few lines above.

### Scope and tests

452 insertions, 5 deletions, 11 files, 202 of the insertions tests. No default plugin
touched, so nothing changes for anyone who does not call it.

Eight tests in `tab_tests.rs` cover the neighbor absorbing the row, expanding
restoring the exact `PaneGeom`, both no-op cases, surviving a fullscreen round trip,
forgetting state on close, and following a tab resize. 

```
cargo test -p zellij-server --lib   1584 pass
cargo test -p zellij-utils --lib     450 pass
cargo fmt --all --check             clean
```

### Origin
I discovered the need for this after finishing the work to resolve #5561 and updating
my https://github.com/myah-mitchell/zjstatus-hints to use the new API. I was able
to hide hints inside nested panes but then if you made the nested zellij fullscreen
you no longer had a bottom bar. Additionally, when I hid the bottom bar there was
still an empty pane taking up the space that had been used, negating the whole reason
to hide the nested hints pane. I've already put this API to use with the work at
https://github.com/myah-mitchell/zjstatus-hints/tree/feat/collapse-when-empty.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 
 ## [Unreleased]
+* feat: let a plugin give its layout space back while it has nothing to draw, so a bar plugin does not hold an empty row (https://github.com/zellij-org/zellij/issues/5588)
 
 ## [0.45.1] - 2026-08-28
 * fix: nested-session detection over SSH (https://github.com/zellij-org/zellij/pull/5522)
--- a/zellij-server/src/panes/tiled_panes/mod.rs
+++ b/zellij-server/src/panes/tiled_panes/mod.rs
@@ -3043,11 +3043,11 @@ impl TiledPanes {
     /// anything.
     ///
     /// A collapsed pane drops out of the constraint solve, so its neighbors grow over the
-    /// space it held. It keeps its geometry untouched while it is out, since the solver only
-    /// ever writes to the panes it can see, which is what lets expanding restore exactly the
-    /// size the layout asked for rather than an approximation of it.
+    /// space it held. It keeps its size constraint the whole time, so expanding gives back
+    /// exactly the row or column the layout asked for rather than an approximation of it.
     ///
-    /// The caller relays out the tab afterwards.
+    /// The caller relays out the tab afterwards, which is also where the pane's geometry is
+    /// kept current: see `take_collapsed_panes`.
     pub fn set_pane_collapsed(&mut self, pane_id: PaneId, collapsed: bool) -> bool {
         let changed = if collapsed {
             self.collapsed_panes.insert(pane_id)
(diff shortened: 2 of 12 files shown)


imsnif:
Hey @myah-mitchell - I think I understand your need with this API. I wonder though, do you think we might be able to do this with the existing `suppressed_panes` in tab/mod.rs? (I'm not sure if there's an API to put a pane there - if not I'd be willing to add one)

These panes are panes that are still open but are not rendered anywhere. I'm leaning toward using these somehow because then we'd be hooking up into existing machinery rather than creating something new. What do you think?

PR #5611, opened 2026-09-12
`NestedListItem::new` require `Text` instead of `&str`

### LINKED ISSUE #5411: `NestedListItem::color_substring` does not exist

Zykino:
This make the `NestedListItem` construction consistent with `Table` and `Ribbon`.

I only fixed some examples usages to discuss in PR, full 400 errors correction will come later.

Fix #5411 more permanently
Supersede #5448

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 
 ## [Unreleased]
+* (BREAKING CHANGE zellij-tile) `NestedListItem` now require a `Text` instead of a `&str` (https://github.com/zellij-org/zellij/pull/5611) (Thanks @Zykino)
 
 ## [0.45.1] - 2026-08-28
 * fix: nested-session detection over SSH (https://github.com/zellij-org/zellij/pull/5522)
--- a/default-plugins/about/src/main.rs
+++ b/default-plugins/about/src/main.rs
@@ -400,17 +400,22 @@ impl App {
             .with_title(Text::new(format!("{}", error)).color_range(3, ..))
             .with_paragraph(vec![
                 ComponentLine::new(vec![ActiveComponent::new(TextOrCustomRender::Text(
-                    Text::new("Unable to permanently dismiss tips."),
+                    Text::from("Unable to permanently dismiss tips."),
                 ))]),
                 ComponentLine::new(vec![ActiveComponent::new(TextOrCustomRender::Text(
-                    Text::new("You can do so manually by adding the following to your config:"),
+                    Text::new(
+                        "You can do so manually by adding the following to your config:"
+                            .to_string(),
+                    ),
                 ))]),
             ])
             .with_paragraph(vec![ComponentLine::new(vec![ActiveComponent::new(
-                TextOrCustomRender::Text(Text::new("show_startup_tips false").color_range(0, ..)),
+                TextOrCustomRender::Text(
+                    Text::new("show_startup_tips false".into()).color_range(0, ..),
+                ),
             )])])
             .with_help(Box::new(|_hovering_over_link, _menu_item_is_selected| {
-                Text::new("<ESC> - dismiss").color_range(1, ..=4)
+                Text::new("<ESC> - dismiss".to_owned()).color_range(1, ..=4)
             }));
         error_page.render(rows, cols, &None)
     }
--- a/default-plugins/configuration/src/presets_screen.rs
+++ b/default-plugins/configuration/src/presets_screen.rs
@@ -449,27 +449,31 @@ impl PresetsScreen {
         let default_text = "1. Default";
         let (mut list_items, max_width) = if cols >= widths.0 {
             let list_items = vec![
-                NestedListItem::new(default_text).color_range(1, ..),
-                NestedListItem::new("All modes available directly from the base mode, eg.:")
+                NestedListItem::new(Text::from(default_text).color_range(1, ..)),
+                NestedListItem::new("All modes available directly from the base mode, eg.:".into())
                     .indent(1),
-                NestedListItem::new(format!(
-                    "{} p - to enter PANE mode",
-                    primary_modifier_key_text
-                ))
-                .color_range(3, ..primary_modifier_key_text_len + 3)
-                .color_range(
-                    2,
-                    primary_modifier_key_text_len + 14..primary_modifier_key_text_len + 18,
+                NestedListItem::new(
+                    Text::from(format!(
+                        "{} p - to enter PANE mode",
+                        primary_modifier_key_text
+                    ))
+                    .color_range(3, ..primary_modifier_key_text_len + 3)
+                    .color_range(
+                        2,
+                        primary_modifier_key_text_len + 14..primary_modifier_key_text_len + 18,
+                    ),
                 )
                 .indent(1),
-                NestedListItem::new(format!(
-                    "{} t - to enter TAB mode",
-                    primary_modifier_key_text
-                ))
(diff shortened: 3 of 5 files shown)


zellij-tile/src/ui_components/nested_list.rs
```diff
@@ -9,118 +8,16 @@ pub struct NestedListItem {
 }
 
 impl NestedListItem {
-    pub fn new<S: AsRef<str>>(text: S) -> Self
-    where
-        S: ToString,
-    {
+    pub fn new(text: Text) -> Self {
```
Zykino, line 11: Be honest, require a `Text` like `Ribbon` and `Table` already does.

zellij-tile/src/ui_components/nested_list.rs
```diff
@@ -9,118 +8,16 @@ pub struct NestedListItem {
...
-        self
-    }
-    pub fn success_color_last_substring<S: AsRef<str>>(mut self, substr: S) -> Self {
-        self.content = self.content.success_color_last_substring(substr);
-        self
-    }
-    pub fn success_color_all(mut self) -> Self {
-        self.content = self.content.success_color_all();
-        self
-    }
```
Zykino, line 123: Do not duplicate `Text` API without adding any value -> Prevent it from being out of sync

zellij-tile/src/ui_components/nested_list.rs
```diff
@@ -9,118 +8,16 @@ pub struct NestedListItem {
...
         self
     }
-    pub fn selected(mut self) -> Self {
-        self.content = self.content.selected();
-        self
-    }
-    pub fn opaque(mut self) -> Self {
-        self.content = self.content.opaque();
-        self
-    }
```
Zykino, line 32: > Do not duplicate Text API without adding any value -> Prevent it from being out of sync

I am less sure about theses, they may have a bit of value if you keep an array of `NestedListItem` and want to select a line already created.

I'm also open to propose a way to get/set `self.content` so a plugin author could change the underlying `Text` as needed.

zellij-tile/src/ui_components/text.rs
```diff
@@ -23,11 +23,20 @@ impl From<StyledText> for Text {
...
+        Text::new(value.to_owned())
+    }
+}
+
 impl Text {
-    pub fn new<S: AsRef<str>>(content: S) -> Self
-    where
-        S: ToString,
-    {
+    pub fn new(content: String) -> Self {
```
Zykino, line 39: Be honest about needing a String. So if someone want to give their Owned string don't force a clone. (At least that my rust's understanding.)

zellij-tile/src/ui_components/text.rs
```diff
@@ -23,11 +23,20 @@ impl From<StyledText> for Text {
...
+    fn from(value: String) -> Self {
+        Text::new(value)
+    }
+}
+
+impl From<&str> for Text {
+    fn from(value: &str) -> Self {
+        Text::new(value.to_owned())
+    }
+}
```
Zykino, line 36: Provide String/&str `.into()`.

CHANGELOG.md
```diff
@@ -5,6 +5,7 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 
 ## [Unreleased]
+* (BREAKING CHANGE zellij-tile) `NestedListItem` now require a `Text` instead of a `&str` (https://github.com/zellij-org/zellij/pull/5611) (Thanks @Zykino)
```
Zykino, line 8: I saw some project celebrating PR commiter. What do you think about it? (Happy to remove if you don't want to do that for all changes.)

Also its been a long time since the last noted "BREAKING CHANGE", so I'm not sure you want to keep this?

default-plugins/about/src/main.rs
```diff
@@ -400,17 +400,22 @@ impl App {
...
             ])
             .with_paragraph(vec![ComponentLine::new(vec![ActiveComponent::new(
-                TextOrCustomRender::Text(Text::new("show_startup_tips false").color_range(0, ..)),
+                TextOrCustomRender::Text(
+                    Text::new("show_startup_tips false".into()).color_range(0, ..),
+                ),
             )])])
             .with_help(Box::new(|_hovering_over_link, _menu_item_is_selected| {
-                Text::new("<ESC> - dismiss").color_range(1, ..=4)
+                Text::new("<ESC> - dismiss".to_owned()).color_range(1, ..=4)
```
Zykino, line 418: What do you prefer?
1) Using the new `Text::from`?
1) Keeping `Text::new` with
    1) `.to_string()`
    1) `.into()`
    1) `.to_owned()`

default-plugins/configuration/src/presets_screen.rs
```diff
@@ -449,27 +449,31 @@ impl PresetsScreen {
...
+                NestedListItem::new(
+                    Text::from(format!(
+                        "{} t - to enter TAB mode",
+                        primary_modifier_key_text
+                    ))
+                    .color_range(3, ..primary_modifier_key_text_len + 3)
+                    .color_range(
+                        2,
+                        primary_modifier_key_text_len + 14..primary_modifier_key_text_len + 17,
+                    ),
```
Zykino, line 476: Example change for `NestedListItem` fixed as best as I could. Note the 2nd line where for fixed `&str` without modifier the change is only an `into()` away: `"[…]".into()`. This may also simplify Ribbon/Table constructions… but I won't touch them if they are not in the compilation error list.

CHANGELOG.md
Zykino: Note: the breaking change is only for `zellij-tile`, I did not touch the struct nor the protobuf serialized string/API.

Zykino (commented):
I wanted to comment/explain my choices and question here. What do you think about this change?

Remember I have ~400 compilation errors to fix once we are OK with the API. So the full change will touch a lot more files. I wanted to show a small portion of the changes to agree on something before doing the rest.

default-plugins/about/src/main.rs
imsnif: I think using the new `Text::from` would be the most idiomatic, but I don't have strong feelings about this.

karpovantonme:
Happy to. Taking the `Text` instead of forwarding each builder is the better fix, #5448 was patching the symptom and this removes the reason the API can drift at all. Feel free to close mine once this lands.

Two things I would look at.

**Narrowing `Text::new` to `String` is not needed for any of it.** `NestedListItem::new(text: Text)` works the same whether `Text::new` stays generic or not, and the two are independent changes riding in one PR. The cost is not: `NestedListItem::new` has 42 call sites in the repo, `Text::new` has 432 across 38 files, 35 of them in `default-plugins`. That is the 400 errors. Keeping `Text::new<S: AsRef<str>>` and adding the `From` impls costs nothing, `impl From<&str> for Text` does not conflict with a generic inherent `new`, and then the breakage is the 42 sites this PR is actually about.

If the intent is to make `Text::from` the one way in, that is a defensible call, but it is a separate deprecation with a much wider blast radius and the CHANGELOG entry currently only flags `NestedListItem`.

**The diff picks four different spellings for the same thing**, all in two files: `Text::from(default_text)`, `Text::new("...".to_string())`, `Text::new("...".into())`, `Text::new("...".to_owned())`. Worth settling on one before the 400 fixes, otherwise whichever mix lands is what the next contributor copies.

Two smaller ones:

- `pub fn new(content: String)` still does `text: content.to_string()`, which clones a `String` you already own. `text: content`.
- `content` on `NestedListItem` is private, and with the forwards gone there is no `content()` or `len()` either, so a built item is write-only. The builders are covered by taking a `Text`, the two getters are not. `pub fn content(&self) -> &Text` would close that.

PR #5630, opened 2026-09-18
Yazelix fixes (sixel/kitty correctness, explicit-swap-layout change rather than toggle, border styling)

imsnif:
This PR mixes a few different fixes and two new features, all derived from the Yazelix Zellij fork (nova) and intended to allow Zellij to provide a full solution for Yazelix, negating the need for a fork, while also providing benefits for other Zellij users and derivative creators.

Note: Most of these fixes were issued by @luccahuguet on the fork itself. I only adapted them to main and made some adjustments.

### The fixes
1. Sixel: a monotonic counter is used rather than reusing image IDs, which as a result of some drift since the feature was implemented are no longer app-unique. Additionally, some correctness fixes were issued for erase detection and reaping.
2. Kitty graphics: correctness fixes with regards to image scaling and eviction.
3. Stacked panes: use logical position ordering when closing and rebuilding stacked members so as to avoid order changes.

### The features
1. It's now possible to apply swap layouts by name specifically, on both planes (tiled + floating) regardless of which ones are visible. This is possible through the cli with a new `zellij action apply-[floating|tiled]-swap-layout [vertical|horizontal|etc.]` command, through the plugin API (see generated API docs, more will be written upon release) and through a keybinding.
2. It's now possible to set double/heavy/dashed/heavy-dashed borders to tiled/floating panes. Either in a global config, a layout or when opening panes from a plugin/cli. It's also possible to mix and match them (eg. top dashed, bottom heavy, etc.) I think this can be cool for plugins wanting to create "read", "needs attention" etc. panes without relying solely on frame color.

@luccahuguet - could you please give this a test to see if I missed anything? If all of this is good, please make a PR to this branch adding this PR to the CHANGELOG.md file so that you'll get credit for these fixes.


luccahuguet:
So happy to be a part of this!

Ok, I will test it

luccahuguet:
I fixed the validation gaps in #5631: the affected snapshots, unknown-pane error reporting for `set-pane-border-style`, tab-ID layout error propagation, and the requested changelog entry.

Locally verified at `3abfd1952`: formatting, 489 `zellij-utils` tests, 1,629 `zellij-server` tests, and all eight affected ignored plugin tests (nine snapshot files). Direct push to this branch was permission-blocked, so #5631 targets `yazelix-fixes`.

imsnif:
Thanks @luccahuguet ! 

I looked at the status-bar change and decided it's a bit too much risk given how specific some people's setups are. I don't want to break anything for anyone. Will it work for you to include it as a separate plugin?

Otherwise, I think everything you asked for is in. What do you think?

luccahuguet:
> Thanks @luccahuguet !
> 
> I looked at the status-bar change and decided it's a bit too much risk given how specific some people's setups are. I don't want to break anything for anyone. Will it work for you to include it as a separate plugin?
> 
> Otherwise, I think everything you asked for is in. What do you think?

Yes, that works for me!

The change is already confined to the status-bar plugin and uses existing plugin events, so we’ll package it as a Yazelix-owned status-bar variant and remove that delta from our Zellij fork. 

Thanks!

imsnif:
> > Thanks @luccahuguet !
> > I looked at the status-bar change and decided it's a bit too much risk given how specific some people's setups are. I don't want to break anything for anyone. Will it work for you to include it as a separate plugin?
> > Otherwise, I think everything you asked for is in. What do you think?
> 
> Yes, that works for me!
> 
> The change is already confined to the status-bar plugin and uses existing plugin events, so we’ll package it as a Yazelix-owned status-bar variant and remove that delta from our Zellij fork.
> 
> Thanks!

Lovely! So... anything else you need before we merge? Does this remove your need for the fork?

luccahuguet:
Thanks! Things are looking good. I tested Nova with #5630 and without our other fork patches; startup and attach/reattach worked. I don’t need anything else in this PR before it merges.

It gets us close, but I can’t retire the fork yet. We still carry #5272, which we reproduced in targeted tests, and #5481, which we reproduced with a stuck server. We can handle both outside this PR.

The other issue is first-launch permissions. Nova ships plugins we created or reviewed as part of its built-in experience, but users have to approve numerous permission prompts across different parts of the UI. Either of these would work for us:

- One prompt showing each plugin and the permissions it requests, with a single approval.
- A way for Yazelix to select its own permission-cache file, perhaps through an env var, so we can pre-seed grants for the exact plugins we ship. I can open a separate PR for this if you’d like.

Thanks again for bringing these changes upstream!

luccahuguet:
(I also had permission bugs in the path, but I don't know if they are still applicable, but they were painful)

imsnif:
Alright, thanks @luccahuguet . I think we might have had a miscommunication, so just to stress: I want to do one go and solve all of these problems together along with retiring the fork. I feel that otherwise this will drag on and runs the risk of never ending.

Could you please take a moment and investigate everything you would need for this to happen then give me a bulletin list? I'll design solutions and implement them as part of this PR. Thanks!

luccahuguet:
Ah, sure @imsnif, that works too. Lets bundle everything here.

This is the complete list required to retire the fork:

- Client and plugin lifecycle: https://github.com/zellij-org/zellij/pull/5272
- Bounded session discovery: https://github.com/zellij-org/zellij/pull/5481
- First-launch plugin permissions: Nova’s packaged background plugins must have a way to be vendored / built-in by Nova, instead of the user having to accept numerous permission prompts in different parts of the UI for plugins that were already screened (or created by) Nova. The smallest solution for us is allowing an isolated permission-cache path through an environment variable when no explicit path is supplied. So an explicit path should win and the existing default should remain unchanged. Nova can then pre-seed grants only for the exact plugins it ships. (either opt-in or opt-out, it does not matter) Our current fork implementation is available here for reference: https://github.com/Yazelix/nova-zellij/commit/106d9ccda56eed19fdc329b732532ca401071d67

imsnif:
Hey @luccahuguet - all issues addressed:

1. For the first issue, I issued a fix which I see as more holistic: namely not re-using client ids (as well as fixing the cleanup of plugin assets on detach, as well as other issues in that area that I found along the way).
2. I merged the linked PR regarding the received timeout.
3. Regarding the plugin permission issues you raised, I felt the best solution is not running away from the problem, but rather giving you the power to bundle Yazelix properly as a Zellij distribution. More below.

### What are Zellij Distributions

Zellij Distributions allow you to do what up until now you've had to hack your way around. Use Zellij as a library. You'll now be able to import Zellij, include your own plugins, layouts and configuration, give it your own name (eg. Yazelix) and then create and distribute your own executable. Your users don't even have to know they're running Zellij.

Naturally, this would also mean that any built-in plugins you include will not require user permission (they are already trusting your executable, just like they would otherwise trust the Zellij executable).

A brief example:

```toml
# Cargo.toml
[package]
name = "yazelix"
version = "0.9.0"
edition = "2021"

[dependencies]
zellij = { git = "https://github.com/zellij-org/zellij", tag = "v0.46.0" }
```
```rust
// src/main.rs
use zellij::{Distribution, DistributionLayout};

fn main() {
    zellij::run(
        Distribution::new("yazelix", env!("CARGO_PKG_VERSION"))
            .with_display_name("Yazelix")
            .with_plugin("yazelix-sidebar", include_bytes!("../plugins/sidebar.wasm"))
            .with_plugin("yazelix-status", include_bytes!("../plugins/status.wasm"))
            .with_plugin("zjstatus", include_bytes!("../plugins/zjstatus.wasm"))
            .without_plugin("status-bar")
            .with_config(include_str!("../assets/config.kdl"))
            .with_layout(DistributionLayout::new(
                "yazelix",
                include_str!("../assets/layouts/yazelix.kdl"),
            )),
    )
}
```

There's a fuller working example in the `example/distribution` folder to get you started.

What do you think? Want to give this a test ride to see if I missed anything?

luccahuguet:
> Hey @luccahuguet - all issues addressed:
> 
>     1. For the first issue, I issued a fix which I see as more holistic: namely not re-using client ids (as well as fixing the cleanup of plugin assets on detach, as well as other issues in that area that I found along the way).
> 
>     2. I merged the linked PR regarding the received timeout.
> 
>     3. Regarding the plugin permission issues you raised, I felt the best solution is not running away from the problem, but rather giving you the power to bundle Yazelix properly as a Zellij distribution. More below.
> 
> 
> ### What are Zellij Distributions
> 
> Zellij Distributions allow you to do what up until now you've had to hack your way around. Use Zellij as a library. You'll now be able to import Zellij, include your own plugins, layouts and configuration, give it your own name (eg. Yazelix) and then create and distribute your own executable. Your users don't even have to know they're running Zellij.
> 
> Naturally, this would also mean that any built-in plugins you include will not require user permission (they are already trusting your executable, just like they would otherwise trust the Zellij executable).
> 
> A brief example:
> 
> ```toml
> # Cargo.toml
> [package]
> name = "yazelix"
> version = "0.9.0"
> edition = "2021"
> 
> [dependencies]
> zellij = { git = "https://github.com/zellij-org/zellij", tag = "v0.46.0" }
> ```
> 
> ```rust
> // src/main.rs
> use zellij::{Distribution, DistributionLayout};
> 
> fn main() {
>     zellij::run(
>         Distribution::new("yazelix", env!("CARGO_PKG_VERSION"))
>             .with_display_name("Yazelix")
>             .with_plugin("yazelix-sidebar", include_bytes!("../plugins/sidebar.wasm"))
>             .with_plugin("yazelix-status", include_bytes!("../plugins/status.wasm"))
>             .with_plugin("zjstatus", include_bytes!("../plugins/zjstatus.wasm"))
>             .without_plugin("status-bar")
>             .with_config(include_str!("../assets/config.kdl"))
>             .with_layout(DistributionLayout::new(
>                 "yazelix",
>                 include_str!("../assets/layouts/yazelix.kdl"),
>             )),
>     )
> }
> ```
> 
> There's a fuller working example in the `example/distribution` folder to get you started.
> 
> What do you think? Want to give this a test ride to see if I missed anything?

Wow, awesome.

Yes,  I shall test this soon and report back.

Just a comment, but I also do like this approach a lot: Looking at a problem from the lens of "how can I one-up this problem to not only to fix it, but to make this part of my project even better?"

luccahuguet:
@imsnif Yazelix Nova edge is already using this branch, things are working remarkably well. 

I already solved the bottom-bar problem as well

imsnif:
Great to hear @luccahuguet ! Are you working with the distributions API? I hope to have this released soon (and would generally recommend you work with a release rather than main or a derivative thereof, since those represent ongoing work that tends to break).

Otherwise, I hope I've shown enough responsiveness for you to reach for collaboration rather than immediate forking in the future. :wink:

luccahuguet:
> Great to hear @luccahuguet ! Are you working with the distributions API? 

yes!

> I hope to have this released soon (and would generally recommend you work with a release rather than main or a derivative thereof, since those represent ongoing work that tends to break).

Yeah, I agree. I am using main now only in the `edge` branch of nova (we have edge, main and stable)
and will wait for a release before promoting edge to main and stable.

it's just this PR is a bit special and i wanted nova to keep up with it, tightly


> Otherwise, I hope I've shown enough responsiveness for you to reach for collaboration rather than immediate forking in the future. 😉

Aye, that's for sure! ❤️

